PR #53, opened 2021-08-27
Initial downstairs state work.

leftwo:
The first of probably many change to support keeping a state for each
downstairs in the upstairs.  It is mostly comments at this point with
only some states set.

Added a panic hook to the upstairs-client and the downstairs binaries, so
any async task that panics will panic the entire program.  The upstairs
lib does not set it, as that is left to the user of the library to
decide what they want to do.

Added more comments in the downstairs around the ExtentMeta format.

Updated the names of several channels used for communication between
the upstairs tasks to better reflect their roles.  Updated comments
around the channels to describe what they do.

Broke apart the up_listen() function into two parts, and added a loop
and check in the beginning of up_listen() that waits for all three
downstairs tasks to start, contact their respective downstairs, and
get status back, before up_listen() will start taking work off the
queue from the guest.  Moved the rest of up_listen() into a function
that does the work of Guest->Upstairs IO translation.

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -48,6 +48,17 @@ pub fn opts() -> Result<Opt> {
  */
 fn main() -> Result<()> {
     let opt = opts()?;
+
+    /*
+     * If any of our async tasks in our runtime panic, then we should
+     * exit the program right away.
+     */
+    let default_panic = std::panic::take_hook();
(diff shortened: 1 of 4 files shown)


jmpesp (approved):
Looks good, one typo

PR #54, opened 2021-08-31
increment cur_offset correctly

jmpesp:
unfortunately this code works until the operation spans more than two extents

luckily, f3write + f3read found it


leftwo (approved):
Seems like we should have found this sooner.

I know the extent file size max limitation is out there and we will have to solve it at some point.  For now, if you run into it, we can just make the extents larger :)

I'm working on a simple data verification test all in crucible that should help us catch issues like this in the future.

PR #55, opened 2021-09-07
Use only block sized operations

jmpesp:
Perform RMW Upstairs if read and write operations are not block aligned and block sized. This currently means an atomicity failure, but in testing with propolis I haven't seen this yet. I have seen it with nbd and hammer testing though.

Put all the logic that handles turning arbitrary reads and writes into block aligned and block sized operations into a new struct IOSpan. Guest::read and Guest::write now uses IOSpan.

This commit also adds more instrumentation, and switches to non-blocking random to generate data for the C version of hammer.

### DIFF
--- a/hammer/src/hammer.c
+++ b/hammer/src/hammer.c
@@ -9,7 +9,11 @@
 int main() {
   srandom(clock());
 
-  const size_t total_size = 512*100*1000; // TODO: get from region.json
+  // TODO: get from region.json
+  const size_t block_size = 512;
+  const size_t extent_size = 512;
+  const size_t extent_count = 32768;
+  const size_t total_size = block_size * extent_size * extent_count;
 
   while (1) {
     size_t offset = random() % total_size;
@@ -28,7 +32,7 @@ int main() {
     size_t r;
 
     // get some random data
-    fp = fopen("/dev/random", "rb");
+    fp = fopen("/dev/urandom", "rb");
     fn = fileno(fp);
 
     if (sz != read(fn, buf, sz)) {
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -1770,9 +1770,109 @@ impl BlockReqWaiter {
         Self { recv }
     }
 
+    /*
+     * A no-op BlockReqWaiter returns immediately.
+     */
+    fn noop() -> BlockReqWaiter {
+        let (send, recv) = std_mpsc::channel();
+        let _ = send.send(0);
+        BlockReqWaiter::new(recv)
+    }
+
     pub fn block_wait(&mut self) {
         // TODO: Instead of i32, errors should be bubbled up.
         let _ = self.recv.recv();
+    }
+}
+
+/*
+ * IO operations are ok to submit directly to Upstairs if:
+ *
+ * - the offset is block aligned, and
+ * - the size is a multiple of block size
+ *
+ * If either of these is not true, then perform some fixup here.
+ */
+#[derive(Debug)]
+struct IOSpan {
+    // IOP details
+    offset: u64,
+    sz: u64,
+
+    // Block and Guest details
+    block_size: u64,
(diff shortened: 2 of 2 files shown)


leftwo (commented):
I have some nit worthy comments, but before I do that I wanted to discuss what we get with these changes.  If I understand correctly, this will add the ability to send non block aligned size/offset reads or writes into Crucible, but, in the case of a write, won't enforce atomicity?

If we can't enforce it, then I wonder if we need a bit more here to actually do the enforcement.  Even if the enforcement involves terrible performance.  As it stands, we could get corruption if the right set of IO's came through.  The enforcement could involve pre-registering some list of job IDs for the future write portion of the operation, and make sure these job ID dependencies are populated on any IO that might be submitted after the RMW IO is accepted.

I considered the idea of moving the RMW into crucible internals, but given we think this is not a common operation and that we can take a perf hit to do it, keeping it implemented where you have it now, just above the crucible guest work, is okay (in my opinion, others are welcome).

jmpesp:
> If I understand correctly, this will add the ability to send non block aligned size/offset reads or writes into Crucible, but, in the case of a write, won't enforce atomicity?

With this PR, Guests that send non block sized or non block aligned reads or writes will have those transformed into block sized block aligned ops before they go Downstairs. If the Guest already sends block sized and block aligned reads and writes, nothing will change :)

Given that propolis' current [virtio](https://github.com/oxidecomputer/propolis/blob/c77c43ad2ab3a0d39759061948cbf2234a04287e/propolis/src/hw/virtio/block.rs#L111) and [nvme](https://github.com/oxidecomputer/propolis/blob/c77c43ad2ab3a0d39759061948cbf2234a04287e/propolis/src/hw/nvme/ns.rs#L160) implementations multiply by sector size when creating a BlockReq, we won't see non block aligned or non block sized reads or writes hit Crucible from there.

> If we can't enforce it, then I wonder if we need a bit more here to actually do the enforcement. Even if the enforcement involves terrible performance. As it stands, we could get corruption if the right set of IO's came through. The enforcement could involve pre-registering some list of job IDs for the future write portion of the operation, and make sure these job ID dependencies are populated on any IO that might be submitted after the RMW IO is accepted.

I took a crack at this idea but wasn't satisfied with the way I did it. I'll try again tomorrow.

jmpesp:
A thought: it's possible the right move here is to undo https://github.com/oxidecomputer/crucible/pull/44? That way Crucible only allows block sized block aligned ops, without the need for fixup. NBD server and Hammer would have to be fixed / changed to support this but that's fine.

leftwo:
> A thought: it's possible the right move here is to undo #44? That way Crucible only allows block sized block aligned ops, without the need for fixup. NBD server and Hammer would have to be fixed / changed to support this but that's fine.

By undo, you mean just put back the assertion that IO only comes in block sizes?  If so, then that would work. 

I guess what I'm looking for here is either we go all in and support RMW, or explicitly block it.  I would hate to debug some corruption down the road because we left a door open :)

I was also thinking about our current 512 byte block size, and that may be why we don't see many I/Os that are not aligned.  If we move to a 4k block size, I wonder if that will turn the tables and force us to handle I/Os of 512 bytes that currently just happen to be our selected block size.

jmpesp:
> By undo, you mean just put back the assertion that IO only comes in block sizes? If so, then that would work.

I mean change the API back to accepting block size / offset instead of byte size / offset :)

leftwo:
> > By undo, you mean just put back the assertion that IO only comes in block sizes? If so, then that would work.
> 
> I mean change the API back to accepting block size / offset instead of byte size / offset :)

So, by default we could never receive a bad offset (unless it was too big). 
Same thing with size, if the size is reported as number of blocks.

Yeah, if that is what NVMe will be giving us, then that should be good.

jmpesp:
Ok, I'm going to close this then.

PR #56, opened 2021-09-08
Add simple integrity test to crucible client

leftwo:
Updated the crucible-client program to have a real simple data integrity
test.  Added a random and a sequential tests as options to the client
program.

Updated the workload option to expect an enum of valid test names using
the arg_enum macro.

Fixed a bug where the dep_workload test would create offsets too large.

Added tests for the integrity checker.

Fixed a bug where sometimes the downstairs region test would fail.

Made a silly test script to start up three downstairs, then run through
the tests in crucible-client.

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -5,19 +5,35 @@ use std::sync::Arc;
 use anyhow::{bail, Result};
 use bytes::{BufMut, Bytes, BytesMut};
 use rand::prelude::*;
+use rand_chacha::rand_core::SeedableRng;
+use structopt::clap::arg_enum;
 use structopt::StructOpt;
 use tokio::runtime::Builder;
 
 use crucible::*;
 
+/*
+ * The various tests this program supports.
+ */
+arg_enum! {
+    #[derive(Debug, StructOpt)]
+    enum Workload {
+        One,
+        Big,
+        Dep,
+        Rand,
+        Baloon,
+    }
+}
+
 #[derive(Debug, StructOpt)]
-#[structopt(about = "volume-side storage component")]
+#[structopt(about = "crucible upstairs test client")]
 pub struct Opt {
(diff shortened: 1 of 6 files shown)


client/src/main.rs
```diff
@@ -5,19 +5,35 @@ use std::sync::Arc;
...
+ * The various tests this program supports.
+ */
+arg_enum! {
+    #[derive(Debug, StructOpt)]
+    enum Workload {
+        One,
+        Big,
+        Dep,
+        Rand,
+        Baloon,
```
jmpesp, line 25: typo

client/src/main.rs
```diff
@@ -142,6 +183,299 @@ fn main() -> Result<()> {
...
+    for (i, wc) in write_count.iter().enumerate().take(total_blocks) {
+        print!("{:5} ", wc);
+        if (i + 1) % (extent_size as usize) == 0 {
+            println!();
+        }
+    }
+}
+
+/*
+ * Generate a random offset and length, and write to then read from
```
jmpesp, line 396: this is what hammer does :) should we fold hammer into this?

jmpesp (changes requested):
:shipit: except for two comments

client/src/main.rs
leftwo: I got the feeling hammer was a bit better test than this, but maybe in the long run they can come together.  I think hammer uses a different interface and that has value as well.  

And, we should at some point come up with a more formal test plan for Crucible, but it's probably beyond the scope of this PR.

client/src/main.rs
leftwo: fixed in 78fd508

And, perhaps it's time for me to return to 7th grade and finish my spelling assignments.

jmpesp (approved):

PR #57, opened 2021-09-10
Revert Crucible back to only block operations

jmpesp:
Revert https://github.com/oxidecomputer/crucible/pull/44/, and go back to only block sized + block offset operations. Importantly, introduce the "Block" type. Attempting to pass just a u64 into Upstairs is no longer supported - users must clearly state they want block X where blocks are size Y in order to avoid passing the wrong offset.

Downstairs will register with a certain Block shift, and then Upstairs operations will fail if passed a different Block shift.

CruciblePseudoFile now uses IOSpan and performs RMW. It uses RwLock so that there isn't an atomicity failure.

Tested with:
- hammer
- import + nbd + md5sum
- all client workloads
- propolis (patched)

Alan and I talked about error handling today and I think the next PR will be to clean up some of the error related TODOs.

### DIFF
--- a/common/src/region.rs
+++ b/common/src/region.rs
@@ -2,17 +2,64 @@
 use anyhow::{bail, Result};
 use serde::{Deserialize, Serialize};
 
+/*
+ * Where the unit is blocks, not bytes, make sure to reflect that in the types used.
+ *
+ * Consumers of this API should know when to use bytes (rarely), and when to use blocks.
+ *
+ * Blocks have a shift field to ensure that consumers and the upstairs agree on what a block is. It
+ * wouldn't make sense to pass Block { 2, 9 } when the downstairs expects Block { 2, 12 }.
+ */
+#[derive(Deserialize, Serialize, Copy, Clone, Debug, PartialEq, PartialOrd)]
+pub struct Block {
+    // Value could mean a size or offset
+    pub value: u64,
+
+    // block size as a power of 2
+    // shift  9 -> 512
+    // shift 12 -> 4096
+    pub shift: u32,
+}
+
+impl Block {
+    pub fn new(value: u64, shift: u32) -> Block {
+        // are you sure you need blocks that big?
+        assert!(shift <= 16);
+
+        Block { value, shift }
+    }
+
+    pub fn new_512(value: u64) -> Block {
+        Block::new(value, 9)
+    }
+
+    pub fn new_with_ddef(value: u64, ddef: &RegionDefinition) -> Block {
+        Block {
+            value,
+            shift: ddef.block_size().trailing_zeros(),
+        }
+    }
+
+    pub fn block_size_in_bytes(&self) -> u32 {
+        1 << self.shift
+    }
+
+    pub fn byte_value(&self) -> u64 {
+        self.value * self.block_size_in_bytes() as u64
+    }
+}
+
 #[derive(Deserialize, Serialize, Copy, Clone, Debug, PartialEq)]
 pub struct RegionDefinition {
     /**
-     * The size of each block in bytes.  Must be a power of 2, minimum 512.
+     * The size of each block in bytes. Must be a power of 2, minimum 512.
      */
     block_size: u64,
 
     /**
      * How many blocks should appear in each extent?
      */
-    extent_size: u64,
+    extent_size: Block,
 
     /**
      * How many whole extents comprise this region?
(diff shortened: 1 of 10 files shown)


upstairs/src/lib.rs
```diff
@@ -2407,22 +2445,126 @@ fn show_guest_work(guest: &Arc<Guest>) {
...
+            block_size,
+            phase: offset % block_size,
+            buffer: Buffer::new(
+                affected_block_numbers.len() * block_size as usize,
+            ),
+            affected_block_numbers,
+        }
+    }
+
+    fn is_block_aligned_and_block_sized(&self) -> bool {
```
leftwo, line 2490: This, and a few of the other function names here I think might be too long.  What if it was named span_check_alignment, or span_alignment, or span_align_check?  Or maybe span_align_ok?

upstairs/src/lib.rs
```diff
@@ -2407,22 +2445,126 @@ fn show_guest_work(guest: &Arc<Guest>) {
...
+        let is_block_aligned = (self.offset % self.block_size) == 0;
+        let is_block_sized = (self.sz % self.block_size) == 0;
+
+        is_block_aligned && is_block_sized
+    }
+
+    #[cfg(test)]
+    fn number_of_affected_blocks(&self) -> usize {
```
leftwo, line 2498: Possibly affected_block_count()

upstairs/src/lib.rs
```diff
@@ -2437,35 +2579,50 @@ impl CruciblePseudoFile {
...
+
+            let mut waiter = span.read_affected_blocks_from_guest(&self.guest);
+            waiter.block_wait();
+
+            span.write_from_buffer_into_blocks(buf);
+
+            let mut waiter = span.write_affected_blocks_to_guest(&self.guest);
+            waiter.block_wait();
+        } else {
+            let offset = Block::new(
```
leftwo, line 2618: Does the else path need to get the lock?

leftwo (approved):
Thanks for the work here, sprinkled all over the place.
I have a few function name comments and one question about a lock, but it looks good and make whatever changes you feel like and push this to main.

common/src/region.rs
```diff
@@ -2,17 +2,64 @@
...
+    // block size as a power of 2
+    // shift  9 -> 512
+    // shift 12 -> 4096
+    pub shift: u32,
+}
+
+impl Block {
+    pub fn new(value: u64, shift: u32) -> Block {
+        // are you sure you need blocks that big?
+        assert!(shift <= 16);
```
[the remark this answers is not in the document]
jmpesp: Done in b8331bf.

protocol/src/lib.rs
```diff
@@ -117,16 +119,18 @@ impl Encoder<Message> for CrucibleEncoder {
 
                 Ok(())
             }
-            Message::Write(rn, eid, dependencies, byte_offset, data) => {
+            Message::Write(rn, eid, dependencies, offset, data) => {
```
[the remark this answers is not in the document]
jmpesp: Next PR :)

upstairs/src/lib.rs
```diff
@@ -1875,16 +1882,47 @@ impl Guest {
...
         self.send(wio)
     }
 
+    /*
+     * `bread` and `bwrite` accept a byte offset, and data must be a multiple of block size.
+     *
+     * note these functions assert that the byte offset is divisible by the block size and should
+     * only be used in non-production code.
+     */
+    pub fn bread(&self, offset: u64, data: Buffer) -> BlockReqWaiter {
```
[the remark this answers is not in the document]
jmpesp: Done in 9d46f3d

upstairs/src/lib.rs
jmpesp: :+1: I named it `is_block_regular` in 92eb012, but I'm not sure how to make the other function names shorter.

upstairs/src/lib.rs
jmpesp: Done in 07ece49

upstairs/src/lib.rs
jmpesp: Yes, good catch :) I also wrote a comment to clarify. Fixed in 3c94740, please double check the comment's logic :)

upstairs/src/lib.rs
```diff
@@ -2437,35 +2588,59 @@ impl CruciblePseudoFile {
...
     fn write(&mut self, buf: &[u8]) -> IOResult<usize> {
-        let mut data = BytesMut::with_capacity(buf.len());
-        data.put_slice(buf);
+        let mut span =
+            IOSpan::new(self.offset, buf.len() as u64, self.block_size);
 
-        let mut waiter = self.guest.write(self.offset, data.freeze());
-        waiter.block_wait();
+        /*
+         * Crucible's dependency system will properly resolve requests in the order they're sent
```
leftwo, line 2616: I would word this as:

"Crucible's dependency system will properly resolve requests in the order they are received.."

We don't actually know who "sent" it first, all Crucible knows is when it received them.  It may
be exactly the same.

upstairs/src/lib.rs
leftwo: Oh, added a comment at the comment.

leftwo (approved):
Still approved.  I added a comment on the comment

jmpesp:
> Still approved. I added a comment on the comment

:+1: fixed that, merging now

PR #58, opened 2021-09-14
Serialize and deserialize Message using serde

jmpesp:
Derive Serialize and Deserialize for Message, and put that on the wire
automatically. Keep the length part of the frame.

I chose bincode after benchmarking[1] each binary based encoding that serde
supports (and JSON for reference):

    $ ministat -s -w 70 bench_*
    x bench_bendy
    + bench_bincode
    * bench_cbor
    % bench_flexbuffers
    # bench_postcard
    @ bench_rmp_serde
    O bench_serde_json
    +----------------------------------------------------------------------+
    |    +OO O@x                                           O    O          |
    |+ O*OOO*OOOO# * xO+x  %    *   #@   %    x     O     OO OOOO         O|
    |     |______M___A__________|                                          |
    ||_____M_A________|                                                    |
    |  |_____MA______|                                                     |
    |   |______M__A________|                                               |
    | |____M___A_______|                                                   |
    |  |______M_A_______|                                                  |
    |                                                   |_____A____|       |
    +----------------------------------------------------------------------+
	N           Min           Max        Median           Avg        Stddev
    x  10          9.24          9.99          9.36         9.443    0.23466525
    +  10           9.1          9.68          9.23         9.283    0.17346469
    No difference proven at 95.0% confidence
    *  10          9.15          9.69          9.27         9.294    0.15882905
    No difference proven at 95.0% confidence
    %  10           9.2          9.87          9.32         9.376    0.20233086
    No difference proven at 95.0% confidence
    #  10          9.18          9.76          9.23         9.306    0.18025907
    No difference proven at 95.0% confidence
    @  10          9.15           9.8          9.29         9.327    0.18427335
    No difference proven at 95.0% confidence
    O  10         10.12         10.59         10.34        10.325    0.12167808
    Difference at 95.0% confidence
	    0.882 +/- 0.175623
	    9.34025% +/- 1.85982%
	    (Student's t, pooled s = 0.186914)

It looks like each encoding is the same (according to ministat), so I chose
bincode because it's "zero-fluff" and should perform some simple compression.

[1]: Using m5d.xlarge, one upstairs, three downstairs, running 5000 iterations
     of hammer and measuring the time in seconds required.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -119,6 +119,9 @@ name = "bytes"
 version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b700ce4376041dcd0a327fd0097c41095743c4c8af8887265942faf1100bd040"
+dependencies = [
+ "serde",
+]
 
 [[package]]
 name = "cfg-if"
@@ -299,8 +302,10 @@ name = "crucible-protocol"
 version = "0.0.0"
 dependencies = [
  "anyhow",
+ "bincode",
  "bytes",
  "crucible-common",
+ "serde",
  "tokio-util",
 ]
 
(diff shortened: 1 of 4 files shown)


leftwo (commented):
This all looks good, but the one question I think is important is, do we get a zero copy out of this?  Meaning we don't have to copy the data buffers that we are sending over the wire when this is serialized?  Apologies if my ignorance on how this works means my question makes no sense.

I know there is still some work to do on the downstairs side as it does some buffer pre-fills and other such nonsense, but we will get to that later and it's not any part of this review.

jmpesp:
> do we get a zero copy out of this?

No, but we do from a52e9ca :)

leftwo (approved):
The other thing to consider here, and we don't have to handle it now, is future versions of the message format.  At some point we may add commands to the protocol, and we want the new version of the software to understand old commands.  

However, that's not part of this review, so that comment is probably not relevant here.

jmpesp:
I have some ideas for versioning but won't gate this PR on that.

PR #59, opened 2021-09-14
Implement XTS encryption

jmpesp:
Adds UpstairsEncryptionContext and uses it when encrypting and
decrypting buffers before sending them downstairs. The benefit here is
that AES operations only happen once, and each downstairs mirror
contains the same data.

Accept a key in CrucibleOpts.


leftwo (commented):
I think this is a great start.
And, we should record all the perf work you were posting in chat somewhere, which leads me to a big picture question.   We might even need an RFD for this, or if we don't then put it into 177 under encryption. 
The question is: deciding on this type of encryption, or do we want to support Authenticated Encryption, which would mean the size of IO between upstairs and downstairs will be larger than a block size.  I think we should put some work into determining that before we proceed here.

jmpesp:
> We might even need an RFD for this, or if we don't then put it into 177 under encryption. The question is: deciding on this type of encryption, or do we want to support Authenticated Encryption, which would mean the size of IO between upstairs and downstairs will be larger than a block size. I think we should put some work into determining that before we proceed here.

Sounds good, let's chat at some point today :)

upstairs/src/lib.rs
```diff
@@ -877,6 +983,21 @@ impl Upstairs {
          */
         assert_eq!(opt.target.len(), 3);
 
+        // create an encryption context if a key is supplied.
+        let encryption_context = opt.key_bytes().map(|key| {
+            UpstairsEncryptionContext::new(
+                key,
+                /*
+                 * XXX: It would be good to do BlockOp::QueryBlockSize here, but
```
leftwo, line 991: I guess, yeah, this has to be an XXX now.  The block size is not available to up_main until the downstairs have all come online and are receiving IO.  We will have to see if the block size can be passed from Nexus to propolis to crucible.  Otherwise we have to wait to setup encryption till we have received the info from downstairs (Which may mean some other state transition stuff) and have the upstairs only allow query commands until the encryption is setup.

upstairs/src/lib.rs
```diff
@@ -817,6 +842,78 @@ impl Work {
     }
 }
 
+/// Implement XTS encryption
+/// See: https://en.wikipedia.org/wiki/Disk_encryption_theory#XEX-based_tweaked-codebook_mode_with_ciphertext_stealing_(XTS)
+pub struct UpstairsEncryptionContext {
```
leftwo, line 847: Nit: could we call this `EncryptionContext`?  There is no downstairs encryption context, and it's less to type :)

upstairs/src/main.rs
```diff
@@ -70,13 +74,47 @@ mod tests {
...
+        )
+        .unwrap();
+
+        if let Some(key) = &opt.key {
+            assert_eq!(
+                base64::decode(key).expect("base64 decode failed"),
+                key_bytes
+            );
+        } else {
+            panic!("bad news");
```
leftwo, line 95: maybe panic with `failed to encode key option` or some more info on why the news is bad :)

upstairs/src/main.rs
```diff
@@ -70,13 +74,47 @@ mod tests {
...
+        let crucible_opts = crucible::CrucibleOpts {
+            target: opt.target,
+            lossy: false,
+            key: opt.key,
+        };
+
+        if let Some(key) = crucible_opts.key_bytes() {
+            assert_eq!(key_bytes, key);
+        } else {
+            panic!("bad news");
```
leftwo, line 107: more bad news?

leftwo (approved):
This all looks good.

You did test with hammer or something that will produce IOs large enough to span multiple extents, correct?  So we know the downstairs buffer sector index stuff is working correctly?  It looks okay, but it's always nice to have the tests agree :)

One other thing to consider, would we ever want a downstairs to be able to decode/encode the data directly (if it had the key) so the import/export stuff could take advantage of it?  I'm not sure if we do, but just thinking about it.

upstairs/src/lib.rs
jmpesp: Changed

upstairs/src/main.rs
jmpesp: :+1:

upstairs/src/main.rs
jmpesp: :+1:

PR #61, opened 2021-09-16
Ack early for upstairs.

leftwo:
Crucible lib:
Enabled the upstairs to send completion for an IO when one read or 2/3
writes/flushes have completed from the downstairs.  This uncovered some flow
control challenges to work on next.

Added to the test show_work operation to return the count of active work in the
upstairs, which allows a test to keep waiting until it finds that all work is
completed.

client test:
Renamed test "one" to "span".

Made new "one" test that just does a single random w/f/r

Changed client test to keep itself alive until all downstairs have completed
their respective IOs, making use of the new test show_work return status.
Updated rand test function to take a count of operations to perform.

Downstairs:
Added "lossy" option to downstairs that simulates skipping work, and makes
some options take much longer than they would otherwise.  This allows testing
of the upstairs ACK to the guest when not all downstairs have completed.
This also uncovered some flow control issues in the upstairs, which will need
to be addressed.

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -20,6 +20,7 @@ arg_enum! {
     #[derive(Debug, StructOpt)]
     enum Workload {
         One,
+        Span,
         Big,
         Dep,
         Rand,
@@ -48,9 +49,7 @@ pub struct Opt {
     lossy: bool,
 
     /*
-     * quit after tests finish.
-     * By default, the client remains running and sends a flush
-     * request every so often to the downstairs.
+     * quit after all crucible work queues are empty.
      */
     #[structopt(short, long)]
     quit: bool,
@@ -122,7 +121,7 @@ fn main() -> Result<()> {
     runtime.spawn(up_main(crucible_opts, guest.clone()));
     println!("Crucible runtime is spawned");
 
-    std::thread::sleep(std::time::Duration::from_secs(5));
+    std::thread::sleep(std::time::Duration::from_secs(2));
 
     println!("Wait for a show_work command to finish before sending IO");
     guest.show_work();
@@ -139,7 +138,11 @@ fn main() -> Result<()> {
     match opt.workload {
         Workload::One => {
             println!("One test");
-            single_workload(&guest)?;
+            runtime.block_on(rand_workload(&guest, 1))?;
+        }
+        Workload::Span => {
+            println!("Span test");
+            span_workload(&guest)?;
         }
         Workload::Big => {
             println!("Run big test");
@@ -163,7 +166,7 @@ fn main() -> Result<()> {
         }
         Workload::Rand => {
             println!("Run random test");
-            runtime.block_on(rand_workload(&guest))?;
+            runtime.block_on(rand_workload(&guest, 50))?;
         }
         Workload::Balloon => {
             println!("Run balloon test");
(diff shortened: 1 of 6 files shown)


client/src/main.rs
```diff
@@ -457,12 +459,17 @@ async fn rand_workload(guest: &Arc<Guest>) -> Result<()> {
...
+            // When count is small, print out the IO info
+            println!("IO at block {}, len:{}", offset.value, data.len());
+        }
         let mut waiter = guest.write(offset, data);
         waiter.block_wait();
 
         let mut waiter = guest.flush();
         waiter.block_wait();
-
+        /*
```
jmpesp, line 472: no more verification?

upstairs/src/lib.rs
```diff
@@ -266,19 +266,23 @@ async fn io_completed(
...
+    #[allow(unused_assignments)]
     let mut gw_work_done = false;
-
     /*
+     * Mark this ds_id for the client_id as completed.
      * We can't call .send with the lock held, so we check to see
-     * if we do need to notify the up_ds_listen task that all work
-     * is finished for a ds_id.
+     * if we do need to notify the up_ds_listen task that enough work
+     * is finished for a ds_id.  We also need to get sta
```
jmpesp, line 281: EOF? :)

upstairs/src/lib.rs
```diff
@@ -266,19 +266,23 @@ async fn io_completed(
     data: Option<Bytes>,
     ds_done_tx: mpsc::Sender<u64>,
 ) -> Result<()> {
+    /*
+     * Having to put scope around the ds_work.lock call to avoid the
+     * compiler complaining (error[E0277]) about holding the lock while
+     * calling send (even if we drop) then creates a warning for the
+     * gw_work_done variable.
+     */
+    #[allow(unused_assignments)]
```
jmpesp, line 275: to avoid that warning, this could be:
```
let mut gw_work_done = {
  let mut work = up.ds_work.lock().unwrap();
  work.complete(ds_id, client_id, data)?
};
```

upstairs/src/lib.rs
```diff
@@ -394,7 +408,7 @@ async fn proc(
     lossy: bool,
 ) -> Result<()> {
     let (r, w) = sock.split();
-    let mut fr = FramedRead::new(r, CrucibleDecoder::new());
+    let mut fr = FramedRead::with_capacity(r, CrucibleDecoder::new(), 9999);
```
jmpesp, line 411: is that unit bytes? what was the default?

upstairs/src/lib.rs
```diff
@@ -757,7 +781,9 @@ impl Work {
         for state in job.state.values() {
             match state {
                 IOState::New | IOState::InProgress => wc.active += 1,
-                IOState::Done | IOState::Skipped | IOState::Error => {
+                IOState::AckReady => wc.ack_ready += 1,
+                IOState::Error => wc.error += 1,
+                IOState::Acked | IOState::Done | IOState::Skipped => {
```
jmpesp, line 786: `acked` is a field in WorkCounts, why not add to that?

jmpesp (approved):
:+1:

upstairs/src/lib.rs
leftwo: Yeah, I struggle with this method.
Its intent was to provide an easy summary of IO states that allow a caller to act on.
The reason I don't have a count for just acked is that the place that is looking for acked also wants to count the "Done" and "Skipped" IO states.

leftwo (commented):
Added response comments, changes coming next.

client/src/main.rs
leftwo: Fixed in 960fc31

upstairs/src/lib.rs
leftwo: Fixed in 960fc31

PR #64, opened 2021-09-22
UUIDs for downstairs, and new arg parsing

leftwo:
Updated the way the downstairs is started.  You now provide a subcommand
and then options for that subcommand, which results in less global
options that are silently ignored when not used.  This makes both the
starting syntax (and the code I stole, ah, was inspired by) similar to
Omicron sled-agent.

Added a UUID to the downstairs region struct.  This is recorded in the
region.json file for each downstairs.

Added two new messages to the upstairs/downstairs initial communication
that requests the RegionDefinition information, including the new
UUID from each downstairs.  Each downstairs UUID is now recorded in the
Upstairs struct.  More coming here later along with processing the rest
of the region information.

Updated all the READMEs I could find that show how to run the
downstairs.  Updated any test scripts or tools I could find as well.

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -25,6 +25,7 @@ use tokio::time::{sleep_until, Instant};
 use tokio_util::codec::{FramedRead, FramedWrite};
 use tracing::{instrument, span, Level};
 use usdt::register_probes;
+use uuid::Uuid;
 
 use aes::cipher::generic_array::GenericArray;
 use aes::{Aes128, NewBlockCipher};
@@ -281,7 +282,6 @@ fn process_downstairs(
  * This function is called when the upstairs task is notified that
  * a downstairs operation has completed.  We add the read buffer to the
  * IOop struct for later processing if required.
- *
  */
 #[instrument]
 async fn io_completed(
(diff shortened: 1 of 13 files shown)


upstairs/src/lib.rs
```diff
@@ -1404,6 +1418,46 @@ impl Upstairs {
...
+            if *uuid != rdef.uuid() {
+                panic!(
+                    "New client:{} uuid:{}  does not match existing {}",
+                    client_id,
+                    rdef.uuid(),
+                    uuid
+                );
+            } else {
+                println!(
+                    "Returning clientL{} UUID:{} matches",
```
jmpesp, line 1445: `client:`

upstairs/src/lib.rs
```diff
@@ -1010,6 +1018,11 @@ pub struct Upstairs {
      */
     ds_state: Mutex<Vec<DsState>>,
 
+    /*
+     * UUID for each downstairs, index by client ID
```
jmpesp, line 1022: the RFD mentions:

> Nexus sends to each Crucible Agent the size of the segment along with UUIDs describing both the overall storage volume and a UUID for this specific downstairs.

but I don't see a distinction in this PR between the volume UUID and downstairs UUID. We might even want different types for that so there won't be confusion in the code.

jmpesp (changes requested):

upstairs/src/lib.rs
leftwo: For sure there will be a pool wide UUID.  There will possibly be certificates or keys or other bits of info that the upstairs will have.  From the  downstairs point of view, I'm not sure if it needs to know the UUID of the overall pool, but if it does we can add it (along with the rest) when we decide what it should have.

I'm still considering making some struct that holds a bunch of information about each downstairs, but as I don't know know exactly what we need yet, I'm holding off on that.

For this change, I'm just handling the upstairs/downstairs UUID passing.  There is a bunch more work here, but I wanted to prevent this review from turning into a monster, and this seemed like a good place to pause.  I'm already a few days into the next part of this, so, don't worry, more is coming :)

upstairs/src/lib.rs
leftwo: Ha,  I just saw that typo last night, and almost pushed an update.  Will fix.

upstairs/src/lib.rs
jmpesp: :+1:

jmpesp (approved):

Issue #63, opened 2021-09-20
Investigate a database for downstairs metadata


leftwo:
Currently we just use the first block of all extents for metadata.  We should investigate using a small local database for storage of metadata.

Some possible metadata could be:
* Dirty bits
* block level checksums
* Recovery data or progress

The actual database would be another file in the filesystem that holds the extent.
The downstairs would make use of a database library crate to interact with the file.

andrewjstone:
Here are a few possible choices for embedded databases. We should probably do an RFD for this with some investigation similar to https://rfd.shared.oxide.computer/rfd/0053. We can assume we are going to use an existing library rather than something we build ourselves. The list below is limited to actively maintained databases in systems languages.

[SQLite](https://www.sqlite.org/index.html) - Small single file SQL database. Minimal configuration. Rock Solid. Written in C.
[RocksDB](http://rocksdb.org/) - LSM tree based Key-value store. Very configurable. Written in C++.
[LMDB](http://www.lmdb.tech/doc/) - B+ tree mmap based key-value store. Minimal configuration. Written in C.
[Sled](http://sled.rs/) -Bw tree based key-value store. Minimal configuration. Written in Rust.

Ask me about my biases!

jclulow:
SQLite is certainly my bias haha.  What's yours?

As an aside, for SQLite I think we'd want to look seriously at the [write-ahead logging](https://sqlite.org/wal.html) mode for these files, and look at the checkpointing settings, etc.  There is some question about whether we would want a database file per `Extent`, which would result in many files; or per `Region`, with the extent ID being a column in some or many of the tables, which may end up as a write bottle-neck depending on how we use the database.

andrewjstone:
My bias is also sqlite :)

The only other thing I think looking into, perhaps for building things with
more perf emphasis long term is sled, which apparently is going 1.0 soon.

Fuck RocksDB. I haven't used LMDB, but I don't really see a point. For now,
and maybe forever, SQLite makes sense.

PR #66, opened 2021-09-24
Use sqlite db instead of metadata block

jmpesp:
Remove the serialized metadata from the front of the extent, and use a
separate sqlite database instead. Enable WAL so that writes are
journaled and crash-safe.

Lots of XXX here, and lots more to do.

I'm running benchmarks now to see if there is a performance degradation.

Addresses #63.


jclulow:
I see you have a table for the metadata which is going to contain a single row, where the properties we want to set for the extent are columns on that table.

Would it make more sense to have a slightly more generic table, like:

```sql
CREATE TABLE IF NOT EXISTS metadata (
    name TEXT PRIMARY KEY,
    value INTEGER NOT NULL,
)
```

Then we would insert a `dirty`, a `flush_number`, a `version`, etc, as rows.  A flush would be a transaction with two statements:

```sql
UPDATE metadata SET value = ? WHERE name = 'metadata';
UPDATE metadata SET value = 0 WHERE name = 'dirty';
```

jmpesp:
A few thousand hammer runs show that there's no statistically significant performance degradation before or after this patch.

I'll try the more generic metadata table.

jclulow:
No big difference is a good sign!

jmpesp:
I had to rebase and fix up some stuff, so this PR has a few unrelated commits now.

downstairs/src/region.rs
```diff
@@ -24,13 +23,66 @@ pub struct Extent {
...
+        for flush_number_value in flush_number_iter {
+            flush_number_values.push(flush_number_value?);
+        }
+
+        assert!(flush_number_values.len() == 1);
+
+        Ok(flush_number_values[0])
+    }
+
+    fn set_new_flush_number(&self, new_flush: u64) -> Result<()> {
```
leftwo, line 46: Nit: can we call this "set_flush_number".  Plus it is shorter and I like shorter.

downstairs/src/region.rs
```diff
@@ -24,13 +23,66 @@ pub struct Extent {
...
+        /*
+         * When we write out the new flush number, the dirty bit in the file
```
leftwo, line 48: Nit: Is this "the dirty bit in the record" now?  We don't have a file for each extent any longer.
I'm not sure what the proper database term is here, if it's not record.

Also, do we have TRUE and FALSE in database language, or do we just have to know that 0 is false
and 1 is true?

downstairs/src/region.rs
```diff
@@ -24,13 +23,66 @@ pub struct Extent {
...
+        let dirty_iter = stmt.query_map([], |row| row.get(0))?;
+
+        let mut dirty_values: Vec<bool> = vec![];
+        for dirty_value in dirty_iter {
+            dirty_values.push(dirty_value?);
+        }
+
+        assert!(dirty_values.len() == 1);
+
+        Ok(dirty_values[0])
```
leftwo, line 75: Question: Is the way the database query works such that it has to be iterated over, even though we will only ever have one item (hence the assert)?

downstairs/src/region.rs
```diff
@@ -24,13 +23,66 @@ pub struct Extent {
...
+
+    fn set_dirty(&self) -> Result<()> {
+        let _rows_affected = self
+            .metadb
+            .execute("UPDATE metadata SET value=1 WHERE name='dirty'", [])?;
+        Ok(())
+    }
 }
 
-/*
```
leftwo, line 30: Is this comment no longer true?  Or just too obvious to leave here?  Just wondering why it was removed.

downstairs/src/region.rs
```diff
@@ -197,12 +232,36 @@ impl Extent {
...
-         * Write out the new inner.meta structure for this new extent.
+         * Create the metadata db
          */
+        path.set_extension("db");
+        let metadb = Connection::open(&path)?;
+        assert!(metadb.is_autocommit());
+        metadb.pragma_update(None, "journal_mode", &"WAL")?;
+
+        /*
+         * Create tables and insert base data
```
leftwo, line 243: Is there any automatic way to go between the internal ExtentMeta type and the database table, or is that path filled with peril?  If we ever change one, we have to change the other.  Maybe that is something for down the road (if at even possible).

leftwo (approved):
Seems like a good first pass on using the database instead of the file method.

Do we have any way to test this yet?  I know that the upstairs will look at the flush
numbers and report if they are different, but I don't think we look at the dirty bit anywhere yet.

downstairs/src/region.rs
jclulow: > Also, do we have TRUE and FALSE in database language, or do we just have to know that 0 is false
and 1 is true?

On the subject of types in SQLite, it is indeed pretty loosey-goosey.  There is a [Datatypes in SQLite Version 3](https://www.sqlite.org/datatype3.html) document on their site that describes the situation in some detail.

In short, for booleans, `1` and `0` are indeed what it has available, and you're expected to convert them (however you do that) into the representation you need.

downstairs/src/region.rs
jmpesp: Shortened in 24da54d

downstairs/src/region.rs
jmpesp: Fixed comment in 24da54d, and yeah - SQLite is pretty "flexible" :)

downstairs/src/region.rs
jmpesp: Unfortunately (as I understand it) anything that maps to SELECT always conceptually returns an array.

downstairs/src/region.rs
jmpesp: I think it's no longer true, we're not reading serialized bytes anymore and don't need to version the struct or something like that. If ExtentMeta changes now we'll need something like a schema upgrade. This is especially true if we change the type on the RHS, we would need a new table for that.

downstairs/src/region.rs
jmpesp: We could use an ORM like [Diesel](https://github.com/diesel-rs/diesel) (which supports sqlite) but it may be too heavy for our uses.

leftwo: Yeah, let's not get heavy until we know we need to.

leftwo (approved):
I'm good to go here, and I clicked on the wrong thing and re-requested @jclulow to review it, so ignore that as I try and figure out how to un-request my re-review.

downstairs/src/region.rs
jclulow: You might want to look at `Statement#query_row()` if the query is structured such that it intends to return just one row: https://docs.rs/rusqlite/0.25.3/rusqlite/struct.Statement.html#method.query_row

PR #71, opened 2021-09-27
Transition to Failed immediately on an error

jmpesp:
Previous to this commit, there would be time between when Work records
errors for a downstairs client, and when that client would be
transitioned to DsState::Failed. This commit adds a Weak pointer in Work
to Upstairs so that transition can occur "immediately".

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -7,7 +7,7 @@ use std::fmt::{Debug, Formatter};
 use std::io::{Read, Result as IOResult, Seek, SeekFrom, Write};
 use std::net::SocketAddrV4;
 use std::sync::mpsc as std_mpsc;
-use std::sync::{Arc, Mutex, MutexGuard, RwLock, Weak};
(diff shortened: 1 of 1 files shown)


leftwo (commented):
What if we look at error state in io_completed() and move the downstairs directly
to DsState::Failed there?  Then we don't have to have Work have a Weak field, and
other changes.

I do like the defaults for things though, those changes are good.

upstairs/src/lib.rs
```diff
@@ -1740,6 +1755,51 @@ impl Upstairs {
 
         Ok(())
     }
+
+    /*
+     * Complete a downstairs operation.
+     *
+     * Returns true if the guest should be notified.
+     */
+    fn complete(
```
leftwo, line 1764: It's my fault.
We have too many things named complete.  I'm going to try and do a better job of naming things without so much overlap.  Nothing for this review, I'm really just leaving this comment as a shame reminder to myself to go fix the problem I have created.

leftwo (approved):
Yeah, yeah, good to go

PR #86, opened 2021-10-14
Break apart upstairs lib into modules

jmpesp:


jmpesp:
:+1:

jmpesp:
I rebased and pushed only breaking out into test and pseudo_file

PR #88, opened 2021-10-19
Updates to dumping out metadata and comparing data for regions.

leftwo:
Moved the dump command to a new file.

Updated the dump command to print out genration number, flush number
and if the dirty bit set for each extent.

The dump command can now take up to three region directories and will
print out the information for all three.

If the dump command is passed a specific extent to look at, it will
dump the metadata, but also compare up to three regions block by block
and print a summary for each block in an extent which region matches
which other regions.

Updated the downstairs to store the generation number and the version
in the database.  The upstairs does not send them yet, but when it
does we will be ready.

Added some tests for the dump command.

Here is some example output.

First, we just dump out three region directories.  The tool shows us the
generation number, the flush ID, and a "D" if the dirty bit is set.
Each row is an extent number.
```
$ cargo run -q -p crucible-downstairs -- dump -d var/3801 -d var/3802 -d var/3803 
N      GEN FLUSH_ID D      GEN FLUSH_ID D      GEN FLUSH_ID D
0        0    21824          0    21824          0    21824   
1        0    21824          0    21824          0    21824   
2        0    21824          0    21824          0    21824   
3        0    21824          0    21824          0    21824   
4        0    21824          0    21824          0    21824   

```
In this next example, we are looking at a specific extent.  The three regions are now
represented as columns, and each row is a block in the given extent.

```
$ cargo run -q -p crucible-downstairs -- dump -d var/3801 -d var/3802 -d var/3803 -e 2
           Extent 2
FLUSH_ID    21824    21824    21824 
DIRTY                               
Block   0       A        A        A
Block   1       A        A        A
Block   2       A        A        A
Block   3       A        A        A
Block   4       A        A        A
Block   5       A        A        A
Block   6       A        A        A
Block   7       A        A        A
Block   8       A        A        A
Block   9       A        A        A
```
When comparing data, each column is a region.
The first column is always `A`.  If the 2nd column matches the first, then
it will be `A` as well.  If it is different, it will be `B`.   The third column will
display what it matches, or `C` if it is unique.

A block is only compared with the same block in other regions, not
with any block before or after it.

### DIFF
--- a/downstairs/src/dump.rs
+++ b/downstairs/src/dump.rs
@@ -1,0 +1,267 @@
+// Copyright 2021 Oxide Computer Company
+use super::*;
+use crate::region::ExtentMeta;
+
+#[derive(Debug, Default)]
+struct ExtInfo {
+    ei_hm: HashMap<u32, ExtentMeta>,
+}
+
+/*
+ * Dump the metadata for one or more region directories.
+ *
+ * If a specific extent is requested, only dump info on that extent.
+ */
+pub fn dump_region(
+    region_dir: Vec<PathBuf>,
+    cmp_extent: Option<u32>,
+) -> Result<()> {
+    /*
+     * We are building a two level hashmap.
+     * The first level index is the extent number.
+     * The second level index is the region_dir index in the region_dir
+     * Vec passed to us.
+     * We build this first because it makes it easier to print it all out
+     * at the end.
+     */
+    let mut all_extents: HashMap<u32, ExtInfo> = HashMap::new();
+    let dir_count = region_dir.len();
+    let mut blocks_per_extent = 0;
+    let mut total_extents = 0;
+    let mut block_size = 0;
+
+    for (index, dir) in region_dir.iter().enumerate() {
+        let region = Region::open(&dir, Default::default(), false)?;
+
+        block_size = region.def().block_size() as usize;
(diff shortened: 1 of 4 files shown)


downstairs/src/dump.rs
```diff
@@ -0,0 +1,267 @@
...
+            println!("No data for {}", en);
+        }
+        println!();
+    }
+
+    Ok(())
+}
+
+/*
+ * Show the metata and a block by block diff of a single extent
```
jmpesp, line 145: metadata

downstairs/src/dump.rs
```diff
@@ -0,0 +1,267 @@
...
+            print!("- ");
+        }
+    }
+    println!();
+
+    print!("FLUSH_ID ");
+    for dir_index in 0..dir_count {
+        if let Some(em) = ei_hm.get(&(dir_index as u32)) {
+            print!("{:8} ", em.flush_number);
+            // Read buffer here. Make a buffer for each extent, then dump?
```
jmpesp, line 178: is this a TODO?

downstairs/src/dump.rs
leftwo: Fixed in 4dd9a19ac0e101bdd9a8c14f01b7deaef775d55f

PR #91, opened 2021-10-20
improve image import

jclulow:
This change allows the creation of a region from an image file that is
smaller than the target region size.  If the file is not an exact
multiple of the target block size, the final short block will be
extended with zeroes.

The change also improves the performance of image import by reading a
32MB chunk of the import file at a time, and writing as much of that as
we are allowed into the target extents.

Finally, a small improvement to type safety: using the Block object to
represent the block count we pass to "extent_from_offset()" and adding
some routines for converting to and from byte sizes.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -10,7 +10,9 @@ use serde::{Deserialize, Serialize};
 use tempfile::NamedTempFile;
 
 mod region;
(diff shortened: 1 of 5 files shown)


jclulow:
NB: This is a reworking of the prototype image import improvements I made in the `demo_m2` branch.

leftwo (approved):
Yeah, looks good.

PR #96, opened 2021-11-01
Some fixes

jmpesp:
Push up some fixes that were made trying to debug a stall(s?).

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -485,6 +485,10 @@ async fn proc(
      */
     let mut negotiated = 0;
 
+    // XXX figure out what deadlines make sense here
+    let mut ping_interval = deadline_secs(5);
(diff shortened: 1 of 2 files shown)


upstairs/src/lib.rs
```diff
@@ -3815,9 +3849,14 @@ async fn up_listen(
          * a upstairs generated flush.
          */
         let mut flush_check = deadline_secs(5);
+        let mut show_work_interval = deadline_secs(5);
 
         loop {
             tokio::select! {
+                _ = sleep_until(show_work_interval) => {
```
leftwo, line 3856: We probably don't want this always on, but go ahead and leave this for now and perhaps when the dtrace stuff comes in we might turn this off or make it optional.

leftwo (approved):

PR #99, opened 2021-11-03
This PR is not a PR.  WIP for the ack task

leftwo:
The two commits here are where I got to with this.

The first one, I made the framed write an onwed frame write and then passed that
everywhere the old one was.

I also broke proc into to functions, the first that did the negotiation, and the 2nd that
does the main work loop.  With that commit, everything still worked.

The 2nd commit is more progress breaking out the ack into a separate task with the `fw` it will need, but it's not compete, and if you run it now, it will hang as it is not actually sending
an ack back to the upstairs.  

To continue down this path, I'm considering making a "work to ack" hashmap, and using a new struct to create the work that goes on it.  When an IO is finished, the work would
come off active, then go onto a new queue to ack.  Having a different queue would also
enable us to add ping responses to the work the ack task should do, and possibly the UUID
mismatch message as well.

### DIFF
--- a/downstairs/src/main.rs
+++ b/downstairs/src/main.rs
@@ -18,8 +18,9 @@ use bytes::BytesMut;
 use futures::{SinkExt, StreamExt};
 use rand::prelude::*;
 use structopt::StructOpt;
-use tokio::net::tcp::WriteHalf;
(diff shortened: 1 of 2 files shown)


leftwo:
Discarding this now that we have the merged the stall fixes

PR #100, opened 2021-11-03
Downstairs negotiation breakout

leftwo:
This is the negotiation work form the PR that is not a PR.  I left all the ack task
stuff out of this, to hopefully improve the merge mess.

Made the downstairs negotiation phase happen first in proc(),
then move to a new function that handles IO going forward.

Updated the socket split to be Owned, which will be used in
future updates to split ACKs to a different task.

### DIFF
--- a/downstairs/src/main.rs
+++ b/downstairs/src/main.rs
@@ -659,8 +659,13 @@ async fn proc(ads: &mut Arc<Mutex<Downstairs>>, sock: TcpStream) -> Result<()> {
     /*
      * See the comment in the proc() function on the upstairs side that
      * describes how this negotiation takes place.
+     *
+     * The final step in negotiation (as dictated by the upstairs) is
+     * either LastFlush, or ExtentVersionsPlease.  Once we respond to
+     * that message, we can move forward and start receiving IO from
+     * the upstairs.
      */
-    while negotiated < 5 {
+    while negotiated < 4 {
         tokio::select! {
             /*
              * Don't wait more than 50 seconds to hear from the other side.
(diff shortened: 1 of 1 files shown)


downstairs/src/main.rs
```diff
@@ -790,15 +761,28 @@ async fn proc(
...
+                    }
                     Some(Message::LastFlush(last_flush)) => {
-                        // TODO: Make a proper negotiation connect flow.
-                        if !negotiated {
-                            bail!("expected HereIAm first");
+                        if negotiated != 3 {
+                            bail!("Received LastFlush out of order {}",
+                                negotiated);
                         }
+                        negotiated = 5;
```
jmpesp, line 785: 4?

downstairs/src/main.rs
```diff
@@ -809,14 +793,122 @@ async fn proc(
                         }
                         fw.send(Message::LastFlushAck(last_flush)).await?;
                     }
-                    Some(msg) => {
-                        if !negotiated {
-                            bail!("expected HereIAm first");
+                    Some(Message::ExtentVersionsPlease) => {
+                        if negotiated != 3 {
```
jmpesp, line 797: 4?

jmpesp (approved):
Approved, with a question. I have a big merge ahead haha

downstairs/src/main.rs
leftwo: I'll add a comment to explain it, but it's because..
Well, wait. 
I think you are correct here, I could just go to 4 and since there are no more commands
coming after that, stop the loop on 4.

The final negotiation message should either be LastFlush or ExtentVersionsPlease, so, yeah.
I'll update that before I merge.

downstairs/src/main.rs
leftwo: The upstairs does a 1-2-3-5 or 1-2-4-5, and I confused myself in thinking I needed it here as well.  Fixed in latest.

PR #103, opened 2021-11-05
Limit test client program to IOs of 1M max.

leftwo:
Added a simple script to create three regions.

### DIFF
--- a/tools/create-generic-ds.sh
+++ b/tools/create-generic-ds.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+#
+# A hack of downstairs create tool
+
+ROOT=$(cd "$(dirname "$0")/.." && pwd)
+
+cd "$ROOT" || (echo failed to cd "$ROOT"; exit 1)
+
+if pgrep -fl crucible-downstairs; then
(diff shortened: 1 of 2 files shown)


tools/create-generic-ds.sh
```diff
@@ -0,0 +1,37 @@
...
+
+cds="./target/debug/crucible-downstairs"
+if [[ ! -f ${cds} ]]; then
+    echo "Can't find crucible binary at $cds"
+    exit 1
+fi
+
+res=0
+for port in 3801 3802 3803; do
+    if ! cargo run -q -p crucible-downstairs -- create -u 12345678-"$port"-"$port"-"$port"-00000000"$port" -d var/"$port" --extent-count 20 --extent-size 100; then
```
jmpesp, line 32: could use uuidgen program to make a random one

jmpesp (approved):
Approved with suggestion

tools/create-generic-ds.sh
leftwo: I like knowing from the uuid which downstairs I'm looking at, but yeah uuidgen is a good thing to do in some cases.

PR #105, opened 2021-11-08
Scope USDT probes in Upstairs

pfmooney:
The latest version of `usdt` allows probes to be module-scoped, rather
than calling them from a global namespace.


leftwo (approved):
Yeah, this all looks good Patrick.  Thanks for fixing it.

PR #107, opened 2021-11-09
More stall (+ panic) fixes

jmpesp:
Summary of downstairs changes:

- combine do_work_task and ack_sender task
- only unblocks jobs for the active UUID
- removed "Unblock any stuck jobs. XXX how does this happen?" branch
- unblock jobs if their deps are about to run (+ add verification that deps are fulfilled)
- unit tests!

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -310,8 +310,6 @@ fn main() -> Result<()> {
         }
 
         Workload::Generic => {
-            println!("Run Generic test in 5 seconds");
-            std::thread::sleep(std::time::Duration::from_secs(5));
             runtime.block_on(generic_workload(
                 &guest,
                 5000,
@@ -1113,7 +1111,7 @@ async fn dep_workload(guest: &Arc<Guest>, ri: &mut RegionInfo) -> Result<()> {
         /*
          * Generate some number of operations
          */
-        for ioc in 0..20 {
+        for ioc in 0..200 {
             my_offset = (my_offset + ri.block_size) % final_offset;
             if random() {
                 /*
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -1107,7 +1107,7 @@ async fn dep_workload(guest: &Arc<Guest>, ri: &mut RegionInfo) -> Result<()> {
     let final_offset = ri.total_size - ri.block_size;
 
     let mut my_offset: u64 = 0;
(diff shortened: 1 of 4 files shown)


leftwo:
I took the downstairs file from this change (and the one upstairs diff) and merged them into my
upstairs restart changes (also a WIP), and it works there as well.

jmpesp:
Note for posterity: 66af7cd642a2cf9589c0cba8ccc5eb672bc869e9 contains Alan's changes mentioned in https://github.com/oxidecomputer/crucible/pull/107#pullrequestreview-802044751

client/src/main.rs
```diff
@@ -1153,9 +1151,11 @@ async fn dep_workload(guest: &Arc<Guest>, ri: &mut RegionInfo) -> Result<()> {
         }
 
         guest.show_work()?;
-        println!("Loop:{} send a final flush and wait", my_count);
-        let mut flush_waiter = guest.flush()?;
-        flush_waiter.block_wait()?;
+        if random() && random() {
```
leftwo, line 1154: The final flush is to help prevent the pause that we get when the last command is a write or read and we have to wait x seconds for the flush check to trigger.  Maybe take the rand out of that?

upstairs/src/lib.rs
```diff
@@ -834,6 +834,40 @@ async fn cmd_loop(
...
 
+    let (tx, mut rx) = mpsc::channel(100);
+
+    {
+        let up_c = up.clone();
+        let up_coms_c = up_coms.clone();
+
+        tokio::spawn(async move {
+            loop {
+                tokio::select! {
```
leftwo, line 845: Maybe make this a while on the rx.recv() like we did for the downstairs.
We don't need the select or the loop then, so it's two less indents.  Just think of all
the whitespace this change could save.

leftwo (approved):
Yeah, let's get this in and we can iterate where needed.

I would not have thought that breaking out a task for pulling things off the wire
would help with the stall, but clearly your testing shows otherwise.

client/src/main.rs
jmpesp: done in 830187a

upstairs/src/lib.rs
jmpesp: done in c6fe5c96b929c1288bd2f8c7c87d0b64c8d893fb

PR #112, opened 2021-11-14
Work on reconnecting downstairs and handling upstairs migration

leftwo:
Better support for downstairs that disconnect and reconnect, as well as paving more path toward support of resolving downstairs mismatch on initial startup.

We now wait for all three downstairs to get to a specific state before moving forward with IO. 
A downstairs disconnect/reconnect is treated differently if it happens before this state and after.

Improvements are made on the negotiation phase to cover both what happens if the downstairs goes
away, and how the downstairs behaves if the upstairs goes away.

Some better than nothing tests were added.
I ran the hammer test 1000 times in a loop while restarting the downstairs.
I ran the reconnect test 1000 times in a loop and saw no errors.
The tools/test_up.sh test runs without errors.

Added some more logic to cover the different tasks discovering that they are no longer needed and having them exit without a panic (when the situation is appropriate).


upstairs/src/lib.rs
```diff
@@ -1628,6 +1787,7 @@ impl Downstairs {
             for id in kvec.iter() {
                 assert!(*id <= ds_id);
                 let wc = self.state_count(*id).unwrap();
+                // ZZZ
```
jmpesp, line 1790: ?

upstairs/src/lib.rs
```diff
@@ -1072,28 +1218,28 @@ async fn looper(
             eprintln!("ERROR: {}: proc: {:?}", target, e);
         }
 
-        /*
-         * Do not attempt to reconnect if we were marked inactive.
-         */
-        if !up.is_active() {
-            drop(up_coms.ds_status_tx);
```
jmpesp, line 1079: When I was testing, I saw that we have to do these drops if we want the Upstairs to stop cleanly

upstairs/src/lib.rs
```diff
@@ -3512,26 +3843,24 @@ impl Guest {
...
         let mut waiter = self.send(BlockOp::GoActive { gen });
+        println!("The guest is requesting activation");
         waiter.block_wait()?;
 
-        // XXX Is this the right number of retries? The right delay between
-        // retries?
-        for _ in 0..10 {
+        // XXX This currently blocks forever, but perhaps it should
+        // give up after some time?
+        loop {
```
jmpesp, line 3851: Why did you change this? It should give up after some time :)

upstairs/src/lib.rs
```diff
@@ -3608,7 +3937,9 @@ impl Guest {
      */
     pub fn show_work(&self) -> Result<WQCounts, CrucibleError> {
         if !self.is_active() {
-            return Err(CrucibleError::UpstairsInactive);
+            println!("Request for work from inactive upstairs");
+            // XXX Test access is allowed for now, but not forever.
```
jmpesp, line 3941: When is this needed?

upstairs/src/lib.rs
```diff
@@ -4016,9 +4348,13 @@ async fn up_listen(
...
                                         returning gracefully"
                                     );
-                                    return;
                                 }
                             }
+                            // XXX Do we break here?
+                            // what do I do?  Exit?
+                            // Maybe all current downstairs need to go back
+                            // to NEW?
+                            panic!("Write more code");
```
jmpesp, line 4357: This is the graceful shutdown path (entered when we dropped `ds_status_tx` in the `Do not attempt to reconnect if we were marked inactive` code) which means we should have returned above?

jmpesp (changes requested):
Some comments, plus I don't see a way of graceful shutdown anymore. One of the commit messages is `Made upstairs keep "looper" task always running.  The only way out is if there is a panic.` so maybe this is a goal now?

It's a shame that https://github.com/oxidecomputer/crucible/pull/111/commits/8bddd2675b7d5df3d241187c08610a947648d1bc wasn't in already to disambiguate some of the questions here.

upstairs/src/lib.rs
```diff
@@ -95,6 +95,7 @@ async fn process_message(
                     "u.uuid {:?} != job uuid {:?} on WriteAck",
                     u.uuid, *uuid
                 );
+                u.set_inactive();
```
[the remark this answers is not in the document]
leftwo: I believe early on I saw this.  However, after a few rounds of merging and better handling on the downstairs side, I can't say for sure this is still a problem.  
I'll switch it to a panic and see if I can get it to happen again, and if not, then I can leave it as a warning message and not disable this upstairs

upstairs/src/lib.rs
leftwo: If a downstairs connects, then disconnects before we go active, then it would exit here and can never re-join (this one took me a while to figure out).   

We may need some other global state to tell an upstairs that it should gracefully shutdown though.

upstairs/src/lib.rs
leftwo: Yeah.. yeah.. It's like the timeout issue.  How long should the activate wait before it gives up?
What happens if it gives up right before the upstairs returns?  I don't have any answers for these questions.  

Also, the time it takes to activate will be influenced by how long repair takes in the event that the downstairs don't agree.  I can switch it back to 10 seconds, but I'm sure we will be back here 
fussing with this in the future.

upstairs/src/lib.rs
leftwo: When I'm testing two clients at the same time.  The non-active client can still show some status.  Before todays storage meeting I was thinking we would support failover/failback situations, and this was part of that.

This will probably all go away at some point, the guest will not have access to the command that calls this, so we won't need to check for is_active.

upstairs/src/lib.rs
leftwo: The drop of ds_status_tx and exiting of looper was causing a problem if a downstairs connected then disconnected before we went active.

To support the graceful shutdown I think there is a bit more work to do.  This is already a big set of changes and I'm wondering if it is enough to polish this and do the full support for graceful shutdown in the next PR.

downstairs/src/main.rs
```diff
@@ -769,29 +769,50 @@ async fn resp_loop(
...
+     * Upstairs connection going away at any time, as well as a forced
+     * migration where a new Upstairs connects and the old (current from
+     * this threads point of view) work is discarded.
+     * As migration or upstairs failure can happen at any time, this
+     * function must watch for tasks going away and handle that
+     * gracefully.  By exiting the loop here, we allow the calling
+     * function to take over and handle a reconnect or a new upstairs
+     * takeover.
+     */
+    let dw_task;
```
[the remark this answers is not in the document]
leftwo: Done in a663fa3

upstairs/src/lib.rs
```diff
@@ -609,10 +646,25 @@ async fn proc(
 
                 match f.transpose()? {
                     None => {
-                        // hung up
+                        println!(
+                            "[{}] client got null message",
```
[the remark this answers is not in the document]
leftwo: Done in a663fa3

upstairs/src/lib.rs
```diff
@@ -609,10 +646,25 @@ async fn proc(
...
+                            up_coms.client_id
+                        );
                         up.ds_missing(up_coms.client_id);
                         return Ok(())
                     }
+                    Some(Message::Imok) => {
+                        // This is a ping ack.
+                        /*
```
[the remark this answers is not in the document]
leftwo: Gone!  a663fa3

upstairs/src/lib.rs
leftwo: Note to self cruft left over.  Gone in a663fa3

upstairs/src/lib.rs
leftwo: Switched back in a663fa3

leftwo:
I made some of your requested changes with a663fa3.
I think we can discuss next steps with this after merging with #111

upstairs/src/lib.rs
jmpesp: > Also, the time it takes to activate will be influenced by how long repair takes in the event that the downstairs don't agree.

This is a great point I didn't consider at all :( Something will have to be done, yeah.

upstairs/src/lib.rs
jmpesp: :+1:

upstairs/src/lib.rs
jmpesp: :+1:

upstairs/src/lib.rs
jmpesp: Something for a future PR though.

jmpesp (approved):

PR #120, opened 2021-11-17
Add nonces and tags to protocol

jmpesp:
Add nonce and tag to both reads and writes. A separate ReadResponse is
created to differentiate between a request and response.

ReadResponse now returns a Result<Vec<Response>> instead of returning
result separately in another field. This obviates some code and makes
one of the failures currently being tested unable to happen. Empty
vectors are used in io_complete for WriteAck and FlushAck.

Nonces and tags are stored in SQLite as blobs. SQLite only uses variable
length records [1] so no optimization can occur by fixing the byte size
here.

If performing authenticated encryption, nonce and tags must be sent per
block. This means `extent_from_offset` must return a list of single
blocks only, and has had an extra boolean added to its arguments as a
result.

Encryption context is now passed around as an Arc, obviating the clone
code that was there before.

Note that this commit *does not add authenticated encryption*, that will
come separately.

[1]: https://www.sqlite.org/different.html

### DIFF
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -460,7 +460,6 @@ impl Extent {
 
         inner.file.seek(SeekFrom::Start(byte_offset))?;
         inner.file.write_all(&write.data)?;
-        inner.file.flush()?;
 
         if write.nonce.is_some() && write.tag.is_some() {
             inner.set_encryption_context(
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -6,7 +6,6 @@ use std::path::{Path, PathBuf};
 use std::sync::{Mutex, MutexGuard};
 
 use anyhow::{bail, Result};
-use bytes::BytesMut;
 use crucible_common::*;
 use rusqlite::{params, Connection};
 use serde::{Deserialize, Serialize};
@@ -722,11 +721,11 @@ impl Region {
     pub fn single_block_region_read(
         &self,
         request: crucible_protocol::ReadRequest,
-    ) -> Result<BytesMut, CrucibleError> {
+    ) -> Result<crucible_protocol::ReadResponse, CrucibleError> {
         let mut responses = self.region_read(&[request])?;
         let response = responses.pop().unwrap();
         drop(responses);
-        Ok(response.data)
+        Ok(response)
     }
 
     #[instrument]
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -760,10 +760,9 @@ impl Region {
 
 #[cfg(test)]
 mod test {
-    use super::extent_path;
     use super::*;
     use crate::dump::dump_region;
-    use bytes::BufMut;
+    use bytes::{BufMut, BytesMut};
     use rand::Rng;
     use std::path::PathBuf;
     use tempfile::tempdir;
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -109,6 +109,57 @@ impl Inner {
             .execute("UPDATE metadata SET value=1 WHERE name='dirty'", [])?;
         Ok(())
     }
+
+    fn get_encryption_context(
+        &self,
+        block: u64,
+    ) -> Result<Option<(Vec<u8>, Vec<u8>)>> {
+        let mut stmt = self.metadb.prepare(
+            &[
+                "SELECT nonce, tag FROM encryption_context",
+                "where block=?1",
+            ]
+            .join(" "),
+        )?;
+
+        let stmt_iter = stmt.query_map(params![block], |row| {
+            Ok(Some((row.get(0)?, row.get(1)?)))
+        })?;
+
+        let mut results = Vec::new();
+
+        for row in stmt_iter {
+            results.push(row);
+        }
+
+        if results.is_empty() {
+            Ok(None)
+        } else {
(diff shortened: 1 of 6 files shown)


downstairs/src/region.rs
```diff
@@ -109,6 +109,57 @@ impl Inner {
             .execute("UPDATE metadata SET value=1 WHERE name='dirty'", [])?;
         Ok(())
     }
+
+    fn get_encryption_context(
```
leftwo, line 113: If/when we turn this on for real, we should add to the `dump` command something that will print this out.  I don't know exactly how that would look yet, but I know when building and debugging the reconciliation, we may want to look at and compare the nonce/tags between different downstairs.

downstairs/src/region.rs
```diff
@@ -378,21 +448,28 @@ impl Extent {
...
 
         inner.set_dirty()?;
 
-        let byte_offset = offset.value * self.block_size;
+        let byte_offset = write.offset.value * self.block_size;
 
         inner.file.seek(SeekFrom::Start(byte_offset))?;
-        inner.file.write_all(data)?;
+        inner.file.write_all(&write.data)?;
         inner.file.flush()?;
```
leftwo, line 463: I'm now really wondering what this is here for.  Also, should it come after the set_encryption_context below?

leftwo (approved):

upstairs/src/lib.rs
```diff
@@ -1672,10 +1688,10 @@ impl Downstairs {
                     dependencies: _dependencies,
                     requests: _,
                 } => {
-                    assert!(read_data.is_some());
+                    assert!(!read_data.is_empty());
```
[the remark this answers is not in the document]
jmpesp: :+1:

upstairs/src/lib.rs
```diff
@@ -168,6 +177,7 @@ pub fn extent_from_offset(
     ddef: RegionDefinition,
     offset: Block,
     num_blocks: Block,
+    single_blocks_only: bool,
```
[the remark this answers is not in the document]
jmpesp: Good idea, added in 9edf184

downstairs/src/region.rs
jmpesp: Yeah, I removed it in 7e263d4. Flushing that file is independent of flushing `set_encryption_context` (which I hope sqlite does unprompted every time!)

upstairs/src/lib.rs
```diff
@@ -318,11 +333,10 @@ async fn io_completed(
     up: &Arc<Upstairs>,
     ds_id: u64,
     client_id: u8,
-    responses: Option<Vec<(ReadRequest, bytes::Bytes)>>,
+    responses: Result<Vec<ReadResponse>, CrucibleError>,
     ds_done_tx: mpsc::Sender<u64>,
-    result: Result<(), CrucibleError>,
 ) -> Result<()> {
-    if up.complete(ds_id, client_id, responses, result)? {
+    if up.complete(ds_id, client_id, responses)? {
```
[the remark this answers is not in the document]
jmpesp: Removed in 39d71dc

downstairs/src/region.rs
jmpesp: I missed adding this and did so in 27b2365. `dump` now looks at nonces and tags, and also has a "DIFF" column to highlight anything it found. Here's an example:

```
$ ./dump.sh -d disks/d1/ -d disks/d2/ -d disks/d3/ -e 0 | head -n20
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
     Running `target/debug/crucible-downstairs dump -d disks/d1/ -d disks/d2/ -d disks/d3/ -e 0`
           Extent 0
FLUSH_ID       78       78       78 
DIRTY           D        D        D 

             Data 0      Data 1      Data 2      Nonce 0     Nonce 1     Nonce 2      Tag 0       Tag 1       Tag 2     DIFF  
Block    0     A           A           A           A           A           A           A           A           A             
Block    1     A           A           A           A           B           B           A           B           B      <-------
Block    2     A           A           A           A           A           A           A           A           A             
Block    3     A           A           A           A           A           A           A           A           A             
Block    4     A           A           A           A           A           A           A           A           A             
Block    5     A           A           A           A           A           A           A           A           A             
Block    6     A           A           A           A           A           A           A           A           A             
Block    7     A           A           A           A           A           A           A           A           A             
Block    8     A           A           A           A           A           A           A           A           A             
Block    9     A           A           A           A           A           A           A           A           A             
Block   10     A           A           A           A           A           A           A           A           A             
Block   11     A           A           A           A           A           A           A           A           A             
Block   12     A           A           A           A           A           A           A           A           A             
Block   13     A           A           A           A           A           A           A           A           A
```

upstairs/src/lib.rs
```diff
@@ -88,74 +88,40 @@ async fn process_message(
...
-            } else {
-                Err(result.as_ref().err().unwrap().clone())
-            };
-
-            if u.complete(*ds_id, up_coms.client_id, result)? {
-                up_coms.ds_done_tx.send(*ds_id).await?;
-            }
-
-            Ok(())
+            (*uuid, *ds_id, result.clone().map(|_| Vec::new()))
```
leftwo, line 94: Yeah, this is awesome.  Much less duplication of code here.

leftwo (approved):

PR #122, opened 2021-11-17
Fix broken import and broken import test

jmpesp:


leftwo (approved):
I wrote a test that I forgot about that tested this: ./tools/test_ds.sh
Too bad we never ran it, it would have caught this :)

jclulow:
I'm not sure this is correct?  The read into the buffer needs to start at `total` or it will overwrite things we read on the last turn?

leftwo:
> I'm not sure this is correct? The read into the buffer needs to start at `total` or it will overwrite things we read on the last turn?

Nice josh, you are correct:  this still fails:
```
HEAD is now at 8227c0a fix broken import
final:jmpcru$ ./tools/test_ds.sh 
/Users/alan/Oxide/ws/jmpcru
   Compiling crucible-protocol v0.0.0 (/Users/alan/Oxide/ws/jmpcru/protocol)
   Compiling crucible v0.0.1 (/Users/alan/Oxide/ws/jmpcru/upstairs)
   Compiling crucible-nbd-server v0.1.0 (/Users/alan/Oxide/ws/jmpcru/nbd_server)
   Compiling crucible-hammer v0.1.0 (/Users/alan/Oxide/ws/jmpcru/hammer)
   Compiling crucible-client v0.1.0 (/Users/alan/Oxide/ws/jmpcru/client)
   Compiling crucible-downstairs v0.0.1 (/Users/alan/Oxide/ws/jmpcru/downstairs)
    Finished dev [unoptimized + debuginfo] target(s) in 13.30s
Create file for import
300+0 records in
300+0 records out
153600 bytes transferred in 0.002113 secs (72689281 bytes/sec)
Import region
Created new region file "/tmp/ds_test/export/region.json"
Import file_size: 153600  Extent size: 51200  Needed extents: 3
Region already large enough for image
Importing "/tmp/ds_test/import" to region
Populated 3 extents by copying 153600 bytes (300 blocks)
UUID: 12345678-1234-1234-1234-000000000001
Blocks per extent:100 Total Extents: 15
Export region
Opened existing region file "/tmp/ds_test/export/region.json"
Export total_size: 768000  Extent size:51200  Total Extents:15
Exporting from start_block: 0  count:280576
Read and wrote out 1500 blocks
Binary files /tmp/ds_test/import and /tmp/ds_test/exported_file differ
final:jmpcru$ echo $?
1
```

jclulow:
FWIW, though I did not run the test case you mention at the time, I did at least test it by hand.

leftwo:
> FWIW, though I did not run the test case you mention at the time, I did at least test it by hand.

The test may not have existed at that time.  I wrote that, then promptly forgot about it, so I've not been running it either.

jmpesp:
> FWIW, though I did not run the test case you mention at the time, I did at least test it by hand.

Hey now, I resent that implication :) I tested like so:

```
$ rm -rf "/tmp/james/disks/alpine1/

$ $ cargo run -p crucible-downstairs -- create -d "/tmp/james/disks/alpine1/" --import-path ~/Downloads/alpine-extended-3.14.2-x86_64.iso --uuid "$(uuidgen)" --block-size 512 --extent-size $((10 * 1024 * 1024 / 512))
   Compiling crucible-downstairs v0.0.1 (/home/jwm/src/oxidecomputer/crucible/downstairs)
    Finished dev [unoptimized + debuginfo] target(s) in 2.19s
     Running `target/debug/crucible-downstairs create -d /tmp/james/disks/alpine1/ --import-path /home/jwm/Downloads/alpine-extended-3.14.2-x86_64.iso --uuid 7eea574e-505b-4f49-babd-3ea3a4ff5c40 --block-size 512 --extent-size 20480`
Created new region file "/tmp/james/disks/alpine1/region.json"
Import file_size: 627048448  Extent size: 10485760  Needed extents: 60
Extending region to fit image
Importing "/home/jwm/Downloads/alpine-extended-3.14.2-x86_64.iso" to region
Populated 60 extents by copying 627048448 bytes (1224704 blocks)
UUID: 7eea574e-505b-4f49-babd-3ea3a4ff5c40
Blocks per extent:20480 Total Extents: 60

$ cargo run -p crucible-downstairs -- export --count 1224704 -d /tmp/james/disks/alpine1/ -e test.iso
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
     Running `target/debug/crucible-downstairs export --count 1224704 -d /tmp/james/disks/alpine1/ -e test.iso`
Opened existing region file "/tmp/james/disks/alpine1/region.json"
Export total_size: 629145600  Extent size:10485760  Total Extents:60
Exporting from start_block: 0  count:1224704
Read and wrote out 1224704 blocks

$ md5sum ~/Downloads/alpine-extended-3.14.2-x86_64.iso test.iso
fcd3dc6a1b0c6c518f45f69b8b1f383c  /home/jwm/Downloads/alpine-extended-3.14.2-x86_64.iso
fcd3dc6a1b0c6c518f45f69b8b1f383c  test.iso
```

Plus I've been using that imported alpine ISO on broadway all afternoon as part of my propolis testing.

Looking at the `./tools/test_ds.sh` output you pasted Alan:

```
Populated 3 extents by copying 153600 bytes (300 blocks)
<ship>
Read and wrote out 1500 blocks
Binary files /tmp/ds_test/import and /tmp/ds_test/exported_file differ
```

Those binary files differ because import is 300 blocks and exported_file is zero padded to 1500 blocks. The count argument on the export is wrong - I fixed it in 7656d81 and now it passes.

jclulow:
> > FWIW, though I did not run the test case you mention at the time, I did at least test it by hand.
> 
> Hey now, I resent that implication :) I tested like so:

Definitely no implication intended on my part!  I was just referring to the testing I did for #91 in response to Alan pointing out the test script, which I don't believe I ran at the time.

jmpesp:
> I'm not sure this is correct? The read into the buffer needs to start at `total` or it will overwrite things we read on the last turn?

Yeah I messed this up. I assumed that Rust's read here would either return `Ok(buffer.len())` or `Err(...)`, hiding the caller from the underlying POSIX (or whatever) read not filling the buffer. From the [documentation](https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read):

> If the return value of this method is Ok(n), then implementations must guarantee that 0 <= n <= buf.len(). A nonzero n value indicates that the buffer buf has been filled in with n bytes of data from this source.

I put the first part of the slice back in https://github.com/oxidecomputer/crucible/pull/122/commits/c2cb3ceca49a1676b761c59f69d760365502da07.

Note if I run `./tools/test_ds.sh` on main (27b3497d8948456ea9a296b66451a944529beda8 right now for me), I see:

```
Create file for import
300+0 records in
300+0 records out
153600 bytes (154 kB, 150 KiB) copied, 0.000680805 s, 226 MB/s
Import region
Created new region file "/tmp/ds_test/export/region.json"
Import file_size: 153600  Extent size: 51200  Needed extents: 3
Region already large enough for image
Importing "/tmp/ds_test/import" to region
thread 'main' panicked at 'source slice length (51200) does not match destination slice length (0)', /rustc/ac2d9fc509e36d1b32513744adf58c34bcc4f43c/library/core/src/slice/mod.rs:3055:13
```

which prompted this PR.

It's too bad there isn't a unit test reader that will incompletely fill the buffer in order to test this...

jmpesp:
I wrote https://github.com/jmpesp/incomplete_file this morning and tested with it:

```
diff --git a/Cargo.lock b/Cargo.lock
index dd956fb..1cc7bd3 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -298,6 +298,7 @@ dependencies = [
  "crucible-protocol",
  "futures",
  "futures-core",
+ "incomplete_file",
  "opentelemetry",
  "opentelemetry-jaeger",
  "rand",
@@ -590,6 +591,14 @@ dependencies = [
  "libc",
 ]
 
+[[package]]
+name = "incomplete_file"
+version = "0.0.1"
+source = "git+https://github.com/jmpesp/incomplete_file#af7bda622564e842d3f6c8a18aa963f929616ef1"
+dependencies = [
+ "rand",
+]
+
 [[package]]
 name = "instant"
 version = "0.1.11"
diff --git a/downstairs/Cargo.toml b/downstairs/Cargo.toml
index ed9a63b..a9ccd24 100644
--- a/downstairs/Cargo.toml
+++ b/downstairs/Cargo.toml
@@ -28,6 +28,7 @@ tracing-subscriber = "0.2.19"
 tracing-opentelemetry = "0.14.0"
 uuid = { version = "0.8", features = [ "serde", "v4" ] }
 rusqlite = { version = "0.25" }
+incomplete_file = { git = "https://github.com/jmpesp/incomplete_file" }
 
 [dev-dependencies]
 tempfile = "3"
diff --git a/downstairs/src/main.rs b/downstairs/src/main.rs
index 8dcea84..7f0d34c 100644
--- a/downstairs/src/main.rs
+++ b/downstairs/src/main.rs
@@ -202,7 +202,7 @@ fn downstairs_import<P: AsRef<Path> + std::fmt::Debug>(
      * Open the file to import and determine how many extents we will need
      * based on the length.
      */
-    let mut f = File::open(&import_path)?;
+    let mut f = incomplete_file::IncompleteFile::open(&import_path)?;
     let file_size = f.metadata()?.len();
     let (_, extent_size, _) = region.region_def();
     let space_per_extent = extent_size.byte_value();
```

All the testing passed.

leftwo (approved):
I like the built in tests too!

Issue #109, opened 2021-11-11
Prototype replacing XTS with authenticated encryption


jmpesp:
Guido's latest suggestion is to forgo XTS:

> From this numbers I wouldn't consider AES XTS plus a data integrity mode plus a diffusion layer.

Prototype using AES-GCM-SIV instead.

rmustacc:
Can we please make sure the why is captured in an rfd?

jmpesp:
100% but at this stage I'm still asking the question of whether or not AES-GCM-SIV incurs too large of a performance hit. If we end up switching, I will detail why.

leftwo:
> 100% but at this stage I'm still asking the question of whether or not AES-GCM-SIV incurs too large of a performance hit. If we end up switching, I will detail why.

I think either way we should document the decision, just so our future selves will know what we were thinking when we did what we did.

jmpesp:
> I think either way we should document the decision, just so our future selves will know what we were thinking when we did what we did.

:+1:

jmpesp:
I ran some benchmarks comparing the current encryption (AES-XTS-128) with both AES-GCM-SIV and Deoxys-II-256 (as recommended by Guido's latest email).

All benchmarks were taken branched off 72e94ad19d82e1d03cf59a3ff8af7aceab6d02e3, where that commit represents the baseline. I benchmarked using the hammer executable from that commit running 3 times for warm up, then recording ptime output for another 10 runs.

This report compares running AES-XTS-128 with a run that generates a fake nonce and tag. A "real" slowdown (wall clock time as per https://illumos.org/man/1/ptime) of 11.4417% +/- 0.448761% was seen, along with sys and user slowdowns, representing only the additional time required to store the nonce and tag.

```
$ ./bench_values.sh before_aes_xts before_aes_xts_with_fake_nonce_and_tag 
real
x before_aes_xts.val
+ before_aes_xts_with_fake_nonce_and_tag.val
+----------------------------------------------------------------------+
|   xx                                                      +  + +     |
|x xxx                                                      ++ + ++ + +|
|  |A|                                                                 |
|                                                            |__AM_|   |
+----------------------------------------------------------------------+
    N           Min           Max        Median           Avg        Stddev
x  10     35.940334     36.207109      36.14301     36.127949   0.076867549
+  10     39.963304     40.664841     40.312214      40.26159    0.23160112
Difference at 95.0% confidence
	4.13364 +/- 0.162128
[...]

PR #124, opened 2021-11-22
[#109] Use Authenticated Encryption

jmpesp:
Replace the current AES-XTS-128 with AES-256-GCM-SIV. This builds off
the previous work to add nonces and tags to the protocol.

With authenticated encryption, decryption operations can fail now.
Decrypt in Downstairs::complete and set the job status to a new
DecryptionError if it fails. In the future, maybe we can mark that
Downstairs as untrustworthy? GtoS no longer needs an encryption context.

This means that Downstairs::complete only has to copy Bytes into the
job, not ReadResponses anymore.

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -2813,7 +2813,8 @@ impl Upstairs {
                 );
 
                 // XXX reconciliation needs to occur, but do we trust that
-                // Downstairs anymore? once could imagine:
+                // Downstairs anymore? One could imagine setting that untrusted
+                // here:
                 //
                 // self.ds_transition(client_id, DsState::Untrusted);
             }
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -30,9 +30,9 @@ use tracing::{instrument, span, Level};
 use usdt::register_probes;
 use uuid::Uuid;
 
-use aes::cipher::generic_array::GenericArray;
-use aes::{Aes128, NewBlockCipher};
-use xts_mode::{get_tweak_default, Xts128};
+use aes_gcm_siv::aead::{AeadInPlace, NewAead};
+use aes_gcm_siv::{Aes256GcmSiv, Key, Nonce, Tag};
+use rand_chacha::ChaCha20Rng;
 
 mod pseudo_file;
 mod test;
@@ -1544,7 +1544,8 @@ impl Downstairs {
         &mut self,
         ds_id: u64,
         client_id: u8,
-        read_data: &Result<Vec<ReadResponse>, CrucibleError>,
+        responses: Result<Vec<ReadResponse>, CrucibleError>,
+        encryption_context: &Option<Arc<EncryptionContext>>,
     ) -> Result<bool> {
         /*
          * Assume we don't have enough completed jobs, and only change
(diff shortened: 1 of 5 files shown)


upstairs/src/lib.rs
```diff
@@ -2735,6 +2806,16 @@ impl Upstairs {
...
                 self.ds_transition(client_id, DsState::Deactivated);
                 self.set_inactive();
+            } else if err == CrucibleError::DecryptionError {
+                println!(
+                    "Authenticated decryption failed from client id {}!",
+                    client_id
+                );
+
+                // XXX reconciliation needs to occur, but do we trust that
+                // Downstairs anymore? once could imagine:
```
leftwo, line 2816: Can we tell if this is malicious or just regular corruption?

leftwo (commented):
Were the previous performance numbers you shared based on the prototype that this PR became?

jmpesp:
> Were the previous performance numbers you shared based on the prototype that this PR became?

Yep, the numbers in the linked issue are from this PR :)

upstairs/src/lib.rs
jmpesp: There is a distinction between an Err(...) response from the Downstairs, and an Ok(ReadResponse) where the data + nonce + tag does not decrypt properly. Using zfs should prevent the scenario where the read reports Ok but it's actually returned corrupted bytes.

Reaching this scenario should then mean either:

- an error has written junk Downstairs, or
- using the wrong key, or
- malicious tampering, or
- a corruption that zfs didn't catch, or
- something else?

leftwo: Using the wrong key is probably the most common one to run into.

upstairs/src/lib.rs
```diff
@@ -1544,7 +1544,8 @@ impl Downstairs {
         &mut self,
         ds_id: u64,
         client_id: u8,
-        read_data: &Result<Vec<ReadResponse>, CrucibleError>,
+        responses: Result<Vec<ReadResponse>, CrucibleError>,
+        encryption_context: &Option<Arc<EncryptionContext>>,
```
leftwo, line 1548: This method, like many, is growing in how much it does under the name of "complete".  I think if we add any more to it, then it might be time to break it up

upstairs/src/lib.rs
leftwo: For the actual line 2816, Nit: "One could imagine"?  Or maybe you mean to say "Once upon a time?"

leftwo (approved):
Currently, I believe, encryption happens when `encryption_context` is `Some()`. That is determined by an option passed to up_main()

I wonder if the expectation of encryption should happen when a region is created and that stored somewhere in the region.  The upstairs on connection would look at the region file and say, hey, I see this is encrypted.  From there it will either expect that it (the upstairs) has been given a key and if not, then just quit trying. 

Maybe all of this is for part 2 of this, or even later.  At some point we will decide either all on on AC, or it's not worth the cost.  Perhaps then we can add expectations.

I bring this up because I know I'll try to ready my own previously encrypted region and waste a bunch of time trying to find a data corruption bug.

upstairs/src/lib.rs
```diff
@@ -1566,6 +1567,52 @@ impl Downstairs {
...
+                            let decryption_result = context.decrypt_in_place(
+                                &mut response.data[..],
+                                nonce,
+                                tag,
+                            );
+
+                            if decryption_result.is_err() {
+                                decryption_error =
+                                    Some(Err(CrucibleError::DecryptionError));
+                                break;
```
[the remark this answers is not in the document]
jmpesp: Correct - we were asked by the Guest to read N blocks and if one or more of them don't decrypt, we don't return any back. I'd rather not support returning only the good blocks, or returning less than N and expecting the Guest to ask for the remaining ones. We have three replicas and another one can satisfy the whole read.

upstairs/src/lib.rs
jmpesp: I meant "one could imagine" haha, fixed in 1b9f56b

jmpesp:
> I wonder if the expectation of encryption should happen when a region is created and that stored somewhere in the region. The upstairs on connection would look at the region file and say, hey, I see this is encrypted. From there it will either expect that it (the upstairs) has been given a key and if not, then just quit trying.

The recommendation is that nothing should be decided upon based on Downstairs. Given that our file formats are all known, an attacker in the Downstairs zone could remove this expectation from the file.

An equivalent problem after this PR exists in that an attacker could remove the nonce and tag rows - the Upstairs would serve junk to the customer instead of the decrypted plaintext.

upstairs/src/lib.rs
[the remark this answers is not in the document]
leftwo: Yeah, I don't think returning a partial read is what we want, but there may be some situation where it would make sense.

PR #125, opened 2021-11-29
Stats

leftwo:
A initial pass at adding some Oximeter stats to the downstairs.  The work here is based off the example code provided in the omicron/oximeter repo.

If requested on startup, a dropshot server is started and we produce stats to the default Oximeter endpoint.  The current stats recorded are:
Number of connection requests from an upstairs.
Number of region reads completed.
Number of region writes completed,
number of flush commands issued.

This work should grow to report more stats, and to take options on the command line for the location of the Oximeter endpoint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,9 @@ Then, go to `http://localhost:16686` to see the Jaeger UI.
 
 # Oximeter #
 Some basic stats have been added to the downstairs that can be sent to Oximeter.
-To enable stats on downstairs, add the `--oximeter` option.
+Currently, only a locally running Oximeter server is supported, and only at
+the default port. To enable stats when running a downstairs, add the
+`--oximeter` option.
 
 To display the stats, you can use the oxdb command from omicron/oximeter
 along with `jq` to make it pretty.
--- a/README.md
+++ b/README.md
@@ -191,16 +191,16 @@ To display the stats, you can use the oxdb command from omicron/oximeter
 along with `jq` to make it pretty.
 
 Replace the UUID below with the UUID for the downstairs you wish to view.
-The available stats are: ds_connect, ds_flush, ds_read, ds_write.
+The available stats are: connect, flush, read, write.
 
 ```
-cargo run --bin oxdb -- query ds_stat:ds_flush ds_uuid=12345678-3801-3801-3801-000000003801 | jq
+cargo run --bin oxdb -- query crucible_downstairs:flush downstairs_uuid=12345678-3801-3801-3801-000000003801 | jq
 ```
 
-Here is a deeper example, to just print the latest count for ds_flush:
+Here is a deeper example, to just print the latest count for flush:
 ```
-LAST_FLUSH=$(cargo run --bin oxdb -- query ds_stat:ds_flush ds_uuid=12345678-3801-3801-3801-000000003801 | jq '.[].measurements[].timestamp '| sort -n | tail -1)
-cargo run --bin oxdb -- query ds_stat:ds_flush ds_uuid=12345678-3801-3801-3801-000000003801 | jq ".[].measurements[] | select(.timestamp == $LAST_FLUSH) | .datum.CumulativeI64.value"
+LAST_FLUSH=$(cargo run --bin oxdb -- query crucible_downstairs:flush downstairs_uuid=12345678-3801-3801-3801-000000003801 | jq '.[].measurements[].timestamp '| sort -n | tail -1)
+cargo run --bin oxdb -- query crucible_downstairs:flush downstairs_uuid=12345678-3801-3801-3801-000000003801 | jq ".[].measurements[] | select(.timestamp == $LAST_FLUSH) | .datum.CumulativeI64.value"
 ```
 
 ## License
--- a/README.md
+++ b/README.md
@@ -185,7 +185,9 @@ Then, go to `http://localhost:16686` to see the Jaeger UI.
 Some basic stats have been added to the downstairs that can be sent to Oximeter.
 Currently, only a locally running Oximeter server is supported, and only at
 the default port. To enable stats when running a downstairs, add the
-`--oximeter` option.
+`--oximeter <IP:Port>` option.  If running locally, the oixmeter IP:Port is
+127.0.0.1:12221
+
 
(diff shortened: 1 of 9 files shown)


upstairs/src/lib.rs
```diff
@@ -1038,6 +1038,7 @@ async fn cmd_loop(
                  * To keep things alive, initiate a ping any time we have
                  * been idle for (TBD) seconds.
                  */
+                println!("[{}] send ping", up_coms.client_id);
```
bnaecker, line 1041: Just making sure you meant to include a new `println!` :)

downstairs/src/stats.rs
```diff
@@ -0,0 +1,196 @@
...
+use omicron_common::api::internal::nexus::ProducerEndpoint;
+use oximeter::{
+    types::{Cumulative, Sample},
+    Error, Metric, Producer, Target,
+};
+use oximeter_producer::{Config, Server};
+
+// These structs are used to construct the required stats for Oximeter.
+#[derive(Debug, Copy, Clone, Target)]
+pub struct DsStat {
```
bnaecker, line 14: I'd probably recommend renaming these structures to something more descriptive. I've been thinking of these as akin to DTrace providers and probes, in which case the names should be meaningful enough that the data they export could be easily interpreted. And ideally, one would be able to search for them without knowing their exact names. Something like `CrucibleDownstairs` for the target and `{Connect,Write,Read,Flush}Count` for the metrics might be a good choice.

downstairs/src/stats.rs
```diff
@@ -0,0 +1,196 @@
...
+    write_count: Vec<DsWrite>,
+    read_count: Vec<DsRead>,
+    flush_count: Vec<DsFlush>,
+}
+
+impl DsCountStat {
+    pub fn new(ds_uuid: Uuid) -> Self {
+        DsCountStat {
+            stat_name: DsStat { ds_uuid },
+            up_connect_count: vec![DsConnect {
```
bnaecker, line 57: Since `Cumulative` implements `Default`, you could add a derive for `Default` to `DsConnect` and friends. Then this would be:

```
DsCountStat {
    stat_name: DsStat { ds_uuid },
    ..default::Default(),
}
```

bnaecker: Nevermind, these are `Vec`s, which would default to empty, I believe.

bnaecker: Actually, why are these vectors? It doesn't seem like there's ever more than one element, but maybe there will be in the future?

downstairs/Cargo.toml
```diff
@@ -9,11 +9,17 @@ edition = "2018"
...
 crucible = { path = "../upstairs" }
 crucible-common = { path = "../common" }
 crucible-protocol = { path = "../protocol" }
+dropshot = { git = "https://github.com/oxidecomputer/dropshot", branch = "main" }
 futures = "0.3"
 futures-core = "0.3"
+omicron-common = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
+oximeter-producer = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
+oximeter = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
+proc-macro2 = "1.0.32"
```
bnaecker, line 22: Is this required? I don't see it used anywhere, so might be leftover from development.

bnaecker (commented):
This is awesome, thanks for helping to test out oximeter! I've got a few small thoughts, and one bigger one around naming the target and metric structs. I could be convinced otherwise, but it seems the names here should err on the side of explicitness and interpretable, since future engineers (ourselves included) may be looking for and processing them to aid debugging, product iteration, customer support, etc.

README.md
```diff
@@ -181,6 +181,28 @@ Pass an option to crucible-downstairs to send traces to Jaeger:
...
+Replace the UUID below with the UUID for the downstairs you wish to view.
+The available stats are: ds_connect, ds_flush, ds_read, ds_write.
+
+```
+cargo run --bin oxdb -- query ds_stat:ds_flush ds_uuid=12345678-3801-3801-3801-000000003801 | jq
+```
+
+Here is a deeper example, to just print the latest count for ds_flush:
+```
+LAST_FLUSH=$(cargo run --bin oxdb -- query ds_stat:ds_flush ds_uuid=12345678-3801-3801-3801-000000003801 | jq '.[].measurements[].timestamp '| sort -n | tail -1)
```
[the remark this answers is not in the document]
leftwo: I hope to have several examples eventually that show how to pick out different aspects of the stats, but thanks for the tip.

downstairs/Cargo.toml
leftwo: At some point during a bunch of trying to get public/private/usdt version change/etc work, I added it, but I just checked and a compile without it seems to work, so I'll pull it out.

downstairs/src/stats.rs
leftwo: I made them `Vec` because I thought I had to :)
I had some trouble when I tried to make them a `Cumulative`.  Now that everything else is working, I'll see if I can get them back as a single element

upstairs/src/lib.rs
leftwo: No, no.  This was debug cruft along with the updates around Ping on the downstairs side.  I'll remove both.

downstairs/src/stats.rs
bnaecker: Cool, making them just `Cumulative` "should" work. It's also probably possible to restructure some of this to avoid the explicit lock as well, since the `Cumulative` is thread-safe. But that's probably not worth it at this point :)

leftwo: I've updated these all to be just a single `Cumulative`

downstairs/src/stats.rs
leftwo: Names have been changed.

upstairs/src/lib.rs
leftwo: Removed!

leftwo (commented):
The latest version should now have all current comments resolved.
Let me know if you see anything else.

bnaecker:
ship it!

PR #126, opened 2021-11-29
Create "biggest" test that sends the largest IO

### LINKED ISSUE #101: Figure out what our max IO size is
Is there a max IO size?  There must be some upper limit.
Figure out what that is (and then write tests for it).

jmpesp:
Create a new "biggest" client test, sending the largest IO our protocol
allows. This required binary searching the number of blocks -
deterministically computing that number didn't work out for some reason.

This uncovered a few issues that are also fixed in this commit:

- the Downstairs would push ping responses onto a queue for task that
  pulled from the message channel. respond to pings instantly instead.

- also downstairs, the lossy_interval check would attempt to lock the
  `Arc<Downstairs>` and lock up the whole select because the task doing
  the actual work would be holding the lock. This would cause pings to
  be dropped, so don't attempt to get that lock.

Fixes #101.

### DIFF
--- a/protocol/src/lib.rs
+++ b/protocol/src/lib.rs
@@ -171,7 +171,9 @@ impl CrucibleEncoder {
      * Binary search to find the maximum number of blocks we can send.
      *
      * Attempts at deterministically computing the number of blocks
-     * repeatedly failed, so binary search instead.
+     * repeatedly failed, so binary search instead. Note that this computes the
+     * maximum size that this Tokio encoding / decoding supports given our
+     * constant MAX_FRM_LEN.
      */
     pub fn max_io_blocks(bs: usize) -> Result<usize, anyhow::Error> {
         let size_of_write_message =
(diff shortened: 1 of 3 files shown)


protocol/src/lib.rs
```diff
@@ -140,6 +142,122 @@ impl CrucibleEncoder {
...
+            nonce: Some(vec![0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]),
+            tag: Some(vec![0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]),
+        }
+    }
+
+    /*
+     * Binary search to find the maximum number of blocks we can send.
+     *
+     * Attempts at deterministically computing the number of blocks
+     * repeatedly failed, so binary search instead.
```
leftwo, line 174: Let's just add a little more to the comment here indicating that we are figuring out the max size that the tokio encode/decode supports

leftwo (approved):

PR #127, opened 2021-11-29
update for nightly-2021-11-24

ahl:
We're moving omicron and all dependents past the asm_sym flag day. See https://github.com/oxidecomputer/omicron/pull/448


leftwo (approved):
LGTM.  I did a quick test on helios and macos and everything worked as expected.

PR #133, opened 2021-12-01
Update oximeter querying comparison operators

bnaecker:


bnaecker:
I updated the way we specify comparison operators in `oximeter`, e.g., `==` for field value equality rather than `=`. I think I caught all use sites.

leftwo (approved):

PR #134, opened 2021-12-02
merge agent/smf from demo_m2 and add agent-client crate

ahl:
The merges agent and the smf crate that @jclulow did in demo_m2; it also adds the generation of an OpenAPI description of the crucible agent API and a progenitor-generated client crate. This work is not yet done in that the agent requires more work; however, I'm proposing that we merge this so that @smklein can start wiring up the agent-client and so that others can potentially help nudge this forward. I don't think the less-than-completed state would be detrimental to other work in this repo.

The next steps to getting this completed:

- fix some of the downstairs command flags as they're invoked from agent/downstairs.xml
- test agent with SMF

... and of course, the door is open to evolve the agent and its API as we see what we need it to do.

### DIFF
--- a/agent/src/model.rs
+++ b/agent/src/model.rs
@@ -1,3 +1,5 @@
+// Copyright 2021 Oxide Computer Company
+
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
--- a/agent/src/model.rs
+++ b/agent/src/model.rs
@@ -63,15 +63,17 @@ impl CreateRegion {
 }
 
 #[derive(
-    Serialize, Deserialize, JsonSchema, Debug, PartialEq, Eq, Clone, Ord,
+    Serialize,
+    Deserialize,
+    JsonSchema,
+    Debug,
+    PartialEq,
+    Eq,
+    Clone,
+    PartialOrd,
+    Ord,
 )]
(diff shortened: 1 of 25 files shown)


smf/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+
+mod transaction;
+pub use transaction::{CommitResult, Transaction, TransactionEntry};
+
+pub type Result<T> = std::result::Result<T, ScfError>;
+
+#[derive(Error, Debug)]
+pub enum ScfError {
+    #[error("no error")]
+    None,
```
leftwo, line 44: My favorite error, the old "Error 0: No error"

agent/src/model.rs
```diff
@@ -0,0 +1,103 @@
...
+pub enum State {
+    Requested,
+    Created,
+    Tombstoned,
+    Destroyed,
+    Failed,
+}
+
+#[derive(Serialize, Deserialize, JsonSchema, Debug, PartialEq, Clone)]
+pub struct Region {
```
leftwo, line 17: Seems like there could be some overlap here with the region struct and the region struct that the upstairs/downstairs uses.  Perhaps in the future (not now) we can merge these two.

PR #142, opened 2021-12-09
add option to only show differences when dumping

### LINKED ISSUE #140: downstairs dump tool, option to only show if different
Add an option to the downstairs dump (and compare) tool to only print a block
if the downstairs differ from each other (omit lines where all three are equal).
Then, possibly, run on all extents and just find the differences?

jmpesp:
when dumping (and comparing) downstairs, add an option to only show the
differences.

also bail with an error if differences are found.

Closes #140

### DIFF
--- a/downstairs/src/dump.rs
+++ b/downstairs/src/dump.rs
@@ -15,6 +15,7 @@ struct ExtInfo {
 pub fn dump_region(
     region_dir: Vec<PathBuf>,
     cmp_extent: Option<u32>,
+    only_show_differences: bool,
 ) -> Result<()> {
     /*
      * We are building a two level hashmap.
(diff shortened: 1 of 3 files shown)


downstairs/src/dump.rs
```diff
@@ -353,14 +394,34 @@ fn show_extent(
...
+                print!("{}", column);
+            }
+            for column in nonce_columns.iter().take(dir_count) {
+                print!("{}", column);
+            }
+            for column in tag_columns.iter().take(dir_count) {
+                print!("{}", column);
+            }
+
+            print!("{0:^7}", if different { "<-------" } else { "" });
```
leftwo, line 415: Nit: If you are only printing differences, we could probably skip the DIFF column all together.

leftwo (approved):

leftwo (approved):

PR #151, opened 2021-12-22
Downstairs reconciliation part 1

leftwo:
These updates are work toward making the downstairs consistent.

Added new downstairs state: `Repair`.  This will be used when a mismatch
is detected and the upstairs is coordinating the repair to make them
all consistent.

Updated the `ds_reconcile()` method to pave the way for actual repair.
The default is still to let mismatches continue, as we don't yet have
a way to repair, but that is coming soon.

The new mend.rs file contains the logic for both finding a mismatch in
downstairs extents and building the list of repair work that will be
required to make all downstairs consistent.  This repair work consists
of a source extent and one or two destination extents.
Keep track of the region info (gen, flush, dirty) from each downstairs
when they first connect.

Removed the `process_downstairs()` function and moved that logic to either
the `proc()` function (to remember the info) or to the `mismatch_list()`
method where we determine where the issues are.

Added a bunch of tests around finding mismatches with region info.

Added some checks to allow IO to work if the downstairs is in Repair
state.

Made `submit_read()` and `submit_write()` methods sender field optional.
This opens the door to the upstairs sending IO on its own behalf
instead of only from a guest.  This is similar to what the flush
operation does now.

Simplified the `submit_flush()` and `submit_read()` methods that no longer
need a vec for new work.


upstairs/src/lib.rs
```diff
@@ -2544,8 +2487,29 @@ impl Upstairs {
...
+                if old_state != DsState::WaitQuorum
+                    && old_state != DsState::Repair
+                {
+                    panic!(
+                        "[{}] {} Invalid transition: {:?} -> {:?}",
+                        client_id, self.uuid, old_state, new_state
+                    );
+                }
+                /*
+                 * Make sure repair happens when offline
```
jmpesp, line 2508: I think this should read "Make sure repair happens when inactive", to avoid confusion with the Offline state.

jmpesp (approved):
:shipit: except for a comment change suggestion :)

upstairs/src/lib.rs
leftwo: Oh yeah, that is a much needed suggestion.

leftwo: Fixed in 2b8628c

PR #158, opened 2022-01-05
Dtrace probes for tracking downstairs IO times from upstairs.

leftwo:
Added some probes that track IO to each downstairs from the upstairs.

I think the dtrace example script can be better, but I'm not sure how to simplify it.
It is currently rather verbose, and I think there could be a way to make it more concise, but I'm not sure how to organize things to make it have less code, so suggestions welcome.


tools/perfdw.d
```diff
@@ -0,0 +1,95 @@
...
+cdt*:::gw_flush_submit_start
+/arg1 == 2/
+{
+    start2[arg0] = timestamp;
+}
+
+/*
+ * When a read ACK comes back, calculate the delta and store it
+ */
+cdt*:::gw_read_submit_end
```
bnaecker, line 34: I think the aggregation for all timestamps could have two keys, one for the operation, one for the client ID. E.g., `@time["read", arg1] = ...`. For reference, [this section](https://illumos.org/books/dtrace/chp-aggs.html#chp-aggs-2) of the DTrace guide discusses aggregations with multiple keys, about 4 examples into the section.

You could also make the start times an associative array, indexed by the client ID and job ID: `start[arg1, arg0] = timestamp;`. Combined with the above, I think you can group all the start actions into one, and the three end actions for each operation type (r/w/f) into one per type.

@ahl Would definitely have some more concrete suggestions here too.

upstairs/src/lib.rs
```diff
@@ -52,6 +52,12 @@ mod cdt {
     fn gw_read_end(_: u64) {}
     fn gw_write_end(_: u64) {}
     fn gw_flush_end(_: u64) {}
+    fn gw_read_submit_start(_: u64, _: u64) {}
+    fn gw_write_submit_start(_: u64, _: u64) {}
+    fn gw_flush_submit_start(_: u64, _: u64) {}
+    fn gw_read_submit_end(_: u64, _: u64) {}
+    fn gw_write_submit_end(_: u64, _: u64) {}
+    fn gw_flush_submit_end(_: u64, _: u64) {}
```
bnaecker, line 60: If you're adding or modifying these, now might be a good time to rename them in accordance with some DTrace conventions: dashes between words (rather than snake case) and start/done to bracket an operation.

E.g., rename `gw_read_submit_end` to `gw__read__submit__done`. If you update the `usdt` dependency to the latest (0.3.1), the double-underscores will be converted to dashes in the probe name. So that would become `gw-read-submit-done` at the end of the day.

bnaecker (commented):
Nice! I've got a few suggestions for simplifying the D script, using multi-dimensional aggregations.

tools/perfdw.d
ahl: Totally agreed, @bnaecker. `start[arg0, arg1] = timestamp` or whatever

tools/perfdw.d
```diff
@@ -0,0 +1,95 @@
...
+/start0[arg0] != 0 && arg1 == 0/
```
ahl, line 35: Perhaps something like this would work:

```dtrace
cdt*:::gw_*_submit_done
/start[arg0, arg1]/
{
    strtok(probename, "_");
    this->cmd = strtok(NULL, "_");

    @[strjoin(this->cmd, " downstairs client"), arg1] = quantize(timestamp - start[arg0, arg1]);
}
```

upstairs/src/lib.rs
ahl: agreed here as well

ahl (commented):
This is great, Alan! I a couple of comments, mostly just agreeing with Ben. Thanks for doing it.

tools/perfdw.d
leftwo: Ah, yes.  Much better

```
  cdt*:::gw_read_submit_start,
  cdt*:::gw_write_submit_start,
  cdt*:::gw_flush_submit_start
  {
      start[arg0, arg1] = timestamp;
  }

  cdt*:::gw_read_submit_end
  /start[arg0, arg1] != 0/
  {
      @time["read  client", arg1] = quantize(timestamp - start[arg0, arg1]);
      start[arg0, arg1] = 0;
  }

```

Still gives me a unique report for each client (though I do have to look to the left to see the client)
```
  read downstairs client                                            0
           value  ------------- Distribution ------------- count
          524288 |                                         0
         1048576 |@@@                                      3
         2097152 |                                         0
         4194304 |@@@@@                                    6
         8388608 |@@@@@@@@@@@                              12
        16777216 |@@@@@@@@@@@@@@@@@@@                      21
        33554432 |@@                                       2
        67108864 |                                         0
```

tools/README.md
```diff
@@ -75,7 +75,147 @@ dtrace: system integrity protection is on, some features will not be available
...
+
+Here is an example of how it might look:
+```
+final:crucible alan$ sudo dtrace -Z -s tools/perfdw.d
+dtrace: system integrity protection is on, some features will not be available
+
+dtrace: script 'tools/perfdw.d' matched 0 probes
+^C
+
+  flush downstairs client 1
```
ahl, line 95: I think the output here may have changd?

ahl:
The update looks great!

tools/README.md
leftwo: Updated!

tools/perfdw.d
leftwo: I rolled that change in, and also for the perfgw.d script as I was updating it as well.

upstairs/src/lib.rs
leftwo: Updated with __ and end->done

leftwo:
Should have all comments addressed now, plus some bonus polishing.

bnaecker (commented):
Looks great to me!

tools/perfgw.d
```diff
@@ -1,27 +1,16 @@
...
-    @time["write"] = quantize(timestamp - start[arg0]);
-    start[arg0] = 0;
-}
-cdt*:::gw_flush_end
-/start[arg0] != 0/
-{
-    @time["flush"] = quantize(timestamp - start[arg0]);
+    strtok(probename, "-");
+    this->cmd = strtok(NULL, "-");
+    @time[this->cmd] = quantize(timestamp - start[arg0]);
```
ahl, line 14: I know I suggested this, but I'm not sure your previous way wasn't clearer. I leave it to you to decide what you prefer.

ahl (commented):
looks great!

PR #160, opened 2022-01-07
Make use of the #derive(JsonSchema) addition to progenitor

smklein:
Relies on https://github.com/oxidecomputer/progenitor/pull/21

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,3 +12,6 @@ members = [
 	"smf",
 	"upstairs",
 ]
+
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -12,3 +12,6 @@ members = [
 	"smf",
 	"upstairs",
 ]
+
+[patch."https://github.com/oxidecomputer/progenitor"]
+progenitor = { path = "../progenitor/progenitor" }
```
smklein, line 17: This obviously need to be removed before merging.

PR #177, opened 2022-01-13
Remove downstairs microbenchmark

jmpesp:
Use a 50k region instead of a 5M region to reduce benchmark time.


jmpesp:
With this change, the "Downstairs Benchmark" Github action step has been reduced from 18m to 9m.

leftwo:
9min is better, for sure.  It still may be too long, but at some point we are trading time for a valid test.

By reducing the size this much, do we lose value in the results of the test?

We may want to consider another "nightly" test that will run longer and might be a better place for this and possibly other tests.  I know the test_up.sh has some overlap with itself.

jmpesp:
@leftwo I'm changing this PR to remove it instead. We may have a downstairs microbenchmark in the future but it will require more thinking about validity.

leftwo (approved):

PR #180, opened 2022-01-19
Deactivate the downstairs from the upstairs

leftwo:
**Support deactivate on the upstairs.**

For a deactivation to complete, we need to:

- Stop all incoming IO.
- Let any outstanding IO finish.
- Submit a final flush to all downstairs.
- Wait for that final flush to finish (on all downstairs).
  
We may be able to take some shortcuts if there is no work
 in the active queue, and our last IO was a flush.
  
We decide what to do while holding the upstairs active state lock.
This prevents any work sneaking in after our switch to deactivation,
and prevents confusion from a flush already in the work queue from
being confused as our deactivation flush.  Since we plan to create a
flush with a real guest job ID and use the ACK of that flush as a
way to notify the guest that their deactivate request is done,
we also need the guest work lock.

By creating a real guest job, we can have the guest wait on completion
of that job as a way to be sure the final flush has been ack'd from
all downstairs (that are present).

At the moment, we don't have any timeout on how long we will try
to clear the outstanding work and final flush.  We try forever and
will only give up if a downstairs goes offline or we finish the
work in the queue.

Details about the changes here:
New client test for deactivation.
New option for client test to retry activation.
Added deactivate test to test_up.sh script.

Simplified the up_listen logic, just one loop instead of two.
New upstairs state, instead of boolean active.
Added some cargo tests for deactivation.
Added new block op for deactivation.

We will need to do some more work to support deactivation when a downstairs is
offline or goes offline while deactivating.


upstairs/src/lib.rs
```diff
@@ -648,9 +629,32 @@ async fn proc(
...
+                             * and is sending us the UUID of the "new"
+                             * downstairs, but that UUID IS our UUID.  So
+                             * clearly the downstairs is confused.
+                             */
+                            bail!(
+                                "[{}] {} bad YouAreNoLongerActive, {:?}!",
+                                up_coms.client_id, up.uuid, new_active_uuid,
+                            );
                         }
+                        return Err(CrucibleError::UuidMismatch.into());
```
jmpesp, line 657: To check my understanding, returning this Err + setting this client_id to DsState::New will cause a whole new connection attempt to that particular downstairs, which means that this upstairs will attempt to negotiate and go active again?

upstairs/src/lib.rs
```diff
@@ -892,20 +919,37 @@ async fn cmd_loop(
...
 
+    tokio::pin!(pm_task);
     loop {
         tokio::select! {
             /*
-             * We set biased so the loop will always start with looking for
-             * an ACK from downstairs to avoid a backup.
+             * We set biased so the loop will:
+             * First make sure the pm task is running.
+             * Second, respond to any ACKs coming back from the downstairs.
```
jmpesp, line 944: here's a comment I didn't update after moving things around - responding to ACKs doesn't occur here anymore

upstairs/src/lib.rs
```diff
@@ -1909,17 +2033,38 @@ impl EncryptionContext {
...
+    /*
+     * The upstairs is online and accepting IO from the guest.
+     */
+    Active,
+    /*
+     * Let in flight IO continue, but don't allow any new IO.  This state
+     * also means that when a downstairs task has completed all the IO
+     * it can, including the final flush, it should reset itself back to
+     * new and let the connection to the downstairs close and let the
+     * loop to reconnect (looper) happen.
```
jmpesp, line 2052: Do we need a state in between when a Guest called deactivate, and when a Guest calls activate? Immediately retrying to reconnect seems problematic here.

upstairs/src/lib.rs
```diff
@@ -2079,34 +2224,273 @@ impl Upstairs {
...
+                active.active_request = true;
+                println!("{} active request set", self.uuid);
+                Ok(())
+            }
+            UpState::Deactivating => {
+                println!("{} active denied while Deactivating", self.uuid);
+                crucible_bail!(UpstairsDeactivating);
+            }
+            UpState::Active => {
+                println!("Request to activate upstairs already active");
```
jmpesp, line 2491: no uuid in print?

upstairs/src/lib.rs
```diff
@@ -2602,8 +3023,18 @@ impl Upstairs {
      * If so, they lose the _  If not, then they will be removed.
      */
     fn ds_reconciliation(&self, _dst: &[Target], _lastcast: &mut u64) -> bool {
+        /*
+         * Reconciliation only happens during initialization.
+         */
+        let active = self.active.lock().unwrap();
+        if active.up_state == UpState::Active
```
jmpesp, line 3030: why not `active.up_state != UpState::Initializing`? if more states are added, then they could be missed being added to the if condition here

upstairs/src/lib.rs
```diff
@@ -2905,7 +3341,27 @@ impl Upstairs {
         client_id: u8,
         read_data: Result<Vec<ReadResponse>, CrucibleError>,
     ) -> Result<bool> {
+        /*
+         * We can't get the upstairs state lock when holding the downstairs
+         * lock, but we need to make decisions about this downstairs work
+         * knowing the upstairs state.  So,
+         *  * get the upstairs state lock,
```
jmpesp, line 3348: grabbing multiple locks in a certain order :grimacing: is there something else we can do here? or am I being too paranoid?

upstairs/src/lib.rs
```diff
@@ -4268,7 +4729,7 @@ async fn process_new_io(
             let _ = req.send.send(Ok(()));
         }
         BlockOp::QueryUpstairsActive { data } => {
```
jmpesp, line 4731: let's change the block op to QueryGuestIoReady?

upstairs/src/lib.rs
```diff
@@ -4299,6 +4776,10 @@ async fn process_new_io(
             *lastcast += 1;
         }
         BlockOp::Flush => {
+            if !up.guest_io_ready() {
```
jmpesp, line 4779: why is this check only here, and not under Read and Write too?

upstairs/src/lib.rs
```diff
@@ -4397,144 +4876,74 @@ async fn up_listen(
...
-         * to change which is our notification that a disconnect has
-         * happened.
-         *
-         * In addition, we also send a periodic flush when we determine it
-         * is time to do so. TODO: Figure out when is the best time to send
-         * a upstairs generated flush.
-         */
-        let mut flush_check = deadline_secs(5);
-        let mut show_work_interval = deadline_secs(5);
+            req = up.guest.recv() => {
```
jmpesp, line 4910: it looks like this now processes post-all-downstairs-coming-online guest requests, so this comment should be removed

jmpesp (approved):
:shipit:

should a call to tools/test_reconnect.sh go in .github/buildomat/jobs/build-and-test.sh?

upstairs/src/lib.rs
leftwo: Yes, it will cause a whole new connection for this downstairs.
Because we `set_inactive()` just above, this upstairs will now require a new activate request from the guest before the downstairs can re-join.

upstairs/src/lib.rs
leftwo: The upstairs will stay in state `Deactivating` until all downstairs have process the deactivation.
During this time, no new commands will be accepted from the Guest (including an activation).
Once all downstairs have completed deactivation, the upstairs will transition back to `Initializing`, which is where the upstairs will be listening for an activation as well as moving downstairs through their negotiation and, if required, reconciliation.

The downstairs themselves will reflect where they are in the path to activation.

upstairs/src/lib.rs
leftwo: I wish I knew a better way to enforce this, and I'm open to suggestions.  It makes me nervous as well as I created several deadlocks while developing this.  The current best I could think of was this comment :)

upstairs/src/lib.rs
leftwo: For read and write, we check in the `submit_` method.  For the flush, since the upstairs will call submit_flush after we have set deactivate, we can't do the check there because it will block our internally generated flush.

Would it make better sense to put the check here for all three, to at least be consistent?

client/src/main.rs
```diff
@@ -791,12 +845,53 @@ async fn one_workload(guest: &Arc<Guest>, ri: &mut RegionInfo) -> Result<()> {
...
+        );
+        let mut retry = 1;
+        while let Err(e) = guest.activate(gen) {
+            println!("{}/{} Retry:{} activate {:?}", c, count, retry, e);
+            std::thread::sleep(std::time::Duration::from_secs(5));
+            if retry > 100 {
+                bail!("Too many retires {} for activate", retry);
+            }
+            retry += 1;
+        }
```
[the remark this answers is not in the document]
leftwo: Done in 9608a17

upstairs/src/lib.rs
leftwo: Done in 9608a17

upstairs/src/lib.rs
leftwo: Good idea, Done in 9608a17.

upstairs/src/lib.rs
leftwo: Yeah, I think this may have been because I had more states for the upstairs, then changed my mind.
Done in 9608a17

client/src/main.rs
```diff
@@ -791,12 +845,53 @@ async fn one_workload(guest: &Arc<Guest>, ri: &mut RegionInfo) -> Result<()> {
...
+        println!(
+            "{}/{} CLIENT: Up:{} ds:{}",
+            c, count, wc.up_count, wc.ds_count
+        );
+        let mut retry = 1;
+        while let Err(e) = guest.activate(gen) {
+            println!("{}/{} Retry:{} activate {:?}", c, count, retry, e);
+            std::thread::sleep(std::time::Duration::from_secs(5));
+            if retry > 100 {
+                bail!("Too many retires {} for activate", retry);
```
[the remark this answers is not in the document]
leftwo: Done in 9608a17

upstairs/src/lib.rs
```diff
@@ -648,9 +629,32 @@ async fn proc(
...
-                        if up.uuid != new_active_uuid {
-                            up.set_inactive();
+                        /*
+                         * XXX If we get this response, but the new uuid is
+                         * actually our uuid, do we just ignore it?
+                         */
+                        println!(
+                            "[{}] {} downstairs self deactivate: new {:?}",
+                            up.uuid,
+                            new_active_uuid,
```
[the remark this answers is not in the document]
leftwo: Yeah, that's backwards..  Done in 9608a17

upstairs/src/lib.rs
```diff
@@ -648,9 +629,32 @@ async fn proc(
...
+                            up_coms.client_id
+                        );
+                        up.ds_transition(up_coms.client_id, DsState::New);
+                        up.set_inactive();
+                        if up.uuid == new_active_uuid {
+                            /*
+                             * Now, this is really going off the rails.  Our
+                             * downstairs things we have a different UUID
```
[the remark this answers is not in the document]
leftwo: Done in 9608a17

upstairs/src/lib.rs
leftwo: Done in 9608a17

upstairs/src/lib.rs
leftwo: Done in 9608a17

leftwo (commented):
Addressed all the specific requested changes.
Added some responses to comments.

upstairs/src/lib.rs
jmpesp: :+1:

upstairs/src/lib.rs
jmpesp: In the migration case, Upstairs 1 will deactivate, which puts it into `Initializing`. This means it will negotate with the downstairs but not activate. Say they are both active at the same time. Upstairs 2 will also negotiate, and then activate. I don't think there's a conflict here like I thought there was. Resolving this.

upstairs/src/lib.rs
jmpesp: I'd go with a comment here instead :)

leftwo: Comment added in a0e30e3

leftwo:
Interesting, this is failing the build-and-test because it's merging with main and running the test there.  I'll have to merge main into this to get the hash changes (I believe), then update, then verify it can compile.

PR #182, opened 2022-01-19
Implement TLS between Upstairs and Downstairs

jmpesp:
Add the option to use authenticated TLS between the Upstairs and
Downstairs. This required generalizing the FramedRead and FramedWrite
types so that either TlsStream or TcpStream could be used, plus wrapping
those streams with an enum in order to pass them around without rustc
complaining :)

When I use the term "authenticated" TLS here, I mean that the server
will check that client certificates are signed by a mutual certificate
authority. Clients do **not** yet check that server certs are also
signed by that authority - aka, this is not mutual TLS. I couldn't find
the appropriate options in rustls to enable this.

Scripts that generate keys and certificates have been added to an x509
directory. You'll need the cfssl tool. First, run "make" there to
generate the appropriate files for Upstairs and Downstairs.

Then, create the Downstairs like normal, but run with new arguments:

    cargo run --release -q -p crucible-downstairs -- \
        run -p "44100" -d "disks/d0/" \
        --cert-pem x509/downstairs0.pem \
        --key-pem x509/downstairs0-key.pem \
        --root-cert-pem x509/ca.pem

Certificate number is important here. Part of the TLS connection is a
hostname specification. The Upstairs will try to connect to the first
client as "downstairs0", the second as "downstairs1", and so on.

The Downstairs will then spawn an listener that only accepts client
certificates from the "root" cert. Connect Upstairs with the same
arguments:

    cargo run --bin=crucible-client -- \
        -q \
        -t 127.0.0.1:44100 \
        -t 127.0.0.1:44101 \
        -t 127.0.0.1:44102 \
        --key "HLyo7ZhAf/E9IdX2DDHPHJO0dLgrRxZabWiTlnoKZXc=" \
        --cert-pem x509/upstairs.pem \
        --key-pem x509/upstairs-key.pem \
        --root-cert-pem x509/ca.pem \
        --workload one

To test that certificates not signed by x509/ca.pem are rejected, try
using x509/selfsigned.pem and x509/selfsigned-key.pem in the client
connection. The client should fail with:

        error: InvalidCertificateData("invalid peer certificate: UnknownIssuer")

And the Downstairs should say:

        rejecting connection from 127.0.0.1:49462: Custom { kind: InvalidData, error: AlertReceived(BadCertificate) }

If you try to connect without certificates in the client, it should
fail, and the Downstairs should say:

        rejecting connection from 127.0.0.1:38116: Custom { kind: InvalidData, error: CorruptMessage }

### DIFF
--- a/common/src/x509.rs
+++ b/common/src/x509.rs
@@ -1,0 +1,82 @@
+// Copyright 2021 Oxide Computer Company
+use std::fs::File;
+use std::io::{self, BufReader};
+
+// Reference tokio-rustls repo examples/server/src/main.rs
+use rustls_pemfile::{certs, rsa_private_keys};
+use tokio_rustls::rustls::{Certificate, PrivateKey, ClientConfig, ServerConfig, RootCertStore};
+use tokio_rustls::rustls::server::AllowAnyAuthenticatedClient;
+
+pub fn load_certs(path: &str) -> io::Result<Vec<Certificate>> {
+    certs(&mut BufReader::new(File::open(path)?))
+        .map_err(|_| {
+            io::Error::new(io::ErrorKind::InvalidInput, "invalid cert")
+        })
+        .map(|mut certs| certs.drain(..).map(Certificate).collect())
+}
+
+pub fn load_rsa_keys(path: &str) -> io::Result<Vec<PrivateKey>> {
+    rsa_private_keys(&mut BufReader::new(File::open(path)?))
+        .map_err(|_| io::Error::new(io::ErrorKind::InvalidInput, "invalid key"))
+        .map(|mut keys| keys.drain(..).map(PrivateKey).collect())
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum TLSContextError {
+    #[error("PKI error")]
+    PKIError(#[from] tokio_rustls::webpki::Error),
+
+    #[error("IO error")]
+    IOError(#[from] std::io::Error),
+
+    #[error("rustls error")]
+    RusTLSError(#[from] tokio_rustls::rustls::Error),
+}
+
+
+#[derive(Debug)]
+pub struct TLSContext {
+    certs: Vec<Certificate>,
+    keys: Vec<PrivateKey>,
+    root_cert_store: RootCertStore,
+}
+
+impl TLSContext {
+    pub fn from_paths(
+        cert_pem_path: &str,
+        key_pem_path: &str,
+        root_cert_pem_path: &str,
+    ) -> Result<Self, TLSContextError> {
+        let mut root_cert_store = RootCertStore::empty();
+        for root_cert in load_certs(&root_cert_pem_path)? {
+            root_cert_store.add(&root_cert)?;
+        }
+
+        Ok(Self {
+            certs: load_certs(cert_pem_path)?,
+            keys: load_rsa_keys(key_pem_path)?,
+            root_cert_store
+        })
+    }
+
+    pub fn get_client_config(&self) -> Result<ClientConfig, TLSContextError> {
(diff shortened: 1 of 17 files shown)


common/src/x509.rs
```diff
@@ -0,0 +1,79 @@
+// Copyright 2021 Oxide Computer Company
```
leftwo, line 1: I probably forgot this in my PR.. need to go check.
New year, 2022!

common/src/x509.rs
```diff
@@ -0,0 +1,79 @@
...
+use std::fs::File;
+use std::io::{self, BufReader};
+
```
leftwo, line 4: I don't know enough about TLS to know if we are doing this correctly.  This is probably fine for our initial prototype, but maybe we want to get Rick or someone who knows this stuff better to verify it?

In the longer term, I wonder if we will have some common library that any Oxide service can use to leverage TLS?  Do we think that might happen?

x509/Makefile
```diff
@@ -0,0 +1,11 @@
+
```
leftwo, line 1: Why the Makefile here?  This won't get built unless someone specifically calls it, right?
My system does not have the binaries to generate the certs, so running the script directly does not
work either.  I think that is okay for now though, until we get serious about requiring TLS.

common/src/x509.rs
jmpesp: Addressed in 274002b :)

common/src/x509.rs
jmpesp: I'm hoping that `with_safe_defaults` delegates some of that responsibility to rustls' maintainers, but I agree that this should be verified. I know enough about TLS to probably be dangerous but can't claim expertise.

upstairs/src/lib.rs
```diff
@@ -1115,12 +1165,35 @@ async fn looper(
             }
         };
 
+        let tcp = {
+            let tls_context = tls_context.lock().await;
+            if let Some(ref tls_context) = *tls_context {
+                // XXX these unwraps are bad!
```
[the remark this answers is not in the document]
jmpesp: Unfortunately I don't know enough to say yes or no. I was wondering how difficult it would be to check the exit status of looper and restart it - I don't think the Upstairs currently does.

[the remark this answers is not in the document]
leftwo: The challenge with restarting looper is it has half the message channels, and the other half live in other tasks.  So, if we exit the looper task, we have to reconnect all those channels somehow.

common/src/x509.rs
leftwo: let's file an issue for ourselves so we don't forget about this.

leftwo:
Your PR description, we should put that in a README somewhere also.  Maybe the top level for Crucible?  Or mabye in the x509 directory?

jmpesp:
> Your PR description, we should put that in a README somewhere also. Maybe the top level for Crucible? Or mabye in the x509 directory?

Done in 7f87990d360e2a495fbaced72661cc29ba358dd3

common/src/x509.rs
jmpesp: Opened https://github.com/oxidecomputer/crucible/issues/187

upstairs/src/lib.rs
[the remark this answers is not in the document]
leftwo: If we just put a TODO: around these unwraps, then we can at least start testing with TLS and figure out if the unwraps do happen under normal operations (as opposed to buggy programs) and fix them later.  There are already a bunch of places where we need to find and handle unwraps().

leftwo (approved):
I think the TODO around the unwrap is my final comment, so do that and we can ship this

PR #188, opened 2022-01-21
derive Default for CrucibleOpts

jmpesp:
Pulled this change out of https://github.com/oxidecomputer/crucible/pull/182 so that the merge of https://github.com/oxidecomputer/propolis/pull/86 will not have to be synchronized.


leftwo (approved):
If that is all we need, great!

PR #189, opened 2022-01-25
Simple CLI for crucible client.

leftwo:
This is very similar to the nbd_server idea, but instead of nbd, I just made a CLI
that you can send specific commands to.  I sometimes what to send just a deactivate, or just
a read of a specific block, and I wanted an interface to do that without having to make a
custom program for it.

This ended up being more work than I thought, and It is still not done, but what I have here
is functional and I can iterate on it to add functionality and make some of the giant loops
into sub-functions or methods or types, depending.  There are more commands to handle
as well.

From the commit:
A super simple crucible-client server and cli program.
This is two programs based on the crucible-client test program.
A cli-server, which starts crucible-client, creates the runtime
and start the Upstairs.  This cli-sever opens a port and listens
for commands to come it, sending them to the upstairs (via the guest
methods).  Results are reported back to the CLI.

Some state is kept in the cli-server side.

The cli-client just takes a list of supported commands and sends
them over the socket to the server, reporting back their result.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 core
 *.swp
 /.*.sock
+.cli_history.txt
 /fio/*.png
 /fio/*.log
 /fio/*.log.hist
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -299,6 +299,17 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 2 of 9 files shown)


jmpesp (approved):
I really like this :ship:

PR #192, opened 2022-01-28
New Dropshot server option to Upstairs to expose some stats.

leftwo:
Added an optional debug Dropshot server to the Upstairs that exposes a
few internal stats about Upstairs state and job queues.  This can grow
into more as we decide what other things we might want to look at.  This
is pretty much a copy of the Dropshot example code with some minor changes.

Fixed a few test warnings that crept in.


leftwo:
This is really just a "am I using it right" PR for Dropshot.

ahl:
This looks good to me. A couple of things to consider as you iterate on this
- do you want it only enabled some of the time? It seems like it might be nice to have this running all of the time in case we might want to grab some info when it's misbehaving
- If you want to have a programmatic client that accesses this, you might want a subcommand to dump out the OpenAPI doc from which we can generate the client

leftwo:
Yeah, I do suspect we will want this on by default.  I had a concern it might cause some conflicts with propolis, but I admit I did not think that through very far.

For the client, yeah, that's probably a good thing to iterate into :)

PR #193, opened 2022-02-04
Create new Volume abstraction

jmpesp:
Create a new volume abstraction that maps logical block addresses to
"sub" volumes, which are anything that implements the new BlockIO trait.
A crucible volume can be composed of multiple region sets and the volume
exposes that as one LBA range. Because Volume also implements the new
BlockIO trait, it can be dropped in whereever Guest was previously used.

Volumes also support read only parents, which enables creating volumes
from snapshots.

To figure out which blocks have been previously written to, jobs now
return ReadResponse instead of Bytes, enabling the transfer function to
tell if blocks have previously been written to before. Buffer now has a
"owned" vector that can mark bytes as having been written to before.

Volume's read function checks if there's a read only parent, and
performs two reads if the LBA range being read overlaps. If the read
only parent owns the appropriate blocks but the sub volumes do not, the
read only parent data is returned.

Volumes can support having only a read only parent, and attempts to
write to them will error.

This functionality was tested by implementing InMemoryBlockIO (which
implements the BlockIO trait) and crafting Volumes from those in test
code.

Anything that implements the BlockIO trait can be used as a sub volume,
so implementations backed by a file path and a HTTP(S) URL were
created.

Still TODO:

- a migration task that will pull blocks from the read only parent,
  write those into the sub volumes, then set the read only parent to
  None.
- parallel dispatch: BlockIO is not an async trait, nor was the Guest
  previously async, so parallel dispatch of requests to subvolumes will
  have to wait.


upstairs/src/info.rs
```diff
@@ -28,7 +28,8 @@ pub async fn start_info(
     let config_dropshot = ConfigDropshot {
         bind_address: addr,
         request_body_max_bytes: 1024,
-        tls: None,
+        // tls: None,
+        ..Default::default()
```
jmpesp, line 32: I changed this because my propolis checkout wouldn't compile without it. TODO for me - pull latest propolis code and try to compile again.

upstairs/src/volume.rs
```diff
@@ -0,0 +1,1228 @@
...
+        } else {
+            None
+        }
+    }
+}
+
+impl BlockIO for Volume {
+    fn activate(&self, gen: u64) -> Result<(), CrucibleError> {
+        for sub_volume in &self.sub_volumes {
+            sub_volume.conditional_activate(gen)?;
```
jmpesp, line 242: I added conditional activation, but we may want to scrap this, and instead have the regular activate function not throw an UpstairsAlreadyActive error?

upstairs/src/volume.rs
```diff
@@ -0,0 +1,1228 @@
...
+}
+
+// Traditional subvolume is just one region set
+impl SubVolume {
+    // Compute sub volume LBA from total volume LBA.
+    //
+    // Example:
+    //
+    // Sub volume LBA range is from [1024, 2048), so total volume LBA of 1234 is
+    // sub volume LBA 210.
```
jmpesp, line 496: This comment makes no sense, TODO for me

volume_integration_tests/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "volume_integration_tests"
```
jmpesp, line 2: TODO this could probably go away?

upstairs/src/volume.rs
leftwo: We have to handle the case where one upstairs activates and the other does not.  A 2nd attempt to activate will then fail on the first upstairs (because it's already active).  So we will have to navigate these sorts of issues.

upstairs/src/volume.rs
```diff
@@ -0,0 +1,1228 @@
...
+                );
+            }
+        }
+
+        let affected_sub_volumes = self.sub_volumes_for_lba_range(
+            offset.value,
+            data.len() as u64 / self.block_size,
+        );
+
+        // TODO parallel dispatch!
```
leftwo, line 331: Let's go make issues for all these TODOs.  We can move forward with this code and be sure to have issues so we don't forget the remaining pieces.

upstairs/src/volume.rs
```diff
@@ -0,0 +1,1228 @@
...
+                    .read(parent_offset, parent_buffer.clone())?;
+                waiter.block_wait()?;
+
+                let mut data_vec = data.as_vec();
+                let sub_data_vec = sub_buffer.as_vec();
+                let sub_owned_vec = sub_buffer.owned_vec();
+                let parent_data_vec = parent_buffer.as_vec();
+
+                // "ownership" comes back from the downstairs per byte but all
+                // writes occur per block. iterate over blocks here.
```
leftwo, line 375: I have questions about this part (this comment is more for me so I can find it)

upstairs/src/volume.rs
```diff
@@ -0,0 +1,1228 @@
...
+        BlockReqWaiter::immediate()
+    }
+
+    fn flush(&self) -> Result<BlockReqWaiter, CrucibleError> {
+        for sub_volume in &self.sub_volumes {
+            let mut waiter = sub_volume.flush()?;
+            waiter.block_wait()?;
+        }
+
+        // XXX no need to flush read only parent?
```
leftwo, line 460: I hope no flush is needed for read only.
However, we do have to make sure that "read only volumes" have all three downstairs consistent.  Not really a topic for this PR though.

upstairs/src/volume.rs
jmpesp: Fixed in 741f9d7

upstairs/src/volume.rs
jmpesp: In `GtoS::transfer`:

```
                        for i in &response.data {                               
                            vec[offset] = *i;                                   
                            owned_vec[offset] = !response.hashes.is_empty();    
                            offset += 1;                                        
                        }
```

owned_vec is populated per byte but the value doesn't change for the whole response.

leftwo (commented):
This is some great work and solves a bunch of problems for us.  I think there are some "git issues" we should make so we get a better idea of what else we need to support this.

* Are there any interface changes to Propolis and Nexus, and what do we need and making sure that what is there now will at least still compile.
* Do we need to move dropshot server and stats collection up to a "volume" layer?
* Create an issue for the background migration task (copy r/o to sub-volume).
* Do we need a new test front end, or can `crucible-client` just have a new layer?
* Adam's concern about the sequential read to find the "owned" block, we should at least make an issue for performance gathering and a plan to improve.

There may be some situations this current version can't handle (like one upstairs comes online and another does not).  If we want to check this in like it is, then we should create issues for those situations (and possibly just panic for now when we hit them).  This can allow us to iterate on work here without making some giant PR that solves all problems.

leftwo (commented):
Oh, thought of another thing we need to support in here, the generation number.  When reconnecting to a region set, Nexus was to supply us with an ever increasing generation number.  We use this to determine which data is the newest and to break ties.    For support here, we need to probably have a generation number for each writable volume. 

The generation number is not plumbed all the way up to Nexus yet, and currently crucible will just ignore it and figure out the largest gen it has on all extents and +1 to that, but we still need to consider it when implementing the volume abstraction.

upstairs/src/volume.rs
jmpesp: Removed this XXX in 80c3b25.

volume_integration_tests/Cargo.toml
jmpesp: Removed in d4d57f2

jmpesp:
Created the following issues:

Volume: migrate blocks from the read-only parent https://github.com/oxidecomputer/crucible/issues/196
Volume: Ensure full activation https://github.com/oxidecomputer/crucible/issues/197
Volume: what is the correct generation number to use? https://github.com/oxidecomputer/crucible/issues/198
Volume: parallel read dispatch https://github.com/oxidecomputer/crucible/issues/199
Do we need to move dropshot server and stats collection up to a "volume" layer? https://github.com/oxidecomputer/crucible/issues/200
Volume: improve performance when there is a read-only parent https://github.com/oxidecomputer/crucible/issues/201

upstairs/src/volume.rs
leftwo: Issue #197 covers this

upstairs/src/volume.rs
leftwo: I think we talked about this only needing to check at a block level, so maybe you already did that and if not we can iterate into it.

leftwo (approved):
Let's get this in and start the iteration!

PR #212, opened 2022-02-20
Crucible Admin API updates

leftwo:
Updated the crucible upstairs admin endpoint (the dropshot server) to generate a openapi JSON file
that can then be used to create an actual admin client.  This is done
through a new upstairs binary that calls the function to display the API.

Added an admin-client directory for the admin-client library that
is generated using the openapi JSON file.  This library can be
used by programs wishing to access the Crucible Admin API.

Added a README in the openapi directory to describe how to create the
json file for crucible upstairs admin.

Added counters for completed and remaining repairs for upstairs
reconciliation. These counters are now part of the info included in
the upstairs stats structure.  Also added the state of each downstairs
to the upstairs stats structure.

Created a simple program (more coming on this later) that uses the
new Crucible Admin API to connect to an upstairs admin port and
request the info stats.  This program (cruadmin) is not part of the
Crucible workspace and has to be built manually.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ core
 /x509/*.pem
 /x509/*.json
 /x509/*.csr
+/cruadmin/target
+/cruadmin/Cargo.lock
--- a/.gitignore
+++ b/.gitignore
@@ -15,5 +15,3 @@ core
 /x509/*.pem
 /x509/*.json
 /x509/*.csr
-/cruadmin/target
-/cruadmin/Cargo.lock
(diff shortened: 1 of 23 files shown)


jclulow:
I wonder if it would be more apt to call this the "control API" rather than "admin API", as none of it should ever really be touched by an administrator but rather by other components in the software stack.

Also, for the engineering tool, CRU is somewhat overloaded as an abbreviation (Customer-Replacable Unit is what comes to mind) so maybe `crucadm`?  Why don't we put this program in the workspace?

jmpesp (approved):

PR #213, opened 2022-02-28
package up for omicron branded zone

jmpesp:
package up smf manifests, binaries, and method scripts for deploying in
an omicron branded zone

### DIFF
--- a/package/src/main.rs
+++ b/package/src/main.rs
@@ -1,0 +1,76 @@
+// Copyright 2022 Oxide Computer Company
+
+use std::fs::OpenOptions;
+use std::path::Path;
+
+use serde::Serialize;
+use tar::{Builder, Header};
+
+#[derive(Serialize)]
+pub struct OmicronZoneMetadata {
+    #[serde(rename = "v")]
+    version: usize,
+
+    #[serde(rename = "t")]
+    type_string: String,
+}
+
+fn main() -> anyhow::Result<()> {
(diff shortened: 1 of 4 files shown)


package/src/main.rs
```diff
@@ -0,0 +1,76 @@
...
+    )?;
+
+    // Binaries
+    archive.append_path_with_name(
+        "target/release/crucible-downstairs",
+        "root/opt/oxide/crucible/bin/downstairs",
+    )?;
+    archive.append_path_with_name(
+        "target/release/crucible-agent",
+        "root/opt/oxide/crucible/bin/agent",
```
smklein, line 67: Could we add some docs to this binary, at least?

Seems like it expects that:
- A caller has *manually* run `cargo build --release` prior to running `package`
- If we want to update the package, the old one must explicitly be deleted

smklein (approved):

smklein:
Okay, I actually don't think this will work as-is - inspecting your tarball, I think you aren't making intermediate paths, which means this won't be parsable by the omicron-branded zone.

I can try to update this PR to use  https://github.com/oxidecomputer/omicron-package , which should deal with zone creation for you.

smklein:
I just uploaded https://github.com/oxidecomputer/crucible/pull/217/files , which should do *functionally* the same thing as this PR, but using a `package-manifest.toml` file instead.

(Also re-shuffles some files around to make my life easier)

PR #229, opened 2022-03-07
Updates for Crucible CLI and a few test scripts

leftwo:
Crucible client, CLI, and tools updates

Updated a bunch of loops in crucible-client from 0..end to 1..=end   This looks much nicer
when I print out current loop index.

CLI updates:
   * Added a quit command to the CLI
   * Print the block read as part of read results.  Include the read block
     number in the ReadResponse protocol message.
   * Moved server side processing of client commands to a new function.
   * Add support for having the cli-server read in previous write counts
     from a file.
  *  Include the generic test as a CLI option.

Updates for tools/create-generic-ds.sh:
   * Allow deletion of region files if they exist.
   * Supply extent size and extent count.
   * Enable and expect encryption on the region.

Added a check in test_up.sh to verify command line args are valid.

### DIFF
--- a/client/src/cli.rs
+++ b/client/src/cli.rs
@@ -40,6 +40,8 @@ enum CliCommand {
     Info,
     /// Report if the Upstairs is ready for guest IO
     IsActive,
+    /// Quit the CLI
+    Quit,
     /// Read from a given block offset
     Read {
         #[structopt(long, short)]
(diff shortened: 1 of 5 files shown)


client/src/cli.rs
```diff
@@ -370,6 +369,135 @@ pub async fn start_cli_client(attach: SocketAddr) -> Result<()> {
...
+/**
+ * Process a CLI command from the client, we are the server side.
+ */
+async fn process_cli_command(
+    guest: &Arc<Guest>,
+    fw: &mut FramedWrite<tokio::net::tcp::OwnedWriteHalf, CliEncoder>,
+    cmd: protocol::CliMessage,
+    ri: &mut RegionInfo,
+    wc_filled: &mut bool,
+    vi: Option<PathBuf>,
```
jmpesp, line 381: This should probably be "verify_input" instead, I had to search for what it meant :)

jmpesp (approved):
Ship it! Just two non-blocking things:

client/src/cli.rs
leftwo: Fixed in 263d810

client/src/cli.rs
```diff
@@ -338,6 +334,9 @@ pub async fn start_cli_client(attach: SocketAddr) -> Result<()> {
                     }
                     // TODO: add a quit command
```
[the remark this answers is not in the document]
leftwo: Updated to: TODONE:
263d810

PR #235, opened 2022-03-09
A list of small things:

jmpesp:
Add address to downstairs SMF.

Remove volume id from agent's Region model, it wasn't being used.

Remove aes and xts-mode crate dependencies from upstairs.

Derive Clone and Debug for VolumeConstructionRequest.

Make clippy happy.

### DIFF
--- a/agent/src/model.rs
+++ b/agent/src/model.rs
@@ -17,7 +17,6 @@ pub enum State {
 #[derive(Serialize, Deserialize, JsonSchema, Debug, PartialEq, Clone)]
 pub struct Region {
     pub id: RegionId,
-    pub volume_id: String,
     pub state: State,
 
     // Creation parameters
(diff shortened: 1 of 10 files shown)


agent/src/model.rs
```diff
@@ -17,7 +17,6 @@ pub enum State {
 #[derive(Serialize, Deserialize, JsonSchema, Debug, PartialEq, Clone)]
 pub struct Region {
     pub id: RegionId,
-    pub volume_id: String,
```
leftwo, line 20: I take it the region itself does not need to "look back up" the tree?  Something at a higher level keeps track of what regions are part of a volume and staring with a region and going back is not needed?

leftwo (approved):

agent/src/model.rs
jmpesp: Right - Nexus will keep track of this

PR #242, opened 2022-03-17
Add ds_repair field to downstairs struct

leftwo:
Added a new field to the Downstairs struct.
This filed contains the address that reconciliation should use when
repairing extents.  This will be used in the coming reconciliation
changes.

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -1650,8 +1650,8 @@ impl Downstairs {
         // addresses from each downstairs.
         let mut ds_repair = HashMap::new();
         for (i, addr) in target.iter().enumerate() {
-            assert!(addr.port() < u16::MAX - 4000);
-            let port = addr.port() + 4000;
+            assert!(addr.port() < u16::MAX - REPAIR_PORT_OFFSET);
+            let port = addr.port() + REPAIR_PORT_OFFSET;
             let repair_addr = SocketAddr::new(addr.ip(), port);
             ds_repair.insert(i as u8, repair_addr);
         }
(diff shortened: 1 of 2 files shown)


upstairs/src/lib.rs
```diff
@@ -2903,6 +2926,14 @@ impl Upstairs {
          */
         #[cfg(not(test))]
         assert_eq!(opt.target.len(), 3);
+        /*
+         * The repair port is the downstairs target port + 4000
```
jmpesp, line 2930: Weren't we going to use even + odd ports for this? We didn't record the meeting so I can't say for sure but 1 instead of 4000 would allow the agent to allocate these pairs together.

jmpesp: Not that the agent can't allocate X and X+4000 together, just that it makes packing them easier

leftwo: We decided to use the Fibonacci sequence to select the next port number,   
We did record the meeting, you can find the link to it in the calendar entry for Monday, after the meeting recording is finished, google goes back in time and adds the recording link to the meeting info.

We talk about both +1 and +4000,  +1 at the start of the meeting, and +4000 around 5:00 into the meeting.

The reason I chose +4000 right now is to not have to go move all the tests we have written to skip ports, right now
they all just create downstairs with sequential port numbers.   I do plan to do that, but was hoping to kick that down the road a little longer.  Making the REPAIR_PORT_OFFSET a const allows me to come back later and switch it to +1 after updating all the tests.

jmpesp: :+1: This sounds good. For reference, when I said "We didn't record the meeting" I was referring to the in-person one on Friday of the meetup week.

jmpesp (approved):

upstairs/src/lib.rs
leftwo: Got it about meetings.

PR #257, opened 2022-03-30
Another piece of the reconciliation puzzle

leftwo:
This covers the next piece of reconciliation where we transfer extent files from 
one downstairs to another using a dropshot server and generating the API
to stream the extent files.
  
This builds on several previous parts that did the work of identifying mismatches 
between downstairs and creating a solve list.  

This piece executes the solve, and results in all three downstairs being the same
on completion of the solving.  This also enables a basic functionality that should
work when no outside errors are encountered (like a downstairs reset during
solve).

To get an idea of what happens when a downstairs does an extent repair, I have
copied this block comment from `downstairs/src/region.rs`:

```
     /**
       * Repair an extent from another downstairs
       *
       * We need to repair an extent in such a way that an interruption
       * at any time can be recovered from.
       *
       * Let us assume we are repairing extent 012
       *  1. Make new 012.copy dir  (extent name plus: .copy)
       *  2. Get all extent files from source side, put in 012.copy directory
       *  3. fsync files we just downloaded
       *  4. Rename 012.copy dir to 012.replace dir
       *  5. fsync extent directory ( 00/000/ where the extent files live)
       *  6. Replace current extent 012 files with copied files of same name
       *    from 012.replace dir
       *  7. Remove any files in extent dir that don't exist in replacing dir
       *  8. fsync files after copying them (new location).
       *  9. fsync containing extent dir
       * 10. Rename 012.replace dir to 012.completed dir.
       * 11. fsync extent dir again (so dir rename is persisted)
       * 12. Delete completed dir.
       *
       *  This also requires the following behavior on every extent open:
       *   A. If xxx.Copy directory found, delete it.
       *   B. If xxx.Completed directory found, delete it.
       *   C. If xxx.Replace dir found start at step 4 above and continue
       *      on through 6.
       *   D. Only then, open extent.
       */
```

Testing:  7000 or so loops of: make a downstairs different, then repair it.

Upstairs:
Update the repair logic to now do more than just close and then
re-open extents. The upstairs will now build the complete repair
steps needed to repair an extent.  This includes the logic about
which downstairs should receive which repair commands.

Protocol:
New ExtentFlush to flush a specific extent.
New ExtentRepair command for repairing a specific extent.
Replaced a few specific ACKs with a generic repair ACK.

Downstairs:
Made the repair server always on, and starting at port + REPAIR_PORT_OFFSET.
New repair-api command that can dump a json file for the repair-client API
Added the logic to support the new repair messages.

Downstairs region:
The downstairs logic for execution of a repair lives in this file.
Created new functions to get the names for the new repair directories and
a bunch of tests around them.  We can now get names for specific
parts of the extent, the housing directory, the copy, replace, and
completed directories, as well as provide an optional file extension
to the file name.
During repair, we check for and remove any copy or completed directories.
If there is a replay directory, then the extent open will do the
replay and replace existing extent files, only if the open is read/write.
New command to just flush a specific extent.

Downstairs repair:
This new file is the dropshot server for extent repair requests.
It listens on the specified port and serves (anyone currently) the
list of files for an extent, or one of the actual extent files.

Repair-client:
Generate a library using progenitor that is used by a downstairs to
request repair extents from another downstairs. This library generation
uses the json file that is output from the downstairs repair-api command.

Tests:
More tests for repair
More tests for building repair work list.
A New test that will in a loop:
 Create a workload that will produce downstairs out of sync
 Restart and repair the out of sync downstairs.
Improved usage messages for create-generic-ds.sh

**What work is still outstanding with this**:  
The following are not part of this commit, but will be addressed in a follow up
commit and issues will be made for these.  I felt this PR was big enough as is, and
the work here does allow more testing to start.
- [ ] Error handling when interrupted during download of repair files.
- [ ] Overall error handling if a repair job returns error to the upstairs.
- [ ] Panic if copy directory exists when going to create it.
- [ ] Panic if copy file exists when going to create it.
- [ ] Panic if replacement directory preconditions are not as expected
- [ ] Panic if move_replacement_extent copy fails
- [ ] fsflush of directories after movement.
- [ ] Progenitor needs an auto detect of a change in the API for downstairs-repair.
- [ ] Repair client/server authentication.  Currently anyone can request extent files from the repair server.

TBD: How we handle TLS and keys for ds->ds communication.


downstairs/src/region.rs
```diff
@@ -965,6 +1106,162 @@ impl Region {
...
+            panic!("Unknown extra files on repair list: {:?}", repair_files);
+        }
+
+        // After we have all files: move the repair dir.
+        println!(
+            "Repair files downloaded, move directory {:?} to {:?}",
+            copy_dir, rd
+        );
+
+        // XXX fsync the parent directory (the extent dir)
```
jmpesp, line 1257: https://doc.rust-lang.org/std/fs/struct.File.html#method.sync_all should work here

protocol/src/lib.rs
```diff
@@ -148,16 +149,22 @@ pub enum Message {
      */
     /// Send a close the given extent ID on the downstairs.
     /// We send the downstairs the repair ID (rep_id) and the extent number.
-    ExtentClose(u64, u64),
-    /// Ack the close of an extent from the downstairs using the rep_id.
-    ExtentCloseAck(u64),
+    ExtentClose(u64, usize),
     /// Send a request (with rep_id) to reopen the given extent.
-    ExtentReopen(u64, u64),
```
jmpesp, line 155: why not u64 for extent id?

tools/test_repair.sh
```diff
@@ -0,0 +1,89 @@
...
+fi
+
+# Stop a downstairs, we will restart with lossy in the loop
+kill "$ds3_pid"
+
+# Start loop
+for (( i = 0; i < 30; i += 1 )); do
+
+    # restart downstairs with lossy
+    ${cds} run -d var/8820 -p 8830 --lossy &> /tmp/ds2 &
```
jmpesp, line 50: typo here - directory should also be 8830, and redirect should also be ds3

upstairs/src/lib.rs
```diff
@@ -2425,6 +2497,8 @@ impl Downstairs {
                         self.downstairs_errors.insert(client_id, errors + 1);
 
                         // XXX We don't count read errors here.
```
jmpesp, line 2499: this comment should be in the else block?

upstairs/src/test.rs
```diff
@@ -3668,20 +3668,206 @@ mod test {
...
+                panic!("{:?} not ExtentClose()", m);
+            }
+        }
+        assert_eq!(Some(&IOState::New), rio.state.get(&0));
+        assert_eq!(Some(&IOState::New), rio.state.get(&1));
+        assert_eq!(Some(&IOState::New), rio.state.get(&2));
+    }
+    /*
+     *
+              let result = std::panic::catch_unwind(||
```
jmpesp, line 3769: commented out code :)

jmpesp (approved):
:shipit: :fire:

downstairs/src/region.rs
```diff
@@ -424,6 +489,23 @@ fn config_path<P: AsRef<Path>>(dir: P) -> PathBuf {
     out
 }
 
+/**
+ * Remove directories associated with repair expect for the replace
```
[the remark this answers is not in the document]
leftwo: Yes, will update

downstairs/src/region.rs
```diff
@@ -965,6 +1106,162 @@ impl Region {
...
+     *
+     * Let us assume we are repairing extent 012
+     *  1. Make new 012.copy dir  (extent name plus: .copy)
+     *  2. Get all extent files from source side, put in 012.copy directory
+     *  3. fsync files we just downloaded
+     *  4. Rename 012.copy dir to 012.replace dir
+     *  5. fsync extent directory ( 00/000/ where the extent files live)
+     *  6. Replace current extent 012 files with copied files of same name
+     *    from 012.replace dir
+     *  7. Remove any files in extent dir that don't exist in replacing dir
```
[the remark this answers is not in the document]
leftwo: Yes (and I'll update the comment). 
The replacement extent set of files could be:
```
012
012.db
```
And the current extent (that is old, that we are repairing) has:
```
012
012.db
012.db-wal
012.db-shm
```
We want to (after replacing `012` and `012.db`) Remove `012.db-shm` and `012.db-wal`, otherwise
we risk corrupting our extent.

downstairs/src/region.rs
```diff
@@ -965,6 +1106,162 @@ impl Region {
...
+     *  8. fsync files after copying them (new location).
+     *  9. fsync containing extent dir
+     * 10. Rename 012.replace dir to 012.completed dir.
+     * 11. fsync extent dir again (so dir rename is persisted)
+     * 12. Delete completed dir.
```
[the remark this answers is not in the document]
leftwo: It probably would not hurt to do it, as if things crash after the delete the directory could come back, but in that scenario we will just delete it again.  And, I actually have an XXX to fsync one last time, so when the
fsync changes come in (probably next repair PR) It will be done!

downstairs/src/region.rs
```diff
@@ -965,6 +1106,162 @@ impl Region {
...
+                IoError,
+                "Replace directory: {:?} already exists",
+                rd,
+            );
+        }
+
+        let extent = &self.extents[eid];
+        let copy_dir = extent.create_copy_dir(&self.dir)?;
+
+        // XXX TLS someday?  Authentication?
```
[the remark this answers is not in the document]
leftwo: Probably more than one follow up issue.

downstairs/src/region.rs
```diff
@@ -965,6 +1106,162 @@ impl Region {
...
+
+        println!("Found repair files: {:?}", repair_files);
+
+        // The repair file list should always contain the extent data
+        // file itself, and the .db file (metadata) for that extent.
+        // Missing these means the repair will not succeed.
+        let filename = extent_file_name(eid as u32, None);
+        if !repair_files.contains(&filename) {
+            // XXX Panic now, but this will eventually bail and abort
+            // the repair and let the upper layers handle it.
```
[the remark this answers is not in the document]
leftwo: Yes, sort of.  For some reason the remote downstairs returned a list of files and the generic extent default filename was not one of them.  It could be because the repair source was bogus, but could also be some other failure.  Basically, this repair cannot continue and we are `\o/` throwing up our hands and giving up.  I don't expect to see it during normal operations, and it should not panic, but until I know things are all working, I want to panic as soon as we encounter unexpected behavior.

downstairs/src/region.rs
```diff
@@ -965,6 +1106,162 @@ impl Region {
...
+        // These next two are optional.
+        let filename = extent_file_name(eid as u32, Some("db-shm"));
+        if repair_files.contains(&filename) {
+            let extent_shm = extent
+                .create_copy_file(copy_dir.clone(), Some("db-shm"))
+                .unwrap();
+            let repair_stream =
+                repair_server.get_shm(eid as u32).await.unwrap();
+            save_stream_to_file(extent_shm, repair_stream.into_inner()).await?;
+        } else if repair_files.len() != 2 {
```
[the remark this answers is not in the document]
leftwo: New function created to verify the expected file list.

protocol/src/lib.rs
leftwo: It's a usize on the downstairs side (as it is an index into the vec of extents) and we have to convert it
when it arrives, so I was thinking why not just keep it usize everywhere.  So I changed it here and
in the place where we determine what extents need repair so it's just always usize

upstairs/src/lib.rs
leftwo: You just want the `git blame` on this to point to me now!
(I will move it)

upstairs/src/test.rs
leftwo: Removed.

That was my trying to figure out how to update some of the  #should_panic tests to only accept panic
on the exact place, and not anywhere in the test (like they do now).  I could not get it working so I
punted on fixing that for now, but forgot to remove my sample code.  It will be remove.

downstairs/src/region.rs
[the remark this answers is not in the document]
leftwo: I re-did this with a new function that validates the file list.

upstairs/src/test.rs
leftwo: Gone

jmpesp:
:+1: re-reviewed, :speedboat:

downstairs/src/region.rs
```diff
@@ -424,6 +489,51 @@ fn config_path<P: AsRef<Path>>(dir: P) -> PathBuf {
...
+    }
+    Ok(())
+}
+
+/**
+ * Validate a list of repair files.
+ * There are either two or four files we expect to find, any more or less
+ * and we have a bad list.
+ */
+pub fn validate_repair_files(eid: usize, files: &[String]) -> bool {
```
ahl, line 514: Do I understand this correctly that we expect the `files` list to be equal to either:

```rust
[
    extent_file_name(eid, None),
    extent_file_name(eid, Some("db"),
]
```
or 
```rust
[
    extent_file_name(eid, None),
    extent_file_name(eid, Some("db")),
    extent_file_name(eid, Some("db-shm")),
    extent_file_name(eid, Some("db-wal")),
]
```

If that's right would this make sense?

```rust
    let some = &[
        extent_file_name(eid, None),
        extent_file_name(eid, Some("db"),
    ].into_iter().collect::<BTreeSet<_>>();
    let mut all = some.clone();
    all.extend(&[
        extent_file_name(eid, Some("db-shm")),
        extent_file_name(eid, Some("db-wal")),
   ]);

    let these = files.iter().collect::<BTreeSet<_>>();

    // returns
    these == some || these == all
```
   
This may be overkill / not work at all for this use case (and my apologies), but I'd also note that this function seems fine with the case where e.g. `files` is 4x the same String.

downstairs/src/region.rs
```diff
@@ -965,6 +1134,155 @@ impl Region {
...
+     *  9. fsync containing extent dir
+     * 10. Rename 012.replace dir to 012.completed dir.
+     * 11. fsync extent dir again (so dir rename is persisted)
+     * 12. Delete completed dir.
+     * 13. fsync extent dir again (so dir rename is persisted)
+     *
+     *  This also requires the following behavior on every extent open:
+     *   A. If xxx.Copy directory found, delete it.
+     *   B. If xxx.Completed directory found, delete it.
+     *   C. If xxx.Replace dir found start at step 4 above and continue
```
ahl, line 1166: forgive me if I'm lost, but should these be lower case ("copy")?

downstairs/src/region.rs
```diff
@@ -407,14 +410,76 @@ impl Default for ExtentMeta {
 }
 
 /**
- * Produce a PathBuf that refers to the backing file for extent "number",
- * anchored under "dir".
+ * Produce the string name of the data file for a given extent number
  */
-fn extent_path<P: AsRef<Path>>(dir: P, number: u32) -> PathBuf {
+pub fn extent_file_name(number: u32, extension: Option<&str>) -> String {
```
ahl, line 415: Should extension be an `Option<&str>` or might it make sense to have this be some other enum to limit the set of inputs we might see here?

downstairs/src/region.rs
```diff
@@ -1229,12 +1569,137 @@ impl Region {
...
+
+    // Setup the original and replacement file names.
+    let mut new_file = replace_dir.clone();
+    new_file.push(extent_file_name.clone());
+
+    let mut original_file = destination_dir;
+    original_file.push(extent_file_name);
+
+    // Copy the new file (the one we copied from the source side) on top
+    // of the original file.
```
ahl, line 1599: do we do this rather than linking because we may have the file open?

downstairs/src/region.rs
```diff
@@ -1229,12 +1569,137 @@ impl Region {
...
+
+    // XXX fsync destination_dir;
+
+    // After we have all files: move the copy dir.
+    println!("Move directory  {:?} to {:?}", replace_dir, completed_dir);
+    rename(replace_dir, &completed_dir)?;
+
+    // XXX fsync the parent directory (the extent dir)
+
+    std::fs::remove_dir_all(&completed_dir)?;
```
ahl, line 1657: would we expect `completed_dir` to be empty at this point?

downstairs/src/repair.rs
```diff
@@ -0,0 +1,386 @@
...
+    eid: u32,
+}
+
+/**
+ * Stream the contents of the data file for the given extent ID
+ */
+#[endpoint {
+    method = GET,
+    path = "/extent/{eid}/data",
+    unpublished = false,
```
ahl, line 118: this is fine, but `false` is the default

downstairs/src/repair.rs
```diff
@@ -0,0 +1,386 @@
...
+pub struct Eid {
+    eid: u32,
+}
+
+/**
+ * Stream the contents of the data file for the given extent ID
+ */
+#[endpoint {
+    method = GET,
+    path = "/extent/{eid}/data",
```
ahl, line 117: Am I reading this right that these 4 endpoints are basically the same? What would you think of something like this:

```rust
    #[derive(Deserialize, JsonSchema)]
    pub enum FileType {
        #[serde(rename = "data")
        Data,
        #[serde(rename = "db")
        Database,
        #[serde(rename = "db-shm")
        DatabaseSharedMemory, // ??
        #[serde(rename = "db-wal")
        DatabaseLog,
    }

    // ...

    struct FileSpec {
        eid: u32,
        file_type: FileType,
    }

    #[endpoint {
        method = GET,
        path = "/extent/{eid}/{file_type}",
    }]
    // ...
```

downstairs/src/repair.rs
```diff
@@ -0,0 +1,386 @@
...
+        Err(HttpError::for_bad_request(None, "EBADF".to_string()))
+    } else {
+        let file = tokio::fs::File::open(&path).await.map_err(|_| {
+            HttpError::for_bad_request(None, "EBADF".to_string())
+        })?;
+
+        let file_stream = hyper_staticfile::FileBytesStream::new(file);
+        let content_type = "application/octet-stream".to_string();
+
+        Ok(Response::builder()
```
ahl, line 208: FYI I think that this will simplify this code: https://github.com/oxidecomputer/dropshot/pull/310

downstairs/src/repair.rs
```diff
@@ -0,0 +1,386 @@
...
+            .status(StatusCode::OK)
+            .header(http::header::CONTENT_TYPE, content_type)
+            .body(file_stream.into_body())?)
+    }
+}
+
+#[derive(Deserialize, Serialize, JsonSchema)]
+struct ExtentFiles {
+    files: Vec<String>,
+}
```
ahl, line 218: it's fine either way, but you could (I think??) just return `Vec<String>` directly rather than wrapping them in a struct

downstairs/src/repair.rs
```diff
@@ -0,0 +1,386 @@
...
+    let mut files = Vec::new();
+    // The data file should always exist
+    if !full_name.exists() {
+        return Err(HttpError::for_bad_request(None, "EBADF".to_string()));
+    }
+    files.push(extent_name);
+    // The db file should always exist.
+    full_name.set_extension("db");
+    if !full_name.exists() {
+        return Err(HttpError::for_bad_request(None, "EBADF".to_string()));
```
ahl, line 268: how do we distinguish a client error (i.e. bogus eid) vs server error (i.e. internal inconsistency)?

downstairs/src/repair.rs
```diff
@@ -0,0 +1,386 @@
...
+    full_name.set_extension("db-shm");
+    if full_name.exists() {
+        println!("Exists: {:?}", full_name.file_name().unwrap());
+        files.push(extent_file_name(eid, Some("db-shm")));
+    }
+    // The db-wal file may exist.
+    full_name.set_extension("db-wal");
+    if full_name.exists() {
+        files.push(extent_file_name(eid, Some("db-wal")));
+    }
```
ahl, line 282:
```suggestion
    let files = &[
        (extent_file_name(eid, None), true),
        (extent_file_name(eid, Some("db")), true),
        (extent_file_name(eid, Some("db-shm"), false),
        (extent_file_name(eid, Some("db-wal"), false),
    ].into_iter()
    .filter_map(|(file, required)| {
        if file.exists() {
            Some(Ok(file))
        } else if required {
            Some(Err(HttpError::for_bad_request(None, "EBADF".to_string())))
        } else {
            None
        }
    })
    .collect::<Result<_>>()?;
```

I don't think this is perfect, but a sketch for your consideration. It may be more confusing than simplifying, but I wanted to try it out.

ahl (commented):
This looks great! Please forgive my chattiness, but I haven't been in this code and it's all very cool.

My only lingering question is why the repair API is on a different port rather than being part of the downstairs API? Is there a big theory statement or something that could clarify?

downstairs/src/repair.rs
leftwo: Yeah, that's simpler.  I've updated it a bit and then updated a bit more now that their is an enum for `ExtentExtension`. I can probably polish it a bit more, but at some point I want to get this back under test.

leftwo (commented):
Addressed a bunch of @ahl comments.

downstairs/src/region.rs
ahl: What you said makes sense. I mean like the `link(2)` system call to create a hard link.

downstairs/src/repair.rs
```diff
@@ -109,82 +108,55 @@ pub struct Eid {
...
-    path = "/extent/{eid}/db-shm",
-    unpublished = false,
+    path = "/newextent/{eid}/{fileType}",
 }]
-async fn get_shm(
+async fn get_extent_file(
     rqctx: Arc<RequestContext<FileServerContext>>,
-    path: Path<Eid>,
+    path: Path<FileSpec>,
 ) -> Result<Response<Body>, HttpError> {
```
ahl, line 138: this seems better; agreed?

downstairs/src/region.rs
```diff
@@ -753,12 +785,13 @@ impl Extent {
     fn create_copy_file(
         &self,
         mut copy_dir: PathBuf,
-        extension: Option<&str>,
+        extension: Option<ExtentExtension>,
```
ahl, line 788: Alternatively you could have the "no extension" variant in `ExtentExtension` or maybe just use `FileType` directly.

downstairs/src/region.rs
```diff
@@ -1586,7 +1625,7 @@ pub fn move_replacement_extent<P: AsRef<Path>>(
     assert!(Path::new(&replace_dir).exists());
     assert!(!Path::new(&completed_dir).exists());
 
-    println!("Move stuff from {:?} in {:?}", replace_dir, destination_dir,);
+    println!("Copy files from {:?} in {:?}", replace_dir, destination_dir,);
```
ahl, line 1628: haha.. seems like an improved message ;-)

openapi/downstairs-repair.json
```diff
@@ -5,39 +5,11 @@
...
-        }
-      }
-    },
-    "/extent/{eid}/db": {
+    "/extent/{eid}/files": {
       "get": {
-        "summary": "Stream the contents of the metadata .db file for the given extent ID",
-        "operationId": "get_db",
+        "summary": "For a given extent, return a vec of strings representing the names of",
+        "description": "the files that exist for that extent.",
```
ahl, line 11: Just note that the first line  is presumed to be the terse summary. It's not a big deal for this kind of internal API, but you might want to change the comment to something like:

```rust
/**
 * Get the list of files related to an extent.
 *
 * For a given extent, return a set of strings representing the names of
 * the files that exist for that extent.
 */
```

ahl (commented):
looks great!

leftwo:
> 

Just realized I did not answer this.
The repair and the "main downstairs" run on different ports because the repair data transfer goes between two downstairs and not through the upstairs.  While the upstairs is running (and connected to all downstairs) it will issue internal commands to tell a downstairs that it needs to repair, and that it should get the data from this other downstairs.  Once a downstairs has received a repair message from an upstairs, it will contact the "repair downstairs" directly, (at the repair port).

Let me figure out where I should put that as a comment somewhere so others can find it.  I have lots of smaller comments on specifics, but nothing that ties them all together

downstairs/src/region.rs
leftwo: Maybe correct even..  Or, as @jmc says. "More correct"

downstairs/src/repair.rs
leftwo: I have some changes here in the next set of changes to the repair work that will come after I push this first set, we should look at what I changed there and decide if #310 can make it better.

leftwo (commented):
Should have all of @ahl 's comments addressed now.

PR #270, opened 2022-04-07
Updates to the AWS automated tests

leftwo:
Some updates so MacOS can run the scripts.
Added a missing export for ami_id variable.
Updated the README with a little more info.

### DIFF
--- a/aws_benchmark/driver.sh
+++ b/aws_benchmark/driver.sh
@@ -25,7 +25,6 @@ program_required virtualenv
 program_required python3
 program_required pip
 program_required aws
-program_required ansible-playbook
 
 # bring up resources, run the benchmark, and clean up
 ./bring_up_resources.sh "${OS}" "${REGION}"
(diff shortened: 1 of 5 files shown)


aws_benchmark/driver.sh
```diff
@@ -24,6 +24,8 @@ program_required terraform
 program_required virtualenv
 program_required python3
 program_required pip
+program_required aws
+program_required ansible-playbook
```
jmpesp, line 28: this should be installed by the `pip install "ansible==5.0.1"` line in `bring_up_resources.sh`, not required to run driver.sh

jmpesp (changes requested):

aws_benchmark/driver.sh
jmpesp: specifically I mean ansible-playbook, not aws :)

leftwo: done in cf0b91a

jmpesp (approved):

leftwo:
@gjcolombo I'm going to push these updates, and we can iterate on it if you find more to add.  I want to build some more off it as well.

PR #271, opened 2022-04-08
Allow use of usdt probes with stable Rust

pfmooney:


leftwo (approved):
Yeah, this looks good.
I compiled and ran on the mac.  Tests still pass, and I could see my dtrace probes.

PR #273, opened 2022-04-13
Updates for Crucible client and CLI

leftwo:
Replace the crucible-client write_count with a new type that has a bit more functionality.

Write log has the ability to remember a set write counts from a specific time.  This allows
us to have a range of write counts that a given block could have.

This will allow creation of tests that will verify some failure/recovery situations still have
data that makes sense in terms of what could be possible and compliant with what POSIX
will let us return as valid data after writes with no flush.

Added a commit cli command to force the write log to do a commit.
Added a bunch more tests around verifying that the write log is
doing the proper things in different situations.

Renamed some functions to better reflect their actions.

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -281,8 +281,6 @@ impl WriteLog {
     // want to record the current write log values as a minimum of what
     // we expect the counters to be.
     pub fn commit(&mut self) {
-        // The commit needs to walk the vec and make sure any minimum
-        // values are not lost
         self.count_min = self.count_cur.clone();
     }
 
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -149,7 +149,7 @@ fn history_file<P: AsRef<Path>>(file: P) -> PathBuf {
 
 /*
  * All the tests need this basic info about the region.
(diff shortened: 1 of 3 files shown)


client/src/main.rs
```diff
@@ -196,52 +195,193 @@ fn get_region_info(guest: &Arc<Guest>) -> Result<RegionInfo, CrucibleError> {
...
+    fn get_seed(&self, index: usize) -> u8 {
+        (self.count_cur[index] % 250) as u8
+    }
+
+    // This is called before a test where we expect to be recovering and we
+    // want to record the current write log values as a minimum of what
+    // we expect the counters to be.
+    pub fn commit(&mut self) {
+        // The commit needs to walk the vec and make sure any minimum
+        // values are not lost
```
jmpesp, line 285: is this something that's still TODO?

jmpesp (approved):
Very cool!

client/src/main.rs
leftwo: That comment is old, so it's a TOREMOVE.

Removing it before pushing.

PR #281, opened 2022-04-15
Tools updates, usage, pause

leftwo:
Added a better usage message to create-generic-ds.sh

Added a "pause" option to downstairs_daemon.sh.  You can now create
the file /var/tmp/ds_test/pause and the downstairs daemon will stop
any running downstairs, but not restart them while that file exists.
Removing it will restart them.  This is needed for the repair tests
where we need to have them running sometimes, restarting sometimes,
and stopped sometimes.

### DIFF
--- a/tools/downstairs_daemon.sh
+++ b/tools/downstairs_daemon.sh
@@ -19,6 +19,7 @@ trap ctrl_c INT
 function ctrl_c() {
     echo "Stopping at your request"
     rm -f "$testdir"/up
+    rm -f "$testdir"/pause
     touch "$testdir"/stop
 }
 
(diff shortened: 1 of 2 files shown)


tools/downstairs_daemon.sh
```diff
@@ -140,6 +155,11 @@ if [[ ! -f ${cds} ]]; then
     exit 1
 fi
 
+os_name=$(uname)
+if [[ "$os_name" == 'Darwin' ]]; then
+    # stupid macos needs this to avoid popup hell.
+    codesign -s - -f "$cds"
```
jmpesp, line 161: D:

jmpesp (approved):

PR #287, opened 2022-04-15
correctly set ownership during volume read

jmpesp:
A volume "owns" a block if the subvolume owns a block, not if the read
only parent does. Set or propagate ownership information accordingly.


jmpesp:
Note: Alan and I discussed this out of band and agreed that this fix should go in separate of the integration tests to avoid merge conflicts with #257. Once that goes in, the integration test work can be done on top of that.

leftwo (approved):

PR #297, opened 2022-04-22
use rayon to parallelize open_extents

jmpesp:


leftwo (approved):
This really does seem like magic.
In looking at what happens when we create or open an extent, I can't see anything obvious that would prevent parallel execution.  I am a little nervous because I don't know what
is actually going on under the covers, how things are parallelized, if there is any tuning, etc.
But, it seems to work, it's much faster, and well, let's send it over the wall and see how it works.

PR #302, opened 2022-04-27
Break github build script into four scripts, parallelize. 

leftwo:
## Break apart github buldomat steps

I took the draft work that @jclulow did and polished it.

break build-and-test.sh into four different steps, which
enables us to run three of them in parallel and enable
all the checks to complete in less time.

There is now just a build.sh which does the build and
cargo test.  At the end, it will copy binaries and some
selected scripts for future jobs to use.

test-up-encrypted.sh: just run test_up.sh encrypted

test-up-unencrypted.sh: just run test_up unencrypted

test-ds.sh: just run test_ds.sh

Updated test_up.sh and test_ds.sh to work better with buildomat


jmpesp (approved):
Cool!

.github/buildomat/jobs/build.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+#:
+#: name = "build"
+#: variety = "basic"
+#: target = "helios"
+#: rust_toolchain = "nightly"
```
jclulow, line 6: I think we should keep this in sync with `rust-toolchain.toml`, so in this case perhaps we should:

```
#: rust_toolchain = "nightly-2021-11-24"
```

jclulow (commented):
This is great!  Thanks for carrying the drafts forward.  Make sure you squash out my XXX commit messages! :P

jclulow (approved):
LGTM!

Issue #301, opened 2022-04-27
test_up output should be less


leftwo:
test_up.sh is too chatty, and given how buildomat handles output, the test should say less.

Update test_up (and other build or automated tests) to have less output to STDOUT and log to files if necessary.

leftwo:
Fixed in #304

PR #304, opened 2022-04-28
crucible client perf and test_up updates

leftwo:
Option to crucible-client perf test option to save IO times

Added an option for the crucible-client perf test to save IO times to a file in csv format.  This can (someday) allow for post processing.

Some formatting of comment strings to allow auto generated help output to look better.  Changed a bunch of u32s to usize.

Update test_up.sh to send most output to a log file instead of stdout.  This allows the test itself, and more so the test when running in buildomat to run faster.

Log test output files to /tmp with .txt extension as this is what buildomat archive will be looking for, when that change comes in later.


leftwo:
https://github.com/oxidecomputer/crucible/issues/301

client/src/main.rs
```diff
@@ -1257,7 +1304,11 @@ async fn perf_workload(
     let big_end = big_start.elapsed();
 
     guest.flush(None)?;
-    perf_summary("rwrites", count, io_depth, wtime, big_end);
+    perf_summary("rwrites", count, io_depth, wtime.clone(), big_end);
+    if wtr.is_some() {
```
leftwo, line 1308: @steveklabnik, in addition to explaining to be why `wtr: &mut Option<&mut csv::Writer<File>>` works, pointed out I can `if let ..` this part which I will do.

leftwo:
The reduced logging for test_up.sh means the "build-and-test" takes 15 minutes, down from almost 30.

PR #308, opened 2022-05-03
Enable panic when hash mismatch (or failure)

leftwo:
### Panic when a read hash does not match between downstairs

When a read response comes back from a downstairs, we store
the hash that is returned.  When another downstairs returns
the same read request, we compare the hashes to make sure
that all downstairs are returning the same values.  A failure
of this check is now cause for panic, where before it was just
printing a message.

This is part one of what we want to do if the upstairs finds
data from a read is not what it expects.  Eventually there
will be more action taken.

However, during replay, there are situations where it's possible
to have a read return a different hash (and data) between the 
downstairs.  When we are replaying, we take all jobs since the 
last flush and replay them in order. This situation could result
in the contents of a read being different on a replaying
downstairs vs. the non-replaying downstairs if a write of the 
same block came after the read.

Add a new field to the DownstairsIO struct to indicate when
an IO is a Replay IO and we can ignore a read hash mismatch
from other downstairs.

### Other tidbits with this change

Updated show_all_work to display the replay field of the 
DownstairsIO struct.

Connect up the cli show command to show the work queues.

Some new tests around making sure we panic on hash mismatch
except when under replay.
A few tests were refactored and renamed to better reflect what 
they are now testing.  Some tests were removed that were not
testing what they claimed to be testing.

test_reconnect.sh improvements:  Change the test to do an
initial fill, then run the generic test during the loop.

### DIFF
--- a/client/src/cli.rs
+++ b/client/src/cli.rs
@@ -255,8 +255,7 @@ async fn cmd_to_msg(
             fw.send(CliMessage::IsActive).await?;
         }
         CliCommand::Show => {
-            println!("No support for {:?}", cmd);
-            return Ok(());
+            fw.send(CliMessage::ShowWork).await?;
         }
         CliCommand::Wait => {
             println!("No support for {:?}", cmd);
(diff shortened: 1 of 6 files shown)


jmpesp:
> When we are replaying, we take all jobs since the last flush and replay them in order. This situation could result in the contents of a read being different on a replaying downstairs vs. the non-replaying downstairs if a write of the same block came after the read.
>
> Add a new field to the DownstairsIO struct to indicate when an IO is a Replay IO and we can ignore a read hash mismatch from other downstairs.

I'm wondering if it would make sense to instead *not* replay reads? That might simplify the code a bit here.

leftwo:
> 

Yes.. .but, it would complicate things in other areas, though I think there are situations where we don't need to replay a read.

Right now, the dependencies are super simple, and every job is done in the order it arrives.  A flush will reset the
dependency list.  A downstairs needs to satisfy a read request first before handling what comes after that read (let's say there is a write next).  If, during a replay, we skip the read, then we would need to update the dependencies for the next
IO to remove that read from its dependency list.  Also, if a read has not been ack'd yet to the guest, then even on a 
replay we need that read to happen.

In short, it's possible, but will complicate a bunch of other things.  We can and should eventually do that, but
as things "work" at the moment, I would rather get this piece in, then look at improving the dependency list, or
other ways (I have some ideas) on avoiding re-reading something we know we don't need.

Issue #321, opened 2022-05-05
test_up.sh can fail if no writes during verify test


leftwo:
The verify test section of test_up.sh will do an initial fill, then copy a region aside, do some new IO, then switch back
the copied region.  It then expects to find a difference with the dump command.  But, if the "new IO" does not include
a write, there will be no difference to find.

PR #323, opened 2022-05-06
fix the client repair test to leave things needing repair

leftwo:
It was possible for the repair test to not leave things needing repair.

Updated the test to be sure to not end on a flush, and to bias it toward writes.

Cleaned up some formatting on progress messages.


leftwo:
https://github.com/oxidecomputer/crucible/issues/321
Was the issue I filed when the CI tests hit this.

jmpesp (approved):

PR #325, opened 2022-05-09
Plug in expectorate to generate openapi docs

ahl:
Add openapi-lint linting (and fix lints)


leftwo (approved):
It all looks good.  Thanks for making the world a better place.

I tried changing the API and verified `cargo test` finds it, then verifyed
that `EXPECTORATE=overwrite` produced the changes.

PR #326, opened 2022-05-09
Add a create region perf test in buildomat

leftwo:
Run the region creation test in buildomat.

This involved creating two parallel build tracks, one for debug and one for release.
The debug (original) track does the same as before, a debug build, then runs:
 * test_up.sh encrypted
 * test_up.sh unencrypted
 * test_ds.sh

All in parallel.

The new track starts a release build, then once that is done it runs
the `dsc region-perf` test. 

Updates to the `dsc` binary:
Optionally dump a csv log of the perf test (buildomat will archive that).
The selection of regions that the `region-perf` test creates was updated.
Took out "quick" test, and made what it was the default.  Added a "long" test
that does several loops for each region size (the former default).


leftwo:
Official stamp of approval:

JMC-review

PR #329, opened 2022-05-13
Add a perf test to buildomat

leftwo:
This adds the `crucible-client perf` test to buildomat.

Improved the display output for the perf test, 80 columns, more info, etc.
Added a header column to the generated CSV file.

New test `test_perf.sh` that loops over four different extent size and
extent count values (for the same 100Gib volume size)

Added some tests to the dsc binary.  There should be more here, but
it makes sense to better flush out how dsc will evolve before adding
too much more.
Updated dsc to take block_size, extent_size, extent_count as options
to the create_and_start test.
dsc auto generated UUID is based on the port number, but allows for
port numbers > 4 digits now.  Refactored the functions a bit in dsc to
better reflect what I want this to do, which I'm also still determining.

Perf test output looks like this (ignoring the upstairs log messages):
```                               
    TEST SECONDS COUNT DPTH    IOPS    MEAN    STDV     MIN      MAX    ES    EC
 rwrites  168.79 40000    1  236.98 0.00422 0.07472 0.00112  8.25378 32768   800
  rreads   56.54 40000    1  707.46 0.00141 0.00118 0.00012  0.01906 32768   800
 rwrites  118.69 40000    1  337.01 0.00297 0.00759 0.00105  0.85717 32768   800
  rreads   58.30 40000    1  686.08 0.00146 0.00118 0.00012  0.02447 32768   800
  ```

Still trying to decide how many loops the test should do itself.  That will
probably become another option to pass to the test.

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -686,14 +686,16 @@ fn main() -> Result<()> {
                 "ES",
                 "EC",
             );
-            runtime.block_on(perf_workload(
-                &guest,
-                &mut region_info,
-                &mut opt_wtr,
-                ops,
-                io_depth,
-                io_size,
-            ))?;
+            for _ in 0..2 {
+                runtime.block_on(perf_workload(
+                    &guest,
+                    &mut region_info,
+                    &mut opt_wtr,
+                    ops,
+                    io_depth,
+                    io_size,
+                ))?;
+            }
             if opt.quit {
                 return Ok(());
             }
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -659,7 +659,7 @@ fn main() -> Result<()> {
                     .unwrap();
                 wtr.serialize((
                     "type",
-                    "total_time",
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: client/src/main.rs [hunk 1]
```diff
@@ -1272,25 +1299,35 @@ async fn dirty_workload(
...
+    es: u64,
+    ec: u64,
 ) {
     // Convert all Durations to u64 nanoseconds.
     let times = iotimes
         .iter()
         .map(|x| (x.as_secs() as u64 * 100000000) + x.subsec_nanos() as u64)
         .collect::<Vec<u64>>();
 
+    let time_in_sec =
```
jmpesp, line 1320: time_in_nsec

jmpesp (approved):
:rocket: , with a variable name change

client/src/main.rs
leftwo: done!

PR #334, opened 2022-05-19
update for uuid v1.0.0

ahl:
See https://github.com/oxidecomputer/omicron/pull/1091 for the plan. This will be updated with the omicron rev after that PR is merged.


leftwo:
Looks like fun

Issue #144, opened 2021-12-12
Model dtrace like omicron


leftwo:
Model our dtrace probes like Omicron:

https://github.com/oxidecomputer/omicron/pull/501

Specifically,
Make a `tools/dtrace` directory

Also, see below where I have copied the exact text from 501, just because it's important to have backup copies. ;)

I've been informed that there are a few conventions around DTrace probe names. I'm going to update this PR, but I wanted to record them here for the record. There may be a better place for this institutional knowledge, but somewhere except my head is better than nowhere for now.

Specifically, where possible providers and probes should have dashes in the name, rather than underscores. So we'd prefer the provider diesel-db to diesel_db. This can be done in the usdt crate by naming the provider with double-underscores, which DTrace will translate into a single dash. For example:

#[usdt::provider(provider = "diesel__db")]
mod probes {
...
}
The second convention is that the probe signaling the completion of an operation is usually called done, rather than end or complete. So the diesel-dtrace probes should be query-start and query-done, rather than query_start and query_end.

bnaecker:
Just a note. I _thought_ the renaming of `__` to `-` was done by DTrace, but it appears that's not the case. I've added support for this in a new version of the `usdt` crate, `0.3`.

PR #338, opened 2022-05-26
Updates for upstairs dtrace scripts

leftwo:
Moved the dtrace scripts into tools/dtrace
Updated the dtrace probe name from cdt to crucible_upstairs
Updated the README for both /tools and /tools/dtrace.
Updated the dtrace probes to reflect the new probe name and
fixed a place where things were incorrect.

I'm doing this here because I'm about to add more probes to the downstairs side, and 
rather than have all this churn in that PR, I thought it best to pull out the structure and
name changes first, then add the new stuff after this goes in.


leftwo:
https://github.com/oxidecomputer/crucible/issues/144

PR #339, opened 2022-05-26
Grow the abilities of the dsc program.

leftwo:
This is another gradual update of the `dsc` program that will hopefully
someday, replace all the bash scripts.  There is more work here, but honestly I felt like
this set of changes was already too big to handle and I wish I had submitted it earlier.
There are a few more features I want to implement before we have complete coverage
with what the bash scripts do now, so I'll probably have little more churn in the exact
set of dropshot endpoints we export.

dsc now supports just creating a region then exiting.  This is almost enough to replace
the `tools/create-generic-ds.sh` script, but we still need encryption.

A dropshot server has been added to dsc, and now the "start" command
will start up three downstairs and a dropshot server.  This dropshot
server will take commands from outside and apply them to one or all
of the downstairs processes.  A bunch of internal plumbing was added
to support this including a work queue and several channels to
communicate between the tasks.

Here is the block comment from the main task:

```
/// This task is the "main task" for the dsc server.  It is responsible for
/// the initial startup of all the downstairs, then to respond to commands
/// from the dropshot server (via a channel and work queue) and submit
/// those commands to the downstairs.
///
/// The dropshot server should already be running when this function
/// is called.
///
///                                       +-------------+
///  Http or API requests from outside -> |  Dropshot   |   Async
///                                       |   server    |   task
///                                       +-------------+
///                                              |
///                                           notify_*
///                                           channel
///                  +-------------+             |
///                  |  DSC main   | <-----------+          Async
///                  |    task     |                        task
///                  +---+--+--+---+
///                      |  |  |
///                     action_*x
///                      channels
///                      |  |  |
///        +-------------+  |  +-------------+
///        |                |                |
/// +------+------+  +------+------+  +------+------+
/// |  Client 0   |  |  Client 1   |  |  Client 2   |       Async
/// |             |  |             |  |             |       task
/// |------|------|  |------|------|  |------|------|
/// | Downstairs  |  | Downstairs  |  | Downstairs  |       Spawned
/// |             |  |             |  |             |       process
/// +-------------+  +-------------+  +-------------+
///
/// Each client task spawns a process for their downstairs
/// and watches that process as well as the action channel
/// for changes requested by the main task.
///
/// When a command is received from the dropshot server, it puts that
/// command on a queue, then notifies the main task that new work has
/// arrived.  When the main task receives a notification, it walks
/// the queue and performs all jobs it finds.
```
Cleaned up the names of some things, now that I know better what I'm
trying to do.

Added tools/test_dsc.sh to test the dropshot endpoints, but I believe
this will be short lived as coming next will be a real API user as
well as a test using the API and not curl.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -539,7 +539,7 @@ dependencies = [
  "toml",
  "tracing",
  "usdt",
- "uuid",
+ "uuid 1.0.0",
  "version_check",
 ]
 
@@ -566,7 +566,7 @@ dependencies = [
  "structopt",
  "subprocess",
  "tokio",
- "uuid",
+ "uuid 1.0.0",
 ]
 
 [[package]]
@@ -610,7 +610,7 @@ dependencies = [
  "tokio",
  "tokio-util 0.7.0",
  "toml",
- "uuid",
+ "uuid 1.0.0",
 ]
 
 [[package]]
@@ -627,7 +627,7 @@ dependencies = [
  "tokio-rustls",
  "toml",
  "twox-hash",
- "uuid",
+ "uuid 1.0.0",
 ]
 
 [[package]]
@@ -692,7 +692,7 @@ dependencies = [
  "tracing",
  "tracing-opentelemetry",
  "tracing-subscriber",
- "uuid",
+ "uuid 1.0.0",
  "version_check",
 ]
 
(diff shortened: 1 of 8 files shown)


leftwo:
`./tools/test_perf.sh` and the github `test-region-create` already make use of `dsc`

jmpesp (approved):

PR #348, opened 2022-06-11
 Replace structopt with clap

luqmana:
Structopt has been in [maintenance mode](https://docs.rs/structopt/latest/structopt/#maintenance) with basically all its features in clap. We already use clap in some places so might as well consolidate and prune some deps.


leftwo (approved):
I've been meaning to do this for a while now, thanks!

PR #356, opened 2022-06-13
Upgrade to clap 3.2 and fix deprecation warnings.

luqmana:
Turns out clap 3.2 came out today and with it a bunch of deprecations.

### DIFF
--- a/downstairs/src/main.rs
+++ b/downstairs/src/main.rs
@@ -60,7 +60,12 @@ enum Args {
         #[clap(short, long, name = "UUID", action)]
         uuid: Uuid,
 
-        #[clap(long, default_value = "false", action)]
+        #[clap(
+            long,
+            default_value = "false",
+            default_missing_value = "true",
+            action(clap::ArgAction::Set)
+        )]
         encrypted: bool,
     },
     /*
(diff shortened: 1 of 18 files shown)


downstairs/src/main.rs
```diff
@@ -60,7 +60,12 @@ enum Args {
         #[clap(short, long, name = "UUID", action)]
         uuid: Uuid,
 
-        #[clap(long, default_value = "false", action)]
+        #[clap(
+            long,
+            default_value = "false",
+            default_missing_value = "true",
+            action(clap::ArgAction::Set)
```
luqmana, line 67: So, I had to change how this argument was handled slightly to keep the tests working. We're explicitly passing a value to the argument whereas by default clap flags don't accept values. I opted to not change the test but I can change either

https://github.com/oxidecomputer/crucible/blob/8ea317a9dfee65d7c018a1fda89fdec63a560ef3/tools/test_up.sh#L83-L84

leftwo: Ug, I'm  confused by what clap wants here.
The default value is false, but if the default is missing, then it is true?

luqmana: Yup, exactly:

| Arg | Effect |
| ----- | ------- |
| [no flag] | false |
| --encrypted=false | false |
| --encrypted=true | true |
| --encrypted | true |

Looking again, I guess the last column there is new behaviour I added. I can remove or keep that if you'd like

leftwo: I don't like how Clap has named these things, but I understand it now.  thanks!

leftwo (approved):

PR #358, opened 2022-06-20
Everything now requires an id!

jmpesp:
During construction, volumes now require an id to be set.

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -453,6 +453,7 @@ fn main() -> Result<()> {
         control: opt.control,
         metric_collect: Some(opt.metric_collect),
         metric_register: Some(opt.metric_register),
+        ..Default::default()
     };
 
     /*
--- a/hammer/src/main.rs
+++ b/hammer/src/main.rs
@@ -89,6 +89,7 @@ fn main() -> Result<()> {
         control: opt.control,
         metric_collect: None,
         metric_register: None,
+        ..Default::default()
     };
     let mut generation_number = opt.gen;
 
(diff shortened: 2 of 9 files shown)


leftwo (approved):
Given we have a Volume that can consist of:

RO parent,  SubVolume.

Is it true that  each RO parent and SubVolume all have their own unique UUID?

jmpesp:
With this commit, any part of the topology that requires a unique UUID from Nexus needs to have a Volume. For example:

```
volume id1 (from nexus)
  subvolumes:
    volume id2 (from nexus)
      guest
        upstairs (with random uuid)
    volume id3 (from nexus)
      guest
        upstairs (with random uuid)
  read-only parent:
    volume id 4 (from nexus)
      url backend (with random uuid)
```

The Upstairs still uses a [random uuid](https://github.com/oxidecomputer/crucible/blob/6dac857b553203f8ab938b0acdbbed5bc1d1aa89/upstairs/src/lib.rs#L3224), and so does [file based](https://github.com/oxidecomputer/crucible/blob/6dac857b553203f8ab938b0acdbbed5bc1d1aa89/upstairs/src/block_io.rs#L27) and [url based](https://github.com/oxidecomputer/crucible/blob/6dac857b553203f8ab938b0acdbbed5bc1d1aa89/upstairs/src/block_io.rs#L144) block IO.

If we add uuid (and generation? even though that gets passed through activation?) to CrucibleOpts then Nexus could pass those in as part of the volume construction and it would eventually propagate to `Upstairs::new`. Similarly, more ID specification could be added for file and block (and anywhere that makes a random uuid really).

Then we could trim out the extra volumes:

```
volume id1
  subvolumes:
    guest
      upstairs id2
    guest
      upstairs id3
  read-only parent:
    url backend id4
```

jmpesp:
I'm going to push more commits to this PR to remove the remaining random UUID generation.

PR #361, opened 2022-06-24
remove unnecessary assert in reopen_all_extents

jmpesp:
reopen_all_extents asserts !self.read_only but so does reopen_extent,
and the assert in reopen_all_extents is causing a panic when creating a
volume with a read-only parent that points to a read-only downstairs.

add a integration test to ensure that we can construct these volumes.


leftwo (approved):
I still like d7a01b55-93a6-c873-b341-d03c93385e97, but this is good.

PR #362, opened 2022-06-25
Metric update for the Upstairs

leftwo:
This changes the way the upstairs handles metrics.  The new way is to
provide up_main() with a ProducerRegister handle that it can use to
register whatever stats it so desires.  The caller/user of up_main() is
responsible for creating a dropshot endpoint to serve the collection
requests coming from Oximeter.

Updated the client test to include a dropshot server and register with
Nexus if requested.

Removed the metric_collect and metric_register fields from CrucibleOpts
as they are now obsolete.  This required updating a few other files as well.

The Volume::construct() now also takes a ProducerRegister that it will
eventually pass to up_main().

This is needed before changes in https://github.com/oxidecomputer/propolis/pull/147 can go in.


client/src/stats.rs
```diff
@@ -0,0 +1,60 @@
...
+    };
+
+    let config = Config {
+        server_info,
+        registration_address,
+        dropshot_config,
+        logging_config,
+    };
+
+    match Server::start(&config).await {
```
jmpesp, line 48: this doesn't appear to match the comment `return the not yet started Server`, maybe that should be changed to "not yet running server"?

tools/show_ox_propolis.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+if [[ -n $1 ]]; then
+    UUID=$1
+else
+    # Use this hard coded UUID, which is probably not what you want.
+    UUID=2effac35-a8ba-45ae-966d-4487e0df0ad6
```
jmpesp, line 7: should be `d7a01b55-93a6-c873-b341-d03c93385e97`

tools/show_ox_upstairs.sh
```diff
@@ -1,9 +1,18 @@
 #!/bin/bash
 
-UUID=4f0d2ac9-dde5-44cf-9011-12add9cd975a
-# Show me the stats for a hard coded upstairs UUID
+if [[ -n $1 ]]; then
+    UUID=$1
+else
+    # Use this hard coded UUID, which is probably not what you want.
+    UUID=2effac35-a8ba-45ae-966d-4487e0df0ad6
```
jmpesp, line 7: `d7a01b55-93a6-c873-b341-d03c93385e97`

upstairs/src/volume.rs
```diff
@@ -672,7 +676,10 @@ pub enum VolumeConstructionRequest {
 }
 
 impl Volume {
-    pub fn construct(request: VolumeConstructionRequest) -> Result<Volume> {
+    pub fn construct(
+        request: VolumeConstructionRequest,
+        producer_registry: Arc<Mutex<Option<ProducerRegistry>>>,
```
jmpesp, line 681: this should be `Option<Arc<Mutex<ProducerRegistry>>>` I think - it someone wants to pass None, it doesn't have to be wrapped in an arc and mutex.

jmpesp (approved):
Looks good!

client/src/main.rs
```diff
@@ -487,7 +492,38 @@ fn main() -> Result<()> {
...
     let guest = Arc::new(Guest::new());
 
-    runtime.spawn(up_main(crucible_opts, guest.clone()));
+    let pr;
+    if opt.metrics {
+        // If metrics are desired, we create and register the server
+        // first. Once we have the server, we clone the ProducerRegister
+        // so we can pass that on to the upstairs.
+        // Finally, spin out a task with the server to provide the endpoint
+        // so metrics can e collected by Oximeter.
```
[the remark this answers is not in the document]
leftwo: Done

client/src/stats.rs
```diff
@@ -0,0 +1,60 @@
+// Copyright 2022 Oxide Computer Company
+use anyhow::{bail, Result};
+use dropshot::{ConfigDropshot, ConfigLogging, ConfigLoggingLevel};
+use omicron_common::api::internal::nexus::ProducerEndpoint;
+use oximeter_producer::{Config, Server};
+use std::net::SocketAddr;
+use uuid::Uuid;
+
+pub fn client_ox() -> Result<()> {
```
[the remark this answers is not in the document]
leftwo: Never called.  This was debug code,  removed it.

client/src/stats.rs
```diff
@@ -0,0 +1,60 @@
...
+    registration_address: SocketAddr,
+) -> Result<Server> {
+    println!(
+        "Attempt to register {:?} with Nexus/Oximeter at {:?}",
+        my_address, registration_address
+    );
+
+    let dropshot_config = ConfigDropshot {
+        bind_address: my_address,
+        request_body_max_bytes: 2048,
```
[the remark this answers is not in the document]
leftwo: The current size of UpCountStats is 160, so yeah, I think this should be okay for a while.  Assuming that is what the request body size is for.

client/src/stats.rs
leftwo: Yes, not yet running is "more correct".

tools/show_ox_propolis.sh
leftwo: Removing the default UUID, so we don't have to argue.

tools/show_ox_upstairs.sh
leftwo: Removing the default UUID, so we don't have to argue.

upstairs/src/volume.rs
leftwo: I tried to have it just an Option, but I'm passing the `ProducerRegistry` around on the propolis side and I was not able to get it to work without Arc Mutex.  @luqmana had the same comment on the propolis side PR, and I'm going to try from the other side to see if I can detangle this and if so, I'll come back here and change it as well.

client/src/stats.rs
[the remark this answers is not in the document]
bnaecker: This has to be enough for any JSON-serialized samples that get splatted into the body when `oximeter` hits the collection-endpoint. If 160 is the byte-count of a JSON-serialized sample, then that _should_ be OK, assuming you're not producing them very frequently (i.e., each response contains fewer than ~12 of them).

PR #365, opened 2022-07-02
Make producer_registry an Option, not Arc Mutex

leftwo:
Ok, so I figured out how to address the comment from PR #362
After discussions and hints from @luqmana, I now have a propolis side version
where produce_metric is an `Option` instead of `Arc<Mutex<Option>>`

Having these changes checked back in allows me to point propolis and omicron at a git rev and should
allow for archives to be built.

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -514,12 +514,10 @@ fn main() -> Result<()> {
         {
             Err(e) => {
                 println!("Failed to register with Oximeter {:?}", e);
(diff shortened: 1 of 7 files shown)


leftwo:
@jmpesp, since you already approved my last PR where you had suggested this change, I'm going to be presumptuous and apply that approval here and merge this to main.  Then I can point propolis (and oximeter) at the main commit, as they will both need this update.

jmpesp:
:+1:

Issue #366, opened 2022-07-12
Need IO stats at the volume level


leftwo:
We have dtrace probes and metrics at the Guest level, but
we also need probes and metrics at the Volume level as well.
Some IOs for a Volume may go to a RO parent, and we don't
record or track how long those take.

PR #376, opened 2022-07-20
More dtrace probes at the volume layer

leftwo:
**DTrace probes at the volume layer**

Synopsis updated 8/24/22, now UUIDs are in the probes

This adds additional probes at the crucible volume layer for read/write/write_unwritten/flush operations.
In addition, a probe has been added at the ReqwestBlockIO read call, to allow for timing
of read only parent IOs apart from the generic volume layer IO measurement.

A few new dtrace example scripts were added to the dtrace directory as well.

Here is some sample output from the volume layer:
```
alan@atrium:prescrub$ pfexec dtrace -Z -s tools/dtrace/perfvol.d 
dtrace: script 'tools/dtrace/perfvol.d' matched 19 probes
CPU     ID                    FUNCTION:NAME
 12  84104                         :tick-5s 

 12  84104                         :tick-5s 
  a416a597-5ec5-417f-a913-e2ee78bff1dc                volume-write-done                                 
           value  ------------- Distribution ------------- count    
          524288 |                                         0        
         1048576 |@@@                                      2        
         2097152 |@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@     27       
         4194304 |@                                        1        
         8388608 |                                         0        

  a416a597-5ec5-417f-a913-e2ee78bff1dc                volume-flush-done                                 
           value  ------------- Distribution ------------- count    
          131072 |                                         0        
          262144 |@@@                                      1        
          524288 |@@@@@@@@@@@                              4        
         1048576 |@@@                                      1        
         2097152 |                                         0        
         4194304 |@@@@@@@@@                                3        
         8388608 |@@@@@@@@@                                3        
        16777216 |@@@@@@                                   2        
        33554432 |                                         0        

  a416a597-5ec5-417f-a913-e2ee78bff1dc                volume-read-done                                  
           value  ------------- Distribution ------------- count    
          262144 |                                         0        
          524288 |@                                        2        
         1048576 |@@@@@@                                   9        
         2097152 |@@@@@@@@@@@@@@@                          21       
         4194304 |@@@@@@@@@@@@@@                           20       
         8388608 |@@@                                      4        
        16777216 |   
```

This is: https://github.com/oxidecomputer/crucible/issues/366


jmpesp:
This will end up double counting for deeper topologies, like a volume composed of volumes (composed of...).

To disambiguate, I think we'll need to record UUID too?

leftwo:
> This will end up double counting for deeper topologies, like a volume composed of volumes (composed of...).
> 
> To disambiguate, I think we'll need to record UUID too?

Yes, but I don't believe we currently have a unique UUID for each volume layer, do we?
I believe each layer needs a UUID from Nexus, right?

jmpesp:
> Yes, but I don't believe we currently have a unique UUID for each volume layer, do we?
> I believe each layer needs a UUID from Nexus, right?

Yeah, we will need unique UUIDs at each layer for this to work. Nexus could do this for us, but it currently reuses UUIDs in a few scenarios.

Based on today's storage huddle, perhaps we need to separate the UUID of the connection (maybe sending UUID as part of the activation along with gen number) with the UUID that the struct contains (which could be used here or in log messages to disambiguate the different layers). Or we go back to unique UUID everywhere and figure out how to tie them together afterwards?

jmpesp:
> Yeah, we will need unique UUIDs at each layer for this to work. Nexus could do this for us, but it currently reuses UUIDs in a few scenarios.

This is no longer true, and I think this PR needs to change to use UUIDs in the probe names. Using a u32 count is confusing and doesn't tell us where the trace came from, just that it's unique. With UUID, we'll be able to look at the volume construction request in Nexus and see where the trace is measuring.

leftwo:
> > Yeah, we will need unique UUIDs at each layer for this to work. Nexus could do this for us, but it currently reuses UUIDs in a few scenarios.
> 
> This is no longer true, and I think this PR needs to change to use UUIDs in the probe names. Using a u32 count is confusing and doesn't tell us where the trace came from, just that it's unique. With UUID, we'll be able to look at the volume construction request in Nexus and see where the trace is measuring.

The u32 count is just to tell IOs apart from each other, so we still need that to exist.  However, you are correct in that the count does not identify which volume we are hitting the trace for.  Adding a 2nd field to the probe so that it also includes the UUID should allow us to both know which volume a trace is for, and to tell the IOs to that volume apart from one another.

leftwo:
This is now re-ready for review.  UUIDs are now part of the volume probes.

jmpesp (approved):

leftwo:
Thanks to help from @ahl

PR #377, opened 2022-07-22
Multiple activations to read-only downstairs

jmpesp:
In order to support snapshots, add support for a downstairs booted in
read-only mode to accept multiple read-only activations. This type of
activation only supports reads, and will create one Work struct for each
Upstairs that connects to it. Note that in order for this to not mix job
ids, *each Upstairs must have a different UUID when connecting*. This is
Nexus' responsibility.

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -466,6 +466,7 @@ fn main() -> Result<()> {
         key_pem: opt.key_pem,
         root_cert_pem: opt.root_cert_pem,
         control: opt.control,
+        read_only: false,
     };
 
     /*
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -1138,7 +1138,15 @@ impl Downstairs {
     ) -> Result<()> {
         // The Upstairs will send Flushes periodically, even in read only mode
         // we have to accept them.
-        if self.read_only && matches!(work, IOop::Write { dependencies: _, writes: _ }) {
+        if self.read_only
+            && matches!(
+                work,
+                IOop::Write {
+                    dependencies: _,
+                    writes: _
+                }
+            )
+        {
             eprintln!("read-only but received work {:?}", work);
             bail!(CrucibleError::ReadOnlyMismatch);
         }
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -917,8 +917,8 @@ where
             _ = &mut dw_task => {
                 bail!("do_work_task task has ended");
             }
-            _ = &mut pf_task => {
-                bail!("pf task ended");
+            e = &mut pf_task => {
+                bail!("pf task ended: {:?}", e);
             }
             /*
              * If we have set "lossy", then we need to check every now and
@@ -1136,7 +1136,10 @@ impl Downstairs {
         ds_id: u64,
         work: IOop,
     ) -> Result<()> {
-        if self.read_only && !work.read_only() {
+        // The Upstairs will send Flushes periodically, even in read only mode
+        // we have to accept them.
+        if self.read_only && matches!(work, IOop::Write { dependencies: _, writes: _ }) {
+            eprintln!("read-only but received work {:?}", work);
             bail!(CrucibleError::ReadOnlyMismatch);
         }
 
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -265,55 +265,60 @@ pub fn downstairs_import<P: AsRef<Path> + std::fmt::Debug>(
  * Debug function to dump the work list.
  */
 async fn _show_work(ds: &Downstairs) {
-    println!("Active Upstairs UUID: {:?}", ds.active_upstairs());
-    let work = ds.work.lock().await;
+    let active_upstairs_uuids = ds._active_upstairs();
+    println!("Active Upstairs {:?}", active_upstairs_uuids);
 
-    let mut kvec: Vec<u64> = work.active.keys().cloned().collect::<Vec<u64>>();
+    for uuid in active_upstairs_uuids {
+        let work = ds.work_lock(uuid).await.unwrap();
 
-    if kvec.is_empty() {
-        println!("Crucible Downstairs work queue:  Empty");
-    } else {
-        println!("Crucible Downstairs work queue:");
-        kvec.sort_unstable();
-        for id in kvec.iter() {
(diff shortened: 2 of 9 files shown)


jmpesp:
Note: we discussed the read-only HTTP server in the storage meeting this week, and I implemented it - but ran into several problems, some of which were:

- this is a new service, and would have to be managed either by Nexus or the Crucible Agent
- the volume construction request had to be sent to this new service, meaning we're passing keys around
- what happens when this new service crashes? reboots? if it's a single service, any volume referencing it will start to fail. as well, the initial volume construction request will have to be recorded in order to reconstruct it, meaning we're serializing keys somewhere.

Which lead me back to this PR. If we can make this work, then we bypass all those problems and don't create more work for ourselves.

downstairs/src/lib.rs
```diff
@@ -494,6 +499,7 @@ where
 
 async fn do_work_task<T>(
     ads: &mut Arc<Mutex<Downstairs>>,
+    upstairs_uuid: Uuid,
```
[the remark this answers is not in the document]
jmpesp: The logic should only change for read-only attachments - if the attachments are read-write, then the original logic should stay the same. I can see about an integration test for the migration case though.

downstairs/src/lib.rs
```diff
@@ -1180,92 +1259,122 @@ impl Downstairs {
         uuid: Uuid,
         tx: Arc<Sender<u64>>,
     ) -> Result<()> {
-        let mut work = self.work.lock().await;
+        if self.read_only {
+            // If in read only mode, multiple read-only activations are allowed.
+            self.active_upstairs.insert(
```
leftwo, line 1264: Does this handle collisions?  Meaning two of the same UUIDs connect?

protocol/src/lib.rs
```diff
@@ -118,13 +118,15 @@ pub enum Message {
...
-     * Forcefully tell this downstairs to promote us (an Upstairs) to
-     * active.
+     * If read only is false, forcefully tell this downstairs to promote us
+     * (an Upstairs) to active. Kick out the old Upstairs.
      *
-     * Kick out the old Upstairs.
+     * If read only is true, multiple upstairs can connect ok.
      */
-    PromoteToActive(Uuid),
+    PromoteToActive(Uuid, bool),
```
leftwo, line 129: I wonder if the read_only needs to come from the protocol?  
Meaning, the downstairs knows if it is RO or not (and will be checking itself).  The downstairs does not care if the upstairs wants to promote to active with RO or not, it has the RO/RW setting already.  All the downstairs has to do is decide if it will honor the promote as RO or RW based on information it has locally.

upstairs/src/lib.rs
```diff
@@ -684,6 +688,14 @@ where
...
                     Some(Message::Imok) => {}
+                    Some(Message::ReadOnlyMismatch(downstairs_read_only)) => {
+                        // Upstairs will never be able to connect, bail
+                        bail!(
+                            "downstairs read_only is {}, ours is {}!",
+                            downstairs_read_only,
+                            read_only,
+                        );
+                    }
                     Some(Message::YesItsMe(version)) => {
```
leftwo, line 699: I wonder if `read_only` should be decided here?  If an upstairs wants read write, and the downstairs is read only (or the other way around) we could abort at this point and not even need to wait for an activation request to show up as it will fail.

I also wonder if (in either this case or error after PromoteToActive) if the upstairs/downstairs will just get in a retry loop forever.

client/src/main.rs
```diff
@@ -466,6 +466,7 @@ fn main() -> Result<()> {
         key_pem: opt.key_pem,
         root_cert_pem: opt.root_cert_pem,
         control: opt.control,
+        read_only: false,
```
leftwo, line 469: What are your thoughts on making "read_only" a field of the `Upstairs` struct?
I wonder if making that (and perhaps even lossy) a struct member so anyone can look them up.  Then we don't have to pass that state around everywhere?

downstairs/src/lib.rs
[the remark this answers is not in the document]
leftwo: I think I believe you.. but I'll let the migration test make the final decision :)

leftwo (commented):
With the UUID stuff still settling,  How do we want to handle multiple RO upstairs and UUIDs and Gens? 
I don't think we can share the same UUIDs between two different RO upstairs, as the tasks that run in the downstairs won't be able to tell the jobs from the upstairs apart, right?

I also think there are future opportunities here to make things faster for RO regions.  For one, we don't care about dependencies any more (I think, though I would want to see if we have done things correctly for a read spanning blocks/extents on re-assembly).

protocol/src/lib.rs
jmpesp: I figured it was the same as encrypted status - an Upstairs with an encryption context expects to only connect to a downstairs with the encrypted flag set to true, otherwise there's a `Encryption expectation mismatch!`.

If we're strict about it, only Upstairs that have read_only set can connect to downstairs with read_only set, and this would cause connections to fail during volume initialization instead of later as a response to user IO.

downstairs/src/lib.rs
jmpesp: Currently this block of code does not handle collisions, and this looks a lot like buggy code. I think the fix should be storing generation number and using that to kick out the old Upstairs if a new one connects with the same UUID, meaning the comments (and RFD) have to be updated to include cases where read-only Upstairs **can** be kicked out / taken over.

upstairs/src/lib.rs
jmpesp: I agree with this, and think it should extend to encrypted / non-encrypted status too.

downstairs/src/lib.rs
```diff
@@ -1048,30 +1074,55 @@ impl Downstairs {
      * Grabbing the lock in this way should properly clear out the previously
      * active Upstairs without causing jobs to be incorrectly sent to the
      * newly active Upstairs.
+     *
+     * # Read-only mode
+     *
+     * In read-only mode, multiple Upstairs with different UUIDs can connect
+     * to the same downstairs and
```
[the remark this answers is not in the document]
jmpesp: tune in to c0f5b0a for the next exciting development!

jmpesp:
I'm close to close this, and re-open it when a few other prereq PRs are put in

PR #381, opened 2022-07-27
Scrubber in the downstairs

leftwo:
I'm breaking the scrubber into multiple parts.  This first piece covers just the downstairs side as well as introduces the new `ReadFill` and `ReadFillAck` protocol messages.   The upstairs currently has just stubs for these cases and the work to support ReadFill in the upstairs will come later.

The storage [RFD scrubber section](https://github.com/oxidecomputer/rfd/tree/0177/rfd/0177#142-block-migration-task-aka-the-scrubber) has more details, but a quick synopsis is this:

* The volume layer is where the actual scrubber will live, it constructs the `ReadFill` `IOop`
* Upstairs sends the `ReadFill` to the three downstairs in the normal IO path.
* The `ReadFill` command when it gets to the downstairs is converted to a write with a special flag.
* Downstairs decides to write or not based on each block having a checksum or not.
* Ack is sent back up.

For this all to work, unwritten blocks must have no checksum, as the presence of a checksum is how we tell if a block has been written or not.

The work in this PR just covers the downstairs handling of a `ReadFill` command.  Updates to the region and extent `write` methods now support the distinction between a `Write` and a `ReadFill`.

Added a bunch of tests to make sure writes of various sizes with various written and unwritten blocks behave as expected.

Added some DTrace probes to cover the specific read_fill path.

### DIFF
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -957,6 +957,7 @@ impl Extent {
     pub fn write(
         &self,
         writes: &[&crucible_protocol::Write],
+        read_fill: bool,
     ) -> Result<(), CrucibleError> {
         let mut inner = self.inner();
 
@@ -994,11 +995,44 @@ impl Extent {
          * 1) set the dirty bit
          * 2) gather and write all encryption contexts + hashes
          * 3) write all extent data
+         *
+         * If "read_fill" is true, then we only issue a write for a
+         * block if that block has not been written to yet.  Note that we
+         * can have a write that is "sparse" if the range of blocks it
+         * contains has a mix of written an unwritten blocks.
+         *
+         * We define a block being written to or not has if that block has
+         * a checksum or not.  So it is required that a written block has
+         * a checksum.
          */
+
+        let mut writes_to_skip: Vec<u64> = Vec::new();
+        if read_fill {
+            for write in writes {
+                if !inner.get_hashes(write.offset.value).unwrap().is_empty() {
+                    writes_to_skip.push(write.offset.value);
+                }
+            }
+        }
+        if read_fill && writes_to_skip.len() == writes.len() {
+            // For read fill, if the list of blocks to skip is the same
+            // length as the number of blocks in the write list, then we
+            // have no work to do here.
+            return Ok(());
+        }
+
+        // We know we have at least one block to write.
         inner.set_dirty()?;
 
         let tx = inner.metadb_transaction()?;
         for write in writes {
+            // Since we only add to writes_to_skip if this is a read_fill,
+            // this will be empty if read_fill is false, so we don't need to
+            // check again here for read_fill being true.
+            if writes_to_skip.contains(&write.offset.value) {
+                assert!(read_fill);
+                continue;
+            }
             if let Some(encryption_context) = &write.encryption_context {
                 Inner::tx_set_encryption_context(
                     &tx,
@@ -1011,6 +1045,11 @@ impl Extent {
         tx.commit()?;
 
         for write in writes {
+            if writes_to_skip.contains(&write.offset.value) {
+                println!("skip this write");
+                assert!(read_fill);
+                continue;
+            }
             let byte_offset = write.offset.value * self.block_size;
 
             inner.file.seek(SeekFrom::Start(byte_offset))?;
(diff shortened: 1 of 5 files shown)


protocol/src/lib.rs
```diff
@@ -197,6 +197,10 @@ pub enum Message {
     ReadRequest(Uuid, u64, Vec<u64>, Vec<ReadRequest>),
     ReadResponse(Uuid, u64, Result<Vec<ReadResponse>, CrucibleError>),
 
+    // ReadWithFill: Uuid, job id, dependencies, [ReadRequest], [Write]
+    ReadFill(Uuid, u64, Vec<u64>, Vec<Write>),
```
jmpesp, line 201: this is missing `Vec<ReadRequest>`?

downstairs/src/region.rs
```diff
@@ -957,6 +957,7 @@ impl Extent {
     pub fn write(
         &self,
         writes: &[&crucible_protocol::Write],
+        read_fill: bool,
```
jmpesp, line 960: this could be named something more specific, something like `only_write_if_unwritten`.

downstairs/src/region.rs
```diff
@@ -1011,6 +1045,11 @@ impl Extent {
         tx.commit()?;
 
         for write in writes {
+            if writes_to_skip.contains(&write.offset.value) {
+                println!("skip this write");
```
jmpesp, line 1049: debug println :)

downstairs/src/region.rs
```diff
@@ -1515,7 +1554,7 @@ impl Region {
                 };
 
             if computed_hash != write.hash {
-                println!("Failed write hash validation");
+                println!("Failed write hash validation {}", computed_hash);
```
jmpesp, line 1557: if we're putting in the hash value, let's put more context in here: at least offset, computed hash, and write.hash

downstairs/src/region.rs
```diff
@@ -1528,6 +1567,7 @@ impl Region {
         &self,
         writes: &[crucible_protocol::Write],
         job_id: u64,
+        read_fill: bool,
```
jmpesp, line 1570: same comment as before, though I guess this change also changes the probe name

upstairs/src/lib.rs
```diff
@@ -5045,6 +5087,10 @@ pub enum IOop {
         dependencies: Vec<u64>, // Jobs that must finish before this
         writes: Vec<crucible_protocol::Write>,
     },
+    ReadFill {
+        dependencies: Vec<u64>, // Jobs that must finish before this
```
jmpesp, line 5091: also missing read_requests?

jmpesp (approved):
:rocket:

downstairs/src/region.rs
leftwo: Yes!
I think the "read_fill" name is here stretching it, and I want to come up with a better name not only here but for the `IOop` name as well.

downstairs/src/region.rs
leftwo: That was really a debug additin, so I could print the hash and then hard code the result in the test.  I'll remove it.

protocol/src/lib.rs
jmpesp: ignore this - we discussed my confusion here, and this isn't a Read

upstairs/src/lib.rs
jmpesp: ignore this - we discussed my confusion here, and this isn't a Read

downstairs/src/region.rs
leftwo: `IOop::WriteUnwritten`

And 

`write_if_unwritten`

protocol/src/lib.rs
leftwo: Side channel discussion and rename to WriteUnwritten, so it's clear we don't need a `ReadRequest` here

downstairs/src/region.rs
leftwo: Gone!

upstairs/src/lib.rs
leftwo: Rename and no need for `ReadRequests` here

downstairs/src/region.rs
leftwo: I changed all the probe names as well.

PR #387, opened 2022-07-27
First pass at setup for a nightly test

leftwo:
**This is a first pass at a nightly test for crucible.**

The buildomat build-release step now creates an archive that has all the binaries and scripts required to run the `tools/test_nightly.sh` script, which is the actual test.
 
Update a few tests to run with a minimum of support scripts, moving toward using `dsc` for everything related to test downstairs creation/start/restart/stop/cleanup rather than manual scripts.  Made $BINDIR as the source of truth for locating binaries for tests.

The simple `tools/test_nightly.sh` script just calls other tests and reports success/failure for each.  We will probably want to play around with which tests are run, and how many loops we want to do for each, which may require updating those scripts to take a loop count as input.

There are plenty of things we can do better, but I want to get something out there we can build from.


jmpesp (approved):

jmpesp:
Sorry about the delay. :rocket: it

PR #388, opened 2022-07-27
dsc gets a shutdown endpoint

leftwo:
Added a new endpoint to dsc to allow remote shutdown of all downstairs and then the dsc process itself.  This opens the door to more bash script replacement.

Updated the communication channel between the dsc main task and the sub-tasks for each downstairs to allow more than one outstanding command at a time.

Updated the test_dsc.sh to test (and make use of) the /shutdown endpoint.

### DIFF
--- a/dsc/src/control.rs
+++ b/dsc/src/control.rs
@@ -173,7 +173,7 @@ async fn dsc_get_state(
  * Stop the downstairs at the given client_id
  */
 #[endpoint {
-    method = GET,
+    method = POST,
     path = "/stop/cid/{cid}",
 }]
 async fn dsc_stop(
@@ -215,7 +215,7 @@ async fn dsc_stop_all(
  * Stop a random downstairs
  */
 #[endpoint {
-    method = GET,
+    method = POST,
     path = "/stop/rand",
 }]
 async fn dsc_stop_rand(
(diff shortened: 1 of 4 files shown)


dsc/src/control.rs
```diff
@@ -356,6 +357,23 @@ async fn dsc_enable_restart_all(
     Ok(HttpResponseUpdatedNoContent())
 }
 
+/**
+ * Stop all downstairs, then stop ourselves.
+ */
+#[endpoint {
+    method = GET,
```
jmpesp, line 364: should be POST

dsc/src/main.rs
```diff
@@ -594,7 +619,7 @@ async fn start_dsc(
         println!("start ds: {:?}", ds.port);
         let txc = tx.clone();
         let dsc = ds.clone();
-        let (action_tx, action_rx) = watch::channel(Action::Start);
```
jmpesp, line 597: `Action::Start` initialized the channel before, does it now need to be sent manually? should it be sent on the next line?

jmpesp (approved):
:package:

dsc/src/control.rs
leftwo: done

dsc/src/main.rs
leftwo: The downstairs start happens automatically in the `ds_start_monitor` function.  That the `Action::Start` was being pre-loaded in the channel before was not an actual requirement for starting, and was either a requirement of the `watch::channel` or a bug in my understanding of how to setup the channel.

PR #395, opened 2022-08-01
Refactor some integration tests to remove code duplication

leftwo:
Created a test function to spawn three downstairs and return the
crucible_opts struct pre-filled with a set of default values
suitable for use in tests.

Added a test using the Guest interface.

I'm writing some tests (in another scrubber WIP) and I made these changes there,
but decided it would be better to push these test changes first, so as to not overwhelm
future reviewers of the scrubber changes.   Also, I kept making cut and paste errors
copying the "downstairs + crucible opts" changes for every test and forgetting to 
change the port numbers correctly.

### DIFF
--- a/integration_tests/src/lib.rs
+++ b/integration_tests/src/lib.rs
@@ -454,7 +454,7 @@ mod test {
         // what we expect, and a write and read work as expected
         const BLOCK_SIZE: usize = 512;
 
-        // Spin of three downstairs, build our Crucible struct.
+        // Spin off three downstairs, build our Crucible struct.
         let opts = three_downstairs(54016, 54017, 54018, false).unwrap();
 
         let guest = Arc::new(Guest::new());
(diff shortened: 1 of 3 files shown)


integration_tests/src/lib.rs
```diff
@@ -517,4 +447,50 @@ mod test {
...
         Ok(())
     }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 8)]
+    async fn integration_test_guest_downstairs() -> Result<()> {
+        // Test using the guest layer to verify a new region is
+        // what we expect, and a write and read work as expected
+        const BLOCK_SIZE: usize = 512;
+
+        // Spin of three downstairs, build our Crucible struct.
```
jmpesp, line 457: typo - `of` -> `off`

jmpesp (approved):
I'm going to avoid checking the git blame to see who didn't do this from the very beginning, but this is definitely the way it should have been done :) :ship:

integration_tests/src/lib.rs
leftwo: Fixed!

PR #397, opened 2022-08-02
Scrubber part 2, Guest support for write_unwritten.

leftwo:
**Part 2 of the scrubber work.**

This covers the middle layer of the scrubber, adding all the logic to "the upstairs" to support write_unwritten commands coming through it.  A `write_unwritten` is treated, for the most part, the same as a `write`.  Future PRs will actually make
the scrubber task (which will live at the volume layer) and make use of the functionality added here.

Add write_unwritten to SubVolume, Guest, and BlockIO traits. Since the write_unwritten is an operation that only exists for the scrubber, we don't want to expose it at the Volume layer, only SubVolumes and below.

Connected all the plumbing in the upstairs to treat IOop::WriteUnwritten mostly the same as a IOop::Write. Updated the upstairs write path to handle write_unwritten commands

Update job queue tests to cover write_unwritten much the same as write.
For many tests, this involved just putting a wrapper around them and calling them twice, once with is_write_written true, and once with it false.  This avoids copying all the code to a new test.

Update crucible-client cli with single block support for write_unwritten

Updated the integration tests for write_unwritten.

Added new dtrace probes for write_unwritten IOs


upstairs/src/block_io.rs
```diff
@@ -90,6 +90,14 @@ impl BlockIO for FileBlockIO {
         BlockReqWaiter::immediate()
     }
 
+    fn write_unwritten(
+        &self,
+        _offset: Block,
+        _data: Bytes,
+    ) -> Result<BlockReqWaiter, CrucibleError> {
+        unimplemented!();
```
jmpesp, line 98: `CrucibleError::Unsupported` instead?

upstairs/src/block_io.rs
```diff
@@ -233,6 +241,14 @@ impl BlockIO for ReqwestBlockIO {
         unimplemented!();
     }
 
+    fn write_unwritten(
+        &self,
+        _offset: Block,
+        _data: Bytes,
+    ) -> Result<BlockReqWaiter, CrucibleError> {
+        unimplemented!();
```
jmpesp, line 249: `CrucibleError::Unsupported` instead?

upstairs/src/lib.rs
```diff
@@ -2366,7 +2399,9 @@ impl Downstairs {
                 dependencies: _,
                 writes: _,
             } => {
-                panic!("WriteUnwritten not supported");
+                cdt::gw__write__unwritten__done!(|| (gw_id));
+                // XXX Should the guest see as writes?
```
jmpesp, line 2403: I don't think so - there shouldn't be a way for guests to trigger any `WriteUnwritten`, so it shouldn't show up as part of their traffic stats.

upstairs/src/lib.rs
```diff
@@ -3763,6 +3796,7 @@ impl Upstairs {
         offset: Block,
         data: Bytes,
         sender: Option<std_mpsc::Sender<Result<(), CrucibleError>>>,
+        is_write_unwritten: bool,
```
jmpesp, line 3799: I'd update the comment for this fn to explain what this flag / operation is.

upstairs/src/lib.rs
```diff
@@ -5498,6 +5550,10 @@ impl BlockOp {
             BlockOp::Write { offset: _, data } => {
                 Some(ceiling_div!(data.len(), iop_sz))
             }
+            BlockOp::WriteUnwritten { offset: _, data } => {
+                // XXX maybe don't for the scrubber?
```
jmpesp, line 5554: Probably not - this type of non-Guest traffic shouldn't count towards any of their disk limits, otherwise the background stuff we're doing competes with their real IO.

Also, we want the scrub to complete as quickly as possible :)

upstairs/src/lib.rs
```diff
@@ -6192,6 +6250,30 @@ impl Guest {
         Ok(self.send(wio))
     }
 
+    // Guest does support write_unwritten
```
jmpesp, line 6253: It should :)

upstairs/src/lib.rs
```diff
@@ -6192,6 +6250,30 @@ impl Guest {
...
+    pub fn write_unwritten(
+        &self,
+        offset: Block,
+        data: Bytes,
+    ) -> Result<BlockReqWaiter, CrucibleError> {
+        if !self.is_active() {
+            return Err(CrucibleError::UpstairsInactive);
+        }
+
```
jmpesp, line 6262: With the new read-only Upstairs flag, you'll have to check for that here and return `ModifyingReadOnlyRegion`

upstairs/src/volume.rs
```diff
@@ -473,6 +473,14 @@ impl BlockIO for Volume {
         BlockReqWaiter::immediate()
     }
 
+    fn write_unwritten(
+        &self,
+        _offset: Block,
+        _data: Bytes,
+    ) -> Result<BlockReqWaiter, CrucibleError> {
+        crucible_bail!(Unsupported, "write_unwritten not supported for volumes")
```
jmpesp, line 481: Why not?

upstairs/src/in_memory.rs
```diff
@@ -80,6 +80,16 @@ impl BlockIO for InMemoryBlockIO {
         BlockReqWaiter::immediate()
     }
 
+    fn write_unwritten(
+        &self,
+        _offset: Block,
+        _data: Bytes,
+    ) -> Result<BlockReqWaiter, CrucibleError> {
+        Err(CrucibleError::Unsupported(
+            "write_unwritten unsupported".to_string(),
```
jmpesp, line 89: I'd make this specific - "write_unwritten unsupported for InMemoryBlockIO"

jmpesp (changes requested):
This all looks good, but not :shipit: just yet - see my question about `Volume::write_unwritten` below.

upstairs/src/lib.rs
leftwo: Yeah..
I think the IO itself should be counted somewhere, and dtrace probes are fine for that.

upstairs/src/lib.rs
leftwo: I think for now, yeah.

We may reconsider this at some point.  If the user has a bunch of volumes and they are all scrubbing like crazy, we could possibly swamp the network?  Maybe?  I'll let it become an actual problem before I try to solve it though.

upstairs/src/lib.rs
leftwo: I put this comment there in part because I get lost as to where I am and this reminds me I am in the Guest Impl.

upstairs/src/volume.rs
leftwo: I don't think we want to expose `write_unwritten` to the user.  It's really something that only exists for the scrubber and the scrubber should run "inside" the volume layer.

upstairs/src/lib.rs
leftwo: Yess. well.
At this layer, the Guest does not know about the Upstairs and if it is read only or not.  The check for that would have to come at a higher level. 

We (you added?) do check in `submit_write()` for RO status, but I don't think we check anywhere in the Guest/BlockIO/DownstairsIO path for RO status for even writes yet, right?

upstairs/src/lib.rs
leftwo: done

upstairs/src/lib.rs
leftwo: change made.

upstairs/src/block_io.rs
leftwo: Updated this and the one above it (free unrelated change)

upstairs/src/block_io.rs
leftwo: Done

leftwo (commented):
A little more about exposing the write_unwritten at the volume layer.

I don't think we need to, and I'm not sure who the consumer would be.
From what I can determine, the scrubber itself will live "inside" the volume layer, meaning it will be the one responsible for reading from the RO side, then writing to the RW side.  So, only SubVolumes need to support WriteUnwritten.  And, of those, only the SubVolumes that end up going to a Guest.

Is there a case I'm not considering?

jmpesp:
> From what I can determine, the scrubber itself will live "inside" the volume layer, meaning it will be the one responsible for reading from the RO side, then writing to the RW side. 

That's my understanding too - the scrubber will only operate at the top of the volume tree. One node (the top) and one task is responsible for scrubbing.

> So, only SubVolumes need to support WriteUnwritten. And, of those, only the SubVolumes that end up going to a Guest.

This I don't agree with. If `write_unwritten` is part of BlockIO, then Volume needs to support it. This supports a future where sub volumes are in fact volumes themselves, because the topology could be anything.

Good news is I don't think `Volume::write_unwritten` will look very different from `Volume::write`. The only difference between the two will be changing the line:

```patch
-let mut waiter = sub_volume.write(sub_offset, slice.clone())?;
+let mut waiter = sub_volume.write_unwritten(sub_offset, slice.clone())?;
```

All the other logic that decides where to send the write/write_unwritten will stay the same.

> I don't think we want to expose write_unwritten to the user. It's really something that only exists for the scrubber and the scrubber should run "inside" the volume layer.

Agreed, but right now it's exposed anyway through the BlockIO trait. Maybe the way to do this is an additional private trait that every thing that implements BlockIO has to implement, like PrivateBlockIO, and move `write_unwritten` there. Problem here is private traits can't "leak" by being required by pub functions. Here's a playground link where I couldn't get something working: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=80f1fa363d5863a24ad4ff94ba449468.

I don't think we need to keep this private, I think it's alright if we expose this to users of crucible. If another crate or service (like `pantry`) requires `write_unwritten`, then they should be allowed to use it. If we restrict visibility of `write_unwritten`, then others can't use it.

If we find ourselves using `write_unwritten` in propolis, then I'd have more questions :)

upstairs/src/lib.rs
jmpesp: I thought this was inside Upstairs for some reason, you're totally right. Disregard this comment.

jmpesp:
> Well, maybe. I could the see the argument in the future, but as it stands now for Volume to support a SubVolume of a Volume, we would need to make a compute_sub_volume_lba() and lba_range_coverage() for Volume, which we don' t have now.

There's no battle here! :) SubVolume is only a thin veneer around a `dyn BlockIO` that implements those LBA related functions:

```rust
pub struct SubVolume {
    lba_range: Range<u64>,
    block_io: Arc<dyn BlockIO + Send + Sync>,
}
```

and `impl BlockIO for SubVolume` simply delegates all calls to the `self.block_io`, like

```rust
    fn read(
        &self,
        offset: Block,
        data: Buffer,
    ) -> Result<BlockReqWaiter, CrucibleError> {
        self.block_io.read(offset, data)
    }
```

and all I'm saying is that the `dyn BlockIO` could be a Volume.

> where the top Volume needed to handle scrubbing from the top level RO volume, call write_unwritten into the top level SubVolume that is a Volume?

My rough understanding of the scrub was:

```rust
impl Volume {
    pub fn scrub(&self) -> Result<()> {
        if let Some(read_only_parent) = self.read_only_parent {
            for offset in self.one_megabyte_chunks().iter() {
                let buffer = Buffer::new(1MB);

                let mut waiter = self.read(
                    offset,
                    buffer.clone(),
                )?;
                waiter.block_wait()?;

                let mut waiter = self.write_unwritten(
                    offset,
                    Bytes::from(buffer.as_vec()),
                )?;
                waiter.block_wait()?;
            }
        }

        Ok(())
    }
}
```

It doesn't need to know about SubVolumes or anything like that, `fn write` (and `fn write_unwritten`) would do the necessary LBA calculations. `write` and `write_unwritten` could even share a base implementation, adding a `is_write_unwritten` like `submit_write` has in the Upstairs.

leftwo:
> > Well, maybe. I could the see the argument in the future, but as it stands now for Volume to support a SubVolume of a Volume, we would need to make a compute_sub_volume_lba() and lba_range_coverage() for Volume, which we don' t have now.
> 
> There's no battle here! :) SubVolume is only a thin veneer around a `dyn BlockIO` that implements those LBA related functions:
> 
> ```rust
> pub struct SubVolume {
>     lba_range: Range<u64>,
>     block_io: Arc<dyn BlockIO + Send + Sync>,
> }
> ```
> 
> and `impl BlockIO for SubVolume` simply delegates all calls to the `self.block_io`, like
> 
> ```rust
>     fn read(
>         &self,
>         offset: Block,
>         data: Buffer,
>     ) -> Result<BlockReqWaiter, CrucibleError> {
>         self.block_io.read(offset, data)
>     }
> ```
> 
> and all I'm saying is that the `dyn BlockIO` could be a Volume.

Yes, following so far.  The file and URL based BlockIO impl's have a stub for `write_unwritten` too.
 
> > where the top Volume needed to handle scrubbing from the top level RO volume, call write_unwritten into the top level SubVolume that is a Volume?
> 
> My rough understanding of the scrub was:
> 
> ```rust
> impl Volume {
>     pub fn scrub(&self) -> Result<()> {
>         if let Some(read_only_parent) = self.read_only_parent {
>             for offset in self.one_megabyte_chunks().iter() {
>                 let buffer = Buffer::new(1MB);
> 
>                 let mut waiter = self.read(
>                     offset,
>                     buffer.clone(),
>                 )?;
>                 waiter.block_wait()?;
> 
>                 let mut waiter = self.write_unwritten(
>                     offset,
>                     Bytes::from(buffer.as_vec()),
>                 )?;
>                 waiter.block_wait()?;
>             }
>         }
> 
>         Ok(())
>     }
> }
> ```
> 
> It doesn't need to know about SubVolumes or anything like that, `fn write` (and `fn write_unwritten`) would do the necessary LBA calculations. `write` and `write_unwritten` could even share a base implementation, adding a `is_write_unwritten` like `submit_write` has in the Upstairs.

I think that could work, and if this is the motivation for support, I understand it.

For the actual scrubber though, I don't think I want to re-use the volume level `read` and `write_unwritten`, because the volume read would involve a read to both the RO and the RW regions, then a write_unwritten back to the RW SubVolume, though a volume level `write_unwritten` will essentially go straight to the sub-volume.  

I'm hoping I can be at the volume level, and just walk the LBA range on the RO side, then send `write_unwritten` to the SubVolume side.

jmpesp:
> For the actual scrubber though, I don't think I want to re-use the volume level read and write_unwritten, because the volume read would involve a read to both the RO and the RW regions, then a write_unwritten back to the RW SubVolume, though a volume level write_unwritten will essentially go straight to the sub-volume.

Yeah, my mistake. `self.read` should be `read_only_parent.read`, that'll do the same thing and be faster...

leftwo:
Okay, how about this. 

I'll make a `write_unwritten()` at the volume layer, but not in this PR.
I made the change locally, then realized I need to write another 20 tests for it like I did for the `Guest` layer interface.  I can do that, but let me take that work to the next PR as it's going to be another pile of changes to review.

jmpesp (approved):
:+1:

leftwo:
https://github.com/oxidecomputer/crucible/issues/404

PR #398, opened 2022-08-02
Enforce read-only Upstairs and Downstairs

jmpesp:
Move the `lossy` flag into the Upstairs, and add a `read_only` flag.
Make sure that an Upstairs in read-only mode cannot submit writes.

Supporting configuring Downstairs in read-only mode, which opens Extents
in read-only mode. Return errors if an Upstairs tries to modify on-disk
state.

Negotiation now sends read-only and encryption expectation, and the
Downstairs will reject connections where there is a mismatch.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -1337,17 +1337,16 @@ impl Downstairs {
     ) -> Result<()> {
         // The Upstairs will send Flushes periodically, even in read only mode
         // we have to accept them. But read-only should never accept writes!
-        if self.read_only
-            && matches!(
-                work,
-                IOop::Write {
-                    dependencies: _,
-                    writes: _
-                }
-            )
-        {
-            eprintln!("read-only but received work {:?}", work);
-            bail!(CrucibleError::ModifyingReadOnlyRegion);
+        if self.read_only {
+            let is_write = match work {
+                IOop::Write { .. } | IOop::WriteUnwritten { .. } => true,
+                IOop::Read { .. } | IOop::Flush { .. } => false,
+            };
+
+            if is_write {
+                eprintln!("read-only but received write {:?}", work);
+                bail!(CrucibleError::ModifyingReadOnlyRegion);
+            }
         }
 
         let dsw = DownstairsWork {
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -845,14 +845,39 @@ where
                         upstairs_id,
                         session_id,
                         gen,
+                        read_only,
+                        encrypted,
                     }) => {
                         if negotiated != 0 {
                             bail!("Received connect out of order {}",
                                 negotiated);
                         }
+
                         if version != 1 {
                             bail!("expected version 1, got {}", version);
                         }
+
+                        // Reject an Upstairs connection if there is a mismatch
+                        // of expectation
+                        {
+                            let ds = ads.lock().await;
+                            if ds.read_only != read_only {
+                                let mut fw = fw.lock().await;
+                                fw.send(Message::ReadOnlyMismatch {
+                                    expected: ds.read_only,
+                                }).await?;
+                                continue;
+                            }
+
+                            if ds.encrypted != encrypted {
+                                let mut fw = fw.lock().await;
+                                fw.send(Message::EncryptedMismatch {
+                                    expected: ds.encrypted,
+                                }).await?;
+                                continue;
+                            }
+                        }
+
                         negotiated = 1;
                         upstairs_connection = Some(UpstairsConnection {
                             upstairs_id,
@@ -1085,11 +1110,11 @@ where
     tokio::pin!(pf_task);
     loop {
         tokio::select! {
-            _ = &mut dw_task => {
-                bail!("do_work_task task has ended");
+            e = &mut dw_task => {
+                bail!("do_work_task task has ended: {:?}", e);
             }
-            _ = &mut pf_task => {
-                bail!("pf task ended");
+            e = &mut pf_task => {
+                bail!("pf task ended: {:?}", e);
             }
             /*
              * If we have set "lossy", then we need to check every now and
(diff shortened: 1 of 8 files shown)


downstairs/src/lib.rs
```diff
@@ -845,14 +845,39 @@ where
...
                             bail!("expected version 1, got {}", version);
                         }
+
+                        // Reject an Upstairs connection if there is a mismatch
+                        // of expectation
+                        {
+                            let ds = ads.lock().await;
+                            if ds.read_only != read_only {
+                                let mut fw = fw.lock().await;
+                                fw.send(Message::ReadOnlyMismatch {
```
leftwo, line 866: I think we want to bail here maybe?  Or send back the response, then bail?
If encrypted or read_only or version don't match what this downstairs expects, then there is really nothing more this downstairs can do with this connection to the upstairs.  Does that sound right?

downstairs/src/lib.rs
```diff
@@ -1300,6 +1335,21 @@ impl Downstairs {
         ds_id: u64,
         work: IOop,
     ) -> Result<()> {
+        // The Upstairs will send Flushes periodically, even in read only mode
+        // we have to accept them. But read-only should never accept writes!
+        if self.read_only
+            && matches!(
+                work,
+                IOop::Write {
```
leftwo, line 1343: `IOop::WriteUnwritten` should be treated like a `IOop::Write`

upstairs/src/lib.rs
```diff
@@ -192,6 +192,7 @@ pub struct CrucibleOpts {
     pub key_pem: Option<String>,
     pub root_cert_pem: Option<String>,
     pub control: Option<SocketAddr>,
+    pub read_only: bool,
```
leftwo, line 195: Just a thought to put in the back of your head here... What would we do if we wanted to support a RW upstairs changing it's mind at some point and wanting to go RO?

I was thinking about the situation where there was some failure, and we could no longer do writes,  could the upstairs switch to read only?   There would be a bunch of other stuff.. but.. just thinking..

upstairs/src/lib.rs
```diff
@@ -712,6 +712,22 @@ where
                         return Ok(())
                     }
                     Some(Message::Imok) => {}
+                    Some(Message::ReadOnlyMismatch { expected }) => {
+                        // Upstairs will never be able to connect, bail
+                        bail!(
```
leftwo, line 717: I think this is the right thing to do.
If this happens, does the upstairs and downstairs just go into a loop forever where they retry and abort the connection?

upstairs/src/lib.rs
```diff
@@ -3437,6 +3447,16 @@ pub struct Upstairs {
      * Upstairs stats.
      */
     stats: UpStatOuter,
+
+    /*
+     * Does this Upstairs throw random errors?
+     */
+    lossy: bool,
```
leftwo, line 3454: funny.  I thought lossy was already a field of this struct...

upstairs/src/lib.rs
```diff
@@ -3437,6 +3447,16 @@ pub struct Upstairs {
...
+
+    /*
+     * Operate in read-only mode
+     */
+    read_only: bool,
```
leftwo, line 3459: If read_only is part of the `Upstairs` struct now, how do things work with volumes that have a ReadOnly part and a RW part?  Does the Volume layer handle that, and the upstairs that serves a snapshot will be a different upstairs than what serves the RW part?  I'm just a little confused still, sorry.

leftwo (commented):
Are we ready for more RO specific tests yet?
Like should get error if RO/RW mismatch between up and down?
Maybe even a downstairs test that returns error if it receives a Write or WriteUnwritten IOop?  Maybe we already have that?  Why am I rambling?

upstairs/src/lib.rs
jmpesp: You know, I think we could support something like this. If we send back a `CrucibleError` up to Propolis, it'll bubble up through the device emulation as an IO error of some sort. Some quick searching shows that [devices do set themselves to read-only when they see errors](https://www.reddit.com/r/techsupport/comments/hx3qkw/nvme_ssds_suddenly_marked_as_readonly/). We could set something similar in the device emulation, matching on a certain error variant.

I think there's situations where we would want to switch an Upstairs to read-only. For example, one that comes to mind is if there's only one downstairs that's healthy and active. We can't accept writes because they won't hit at least two downstairs.

It's a good idea :)

downstairs/src/lib.rs
jmpesp: It does sound right - I switched to send response then bail in 9ec0a50

downstairs/src/lib.rs
jmpesp: ah, missed that - I made this a match on the type instead so that in the future if someone adds an `IOop` they will have to add a variant here. Fixed in 56f92bf

upstairs/src/lib.rs
jmpesp: Yep, it loops indefinitely.

jmpesp:
> Are we ready for more RO specific tests yet? Like should get error if RO/RW mismatch between up and down? Maybe even a downstairs test that returns error if it receives a Write or WriteUnwritten IOop? Maybe we already have that?

We don't have any integration test right now that expects or returns an error - this is partially due to the fact that the Upstairs and Downstairs will loop trying to connect in the presence of any error. We'll need to match on errors returned in `proc` Upstairs and consider some fatal for the tests not to loop indefinitely.

jmpesp:
However, added a test to see that a read-only Upstairs rejects a write in 4be47f0.

jmpesp:
> We don't have any integration test right now that expects or returns an error

To clarify - an error during negotiation, or a connection error, or a run-time problem (downstairs goes away). I did just add a test that checks on the error returned by `guest.write` :)

upstairs/src/lib.rs
leftwo: > It's a good idea :)

Crazy that all those people in the past came to the future and stole my original idea then went back to the past and implemented it.

leftwo (approved):
Pull the pin!

Issue #400, opened 2022-08-03
Make a lightweight crucible-for-clients crate


leftwo:
This is the current dependency tree. Arrows mean "depends on".

![crucible-deps (1)](https://user-images.githubusercontent.com/3258857/182639055-1b84197d-e4af-41b2-8e82-b3ef1b66a539.jpg)

The problem here is that **we are building Crucible, and all transitive dependencies of Crucible within Omicron, just to use a couple structs**.

If we clear out the dependencies on `crucible` in `omicron` (which transitively depend on `oximeter-producer`), I think we can solve this issue.

- `omicron-sled-agent` depends on `crucible::VolumeConstructionRequest` in the simulated sled agent, to emulate a server.
- `propolis-client` also depends on `crucible::VolumeConstructionRequest`

If we could pull these structures out of the main Crucible repo, we could produce the following:

![crucible-deps (2)](https://user-images.githubusercontent.com/3258857/182640628-08139acf-7296-4790-b7c8-082a60c7fcf9.jpg)

This would no longer require Omicron to build Crucible, and would fix the warning from the docs generator.

@leftwo , @jgallagher , WDYT?

_Originally posted by @smklein in https://github.com/oxidecomputer/omicron/issues/1537#issuecomment-1204062684_

leftwo:
In a test build using this branch for omicron, I verified that [crucible] no longer shows up in the build process for omicron:

```
alan@atrium:omicron$ cargo build > /tmp/fff 2>&1                                                
alan@atrium:omicron$ grep crucible /tmp/fff                                     
   Compiling crucible-client-types v0.1.0 (https://github.com/oxidecomputer/cruc
ible?branch=clientlib#6f3fe5c9)                                                 
   Compiling crucible-agent-client v0.0.1 (https://github.com/oxidecomputer/cruc
ible?branch=clientlib#6f3fe5c9)                                                 
alan@atrium:omicron$
```

On a stock build: 
```
alan@atrium:ostock$ grep crucible /tmp/fffstock 
   Compiling crucible v0.0.1 (https://github.com/oxidecomputer/crucible?rev=2add0de8489f1d4de901bfe98fc28b0a6efcc3ea#2add0de8)
   Compiling crucible-common v0.0.0 (https://github.com/oxidecomputer/crucible?rev=2add0de8489f1d4de901bfe98fc28b0a6efcc3ea#2add0de8)
[...]

PR #402, opened 2022-08-03
Move crucible-client to crutest

leftwo:
The crucible-client name is confusing for almost everyone.
This changes it to crutest.

Changed all the places where crucible-client was called.


leftwo:
This is partly in prep for https://github.com/oxidecomputer/omicron/issues/1537 and the crucible side https://github.com/oxidecomputer/crucible/issues/400

jmpesp (approved):

Issue #404, opened 2022-08-05
write_unwritten for volumes


leftwo:
Implement write_unwritten for the volume layer.

PR #405, opened 2022-08-05
Scrubber 2.5: write_unwritten at the volume layer

leftwo:
This adds support for write_unwritten at the volume layer and
a bunch of tests to make sure it works.  Abstracted the previous
write method to support both regular write and now write_unwritten.

https://github.com/oxidecomputer/crucible/issues/404


jmpesp (approved):
:rocket:

PR #406, opened 2022-08-05
Identify Downstairs work by more than Upstairs ID

jmpesp:
One more preparation step towards multiple read-only activations is
identifying Downstairs work by Upstairs ID and session ID, not just
Upstairs ID. This required sending session ID in more packets.

The downstairs will now check both Upstairs ID and session ID, and embed
that information in the jobs.

Doing this caught a bug where a rogue Upstairs could identify itself
using the same connection information as the active Upstairs, then
disconnect, leading to `clear_active` nuking the active Upstairs'
in-flight work.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -1212,10 +1212,9 @@ where
                     Some(new_upstairs_connection) => {
                         // another upstairs negotiated and went active after
                         // this one did
-                        println!("Another upstairs promoted to active, \
+                        println!("Another upstairs {:?} promoted to active, \
                             shutting down connection for {:?}",
-                            upstairs_connection,
-                        );
+                            new_upstairs_connection, upstairs_connection);
 
                         let mut fw = fw.lock().await;
                         fw.send(Message::YouAreNoLongerActive {
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -1313,7 +1313,7 @@ impl Downstairs {
     }
 
     /*
-     * Only grab the lock if the Upstairs connection matches.
+     * Only grab the lock if the UpstairsConnection matches.
      *
      * Multiple Upstairs connecting to this Downstairs will spawn multiple
      * threads that all can potentially add work to the same `active` hash
@@ -1326,18 +1326,21 @@ impl Downstairs {
      * function. Let's say `promote_to_active` and `add_work` are racing for
      * the work lock. If `add_work` wins the race it will put work into
      * `active`, then `promote_to_active` will clear it out. If
-     * `promote_to_active` wins the race, it will change the connection and
-     * signal to the previously active Upstairs that it should close this
-     * connection. If `add_work` does fire, it will fail to grab the lock
-     * because the connection is no longer active, and the connection thread
-     * should close.
+     * `promote_to_active` wins the race, it will change the Downstairs'
+     * active UpstairsConnection, and send the terminate signal to the
+     * tasks that are communicating to the previously active Upstairs
+     * (along with terminating the Downstairs tasks). If `add_work` for
+     * the previous Upstairs then does fire, it will fail to
+     * grab the lock because the UpstairsConnection is no longer active, and
+     * that `add_work` thread should close.
      *
      * Let's say `new_work` and `promote_to_active` are racing. If `new_work`
      * wins, then it will return and run those jobs in `do_work_task`.
      * However, `promote_to_active` will grab the lock and change the
-     * connection, causing `do_work` to return UpstairsInactive for the jobs
-     * that were just returned. If `promote_to_active` wins, it will
-     * clear out the jobs of the old connection.
+     * active UpstairsConnection, causing `do_work` to return
+     * UpstairsInactive for the jobs that were just returned. If
+     * `promote_to_active` wins, it will clear out the jobs of the old
+     * Upstairs.
      *
      * Grabbing the lock in this way should properly clear out the previously
      * active Upstairs without causing jobs to be incorrectly sent to the
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -359,18 +359,26 @@ async fn proc_frame<WT>(
 where
     WT: tokio::io::AsyncWrite + std::marker::Unpin + std::marker::Send,
 {
-    let upstairs_uuid = upstairs_connection.upstairs_id;
     let new_ds_id = match m {
         Message::Write {
             upstairs_id,
+            session_id,
             job_id,
             dependencies,
             writes,
         } => {
-            if upstairs_uuid != *upstairs_id {
+            if upstairs_connection.upstairs_id != *upstairs_id {
                 let mut fw = fw.lock().await;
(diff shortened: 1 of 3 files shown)


leftwo:
> One more preparation step towards multiple read-only activations is identifying Downstairs work by Upstairs ID and session ID, not just Upstairs ID. This required sending session ID in more packets.
> 
> The downstairs will now check both Upstairs ID and session ID, and embed that information in the jobs.
> 
> Doing this caught a bug where a rogue Upstairs could identify itself using the same connection information as the active Upstairs, then disconnect, leading to `clear_active` nuking the active Upstairs' in-flight work.

Tell me more about this rogue Upstairs.  Did you see this with hammer test? 
Does it require an upstairs using the same session_id?

Ah, maybe it's because we just switched from using the upstairs_id to session_id, and in the past, the upstairs_id was always, essentially, a session_id, as it was "new" for every upstairs instance.

downstairs/src/lib.rs
```diff
@@ -1275,7 +1313,7 @@ impl Downstairs {
     }
 
     /*
-     * Only grab the lock if the Upstairs UUID matches.
+     * Only grab the lock if the Upstairs connection matches.
```
leftwo, line 1316: I think "connection" here might be confusing, as it's a generic enough name and a reader might get confused.  What if instead this was `UpstairsConnection`.  This might signal to the reader that there are specific things that are checked and gives a little more detail.

downstairs/src/lib.rs
```diff
@@ -1589,11 +1635,11 @@ impl Work {
      * Return a list of downstairs request IDs that are new or have
      * been waiting for other dependencies to finish.
      */
-    fn new_work(&self, upstairs_uuid: Uuid) -> Vec<u64> {
+    fn new_work(&self, upstairs_connection: UpstairsConnection) -> Vec<u64> {
         let mut result = Vec::with_capacity(self.active.len());
 
         for job in self.active.values() {
-            if job.upstairs_uuid != upstairs_uuid {
+            if job.upstairs_connection != upstairs_connection {
```
leftwo, line 1642: This will check both `upstairs_id` and `session_id`?

downstairs/src/lib.rs
```diff
@@ -2129,11 +2179,16 @@ mod test {
         );
     }
 
-    fn add_work_rf(work: &mut Work, uuid: Uuid, ds_id: u64, deps: Vec<u64>) {
+    fn add_work_rf(
```
leftwo, line 2182: Oh, I need to fix that name.  the "rf" was leftover from `ReadFill`

downstairs/src/lib.rs
```diff
@@ -822,10 +860,10 @@ where
                                 upstairs_connection, ds.jobs().await,
                             );
 
-                            if ds.is_active(upstairs_connection.upstairs_id) {
+                            if ds.is_active(upstairs_connection) {
```
leftwo, line 863: We have the ds.lock at this point, right?  So we would have to be active to call `ds.clear.active()` here?

downstairs/src/lib.rs
```diff
@@ -1528,22 +1569,27 @@ impl Downstairs {
...
+    fn active_upstairs(&self) -> Option<UpstairsConnection> {
         self.active_upstairs
             .as_ref()
-            .map(|(upstairs_connection, _tx)| upstairs_connection.upstairs_id)
+            .map(|(upstairs_connection, _tx)| *upstairs_connection)
     }
 
-    async fn clear_active(&mut self) {
+    async fn clear_active(&mut self, upstairs_connection: UpstairsConnection) {
+        // Don't allow another connection to clear us
```
leftwo, line 1588: How can we get into `clear_active()` without being active?

The two places where I see we call `clear_active()` are both holding the downstairs lock, and both call `ds.is_active()` before calling `clear_active()`.

I know we need some better tests in this area.  I think I was able to find some strange situations by doing both failover testing between different upstairs while restarting downstairs at random.

leftwo (commented):
If we have multiple read only connections to a region, will each of those connections be associated with a different "upstairs_id"?

Meaning, at the top, there is a UUID for a volume,
under that, a UUID for the RO portion of that volume (the snapshot part).

In Nexus, there will be different UUIDs for each top level Volume, as they will correspond to different virtual disks.
But, for the RO portion of the volume, will those also get a unique UUID?  It seems like they need to be unique also?

jmpesp:
> > One more preparation step towards multiple read-only activations is identifying Downstairs work by Upstairs ID and session ID, not just Upstairs ID. This required sending session ID in more packets.
> > The downstairs will now check both Upstairs ID and session ID, and embed that information in the jobs.
> > Doing this caught a bug where a rogue Upstairs could identify itself using the same connection information as the active Upstairs, then disconnect, leading to `clear_active` nuking the active Upstairs' in-flight work.
> 
> Tell me more about this rogue Upstairs. Did you see this with hammer test? Does it require an upstairs using the same session_id?
> 
> Ah, maybe it's because we just switched from using the upstairs_id to session_id, and in the past, the upstairs_id was always, essentially, a session_id, as it was "new" for every upstairs instance.

`clear_active` is called in two places:

[proc (before negotiation is complete)](https://github.com/oxidecomputer/crucible/blob/1d67a53042f19ff7ca30dd20a04da94b7715ed7c/downstairs/src/lib.rs#L825-L829):

```rust
                            if ds.is_active(upstairs_connection.upstairs_id) {
                                println!("upstairs {:?} was previously \
                                    active, clearing", upstairs_connection);
                                ds.clear_active().await;
                            }
```

[resp_loop (after negotiation is complete)](https://github.com/oxidecomputer/crucible/blob/1d67a53042f19ff7ca30dd20a04da94b7715ed7c/downstairs/src/lib.rs#L1206-L1210):

```rust
                        if ds.is_active(upstairs_connection.upstairs_id) {
                            println!("upstairs {:?} was previously \
                                active, clearing", upstairs_connection);
                            ds.clear_active().await;
                        }
```

I think I misread one of these as not being guarded by the `is_active` check! Looking at it today, I don't think there's a bug like I thought.

`clear_active` should still take an UpstairsConnection argument - when multiple read-only downstairs are active, we need to know which one to clear.

downstairs/src/lib.rs
jmpesp: You're right, there isn't a bug here (I posted a followup comment too).

downstairs/src/lib.rs
jmpesp: Yes, you're right - there's no bug here. This Upstairs would have to be active.

downstairs/src/lib.rs
jmpesp: That's how my brain thought of this check, but actually, because UpstairsConnection automatically derives `PartialEq`, I believe this is also checking gen number.

downstairs/src/lib.rs
jmpesp: Tried to clarify and re-word in f37ab1a, let me know what you think.

leftwo (approved):
No wolves seen here.  Good to go.

PR #407, opened 2022-08-08
Creating a running snapshot should be idempotent

### LINKED ISSUE #380: Agent endpoint to launch read-only downstairs should be idempotent
`region_run_snapshot` is called from a Nexus saga and therefore needs to be idempotent for the same request. [It currently is not](https://github.com/oxidecomputer/crucible/blob/0f18733ba0136e01d686812893d7a746c1c882d1/agent/src/datafile.rs#L249-L251), so fix that :)

jmpesp:
Also, comparison of running downstairs request should check X509
information too (even though we don't current use it).

Closes #380

### DIFF
--- a/agent/src/datafile.rs
+++ b/agent/src/datafile.rs
@@ -352,7 +352,6 @@ impl DataFile {
             }
         }
 
-        // zfs delete snapshot
         let mut path = self.base_path.to_path_buf();
         path.push("regions");
         path.push(request.id.0.clone());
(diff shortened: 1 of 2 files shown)


leftwo (approved):
The one thing I can think if here is making sure all three downstairs have the same "snapshot" for a given `id`.  This means that if two work, and one fails, we throw the snapshots out and try again? 

I believe that the upstairs will complain and refuse to stop if you do have a RO region where there is a difference in downstairs, but I would like to if possible avoid such a scenario.  Breaking here is better than breaking when you try to read the snapshots(s).

Other than that, everything else here is good.

jmpesp:
> The one thing I can think if here is making sure all three downstairs have the same "snapshot" for a given id. This means that if two work, and one fails, we throw the snapshots out and try again?

I think this is a bug we currently have. Whether or not a Flush has `SnapshotDetails` set, It looks like the Upstairs will be ok with two out of three flushes returning OK. I think this has to change: if `SnapshotDetails` is set, then all three downstairs have to report OK for that Flush for it to be considered OK and acked.

During the snapshot create saga, Nexus will undo the saga if the snapshot request to Propolis fails. This will delete any snapshots created.

leftwo:
> > The one thing I can think if here is making sure all three downstairs have the same "snapshot" for a given id. This means that if two work, and one fails, we throw the snapshots out and try again?
> 
> I think this is a bug we currently have. Whether or not a Flush has `SnapshotDetails` set, It looks like the Upstairs will be ok with two out of three flushes returning OK. I think this has to change: if `SnapshotDetails` is set, then all three downstairs have to report OK for that Flush for it to be considered OK and acked.

To throw some more gas on this fire, If we require all three downstairs to answer before we can take a snapshot, it will essentially prevent taking a snapshot where one downstairs is broken.  I think there are other situations (like deactivate) where this issue is already present.

jmpesp:
I've opened https://github.com/oxidecomputer/crucible/issues/431 to track this issue.

PR #411, opened 2022-08-09
Support multiple read-only activations Downstairs

jmpesp:
Previously, the Downstairs would only accept one active Upstairs at a
time. This commit adds support Downstairs for multiple read-only
Upstairs connecting, and sending reads.

This was done by storing a Work struct per Upstairs connection, and
getting the Downstairs to select the correct one based on an
UpstairsConnection argument.

The biggest change in this commit is to `promote_to_active`, which has
different behaviour now depending on if the Downstairs is in read-only
or read-write mode. The read-write path should be functionally the same
as before this commit. Tests were also added to validate
`promote_to_active`'s new behaviour.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -3555,6 +3555,9 @@ mod test {
 
         assert_eq!(ds.active_upstairs().len(), 1);
 
+        assert!(!ds.is_active(upstairs_connection_1));
+        assert!(ds.is_active(upstairs_connection_2));
+
         Ok(())
     }
 
@@ -3595,6 +3598,9 @@ mod test {
         assert!(matches!(rx2.try_recv().unwrap_err(), TryRecvError::Empty));
 
         assert_eq!(ds.active_upstairs().len(), 1);
+
+        assert!(!ds.is_active(upstairs_connection_1));
+        assert!(ds.is_active(upstairs_connection_2));
 
         Ok(())
     }
@@ -3637,6 +3643,9 @@ mod test {
 
         assert_eq!(ds.active_upstairs().len(), 2);
 
+        assert!(ds.is_active(upstairs_connection_1));
+        assert!(ds.is_active(upstairs_connection_2));
+
         Ok(())
     }
 
@@ -3676,6 +3685,9 @@ mod test {
         assert!(matches!(rx2.try_recv().unwrap_err(), TryRecvError::Empty));
 
         assert_eq!(ds.active_upstairs().len(), 1);
+
+        assert!(!ds.is_active(upstairs_connection_1));
+        assert!(ds.is_active(upstairs_connection_2));
 
         Ok(())
     }
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -1508,14 +1508,21 @@ impl Downstairs {
         // Complete the job
         let is_flush = matches!(m, Message::FlushAck { .. });
 
-        // _ can be None if promote_to_active ran and cleared out active.
-        let _ = work.active.remove(&ds_id);
-
-        if is_flush {
-            work.last_flush = ds_id;
-            work.completed = Vec::with_capacity(32);
-        } else {
-            work.completed.push(ds_id);
+        // If upstairs_connection grabs the work lock, it is the active
+        // connection for this Upstairs UUID. The job should exist in the Work
+        // struct. If it does not, then we're in the case where the same
+        // Upstairs has reconnected and been promoted to active, meaning
+        // `work.clear()` was run. If that's the case, then do not alter the
+        // Work struct, because there's now two tasks running for the same
+        // UpstairsConnection, and we're the one that should be on the way out
+        // due to a message on the terminate_sender channel.
+        if work.active.remove(&ds_id).is_some() {
+            if is_flush {
+                work.last_flush = ds_id;
+                work.completed = Vec::with_capacity(32);
+            } else {
+                work.completed.push(ds_id);
+            }
         }
 
         Ok(())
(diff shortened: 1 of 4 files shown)


downstairs/src/lib.rs
```diff
@@ -1499,106 +1554,231 @@ impl Downstairs {
...
-            work.active = HashMap::new();
-        }
+            Ok(())
+        } else {
+            // Only one active read-write session is allowed. Kick out the
+            // currently active Upstairs session if one exists.
+            let currently_active_upstairs_uuids: Vec<Uuid> =
+                self.active_upstairs.keys().copied().collect();
+
+            match currently_active_upstairs_uuids.len() {
```
leftwo, line 1635: Similar locking question here, thought it seems like we are okay.  The action of getting
currently_active_upstairs_uuids, that is atomic?  Meaning, could two connections both call that `self.active_upstairs.keys().copied().collect();` and come back with `0`?

Actually, I think it could, as the time of check is different that the time of add below, so unless we
lock around this, I suspect it's possible for two connections to sneak in here.

downstairs/src/lib.rs
```diff
@@ -3228,4 +3431,318 @@ mod test {
...
+        ds.promote_to_active(upstairs_connection_1, tx1).await?;
+
+        assert_eq!(ds.active_upstairs().len(), 1);
+        assert!(matches!(rx1.try_recv().err().unwrap(), TryRecvError::Empty));
+        assert!(matches!(rx2.try_recv().err().unwrap(), TryRecvError::Empty));
+
+        ds.promote_to_active(upstairs_connection_2, tx2).await?;
+        assert_eq!(rx1.try_recv().unwrap(), upstairs_connection_2);
+        assert!(matches!(rx2.try_recv().unwrap_err(), TryRecvError::Empty));
+
```
leftwo, line 3542: Could we also test `is_active` in here?  Making sure `upstairs_connection_1` is not active and `upstairs_connection_2` is?

integration_tests/src/lib.rs
```diff
@@ -1432,4 +1432,79 @@ mod test {
 
         Ok(())
     }
+
+    // Test that multiple upstairs can connect to a single read-only downstairs
+    #[tokio::test(flavor = "multi_thread", worker_threads = 8)]
+    async fn integration_test_multi_read_only() -> Result<()> {
+        const BLOCK_SIZE: usize = 512;
+
+        let mut opts = three_downstairs(54040, 54041, 54042, true).unwrap();
```
leftwo: These port numbers are already in use on line 907 and can cause the test to hang.

I wish I had some better way of allocating port numbers, as I do this to myself all the time.

Also, I was trying to group the "volume" tests into one section (ending at line ~986), and having their ports start with 54xxx and the guest section of tests (which is where this is) all start with 55xxx.

leftwo (commented):
Some thoughts and locking concerns below, take a read and let me know what you think.

It may be worth making a higher level test that starts two upstairs talking to the same downstairs RO, then has the downstairs restart while the upstairs continue to send IO.  That sort of test (test_reconnect.sh) found a bunch of edge cases for me when doing the reconnect work.  Maybe make an issue for that as I need to update the dsc program a tiny bit more to allow it to do the work instead of using `downstairs_daemon.sh`.

downstairs/src/lib.rs
```diff
@@ -1168,7 +1188,7 @@ where
              * trigger once then never again.
              */
             _ = sleep_until(lossy_interval), if lossy => {
-                //_show_work(&ds).await;
+                //show_work(&ds).await;
```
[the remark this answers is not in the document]
jmpesp: done in 59de8ef

downstairs/src/lib.rs
```diff
@@ -1349,25 +1373,31 @@ impl Downstairs {
...
+
+        let active_upstairs = self.active_upstairs.get(&upstairs_uuid).unwrap();
+
+        if active_upstairs.upstairs_connection != upstairs_connection {
+            println!(
+                "{:?} cannot grab lock, does not match {:?}!",
+                upstairs_connection, active_upstairs.upstairs_connection,
+            );
+
+            bail!(CrucibleError::UpstairsInactive)
```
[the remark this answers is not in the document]
jmpesp: These should both be returning `UpstairsInactive`, as the Upstairs has to bail either way. Changed it to a print + bail in 89c2384.

downstairs/src/lib.rs
```diff
@@ -1443,8 +1491,13 @@ impl Downstairs {
...
      */
-    async fn complete_work(&mut self, ds_id: u64, m: Message) -> Result<()> {
-        let mut work = self.work.lock().await;
+    async fn complete_work(
+        &mut self,
+        upstairs_connection: UpstairsConnection,
+        ds_id: u64,
+        m: Message,
+    ) -> Result<()> {
+        let mut work = self.work_lock(upstairs_connection).await?;
```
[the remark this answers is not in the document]
jmpesp: You're right, each connection gets a unique work queue.

I tried to address this in aa2b3a5 - there is one scenario I could think of where that could be None, and I think you're right that it shouldn't touch the Work struct. Also added some tests there.

downstairs/src/lib.rs
```diff
@@ -1468,9 +1521,11 @@ impl Downstairs {
      */
     async fn complete_work_stat(
         &mut self,
+        _upstairs_connection: UpstairsConnection,
         m: &Message,
         ds_id: u64,
     ) -> Result<()> {
+        // XXX dss per upstairs connection?
```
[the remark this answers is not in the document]
jmpesp: opened https://github.com/oxidecomputer/crucible/issues/416

downstairs/src/lib.rs
jmpesp: Good idea, done in 39514d5

downstairs/src/lib.rs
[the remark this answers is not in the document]
leftwo: thanks

downstairs/src/lib.rs
jmpesp: I think we're ok here - `promote_to_active` needs a `&mut self`, meaning an exclusive reference, meaning I don't think two calls of this can race.

integration_tests/src/lib.rs
leftwo: Oh, and sorry, but there is some churn in this area with all the scrubber tests I added.

downstairs/src/lib.rs
```diff
@@ -1499,106 +1554,231 @@ impl Downstairs {
...
-         */
-        if work.active.keys().len() > 0 {
-            println!(
-                "Crucible Downstairs promoting {:?} to active, \
-                discarding {} jobs",
-                upstairs_connection,
-                work.active.keys().len()
+            // Insert a new session, overwritting the previous entry if the
+            // Upstairs UUID has an entry already.
+            self.active_upstairs.insert(
```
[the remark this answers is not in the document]
jmpesp: I think we're ok here - `promote_to_active` needs a &mut self, meaning an exclusive reference. I made `work_lock` (and others) require `&mut self` to ensure that callers need an exclusive reference in order to read from or modify `self.active_upstairs`. See: ca64e99

downstairs/src/lib.rs
```diff
@@ -1473,14 +1485,133 @@ impl Downstairs {
...
+            // UpstairsConnection, but the tasks can still be working on
+            // outdated job IDs. If that happens, `get_ready_job` will return a
+            // None, so bail early here.
+            if job.is_none() {
+                return Ok(None);
+            }
+
+            job.unwrap()
+        };
+
```
leftwo, line 1511: The old do_work() had a few asserts here:
```
    assert_eq!(job.state, WorkState::InProgress);
    assert_eq!(job_id, job.ds_id);
```
Would those still make sense?

It also had a check to be sure the deps were satisfied.  Same question for that.

leftwo: Ah, nope, nevermind.
I see this now happens in the `get_ready_job()` method.
So, why is this comment even here still?  Because I want you to see how hard I'm working, that's why.

leftwo (approved):

PR #414, opened 2022-08-11
Scrubber part 3

leftwo:
Part 3 of the scrubber.  Still not done yet, but moving things forward and (should not) break anything existing.

We don't yet run the scrubber, but the parts are all present.  More work to come to determine how big should
the scrubber's IOs be, and how do we control it from the outside.

This introduces the scrub method on the Volume layer.  Calling this will result in a reading of the read only parent
of a volume and issuing a write_unwritten for every block.

A batch of new tests to verify that a scrub of a volume will result in data being copied where unwritten and that future
write_unwrittens will not update the blocks covered by a read only parent.

I also moved volume_write_up up into the same impl that had volume write and volume read.

### DIFF
--- a/upstairs/src/volume.rs
+++ b/upstairs/src/volume.rs
@@ -12,7 +12,7 @@ pub struct Volume {
     uuid: Uuid,
 
     sub_volumes: Vec<SubVolume>,
-    read_only_parent: Mutex<Option<SubVolume>>,
+    read_only_parent: Option<SubVolume>,
 
     /*
      * Each sub volume should be the same block size (unit is bytes)
@@ -38,7 +38,7 @@ impl Volume {
         Self {
             uuid,
             sub_volumes: vec![],
-            read_only_parent: Mutex::new(None),
+            read_only_parent: None,
             block_size,
         }
     }
@@ -63,7 +63,7 @@ impl Volume {
         Ok(Self {
             uuid,
             sub_volumes: vec![sub_volume],
-            read_only_parent: Mutex::new(None),
+            read_only_parent: None,
             block_size,
         })
     }
@@ -156,14 +156,14 @@ impl Volume {
         // None when done - read block by block, don't overwrite if owned by sub
         // volume
 
-        self.read_only_parent = Mutex::new(Some(SubVolume {
+        self.read_only_parent = Some(SubVolume {
             // Read only parent LBA range always starts from 0
             lba_range: Range {
                 start: 0,
                 end: number_of_blocks,
             },
             block_io,
-        }));
+        });
 
         Ok(())
     }
@@ -237,8 +237,7 @@ impl Volume {
         length: u64,
     ) -> Option<Range<u64>> {
         // Does the read only parent apply for this range
-        let read_only_parent = self.read_only_parent.lock().unwrap();
-        if let Some(ref read_only_parent) = *read_only_parent {
+        if let Some(ref read_only_parent) = self.read_only_parent {
             read_only_parent.lba_range_coverage(start, length)
         } else {
             None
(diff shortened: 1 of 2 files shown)


jmpesp (approved):
:+1:

upstairs/src/volume.rs
```diff
@@ -12,8 +12,7 @@ pub struct Volume {
     uuid: Uuid,
 
     sub_volumes: Vec<SubVolume>,
-    read_only_parent: Option<SubVolume>, /* TODO mutex so migration task can
-                                          * remove */
+    read_only_parent: Mutex<Option<SubVolume>>,
```
jmpesp, line 15: Actually, it's also possible that we may not need this to be a Mutex, or even modifiable - the scrubber could run, signal to Nexus that it's done, then Nexus could PUT a new VolumeConstructionRequest without the read-only parent to Propolis, causing the Volume to be reinitialized without the read-only parent. That way all modification is Nexus driven, and there's no drift between what VolumeConstructionRequests are in the DB vs what are live.

leftwo: I like both the RW lock and the no Mutex at all ideas.

I think I'm going to pull the `read_only_parent: Mutex<Option<SubVolume>>>` change out of this PR and just push (after addressing comments) the scrubber part and the tests.

We can determine next steps with RW lock, Pushing it to Nexus, or some other solution.
I also want to implement the scrubber high water mark, so the volume layer can tell where the scrubber has been and reads under that point can go directly to the SubVolume.

upstairs/src/volume.rs
```diff
@@ -237,12 +237,100 @@ impl Volume {
...
+    // If a volume has a read only parent, do the work to read from
+    // it and write_unwritten to the LBA it covers.
+    pub fn scrub(&self) -> Result<(), CrucibleError> {
+        println!("Scrub check for {}", self.uuid);
+        // XXX Can we assert volume is activated?
+
+        let read_only_parent = self.read_only_parent.lock().unwrap();
+        // TODO: Before we can use the scrubber live, we have to
+        // come up with a more granular locking, as this current lock
+        // will just prevent any reads while the scrubber is working.
```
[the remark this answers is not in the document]
leftwo: I pulled the mutex stuff out for now.

upstairs/src/volume.rs
```diff
@@ -237,12 +237,100 @@ impl Volume {
...
+                let block = Block::new(offset, bs.trailing_zeros());
+                let buffer = Buffer::new(bs);
+
+                let mut waiter =
+                    read_only_parent.read(block, buffer.clone())?;
+                waiter.block_wait()?;
+
+                let mut waiter = self.write_unwritten(
+                    Block::new(offset, bs.trailing_zeros()),
+                    Bytes::from(buffer.data.lock().unwrap().clone()),
```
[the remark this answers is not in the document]
leftwo: Done in bd85f07

upstairs/src/volume.rs
```diff
@@ -243,6 +243,89 @@ impl Volume {
...
+            let bs = read_only_parent.get_block_size()? as usize;
+
+            println!("Scrub for total_size:{:?} block_size:{:?}", ts, bs);
+            let start = read_only_parent.lba_range.start;
+            let end = read_only_parent.lba_range.end;
+            println!("Scrub will go from {:?} to {:?}", start, end);
+
+            for offset in start..end {
+                let block = Block::new(offset, bs.trailing_zeros());
+                let buffer = Buffer::new(bs);
```
jmpesp, line 264: I'd put a TODO here (or open an issue) to explore larger buffer sizes.

jmpesp (approved):
LGTM

PR #417, opened 2022-08-15
More features for dsc

leftwo:
These changes mean nearly all of the ./tools/downstairs_daemon.sh script
can be replaced by dsc (though that work is not part of this change).

Dsc can now support random restart of one of the downstairs under
its control. The requester can pick a min and max time to fence
the restarts. This required a little refactor of the dsc main loop
to allow for some persistent state.

Updated dsc to support sending a single command to a running dsc
server, basically making it a client of itself. To support this,
further refactoring of dsc was required.

Added the new dsc commands to the crutest cli.

Wrote a small test function to create a temporary file instead of
creating an empty file in the current directory which was not
removed at the end of the tests.

Updated the top level README.md to make it match the current
names of things.

dsc now looks like this:
```
final:crucible$ ./target/release/dsc
dsc
A downstairs controller

USAGE:
    dsc <SUBCOMMAND>

OPTIONS:
    -h, --help    Print help information

SUBCOMMANDS:
    cmd            Send a command to a running dsc server
    create         Create a downstairs region then exit
    help           Print this message or the help of the given subcommand(s)
    region-perf    Test creation of downstairs regions
    start          Start a downstairs region set This requires the region is
                       already created, unless you include the --create option

```

The new `cmd` looks like this:
```
final:crucible$ ./target/release/dsc cmd
dsc-cmd
Send a command to a running dsc server

USAGE:
    dsc cmd [OPTIONS] <SUBCOMMAND>

OPTIONS:
    -h, --help               Print help information
        --server <SERVER>    URL location of the Crucible control server
                             [default: http://127.0.0.1:9998]

SUBCOMMANDS:
    disable-random-stop    Disable random stopping of downstairs
    disable-restart        Disable auto restart on the given downstairs
                               client ID
    disable-restart-all    Disable auto restart on all downstairs
    enable-random-max      Set the maximum random stop time (in seconds)
    enable-random-min      Set the minimum random stop time (in seconds)
    enable-random-stop     Enable random stopping of downstairs
    enable-restart         Disable restart on the given client ID
    enable-restart-all     Enable auto restart on all downstairs
    help                   Print this message or the help of the given
                               subcommand(s)
    pid                    Get PID of the given client ID
    shutdown               Shutdown all downstairs, then shutdown dsc itself
    start                  Start the downstairs at the given client ID
    start-all              Start all downstairs
    state                  Get the state of the given client ID
    stop                   Stop the downstairs at the given client ID
    stop-all               Stop all the downstairs
    stop-rand              Stop a random downstairs
```

### DIFF
--- a/dsc/src/main.rs
+++ b/dsc/src/main.rs
@@ -1227,7 +1227,7 @@ async fn region_create_test(
 }
 
 // Remove any existing dsc files/regions
-fn cleanup_region(output_dir: PathBuf, region_dir: PathBuf) -> Result<()> {
+fn cleanup(output_dir: PathBuf, region_dir: PathBuf) -> Result<()> {
     if Path::new(&output_dir).exists() {
         println!("Removing existing dsc directory {:?}", output_dir);
         std::fs::remove_dir_all(&output_dir)?;
@@ -1278,7 +1278,7 @@ fn main() -> Result<()> {
             region_dir,
         } => {
             if cleanup {
-                cleanup_region(output_dir.clone(), region_dir.clone())?;
+                crate::cleanup(output_dir.clone(), region_dir.clone())?;
             }
             let dsci =
                 DscInfo::new(ds_bin, output_dir, region_dir, notify_tx, true)?;
@@ -1315,7 +1315,7 @@ fn main() -> Result<()> {
         } => {
             // Delete any existing region if requested
             if cleanup {
-                cleanup_region(output_dir.clone(), region_dir.clone())?;
+                crate::cleanup(output_dir.clone(), region_dir.clone())?;
             } else if create && Path::new(&region_dir).exists() {
                 println!("Removing existing region {:?}", region_dir);
                 std::fs::remove_dir_all(&region_dir)?;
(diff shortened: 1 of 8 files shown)


dsc/src/main.rs
```diff
@@ -1221,14 +1349,22 @@ fn main() -> Result<()> {
 #[cfg(test)]
 mod test {
     use super::*;
-    use tempfile::tempdir;
+    use tempfile::{tempdir, NamedTempFile};
+
+    // Create a temporary file.  Close the file but return the path
+    // so it won't delete be deleted.
```
jmpesp, line 1355: typo: `delete be deleted.`

dsc/src/main.rs
```diff
@@ -1111,8 +1226,21 @@ async fn region_create_test(
     Ok(())
 }
 
+// Remove any existing dsc files/regions
+fn cleanup_region(output_dir: PathBuf, region_dir: PathBuf) -> Result<()> {
```
jmpesp, line 1230: this does more than just cleaning up the region, maybe just `cleanup`?

jmpesp (approved):
Very cool :)

Issue #419, opened 2022-08-16
Measure repair performance


leftwo:
Measure the performance of extent repair.

Roughly, the time it takes to:
1: Stop extent(s).
2: Send repair command, wait for completion.
3: Reopen extent.

A variety of extent sizes and extent counts should be used, similar to what the `./tools/test_perf.sh` test does now.

leftwo:
From tests on on 8/22/22
Extent repair summary from Folgers.
This is using three different SSDs for the downstairs region files.
An upstairs is also running on Folgers along side the three downstairs.
All traffic between downstairs (and upstairs) is localhost.
```
Extent Size  Time to repair
    16 MiB   0.12 seconds
    32 MiB   0.25 seconds
    64 MiB   0.51 seconds
   128 MiB   1.01 seconds
   192 MiB   1.52 seconds
   256 MiB   2.05 seconds
```

Extent repair time summary on Atrium.  This is the total time it takes to repair a full extent, copying from one downstairs to another, as recorded from the upstairs.
```
Extent Size  Time to repair
    16 MiB   0.2  seconds
    32 MiB   0.35 seconds
    64 MiB   0.62 seconds
   128 MiB   1.19 seconds
   192 MiB   1.78 seconds
   256 MiB   2.42 seconds
```

PR #429, opened 2022-08-23
Repair perf

leftwo:
**A region repair perf test**

This is an initial version of an extent repair performance test.  The test does the
following:

- Create a region set with three downstairs. 
- Make an entire downstairs set of extents require repair.
- Run the regular upstairs and let it perform the repair. 
- Use output from the upstairs to determine how long that repair takes, and divide 
   by the number of extents to get an average.

While it's not a perfect test, it does give us an idea of how repair performs.

To support this effort, the `dsc` tool has been updated to allow each downstairs 
to have its own unique full path to the region directory.  This allows downstairs to 
run on different physical disks if desired.  Making this change to `dsc` required 
updating a number of places to support a `Vec` instead of a single item.  There 
was a number of assumptions inside `dsc` on how the regions were setup. 
This reduces that somewhat, but there is still more we could do.

A small update to the upstairs that reports time spent in repair and computation 
of the average repair time.


leftwo:
https://github.com/oxidecomputer/crucible/issues/419

faithanalog:
This looks good to me, though i guess it needs a conflict resolution since I didn't realize I was requested as a reviewer until i was nudged about it a couple days ago.

faithanalog (approved):

Issue #433, opened 2022-09-01
write error deadlocks the upstairs


leftwo:
I was playing around with error injection and forced a write error to happen in the downstairs.
This lead to a hang in the upstairs.

It looks like the error path has a deadlock in it, this task is stuck:

```
------------ thread# 2 / lwp# 2 [crucible-tokio] -------------                                                                                                
 fffffc7feec11207 lwp_park (0, 0, 0)                                                                                                                          
 fffffc7feec08ffd mutex_lock_impl (5ba5cd0, 0) + 14d                                                                                                          
 fffffc7feec09100 mutex_lock (5ba5cd0) + 10                                                                                                                   
 0000000002d7158d std::sys::unix::mutex::Mutex::lock::h8bbee83112999296 () + 1d                                                                               
 0000000002d15ad4 std::sys_common::mutex::MovableMutex::raw_lock::h92b5ae88527e5374 () + 14                                                                   
 0000000002b67cf4 std::sync::mutex::Mutex<T>::lock::hdf7dca704e67e7df () + 24                                                                                 
 0000000002b329e4 crucible::Upstairs::ds_transition::h23040cdf2d76ae16 () + 94                                                                                
 0000000002b37154 crucible::Upstairs::process_ds_operation::h4403d7872444b0cc () + b74         
```

leftwo:
Yes, 
Inside the error path for process_ds_operation we call:
```
self.ds_transition(client_id, DsState::Failed);
```
But, we are already holding the downstairs lock.
Instead, we should call
```
self.ds_transition_with_lock(...)
```

PR #434, opened 2022-09-01
Fix deadlock on IO error, more tests.

leftwo:
Updated process_ds_operation() to call ds_transition_with_lock() because
we grab the ds lock when we enter that method. This was in the
write/flush error path, so not detected during normal operations.

Added more comments in various places to help others (and my future
self) understand what is going on.

Added more tests around process_ds_operation() that would have caught
this.

Cleaned up some minor formatting issues in show_all_work()

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -309,6 +309,10 @@ async fn process_message(
         return Err(CrucibleError::UuidMismatch.into());
     }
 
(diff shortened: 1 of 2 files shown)


leftwo:
https://github.com/oxidecomputer/crucible/issues/433

jmpesp (approved):
Nice catch :)

PR #435, opened 2022-09-02
Add crudd IO tool

faithanalog:
`crudd` is a command-line upstairs guest that performs simple reads and
writes. Think of it as the "dd" of crucible. See the README for more
information. This commit introduces the command with

- read/write support
- non-block-aligned reads/writes
- seeking (offset)
- a `test.sh` script to make sure it's actually working

Currently there's no automated testing of it, and no support for
Volumes.

### DIFF
--- a/crudd/src/main.rs
+++ b/crudd/src/main.rs
@@ -1,11 +1,12 @@
+// Copyright 2022 Oxide Computer Company
+
 use std::collections::VecDeque;
 use std::fs::File;
 use std::io::{Read, Write};
-use std::{cmp, io};
-// Copyright 2021 Oxide Computer Company
 use std::net::SocketAddr;
 use std::os::unix::io::FromRawFd;
 use std::sync::Arc;
+use std::{cmp, io};
 
 use anyhow::{bail, Result};
 use clap::Parser;
--- a/crudd/src/main.rs
+++ b/crudd/src/main.rs
@@ -1,0 +1,510 @@
+use std::collections::VecDeque;
+use std::fs::File;
+use std::io::{Read, Write};
+use std::{cmp, io};
+// Copyright 2021 Oxide Computer Company
+use std::net::SocketAddr;
+use std::os::unix::io::FromRawFd;
+use std::sync::Arc;
+
+use anyhow::{bail, Result};
+use clap::Parser;
+use tokio::runtime::Builder;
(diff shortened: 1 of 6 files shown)


crudd/src/main.rs
```diff
@@ -0,0 +1,510 @@
...
+            w_buf[n_read..n_read + uflow_backfill]
+                .copy_from_slice(&r_bytes[uflow_remainder as usize..]);
+
+            // Issue the write
+            let w_waiter = crucible.write(offset, w_buf.freeze())?;
+            waiters.push_back(w_waiter);
+        }
+
+        // Flush
+        let flush_waiter = crucible.flush(None)?;
```
leftwo, line 385: I don't think you technically need a flush here.  It's nice to do if you like your data, but I also wonder about any impact it has on performance.  As long as you don't have any overlapping IOs, everything should be written out as you expect.

crudd/test.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
```
leftwo, line 1: The `/tools/` directory has a smattering of tests for things that are both in CI and not in CI.  You could put this there, or leave it here, either way is fine.

leftwo (approved):
Yeah, it's great to get this back and you can iterate on it as you so require.

You might want to consider a `-o output_file` option, then you don't have to redirect, but I also understand balancing where you spend your time, and as you are probably the main user of this tool, do as you see fit.

crudd/src/main.rs
```diff
@@ -0,0 +1,510 @@
...
+    };
+
+    // How many full commands we can issue, and then what our remainder bytes
+    // will be. Note we'll round up the remainder later to get the number of
+    // blocks for the last iocmd
+    let cmd_count = num_bytes / (opt.iocmd_block_count * native_block_size);
+    let remainder = num_bytes % (opt.iocmd_block_count * native_block_size);
+
+    // Issue all of our read commands
+    for i in 0..cmd_count {
```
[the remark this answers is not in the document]
leftwo: Not sure how I came to the "block at a time" conclusion here, as that is clearly not the case.

Resolved!

faithanalog:
what do you want from me buildomat i moved a comment. im not even touching code in that test.

PR #436, opened 2022-09-03
Cache sqlite prepared statements

faithanalog:
rusqlite provides an LRU Cache (a cache which, when full, evicts the
least-recently-used value). This caches prepared statements, allowing us
to nullify the cost of parsing and compiling frequently used statements.
I've changed all sqlite queries in Inner to use `prepare_cached` to take
advantage of this cache. I have not done this for region creation, since
it wouldn't be relevant there.

The result is a dramatic reduction in CPU time spent parsing queries.
Prior to this change, sqlite3Prepare was taking 60% of CPU time during
reads and 50% during writes based on dtrace flamegraphs. Afterwards,
they don't even show up on the graph. I'm seeing a minimum doubling of
actual throughput with `crudd` after this change on my local hardware.

However, this does cost some amount of memory per-extent and thus will
scale linearly . This cache size I'm setting now (64 statements) is
chosen somewhat arbitrarily. If this becomes a significant consumer of
memory, we should reduce the size of the LRU, and stop caching the
statements in the coldest paths. At present, it doesn't seem to have any
meaningful impact on memory usage.

We could instead try to pre-generate and hold onto all the statements we
need to use ourselves, but I looked into doing that, and basically the
code required would be a mess due to lifetimes. We shouldn't do that
except as a last resort.

Also, if you're curious, the hashing function used is
https://docs.rs/ahash/latest/ahash/index.html

### DIFF
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -46,9 +46,9 @@ pub struct Inner {
 
 impl Inner {
     pub fn gen_number(&self) -> Result<u64> {
-        let mut stmt = self
-            .metadb
-            .prepare("SELECT value FROM metadata where name='gen_number'")?;
+        let mut stmt = self.metadb.prepare_cached(
+            "SELECT value FROM metadata where name='gen_number'",
+        )?;
         let gen_number_iter = stmt.query_map([], |row| row.get(0))?;
 
         let mut gen_number_values: Vec<u64> = vec![];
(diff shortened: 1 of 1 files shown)


jmpesp (approved):
I tested this in the Canada region :tm: using f3write + f3read, and saw:

```
before:
Average writing speed: 9.07 MB/s
Average reading speed: 16.04 MB/s

after:
Average writing speed: 12.80 MB/s
Average reading speed: 20.94 MB/s
```

:shipit: :shipit: :shipit: :shipit: :shipit:

PR #441, opened 2022-09-12
Remove vestigial num_blocks field from protocol

faithanalog:
ReadRequest and ReadResponse had a num_blocks field that was at one point used, but isn't anymore. Now, one request is one block, and its associated response bundles up all the hash and encryption metadata associated with that block. This commit removes that field.

This field was not actually used in more than a few places in the code, so most of the affected lines of code are in tests. A lot of them were using `num_blocks = 2`, but none of those tests were actually testing behavior that was impacted by the `num_blocks` field, and a couple of them were otherwise acting as if they had requested 1 block of data, so this shouldn't impact the veracity of the tests.


leftwo (approved):
bye bye num_blocks

PR #444, opened 2022-09-13
Batch IO and Sqlite operations in Downstairs

faithanalog:
Previously, Downstairs has done sqlite queries and disk IO one block at a time. This commit changes downstairs to perform larger queries and disk IO for contiguous blocks, when possible, reducing overhead significantly and improving overall throughput. There is potential room for improvement in the future:
- switching to preadv, removing a seek syscall.
- switching the WriteUnwritten implementation to just check for the existence of hashes instead of actually getting their values.

This commit also switches to HashSet for the WriteUnwritten implementation. honestly though in the future we might want to just make that a vec of bools pre-allocated to match the number of writes. needs some testing, which I can do easier once i add WriteUnwritten support to crudd.

### DIFF
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -1,9 +1,9 @@
 // Copyright 2021 Oxide Computer Company
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::convert::TryInto;
 use std::fmt;
 use std::fs::{rename, File, OpenOptions};
-use std::io::{Read, Seek, SeekFrom, Write};
+use std::io::{BufWriter, Read, Seek, SeekFrom, Write};
 use std::net::SocketAddr;
 use std::path::{Path, PathBuf};
 use std::sync::{Mutex, MutexGuard};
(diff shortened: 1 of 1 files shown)


downstairs/src/region.rs
```diff
@@ -895,32 +918,87 @@ impl Extent {
...
+                first_req.offset.value,
+                n_contiguous_requests as u64,
+            )?;
 
-            response.hashes = inner.get_hashes(request.offset.value)?;
+            // Now it's time to put everything into the responses.
+            // We use into_iter here to move values out of enc_ctxts/hashes,
+            // avoiding a clone(). For code consistency, we use iters for the
+            // response and data chunks too. These iters will be the same length
+            // (equal to n_contiguous_requests) so zipping is fine
```
jmpesp, line 980: Will these iterators be the same length if the downstairs is storing unencrypted blocks?

downstairs/src/region.rs
```diff
@@ -1018,34 +1096,62 @@ impl Extent {
...
+                let n_contiguous_writes = writes[write_run_start..]
+                    .windows(2)
+                    .take_while(|wr_pair| {
+                        wr_pair[0].offset.value + 1 == wr_pair[1].offset.value
+                    })
+                    .count()
+                    + 1;
+
+                // Query hashes for the write range.
+                // XXX we should add a query that doesnt actually give us back
```
jmpesp, line 1122: Is this XXX something for the future, or for this PR? :)

downstairs/src/region.rs
```diff
@@ -1057,17 +1163,33 @@ impl Extent {
...
+            let block = write.offset.value;
+            if writes_to_skip.contains(&block) {
+                debug_assert!(only_write_unwritten);
                 continue;
             }
-            let byte_offset = write.offset.value * self.block_size;
 
-            inner.file.seek(SeekFrom::Start(byte_offset))?;
-            inner.file.write_all(&write.data)?;
+            // Seeking flushes the buffer so only do it when we need to
```
jmpesp, line 1179: Do you mean this flushes to the file, or just performs the actual writes?

downstairs/src/region.rs
```diff
@@ -1057,17 +1163,33 @@ impl Extent {
...
+                    .seek(SeekFrom::Start(block * self.block_size))?;
+            }
+
+            buffered_writer.write_all(&write.data)?;
+
+            next_block_in_run = block + 1;
         }
 
+        // writes remaining buffer without calling flush() on the file handle.
+        drop(buffered_writer);
```
jmpesp, line 1191: The following comment from [BufWriter's source](https://doc.rust-lang.org/src/std/io/buffered/bufwriter.rs.html#23-27) makes me very anxious:

```
/// It is critical to call [`flush`] before `BufWriter<W>` is dropped. Though
/// dropping will attempt to flush the contents of the buffer, any errors
/// that happen in the process of dropping will be ignored. Calling [`flush`]
/// ensures that the buffer is empty and thus dropping will not even attempt
/// file operations.
```

Based on that comment, I would even say that this means we can't use `BufWriter` here. Previously, the downstairs would return errors from each individual `inner.file.write_all`, where here they are being ignored. We have to know when there are errors and propagate them up to the Upstairs (and eventually Nexus).

jmpesp (changes requested):
With this PR, I'm seeing (in the bathtub cluster using the same f3write && f3read test) average write speed go from 15.36 MB/s to 21.05 MB/s, and average read speed go from 23.70 MB/s to 52.07 MB/s (!). But there's a problem before it can go in - see the comment about BufWriter.

downstairs/src/region.rs
faithanalog: yeah that's fair, i'll rewrite with my own buffering logic

downstairs/src/region.rs
faithanalog: yes. the function to retrieve tx hashes always returns a `Vec<Vec<>>` that is precisely equal in length to the amount of blocks you requested, regardless of the contents of the sqlite database. If there aren't any encryption contexts, it'll just be a vec full of empty vecs

downstairs/src/region.rs
faithanalog: seek just performs the writes. but per the later comment, i'll rewrite to do my own buffering

downstairs/src/region.rs
faithanalog: Oh I haven't heard of that, no. I like that though, I think I'll try following that convention.

downstairs/src/region.rs
```diff
@@ -1057,15 +1165,47 @@ impl Extent {
         }
         tx.commit()?;
 
+        // Buffer writes for fewer syscalls. The 65536 buffer size here is
+        // chosen somewhat arbitrarily.
+        let mut write_buffer = [0u8; 65536];
```
faithanalog, line 1170: looking for feedback, is this a reasonable size to be allocating on the stack here?

downstairs/src/region.rs
```diff
@@ -136,44 +136,64 @@ impl Inner {
     fn get_encryption_contexts(
         &self,
         block: u64,
-    ) -> Result<Vec<EncryptionContext>> {
+        count: u64,
+    ) -> Result<Vec<Vec<EncryptionContext>>> {
```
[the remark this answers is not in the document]
faithanalog: updated that one and put some more detail in the hashes one to make it clearer about the guarantees you get on the output data.

jmpesp (approved):
Changes look good, but we should add some tests in `downstairs/src/region.rs` for larger contiguous and non-contiguous reads and writes to exercise this logic.

downstairs/src/region.rs
jmpesp: I think it's reasonable, yeah. I'd be curious what size OSes submit on average. The f3 test submits something like 80k sized writes, but even buffering at 64k vastly reduces the syscalls there.

I also suspect this relates to https://github.com/oxidecomputer/crucible/issues/101

jmpesp: As I pressed submit, I realized that you're asking about stack allocation, not how it relates to the writes we receive from the guest. Disregard previous :)

I don't know about if 64k is reasonable on the stack. I'd say if it works then :shipit:

Issue #298, opened 2022-04-25
implement integration tests mentioned in volume.rs


jmpesp:
there's several "can't test this without a running set of downstairs" tests that I left in volume.rs. now that we have the ability to write integration tests, implement them!

Issue #389, opened 2022-07-28
Upstairs misses an activation during hammer loop


leftwo:
During the `./tools/hammer_loop.sh` test, the guest (hammer) will send an activation.
I believe there is some path that exists during the downstairs coming and going that causes the
upstairs to forget that it has received an activation and the guest will end up waiting forever
for an upstairs that will never finish.

You end up seeing messages like these:
```
Upstairs is not yet active, waiting in activate function
[1] Received Imok
[0] Received Imok
Upstairs is not yet active, waiting in activate function
Upstairs is not yet active, waiting in activate function
[0] Received Imok
[1] Received Imok
Upstairs is not yet active, waiting in activate function
Upstairs is not yet active, waiting in activate function
[1] Received Imok
[0] Received Imok
Upstairs is not yet active, waiting in activate function
[0] Received Imok
[1] Received Imok
```
Attached are two log of a hammer test where this happened.

[htl2.log](https://github.com/oxidecomputer/crucible/files/9205617/htl2.log)
[htl.log](https://github.com/oxidecomputer/crucible/files/9205618/htl.log)

leftwo:
More debugging and log messages were added.
I see this message:
```
ERROR: 127.0.0.1:8820: proc: [1] bb1aa090-3748-4744-91fc-69f3452416ef bad YouAreNoLongerActive, same upstairs uuid and our gen 181 >= new gen 180!
```
Which tells me perhaps it's time to get the generation number work connected into Nexus proper.  Without the Upstairs refusing to activate when generation numbers are wrong, the hammer test loop is of limited value as it can get in its own way.

leftwo:
This is fixed with the generation number work done recently in the upstairs.

PR #445, opened 2022-09-15
Implement missing "snapshot" integration test

jmpesp:
There were two tests in upstairs/src/volume.rs that remained unimplemented due to requiring a running set of downstairs:

        construct_snapshot_backed_vol
        construct_read_only_iso_volume

We have a way to write integration tests now, so this commit implements a test for a "snapshot" backed volume. It does not perform a snapshot but simply boots a downstairs in read-only mode and uses that as a read-only parent.

In order to implement this, I wanted to write some data into a set of downstairs, then boot new downstairs read-only but pointing at the same on-disk files. The integration test framework couldn't support this without adding `fn deactivate` to the BlockIO trait - the original read-write downstairs have to be deactivated (and have their tasks stopped) in order for the read-only downstairs to boot ok.

Closes https://github.com/oxidecomputer/crucible/issues/298

### DIFF
--- a/integration_tests/src/lib.rs
+++ b/integration_tests/src/lib.rs
@@ -1888,15 +1888,6 @@ mod test {
 
         volume.deactivate()?.block_wait()?;
 
-        let mut active_check_counter = 0;
-        while volume.query_is_active()? {
-            tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
-            active_check_counter += 1;
-            if active_check_counter > 10 {
-                panic!("took too long to deactivate!");
-            }
-        }
-
         drop(volume);
 
         test_downstairs_set.reboot_read_only()?;
--- a/integration_tests/src/lib.rs
+++ b/integration_tests/src/lib.rs
@@ -18,8 +18,11 @@ mod test {
 
     #[allow(dead_code)]
     struct TestDownstairs {
+        address: IpAddr,
+        port: u16,
         tempdir: TempDir,
         downstairs: Arc<Mutex<Downstairs>>,
+        join_handle: tokio::task::JoinHandle<Result<()>>,
     }
 
     impl TestDownstairs {
@@ -48,7 +51,7 @@ mod test {
             )?;
 
             let adownstairs = downstairs.clone();
-            tokio::spawn(async move {
+            let join_handle = tokio::spawn(async move {
                 start_downstairs(
                     adownstairs,
                     address,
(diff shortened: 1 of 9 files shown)


jmpesp:
Note: this test seems to be looping on my workstation, currently digging into why.

faithanalog:
Yeah it's stuck on my workstation as well. I've noticed a few of the other integration tests will occasionally get stuck for me, but not consistently, because sometimes they'll get stuck and other times they won't. I wonder if it's related though.

upstairs/src/volume.rs
```diff
@@ -369,6 +369,18 @@ impl BlockIO for Volume {
...
+        for sub_volume in &self.sub_volumes {
+            sub_volume.deactivate()?;
+        }
+
+        if let Some(ref read_only_parent) = &self.read_only_parent {
+            read_only_parent.deactivate()?;
+        }
+
+        BlockReqWaiter::immediate()
+    }
```
faithanalog, line 382: on lines 374/378, we should probably block-wait on the BlockReqWaiters returned by the subvolume/parent deactivations before returning a BlockReqWaiter::immediate()

integration_tests/src/lib.rs
```diff
@@ -1764,9 +1852,172 @@ mod test {
...
+        };
+
+        volume
+            .write(
+                Block::new(0, BLOCK_SIZE.trailing_zeros()),
+                Bytes::from(random_buffer.clone()),
+            )?
+            .block_wait()?;
+
+        volume.deactivate()?.block_wait()?;
```
faithanalog, line 1889: Asking primarily for my own benefit - does this need a flush before the deactivation, or does the deactivation send a flush implicitly?

faithanalog (changes requested):

integration_tests/src/lib.rs
jmpesp: It's a good question - the deactivation sends a final flush. Check out the relevant function is [`set_deactivate`](https://github.com/oxidecomputer/crucible/blob/5263f0a147185e63cf64ea7ae05837ef1b501a63/upstairs/src/lib.rs#L3698):

```
     * For a deactivation to complete, we need to:
     * Stop all incoming IO.
     * Let any outstanding IO finish.
     * Submit a final flush to all downstairs.
     * Wait for that final flush to finish (on all downstairs).
```

upstairs/src/volume.rs
jmpesp: Definitely! I totally missed this. Fixed in 0e27718de864ee59b6c1f17b65e0265737818424.

I also added a `#[must_use]` for `BlockReqWaiter` in 523d63e, which caught some other misses.

jmpesp: Update: it was probably this that was causing the looping, I don't see it anymore.

integration_tests/src/lib.rs
```diff
@@ -1764,9 +1852,172 @@ mod test {
...
+
+        let mut active_check_counter = 0;
+        while volume.query_is_active()? {
```
leftwo, line 1892: Did you see this loop activate.  I believe that each upstairs deactivate, if waited on, will not return until the deactivation has completed.  Since I believe that is what you do in the `volume.deactivate`, I would not think it would return from `volume.deactivate()?.block_wait()?;` until the deactivation is complete.  Maybe I'm not understanding how block_wait works?

upstairs/src/volume.rs
faithanalog: oh awesome, I know I definitely almost forgot to wait on a couple while I was writing crudd.

faithanalog (approved):

integration_tests/src/lib.rs
jmpesp: You're not wrong, this code is - this was put in previous to 0e27718de864ee59b6c1f17b65e0265737818424, so I ripped it out in aab698d

jmpesp:
Running for a while with `--nocapture`, I see the following:

```
Upstairs is not yet active, waiting in activate function
127.0.0.1:54002[1] looper connecting to 127.0.0.1:54002
[0] Received Imok
[2] Received Imok
127.0.0.1:54002[1] looper connecting to 127.0.0.1:54002
127.0.0.1:54002[1] looper connecting to 127.0.0.1:54002
Upstairs is not yet active, waiting in activate function
127.0.0.1:54002[1] looper connecting to 127.0.0.1:54002
127.0.0.1:54002[1] looper connecting to 127.0.0.1:54002
127.0.0.1:54002[1] looper connecting to 127.0.0.1:54002
test test::integration_test_region has been running for over 60 seconds
Upstairs is not yet active, waiting in activate function
[0] Received Imok
[2] Received Imok
127.0.0.1:54002[1] looper connecting to 127.0.0.1:54002
127.0.0.1:54002[1] looper connecting to 127.0.0.1:54002
127.0.0.1:54002[1] looper connecting to 127.0.0.1:54002
```

note that this is `integration_test_region`, not the test added with this PR.

This looks a lot like the issue described in https://github.com/oxidecomputer/crucible/issues/389. If this is something that the three of us are intermittently seeing, then I think there's a deeper issue here that needs to be dug out.

leftwo:
When I usually see that a test hass been not making progress for >60 seconds, it's either a port was re-used for two different tests (not the case here) or something panic'd and we did not catch the panic and the test is waiting for a result.  I don't think that's what we are seeing here either.

I am working on some "nightly" tests, and one of them is the hammer loop, and it just hit I believe the hang mentioned in https://github.com/oxidecomputer/crucible/issues/389  I'm going to see if I can get to the bottom of that issue.  I'm also testing on the nightly branch which does not have these changes.

leftwo (approved):

PR #449, opened 2022-09-20
Bite the bullet: make BlockIO async

jmpesp:
BlockIO is now an async trait. This was mostly a rote process of adding async to every fn, and adding await everywhere, with the exception of a few changes that should be **closely** reviewed:

The first pass at this commit *only* changed the BlockIO trait, and saw errors where BlockReq's Sender was being closed without a message being sent. This caused BlockReqWaiter's `recv` to return an error.

In order to fix this, we shouldn't pass the Sender (and clones) around anymore.  Several Upstairs functions were changed to *consume* BlockReq, and make sure that a send is sent before the BlockReq is dropped. In order to make sure that other code was not accessing the fields of these structs incorrectly, they were moved to a new module `block_req`.

Another change here: sending through BlockReq, or receiving through BlockReqWaiter consumes the object. A channel of buffer size 1 is created, and it doesn't make sense that users could send or receive multiple messages through these structs.

Have a look at the changes to process_new_io in upstairs/src/lib.rs first, this is where we were previously accessing BlockReq fields directly and handing out clones of Sender. Now we can lean on Rust's borrow checker and make a little more sure that BlockReq and BlockReqWaiter are being used correctly.

If everything is properly async, block_on and block_in_place should no longer be required. If you see block_in_place now, it should be scrutinized and is probably wrong. Likewise, if you're using std::sync::Mutex or other blocking implementations from std, they should be changed to tokio equivalents.

One place where block_in_place is still required is CruciblePseudoFile, which implements std::io::{Read, Write, Seek}, which are *not* async.

Any std::thread::sleep use is a now a bug!

- extent_from_offset no longer returns Result, because it didn't needto

- block_wait was renamed to wait

- FileBlockIO can use reqwest::Client instead of reqwest::blocking::Client, which is nice.

- Volume::add_read_only_parent_create_guest was removed, it wasn't being used

With this commit, the integration tests no longer intermittently hang, and importantly no longer hang even though `worker_threads = 8` was removed. This is a more constrainted environment, where previous to this commit if we boosted the number of threads we would see the hangs go away.

Note I have not yet tested integrating with Propolis and Omicron. That'll happen soon after this commit goes in.


jmpesp:
I left this running overnight, and unfortunately saw

```
test test::integration_test_volume_write_unwritten_1 has been running for over 60 seconds
```

so it seems the hang investigation isn't over yet.

jmpesp:
I ran the `measure_iops` tooling to see if it measured any difference:

before:
```
IOPS: [1102.4537, 687.5161, 883.2225, 735.6631, 483.2462, 338.77472, 258.18665, 1143.1948, 627.30444, 456.2048, 342.5279, 240.78601, 1080.2274, 654.8024, 459.7501, 365.50443, 252.71648, 1179.285, 638.8333, 442.99002, 344.06207, 240.87225, 1150.1, 671.9266, 442.46338, 343.79578, 240.9019, 1150.1453, 605.75836, 437.2962, 348.30203, 260.74817, 1106.0356, 627.20276, 452.48642, 354.1651, 235.99773, 1183.6311, 651.46454, 452.3445, 347.12186, 317.9987, 1307.056, 589.8102, 455.087, 348.70407, 444.93475, 1303.6804, 582.1082, 398.1684, 322.53357, 446.92947, 1299.4061, 574.12683, 399.68707, 320.6549, 494.99356, 1282.5076, 589.85187, 414.71634, 321.02795, 503.9869, 1298.0137, 578.8945, 404.2403, 319.4599, 445.59302, 1298.6588, 591.82007, 395.1414, 328.23337, 453.98474, 1322.5967, 553.9318, 403.90668, 321.92664, 432.27814, 1268.5223, 588.0486, 400.45773, 319.73666, 463.78415, 1303.0916, 594.43805, 406.28726, 320.3361, 488.89954, 1234.8977, 582.7446, 400.1917, 317.54135, 467.95303, 1288.9108, 573.6066, 402.38147, 318.74295, 501.7006, 1256.6404, 577.0013, 398.6054]
IOPS mean 603.6118 stddev 337.30182
IOPS min 235.99773 max 1322.5967
BW: [564456.3, 352008.25, 452209.9, 376659.5, 247422.05, 173452.66, 132191.56, 585315.75, 321179.88, 233576.86, 175374.28, 123282.44, 553076.44, 335258.84, 235392.05, 187138.27, 129390.836, 603793.94, 327082.66, 226810.89, 176159.78, 123326.59, 588851.2, 344026.4, 226541.25, 176023.44, 123341.77, 588874.4, 310148.28, 223895.66, 178330.64, 133503.06, 566290.25, 321127.8, 231673.05, 181332.53, 120830.836, 606019.1, 333549.84, 231600.39, 177726.39, 162815.33, 669212.7, 301982.8, 233004.55, 178536.48, 227806.6, 667484.4, 298039.4, 203862.22, 165137.19, 228827.89, 665295.94, 293952.94, 204639.78, 164175.31, 253436.7, 656643.9, 302004.16, 212334.77, 164366.31, 258041.3, 664583.0, 296393.97, 206971.03, 163563.47, 228143.63, 664913.3, 303011.88, 202312.39, 168055.48, 232440.19, 677169.5, 283613.1, 206800.22, 164826.44, 221326.4, 649483.44, 301080.88, 205034.36, 163705.17, 237457.48, 667182.9, 304352.28, 208019.08, 164012.08, 250316.56, 632267.6, 298365.25, 204898.16, 162581.17, 239591.95, 659922.3, 293686.6, 206019.31, 163196.39, 256870.7, 643399.9, 295424.66, 204085.97]
BW mean 309049.25 stddev 172698.53
BW min 120830.836 max 677169.5
real 135.25
user 346.64
sys 15.83
```

after:
```
IOPS: [1145.2971, 638.9745, 950.68506, 788.84454, 515.79095, 399.61655, 270.09067, 1159.2004, 650.4437, 451.4079, 412.898, 267.58505, 1161.3591, 616.88, 494.89865, 402.29095, 247.65329, 1185.6207, 678.83435, 476.32355, 386.71942, 245.40657, 1149.8745, 656.9771, 487.4807, 396.9903, 295.96185, 1070.752, 675.8361, 454.61987, 400.90494, 288.97342, 1145.917, 615.1666, 490.56192, 388.68738, 292.89685, 1263.0776, 677.7369, 504.1715, 389.0397, 447.94522, 1323.9509, 649.6847, 461.3178, 361.76117, 516.95105, 1312.4968, 652.6691, 461.9465, 370.60535, 507.80817, 1357.7465, 640.7908, 451.21896, 340.2732, 513.7798, 1298.5195, 633.3233, 450.4259, 362.17343, 475.78543, 1345.8156, 657.3422, 448.29074, 361.78125, 554.98486, 1349.9246, 627.0067, 436.9413, 352.25555, 478.99664, 1347.14, 633.7543, 449.52478, 361.30673, 482.93488, 1264.2683, 667.7689, 465.84363, 365.88394, 392.85434, 1366.7778, 677.0315, 465.5952, 366.8271, 549.9081, 1266.8243, 652.97687, 453.82428, 364.92722, 558.65265, 1320.8761, 546.8262, 456.21576, 367.81235, 560.4613, 1309.85, 622.8308, 449.36227]
IOPS mean 641.85815 stddev 334.1369
IOPS min 245.40657 max 1366.7778
BW: [586392.1, 327154.94, 486750.75, 403888.4, 264084.97, 204603.67, 138286.42, 593510.6, 333027.2, 231120.84, 211403.78, 137003.55, 594615.9, 315842.56, 253388.11, 205972.97, 126798.484, 607037.8, 347563.2, 243877.66, 198000.34, 125648.164, 588735.75, 336372.28, 249590.13, 203259.03, 151532.47, 548225.0, 346028.1, 232765.38, 205263.33, 147954.39, 586709.5, 314965.3, 251167.7, 199007.94, 149963.19, 646695.75, 347001.28, 258135.81, 199188.33, 229347.95, 677862.9, 332638.56, 236194.72, 185221.72, 264678.94, 671998.4, 334166.6, 236516.61, 189749.94, 259997.78, 695166.2, 328084.88, 231024.11, 174219.88, 263055.25, 664842.0, 324261.53, 230618.06, 185432.8, 243602.14, 689057.56, 336559.22, 229524.86, 185232.0, 284152.25, 691161.4, 321027.44, 223713.95, 180354.84, 245246.28, 689735.7, 324482.2, 230156.69, 184989.05, 247262.66, 647305.4, 341897.7, 238511.94, 187332.58, 201141.42, 699790.25, 346640.13, 238384.75, 187815.47, 281552.94, 648614.06, 334324.16, 232358.03, 186842.73, 286030.16, 676288.56, 279975.0, 233582.47, 188319.92, 286956.2, 670643.2, 318889.38, 230073.48]
BW mean 328631.38 stddev 171078.1
BW min 125648.164 max 699790.25
real 139.63
user 291.05
sys 12.56
```

running both IOPs and BW measurements through ministat:

IOPs:
```
+-------------------------------------------------------------------------------------------------------+
|        x      x   +                                                                                   |
|        x x    x   +x                                                                                  |
|        x x +  x   **          xx                                                               x      |
|x       x x++ +*   ** +  +     xx     + +                                                       x      |
|x x  + xx x++ +*   ** ++ +   + xx   +++ +                                         *          +  x   +  |
|x+x+ + xx x++ +*  x***+*x+  ++ xx +*+++ +                                     x   *+ x       +  x++ +  |
|x+x+ + xx**** *** *******+  +*xxxx+*****+x   x    +       x     +          +x x   *+ x+   x x*xx*** +++|
|   |__|_____________M____M_______A___A__________________________|__|                                   |
+-------------------------------------------------------------------------------------------------------+
    N           Min           Max        Median           Avg        Stddev
x 100     235.99773     1322.5967      459.7501     603.61181     337.30183
+ 100     245.40657     1366.7778     507.80817      641.8582     334.13689
No difference proven at 95.0% confidence
```

BW:
```
x before
+ after
+-------------------------------------------------------------------------------------------------------+
|        x      x   +                                                                                   |
|        x x    x   +x                                                                                  |
|        x x +  x   **          xx                                                               x      |
|x       x x++ +*   ** +  +     xx     + +                                                       x      |
|x x  + xx x++ +*   ** ++ +   + xx   +++ +                                         *          +  x   +  |
|x+x+ + xx x++ +*  x***+*x+  ++ xx +*+++ +                                     x   *+ x       +  x++ +  |
|x+x+ + xx**** *** *******+  +*xxxx+*****+x   x    +       x     +          +x x   *+ x+   x x*xx*** +++|
|   |__|_____________M____M_______A___A__________________________|__|                                   |
+-------------------------------------------------------------------------------------------------------+
    N           Min           Max        Median           Avg        Stddev
x 100     120830.84      677169.5     235392.05     309049.25     172698.54
+ 100     125648.16     699790.25     259997.78      328631.4     171078.09
No difference proven at 95.0% confidence
```

jmpesp:
I ran some tests with crudd and there might be a slight performance increase for reads.

for a crudd-based read test, I 

- initialized three blank 1GB downstairs
- wrote zeroes to the whole lba
- then did two warm up reads of the whole lba, then measured four reads of the whole lba

main (d2be3d81):
```
+ cargo run --release -p crudd -- --key HLyo7ZhAf/E9IdX2DDHPHJO0dLgrRxZabWiTlnoKZXc= --byte-offset 0 --gen 1 -t 127.0.0.1:44101 -t 127.0.0.1:44102 -t 127.0.0.1:44103 read
+ pv -rb
1020MiB [66.1MiB/s]

+ cargo run --release -p crudd -- --key HLyo7ZhAf/E9IdX2DDHPHJO0dLgrRxZabWiTlnoKZXc= --byte-offset 0 --gen 1 -t 127.0.0.1:44101 -t 127.0.0.1:44102 -t 127.0.0.1:44103 read
+ pv -rb
1020MiB [65.4MiB/s]

+ cargo run --release -p crudd -- --key HLyo7ZhAf/E9IdX2DDHPHJO0dLgrRxZabWiTlnoKZXc= --byte-offset 0 --gen 1 -t 127.0.0.1:44101 -t 127.0.0.1:44102 -t 127.0.0.1:44103 read
+ pv -rb
1020MiB [69.0MiB/s]

+ cargo run --release -p crudd -- --key HLyo7ZhAf/E9IdX2DDHPHJO0dLgrRxZabWiTlnoKZXc= --byte-offset 0 --gen 1 -t 127.0.0.1:44101 -t 127.0.0.1:44102 -t 127.0.0.1:44103 read
+ pv -rb
1020MiB [67.8MiB/s]
```

PR:
```
+ cargo run --release -p crudd -- --key HLyo7ZhAf/E9IdX2DDHPHJO0dLgrRxZabWiTlnoKZXc= --byte-offset 0 --gen 1 -t 127.0.0.1:44101 -t 127.0.0.1:44102 -t 127.0.0.1:44103 read
+ pv -rbp
1020MiB [70.0MiB/s]

+ cargo run --release -p crudd -- --key HLyo7ZhAf/E9IdX2DDHPHJO0dLgrRxZabWiTlnoKZXc= --byte-offset 0 --gen 1 -t 127.0.0.1:44101 -t 127.0.0.1:44102 -t 127.0.0.1:44103 read
+ pv -rbp
1020MiB [70.1MiB/s]

+ cargo run --release -p crudd -- --key HLyo7ZhAf/E9IdX2DDHPHJO0dLgrRxZabWiTlnoKZXc= --byte-offset 0 --gen 1 -t 127.0.0.1:44101 -t 127.0.0.1:44102 -t 127.0.0.1:44103 read
+ pv -rbp
1020MiB [68.7MiB/s]

+ cargo run --release -p crudd -- --key HLyo7ZhAf/E9IdX2DDHPHJO0dLgrRxZabWiTlnoKZXc= --byte-offset 0 --gen 1 -t 127.0.0.1:44101 -t 127.0.0.1:44102 -t 127.0.0.1:44103 read
+ pv -rbp
1020MiB [69.8MiB/s]
```

ministat on those MiB/s numbers might show a slight improvement, but the +/- is too large to be sure.

```
x before
+ after
+-------------------------------------------------------------------------------------------------------+
|x              x                                  x                  +     x                +   + +    |
| |_________________________________A______________M__________________|     |_____________A______M_____||
+-------------------------------------------------------------------------------------------------------+
    N           Min           Max        Median           Avg        Stddev
x   4          65.4            69          67.8        67.075     1.6317169
+   4          68.7          70.1            70         69.65    0.64549722
Difference at 95.0% confidence
	2.575 +/- 2.14694
	3.83899% +/- 3.20081%
	(Student's t, pooled s = 1.2408)
```

for a crudd-based write test, I similarly:

- initialized three blank 1GB downstairs
- did two warm up writes of the whole lba, then measured four writes of the whole lba

main (d2be3d81):
```
+ pv -rb /dev/zero
+ cargo run --release -p crudd -- --key HLyo7ZhAf/E9IdX2DDHPHJO0dLgrRxZabWiTlnoKZXc= --byte-offset 0 --gen 1 -t 127.0.0.1:44101 -t 127.0.0.1:44102 -t 127.0.0.1:44103 write
1020MiB [45.6MiB/s]

+ pv -rb /dev/zero
+ cargo run --release -p crudd -- --key HLyo7ZhAf/E9IdX2DDHPHJO0dLgrRxZabWiTlnoKZXc= --byte-offset 0 --gen 1 -t 127.0.0.1:44101 -t 127.0.0.1:44102 -t 127.0.0.1:44103 write
1020MiB [47.8MiB/s]

+ pv -rb /dev/zero
+ cargo run --release -p crudd -- --key HLyo7ZhAf/E9IdX2DDHPHJO0dLgrRxZabWiTlnoKZXc= --byte-offset 0 --gen 1 -t 127.0.0.1:44101 -t 127.0.0.1:44102 -t 127.0.0.1:44103 write
1020MiB [45.9MiB/s]

+ pv -rb /dev/zero
+ cargo run --release -p crudd -- --key HLyo7ZhAf/E9IdX2DDHPHJO0dLgrRxZabWiTlnoKZXc= --byte-offset 0 --gen 1 -t 127.0.0.1:44101 -t 127.0.0.1:44102 -t 127.0.0.1:44103 write
1020MiB [46.7MiB/s]
```

this PR:
```
+ pv -rb /dev/zero
+ cargo run --release -p crudd -- --key HLyo7ZhAf/E9IdX2DDHPHJO0dLgrRxZabWiTlnoKZXc= --byte-offset 0 --gen 1 -t 127.0.0.1:44101 -t 127.0.0.1:44102 -t 127.0.0.1:44103 write
1020MiB [46.3MiB/s]

+ pv -rb /dev/zero
+ cargo run --release -p crudd -- --key HLyo7ZhAf/E9IdX2DDHPHJO0dLgrRxZabWiTlnoKZXc= --byte-offset 0 --gen 1 -t 127.0.0.1:44101 -t 127.0.0.1:44102 -t 127.0.0.1:44103 write
1020MiB [46.3MiB/s]

+ pv -rb /dev/zero
+ cargo run --release -p crudd -- --key HLyo7ZhAf/E9IdX2DDHPHJO0dLgrRxZabWiTlnoKZXc= --byte-offset 0 --gen 1 -t 127.0.0.1:44101 -t 127.0.0.1:44102 -t 127.0.0.1:44103 write
1020MiB [46.4MiB/s]

+ pv -rb /dev/zero
+ cargo run --release -p crudd -- --key HLyo7ZhAf/E9IdX2DDHPHJO0dLgrRxZabWiTlnoKZXc= --byte-offset 0 --gen 1 -t 127.0.0.1:44101 -t 127.0.0.1:44102 -t 127.0.0.1:44103 write
1020MiB [47.1MiB/s]
```

ministat on those MiB/s numbers shows no change:

```
x before
+ after
+-------------------------------------------------------------------------------------------------------+
|                                   +                                                                   |
|    x            x                 +   +             x                 +                              x|
||___________________________|__________M____AA_______M________|_________________________|              |
+-------------------------------------------------------------------------------------------------------+
    N           Min           Max        Median           Avg        Stddev
x   4          45.6          47.8          46.7          46.5    0.98319208
+   4          46.3          47.1          46.4        46.525    0.38622101
No difference proven at 95.0% confidence
```

upstairs/src/test.rs
```diff
@@ -506,11 +499,11 @@ mod up_test {
         assert_eq!(ds.completed.len(), 1);
     }
 
-    #[test]
-    fn work_flush_one_error_then_ok() {
+    #[tokio::test]
```
leftwo, line 502: I wonder if there is any impact to all these tests being inside tokio now.  I don't think there is really and choice though.

upstairs/src/test.rs
```diff
@@ -2173,10 +2174,10 @@ mod up_test {
         assert_eq!(state, AckStatus::Acked);
     }
 
-    // ZZZ another test that does read replay for the first IO
+    // XXX another test that does read replay for the first IO
```
leftwo, line 2177: Ah, this was my not to self to make another test for some specific thing.  I'll go figure out what I was talking about and add the test and delete the comment.

upstairs/src/test.rs
```diff
@@ -3256,17 +3257,17 @@ mod up_test {
...
         // Now, we should be empty, so nw is false
-        assert!(!tokio_test::block_on(up.new_rec_work()).unwrap());
+        assert!(!up.new_rec_work().await.unwrap());
     }
 
-    #[test]
+    #[tokio::test]
     #[should_panic]
-    fn reconcile_leave_no_job_behind() {
+    async fn reconcile_leave_no_job_behind() {
```
leftwo, line 3265: Ahh... this.

I think we need to make sure that the test is panic'ing in the place we expect it to, instead of in some earlier setup.

And, for this test... I think it's poorly written in that there are several places where it could have panic'd that were not correct.   Maybe let's just file an issue to come back and fix this test specifically, as none of the async changes I think did anything to change the issue that was already there.

upstairs/src/test.rs
```diff
@@ -3346,14 +3347,14 @@ mod up_test {
         assert!(ds.rep_done(2, rep_id));
     }
 
-    #[test]
+    #[tokio::test]
```
leftwo, line 3350: This test should be on the same list as the issue above.

upstairs/src/test.rs
```diff
@@ -3831,268 +3832,306 @@ mod up_test {
...
-        assert_eq!(*guest.iop_tokens.lock().unwrap(), 0);
+        guest.leak_iop_tokens(16000).await;
+        assert_eq!(*guest.iop_tokens.lock().await, 0);
 
         Ok(())
     }
 
-    #[test]
-    fn test_flush_does_not_consume_iops() -> Result<()> {
+    #[tokio::test(flavor = "multi_thread", worker_threads = 8)]
```
leftwo, line 3932: Why did you add the worker threads here, but not elsewhere?

leftwo (commented):
Wow.. this has touched everything... A lot of work here, nice.

I have some general questions that I thought this box had saved for me, but it looks like it lost them so I'll just attach them as a separate comment (when I remember what they were).

crudd/src/main.rs
```diff
@@ -226,19 +225,89 @@ fn cmd_read<T: BlockIO>(opt: &Opt, crucible: Arc<T>) -> Result<()> {
...
+        let waiter = crucible.read(offset, buffer.clone()).await?;
+        waiter.wait().await?;
+        output.write_all(&buffer.as_vec().await[0..remainder as usize])?;
     }
 
     Ok(())
 }
 
-fn cmd_write<T: BlockIO>(opt: &Opt, crucible: Arc<T>) -> Result<()> {
+// We need to run this at the end to handle the remainder bytes, but we also
```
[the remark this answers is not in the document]
faithanalog: this was perviously a function i defined inside cmd_write to close over a bunch of variables calculater before the write loop gets going. my assumption is async makes that kind of closure not work. all the captured variables would need to be moved into the function since the borrow checker can't guarantee they'll live as long as the async function, whereas previously they were read-only borrowed and everything works out fine. so that's probably why it got moved out into its own thing.

faithanalog:
Something I am wondering is whether BlockIO still needs to return a BlockReqWaiter at all as part of its API? The `.await.wait().await` pattern that results seems rather redundant. Not all BlockIO implementations need to use BlockReqWaiter, so many of them just create immediate()s (such as the Reqwest one). At the same time, the ones that _do_ use BlockReqWaiter could just as well `await` them inside their implementation rather than returning them to the caller.

With the move to async, the caller has the power to decide when they want to block to wait for the operation to finish purely from the use of async (since all async functions return `Future`s), without having to involve the BlockReqWaiter. Is there a compelling reason to keep around BlockReqWaiter in the BlockIO API?

integration_tests/src/lib.rs
```diff
@@ -131,49 +131,50 @@ mod test {
...
+            .await?;
 
-        assert_eq!(vec![0x55_u8; BLOCK_SIZE * 10], *buffer.as_vec());
+        assert_eq!(vec![0x55_u8; BLOCK_SIZE * 10], *buffer.as_vec().await);
 
         Ok(())
     }
 
-    #[tokio::test(flavor = "multi_thread", worker_threads = 8)]
+    #[tokio::test]
```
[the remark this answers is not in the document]
jmpesp: It's a more constrained environment: by default `tokio::test` runs single-threaded. If we had any locking or awaiting problems it would be exposed by that.

upstairs/src/block_io.rs
```diff
@@ -35,63 +35,64 @@ impl FileBlockIO {
...
+        let mut file = self.file.lock().await;
         file.seek(SeekFrom::Start(start))?;
         file.read_exact(&mut data_vec[..])?;
 
         for i in 0..data_vec.len() {
             owned_vec[i] = true;
         }
 
-        BlockReqWaiter::immediate()
+        BlockReqWaiter::immediate().await
```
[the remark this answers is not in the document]
jmpesp: :)

upstairs/src/test.rs
jmpesp: right, yeah. they all require async now.

upstairs/src/block_req.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    /// Return a copy of the block op
+    pub fn op(&self) -> BlockOp {
+        self.op.clone()
+    }
+
+    /// Consume this BlockReq and send Ok to the receiver
+    pub async fn send_ok(self) {
+        // XXX this eats the result!
```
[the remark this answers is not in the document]
jmpesp: It'd have to flow through something else, if we can't send anything through the Sender here. maybe this deserves a followup issue

upstairs/src/lib.rs
```diff
@@ -6066,7 +6101,7 @@ impl GuestWork {
          * A gw_id that already finished and results were sent back to
          * the guest could still have an outstanding ds_id.
          */
-        if let Some(gtos_job) = self.active.get_mut(&gw_id) {
+        if let Some(mut gtos_job) = self.active.remove(&gw_id) {
```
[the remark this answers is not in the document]
jmpesp: have a look at d483a3e, I turned the else branch into a panic and it passed the test suite.

upstairs/src/lib.rs
```diff
@@ -6112,10 +6147,13 @@ impl GuestWork {
              */
             if gtos_job.submitted.is_empty() {
                 if result.is_ok() && gtos_job.guest_buffer.is_some() {
-                    gtos_job.transfer();
+                    gtos_job.transfer().await;
                 }
-                gtos_job.notify(result);
-                self.gw_complete(gw_id);
+
+                assert!(gtos_job.submitted.is_empty());
```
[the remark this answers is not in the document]
jmpesp: in d483a3e, I removed the if here and instead asserted `gtos_job.submitted.is_empty()`. this passed the test suite.

upstairs/src/test.rs
jmpesp: That was a mistake, corrected in ec2207f

upstairs/src/block_req.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// Consume this BlockReq and send a Result to the receiver
+    pub async fn send_result(self, r: Result<(), CrucibleError>) {
+        // XXX this eats the result!
+        let _ = self.sender.send(r).await;
+    }
+}
+
+/**
+ * When BlockOps are sent to a guest, the calling function receives a
+ * waiter that it can block on.
```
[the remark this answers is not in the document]
jmpesp: definitely, done in cde453c01c12a148aa382835830a00a8a2423751

upstairs/src/lib.rs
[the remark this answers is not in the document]
leftwo: Yeah, I tried something similar with the same result. I'll make an issue to fix the comment

jmpesp:
> Something I am wondering is whether BlockIO still needs to return a BlockReqWaiter at all as part of its API? The `.await.wait().await` pattern that results seems rather redundant. Not all BlockIO implementations need to use BlockReqWaiter, so many of them just create `immediate()`s (such as `ReqwestBlockIO` and `InMemoryBlockIO`). At the same time, the ones that _do_ use BlockReqWaiter could just as well `await` them inside their implementation rather than returning them to the caller.
> 
> With the move to async, the caller has the power to decide when they want to block to wait for the operation to finish purely from the use of async (since all async functions return `Future`s), without having to involve the BlockReqWaiter. Is there a compelling reason to keep around BlockReqWaiter in the BlockIO API?

I agree it is ugly, but I don't see another way to preserve the distinction of `"this request was submitted".await`, and `"this request completed".await`. For example, code that originally looked like

```rust
let mut waiters: Vec<BlockReqWaiter> = Vec::with_capacity(MAGIC_NUMBER);

// Submit multiple writes
for buf in bufs {
  let waiter = crucible.write(offset, buf.freeze()).await?; // did the send fail?
  waiters.push(waiter);
}

// wait on all the answers
for waiter in waiters {
  waiter.wait().await?; // did the operation fail?
}
```

would instead look something like

```rust
let mut waiters: Vec<Future<blah blah>> = Vec::with_capacity(MAGIC_NUMBER);

// Submit multiple writes
for buf in bufs {
  let waiter = crucible.write(offset, buf.freeze());
  waiters.push(waiter);
}

// wait on all the answers
let results: Vec<Result<(), CrucibleError> = futures::future::join_all!(waiters).await;
results.collect::<Result<Vec<()>, CrucibleError>()?
```

I guess the question is: do we need this distinction? I'm not sure we do. If there's an error submitting the request, then that should be returned as a variant of `Err(CrucibleError)`. I'll try making this change and see what it looks like.

jmpesp:
@leftwo @faithanalog I made the change to remove BlockReqWaiter (pushed as branch [no_block_req_waiter](https://github.com/jmpesp/crucible/tree/no_block_req_waiter) to my crucible fork), and discovered that preserving the distinction is important, but not for the reason of figuring out what failed like I originally thought. Specifically, it's important when submitting multiple requests and then waiting for their answers.

If there is a BlockReqWaiter, requests are **sent in order** and you can wait on all the answers in parallel:

```rust
let mut waiters: Vec<BlockReqWaiter> = Vec::with_capacity(MAGIC_NUMBER);

// Submit multiple writes
for buf in bufs {
  let waiter = crucible.write(offset, buf.freeze()).await?; // submit requests in order
  waiters.push(waiter);
}

// wait on all the answers in any order
let results: Vec<Result<(), CrucibleError> = futures::future::join_all(waiters).await;
results.collect::<Result<Vec<()>, CrucibleError>()?
```

Without a BlockReqWaiter, you're awaiting on the **combination** of sending requests and waiting for the results:

```rust
let mut futures: Vec<Future<blah blah>> = Vec::with_capacity(MAGIC_NUMBER);

// Submit multiple writes
for buf in bufs {
  let future = crucible.write(offset, buf.freeze()); // request is not submitted until the future is polled!
  futures.push(future);
}

// wait on all the futures in any order
let results: Vec<Result<(), CrucibleError> = futures::future::join_all(futures).await;
results.collect::<Result<Vec<()>, CrucibleError>()?
```

[`futures::future::join_all`](https://docs.rs/futures/0.3.24/futures/future/fn.join_all.html)'s documentation says:

> The returned future will drive execution for all of its underlying futures, collecting the results into a destination Vec<T> in the same order as they were provided.

which I took as it would poll them in the same order they arrive too. I wrote two tests to make sure that sending multiple operations and awaiting on their futures were still correct, and they both failed (without BlockReqWaiter). Have a look at `integration_test_join_all_is_ordered_for_writes` and `integration_test_join_all_is_ordered_for_reads_and_writes` on that other branch I posted above.

Another bug that was exposed by this change, and something to think about: the Guest is the only BlockIO implementation that will queue each request internally, and make sure jobs complete in the right order. Using FileBlockIO or InMemoryBlockIO will not, and for those to remain correct they need to return `BlockReqWaiter::immediate`. Take this example again:

```rust
let mut waiters: Vec<BlockReqWaiter> = Vec::with_capacity(MAGIC_NUMBER);

// Submit multiple writes to in memory block io
for buf in bufs {
  let waiter = in_memory.write(offset, buf.freeze()).await?; // returns BlockReqWaiter::immediate
  waiters.push(waiter);
}

// wait on all the BlockReqWaiter::immediate (in any order :))
let results: Vec<Result<(), CrucibleError> = futures::future::join_all(waiters).await;
results.collect::<Result<Vec<()>, CrucibleError>()?
```

Using BlockReqWaiter in this way means that requests are (big hand wave) "submitted in order to InMemoryBlockIO", because the call to write is actually performing the operation. The code that's calling it still has to wait on the BlockReqWaiters - the mental model remains the same. It's important that each implementation of BlockIO have the same behaviour like this, because otherwise we violate [the Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle).

Based on all this, I think we have to separate sending the request and waiting for the answer, which means we'll be writing `.await.wait().await` a lot unfortunately. I pushed another bunch of commits to this branch, one of which copies the integration tests for "join_all" and instead test `wait_all`, which waits on all BlockReqWaiters in parallel. This has not failed for me after a few hours of running in a loop.

jmpesp:
Ready for re-review now :)

integration_tests/src/lib.rs
```diff
@@ -2080,56 +2085,205 @@ mod test {
...
+        volume.flush(None).await?.wait().await?;
+
+        Ok(())
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 8)]
+    async fn integration_test_wait_all_is_ordered_for_writes() -> Result<()> {
+        // In a loop, test that submitting a big batch of operations and waiting
+        // on their BlockReqWaiters in parallel is correct. It's important that
+        // this test use multiple worker_threads!
```
leftwo, line 2146: To be clear I understand, this test is submitting overlapping writes, but making sure that they
are performed in the order they were submitted?

upstairs/src/lib.rs
```diff
@@ -132,6 +174,22 @@ pub trait BlockIO: Sync {
     }
 }
 
+/// Await on multiple BlockReqWaiters in parallel
+#[inline]
+pub async fn wait_all(
```
leftwo, line 179: By using the `wait_all`, does it mean that the work that was put into the `iter` won't start until the `wait_all` is called?

upstairs/src/lib.rs
```diff
@@ -6986,6 +6916,16 @@ impl BlockIO for Guest {
         }
     }
 
+    async fn deactivate(&self) -> Result<BlockReqWaiter, CrucibleError> {
+        // Disable any more IO from this guest and deactivate the downstairs.
+        // We can't deactivate if we are not yet active.
+        if !self.is_active().await {
```
leftwo, line 6922: This is checking  the `Guest` local `is_active`, right?
I worry that the guest level `is_active` can be out of sync.
Can we just let the IO through here and if the Upstairs is not active have it return the error?

It may mean some more logic changing around this, and maybe candidate for a follow up PR?

upstairs/src/lib.rs
```diff
@@ -7007,34 +6947,99 @@ impl BlockIO for Guest {
         offset: Block,
         data: Buffer,
     ) -> Result<BlockReqWaiter, CrucibleError> {
-        self.read(offset, data).await
+        if !self.is_active().await {
+            return Err(CrucibleError::UpstairsInactive);
+        }
+
+        let bs = self.query_block_size().await?;
```
leftwo, line 6954: Ah, is this why the check for is_active?  Because we need to know the block size?

upstairs/src/volume.rs
```diff
@@ -457,6 +457,18 @@ impl BlockIO for Volume {
         Ok(())
     }
 
+    async fn deactivate(&self) -> Result<BlockReqWaiter, CrucibleError> {
+        for sub_volume in &self.sub_volumes {
```
leftwo, line 461: We wait here on each subvolume and ROP, and only return from this method when all subvolumes and ROPs have returned, right?

upstairs/src/lib.rs
jmpesp: In this case, this is waiting on the operations completing, not starting.

integration_tests/src/lib.rs
jmpesp: Correct.

upstairs/src/volume.rs
jmpesp: Correct

upstairs/src/lib.rs
jmpesp: Yep :)

jmpesp:
As discussed in the storage meeting today, I am going to re-remove `BlockReqWaiter`, and clearly note that a `BlockIO` implementation is going to behave like a physical disk: it will **not** provide any contract about what order operations are performed in. What we've previously implemented regarding flushes and crash safety will still apply.

Here are some relevant lines from the NVMe and Virtio specifications, emphasis mine:

From NVMe 1.0e specification, section 7.2.1, step 4:

> The controller performs command arbitration. **The controller may execute commands out of order**. The command arbitration mechanism selects the next command to execute from commands that have been previously fetched. The controller then proceeds with execution of the command.

From the Virtio 1.1 specification, section 5.2:

> The virtio block device is a simple virtual block device (ie. disk). Read and write requests (and other exotic requests) are placed in the queue, and serviced (**probably out of order**) by the device except where noted.

jmpesp:
Again, ready for re-review! :)

crudd/src/main.rs
```diff
@@ -124,8 +124,11 @@ async fn cmd_read<T: BlockIO>(opt: &Opt, crucible: Arc<T>) -> Result<()> {
     // interested in seeing how the pipeline length affects performance. I
     // don't think it'll make a difference beyond 2, (one reading from
     // crucible, one writing to output), but we'll see!
+    //
```
leftwo, line 127: Maybe @faithanalog has some thoughts here on how this comment should be updated?

leftwo (approved):
Just one nit from me.

crudd/src/main.rs
jmpesp: I opened https://github.com/oxidecomputer/crucible/issues/459 to track this.

crutest/src/main.rs
```diff
@@ -2043,18 +2001,16 @@ async fn repair_workload(
                     c,
                     count,
                     offset.value,
-                    data.len(),
+                    data.len().await,
                     width = count_width,
                 );
-                let waiter = guest.read(offset, data.clone())?;
-                waiterlist.push(waiter);
+                let future = guest.read(offset, data.clone());
```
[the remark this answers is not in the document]
jmpesp: using the "let future =" pattern in 1adf214fd076981c351d7cd45ede88f153992dc9

upstairs/src/lib.rs
jmpesp: I agree. I opened https://github.com/oxidecomputer/crucible/issues/461 to track this

jmpesp:
Note: I'm able to reproduce the `test-repair` failure locally with this PR, but not with the changes that I'm going to push as a follow up to fix the integration test hangs.

faithanalog:
sorry I actually gave the code a look a couple days ago, but should have posted something instead of just waiting ages for my benchmarks to actually finish. Anyway, all looks good to me! See some numbers over in https://github.com/oxidecomputer/crucible/issues/459 .

faithanalog (approved):

PR #453, opened 2022-09-22
Merge buildomat image job with build-release

leftwo:
Combine the two buildomat jobs into one.
Should make the overall CI process for crucible run a little faster, or at least
use less resources.


jclulow (approved):
Seems reasonable!  What does this do to the runtime of the `build-release` job?

leftwo:
> Seems reasonable! What does this do to the runtime of the `build-release` job?

Looking at several recent build times, they were all within 60 seconds of each other (18 - 19 minutes) and the `build-release` with this change was not the longest of the samples I looked at.

The additional work that build-release job now has is to tar up and then transfer the archive, so I would think <30 seconds to do all that.

PR #458, opened 2022-09-23
leave no BlockReqWaiter behind

leftwo:
A Merge missed this new call to flush

### DIFF
--- a/upstairs/src/volume.rs
+++ b/upstairs/src/volume.rs
@@ -380,7 +380,7 @@ impl Volume {
                 total_time.as_secs(),
                 retries,
             );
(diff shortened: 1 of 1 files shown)


jmpesp:
Note https://github.com/oxidecomputer/crucible/pull/449/files#diff-4bb6901cd12d8aa6df6a97b8db98b599d0afbbc80fc95e656bdfa6ae01f7cbf4R360 is the same fix

leftwo:
Ah yeah, I noticed when running your build that I don't get the warning.

leftwo:
Not needed now that async is back

Issue #447, opened 2022-09-20
Add slog to crucible


leftwo:
Tracker issue for slog

- [x] Upstairs and initial changes
- [x] logs in other Upstairs sections (mend, volume)
- [x] Downstairs
- [ ] Tasks have unique log tags
- [ ] Guest specific log (Maybe).
- [ ] DTrace debug log that have value

Some additional discussion should happen around what format we should use and how best to play well with others.

leftwo:
Upstairs now takes a `&Logger` struct on startup.
Downstairs and upstairs now default to the same type of logger config that Propolis uses.

PR #460, opened 2022-09-28
slog in the downstairs

leftwo:
This adds slog logging to the downstairs and replaces most of the println calls with slog.

I chose to add a Logger field to the region structure as well as to the Downstairs structure.  This resulted in the least amount of having to pass around a Logger element.

build_downstairs_for_region uses an Optional log because the downstairs admin dropshot params can't take a log arg through the same way it takes other args.  There may be another way to pass it, but it was not obvious given the wrappers that dropshot provides that does everything for you.

A smattering of tests also need the Logger element, so there are a few csl functions (create simple logger) that is used to satisfy the need for a Logger.

I removed a few println's that were really there just for debugging.

For the upstairs, there were still some println's in the scrubber code that I did not deal with the last time.  The scrub method now requires a Logger so it knows where to send its messages.

There are still a few println's in the BlockIO for Guest section I'll need to figure out how to get a Logger, but I'm doing that in another PR.


leftwo:
Another piece of: https://github.com/oxidecomputer/crucible/issues/447

faithanalog:
This looks good for me but I have some questions about a few things that aren't using the logger, and whether they should. The first is in dump.rs. There's a number of functions that take a logger as input, but do a bunch of print!ing instead. Some lines:

```
src/dump.rs
221:                    println!(
268:                println!();
319:                println!(" <---");
322:                println!("{}", sgr(0, nc));
328:            println!("No data for {}", en);
332:    println!("Max gen: {},  Max flush: {}", max_gen, max_flush);
459:    println!("           Extent {}", cmp_extent);
469:    println!();
479:    println!();
495:    println!();
496:    println!();
514:    println!();
617:            println!();
647:    println!(
651:    println!();
688:        println!("{:>6}  {:<64}  {:3}", "DATA", "SHA256", "VER");
689:        println!(
698:            println!(
705:        println!();
733:        println!();
742:        println!();
765:            println!();
767:        println!();
787:        println!();
796:        println!();
819:            println!();
821:        println!();
846:        println!();
855:        println!();
877:            println!();
879:        println!();
```

The other is in lib.rs, where im specifically wondering
- should show_work should be using a logger instead of print!ing?
- should downstairs_import / downstairs_export use a logger?

My gut feeling is show_work should use the logger. But I don't have opinions on the stuff in dump.rs or the import/export, just wondering whether they're an intentional omission or not.

leftwo:
Thanks for the comments,

> This looks good for me but I have some questions about a few things that aren't using the logger, and whether they should. The first is in dump.rs. There's a number of functions that take a logger as input, but do a bunch of print!ing instead. Some lines:
> 

Yeah, the "dump" command is something we use while debugging at the command line.  I don't see sled-agent ever calling that command.  I added `log` to some functions so I could just pass them on through to other functions that dump uses to get data before printing it.  It was either that, or make the functions all use `Option<Logger>`, and given that this was just a test program I kept it simple and just passed the log through.

> 
> The other is in lib.rs, where im specifically wondering
> 
> * should show_work should be using a logger instead of print!ing?

Ah well, let me try and explain that one also,  I have a few reasons :)
This is another "debug only" sort of tool, and is not part of the normal IO path, so propolis should not be sending the command to execute it (and the other places where it's called are commented out).
Also, I've spent hours (too many, yes) getting the columns to all line up and look pretty, and trying to stuff all that formatting magic into slog, I was unsure if it would be worth it (or even possible given my ability to format slog messages).

Behold it in all its beauty:
```
 Crucible gen:15 GIO:true work queues:  Upstairs:0  downstairs:15
GW_ID      ACK  DSID    TYPE  BLOCKS  DS:0  DS:1  DS:2  REPLAY
  125    Acked  1124    Read     100  Done  Done  Done  false
  126    Acked  1125    Read     100  Done  Done  Done  false
  127    Acked  1126    Read     100  Done  Done  Done  false
  128    Acked  1127    Read     100  Done  Done  Done  false
  129    Acked  1128    Read     100  Done  Done  Done  false
  130    Acked  1129    Read     100  Done  Done  Done  false
  131    Acked  1130    Read     100  Done  Done  Done  false
  132    Acked  1131    Read     100  Done  Done  Done  false
  133    Acked  1132    Read     100  Done  Done  Done  false
  134    Acked  1133    Read     100  Done  Done  Done  false
  135    Acked  1134    Read     100  Done  Done  Done  false
  136    Acked  1135    Read     100  Done  Done  Done  false
  137    Acked  1136    Read     100  Done  Done  Done  false
  138    Acked  1137    Read     100  Done  Done  Done  false
  139    Acked  1138    Read     100  Done  Done  Done  false
   STATES      DS:0   DS:1   DS:2   TOTAL
    New           0      0      0      0
    Sent          0      0      0      0
    Done         15     15     15     45
    Skipped       0      0      0      0
    Error         0      0      0      0
Last Flush: 1123 1123 1123
Downstairs last five completed: 1123 1122 1121 1120 1119 1118
Upstairs last five completed:    139  138  137  136  135  134
CLIENT: Up:0 ds:15
```

> * should downstairs_import / downstairs_export use a logger?

I also don't think these are tools that will ever run through sled-agent or anything automated.  Being a command line only manual tool, I did not think it worth adding them either.

> My gut feeling is show_work should use the logger. But I don't have opinions on the stuff in dump.rs or the import/export, just wondering whether they're an intentional omission or not.

Let me know if my reasons have swayed you at all, and thanks for the review!

faithanalog (approved):

PR #464, opened 2022-09-30
Return JoinHandles when spawning our services

jmpesp:
Previous to this commit, the integration tests would "randomly" hang. This was due to our port selection for downstairs being "fixed" in the integration test (with required collision detection), but but the Upstairs gets a random port when it opens its connection to the downstairs. The hang occurs when the Upstairs is randomly assigned a port that one of the integration tests is going to use to spawn the downstairs.

Say there are two integration tests. Test A uses ports X, X+1, and X+2 for the downstairs, and test B uses X+3, X+4, X+5. The hang occurs when:

- Test A runs and spawns three downstairs successfully at X, X+1, and X+2.

- Test A spawns an upstairs, which is assigned port X+4, to talk to X, X+1, and X+2.

- Test B tries to spawn three downstairs, but can only spawn X+3 and X+5 because X+4 is taken.

- Test B spawns an upstairs, which is assigned some random port, to talk to X+3, X+4, and X+5.

When Test B's upstairs tries to activate, it will send the HereIAm message to the three targets defined in the Crucible opts. Test B sends HereIAm to X+3, X+4, and X+5, where X+4 in this case is another Upstairs, leading to the very confusing error message:

> ERRO 127.0.0.1:55016: proc: [0] unexpected command HereIAm { ... }
> received in state New

Notice in this case the Upstairs' port is 55016, the same as one of the integration tests (before this commit). Activation cannot occur without three downstairs, and the test hangs.

The fix is to not set the downstair's ports and instead grab whatever is available. By passing 0 into start_downstairs, the operating system will choose a port from the ephemeral range. This necessitated storing the SocketAddr the downstairs is listening on in the Downstairs struct for later querying.

I've tested this for 24h and it ran without hanging. But, there is some additional refactoring that this commit introduces to make this scenario less likely in the future.

When calling up_main or start_downstairs, return a JoinHandle that consumers can await on instead of never returning anything. This can be used to signal that the creation of the Upstairs or Downstairs was successful, and if that is not returned then that is an error that can be dealt with accordingly. This would have prevented our integration tasks from hanging: `start_downstairs` would have returned an Err instead of Ok(JoinHandle) because it wouldn't have been able to listen on the port. Notably, if Ok(JoinHandle) was returned, then the address field in Downstairs will be Some.

One additional complication here: having the operating system choose from the ephemeral range means that it may choose a port that is too high to add REPAIR_PORT_OFFSET to in order to choose a repair port.

So this commit **also** changes how repair ports work: start_downstairs now accepts an rport argument. If it's set to non-zero, then use that port for the repair server address. If it's zero, then pick an open port for the repair server. This necessitated the downstairs sending the upstairs what port it chose for the repair server, which caused a bit of a jumble around of things BUT means that the only place where REPAIR_PORT_OFFSET is used now is when running the downstairs from the command line.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -783,6 +783,10 @@ where
         + std::marker::Send
         + 'static,
 {
+    // In this function, repair address should exist, and shouldn't change. Grab
+    // it here.
+    let repair_addr = ads.lock().await.repair_address.unwrap();
+
     let mut negotiated = 0;
     let mut upstairs_connection: Option<UpstairsConnection> = None;
 
@@ -946,7 +950,7 @@ where
                             upstairs_connection.unwrap());
 
                         let mut fw = fw.lock().await;
-                        fw.send(Message::YesItsMe { version: 1, repair_addr: ads.lock().await.repair_address.unwrap() }).await?;
+                        fw.send(Message::YesItsMe { version: 1, repair_addr }).await?;
                     }
                     Some(Message::PromoteToActive {
                         upstairs_id,
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -2307,15 +2307,19 @@ pub async fn start_downstairs(
         IpAddr::V6(ipv6) => SocketAddr::new(std::net::IpAddr::V6(ipv6), rport),
     };
 
-    if let Err(e) = repair::repair_main(&d, repair_address).await {
-        // TODO tear down other things if repair server can't be started?
-        bail!("got {:?} from repair main", e);
-    }
+    let repair_listener = match repair::repair_main(&d, repair_address).await {
+        Err(e) => {
+            // TODO tear down other things if repair server can't be started?
+            bail!("got {:?} from repair main", e);
+        }
+
+        Ok(socket_addr) => socket_addr,
+    };
 
     {
         let mut ds = d.lock().await;
-        ds.repair_address = Some(repair_address);
-        println!("Using repair address: {:?}", repair_address);
+        ds.repair_address = Some(repair_listener);
+        println!("Using repair address: {:?}", repair_listener);
     }
 
     // Optionally require SSL connections
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -946,7 +946,7 @@ where
                             upstairs_connection.unwrap());
 
                         let mut fw = fw.lock().await;
-                        fw.send(Message::YesItsMe { version: 1 }).await?;
+                        fw.send(Message::YesItsMe { version: 1, repair_addr: ads.lock().await.repair_address.unwrap() }).await?;
                     }
                     Some(Message::PromoteToActive {
                         upstairs_id,
@@ -1311,6 +1311,8 @@ pub struct Downstairs {
     dss: DsStatOuter,
     read_only: bool,
     encrypted: bool,
+    pub address: Option<SocketAddr>,
+    pub repair_address: Option<SocketAddr>,
 }
 
 impl Downstairs {
@@ -1334,6 +1336,8 @@ impl Downstairs {
             dss,
             read_only,
             encrypted,
+            address: None,
+            repair_address: None,
         }
     }
 
@@ -2246,15 +2250,21 @@ pub fn build_downstairs_for_region(
     ))))
 }
 
+/// Returns Ok if everything spawned ok, Err otherwise
+///
+/// Return Ok(main task join handle) if all the necessary tasks spawned
+/// successfully, and Err otherwise.
+#[allow(clippy::too_many_arguments)]
 pub async fn start_downstairs(
     d: Arc<Mutex<Downstairs>>,
     address: IpAddr,
     oximeter: Option<SocketAddr>,
     port: u16,
+    rport: u16,
     cert_pem: Option<String>,
     key_pem: Option<String>,
     root_cert_pem: Option<String>,
-) -> Result<()> {
+) -> Result<tokio::task::JoinHandle<Result<()>>> {
     if let Some(oximeter) = oximeter {
         let dssw = d.lock().await;
         let dss = dssw.dss.clone();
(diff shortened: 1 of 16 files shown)


jmpesp:
Note: I've just seen

```
Error: Initial volume verify failed: Error in range 500 -> 600
Exit on verify fail, loop: 0, choice: 0
Check /tmp/verify_out.txt for details
```

locally from `test_repair.sh`, :-1:

leftwo:
> Note: I've just seen
> 
> ```
> Error: Initial volume verify failed: Error in range 500 -> 600
> Exit on verify fail, loop: 0, choice: 0
> Check /tmp/verify_out.txt for details
> ```
> 
> locally from `test_repair.sh`, 👎

Yeah, I see this also.  I'm testing with a stock repo, so it's nothing in this PR.

leftwo:
Do the proposed async changes from todays (10/3/2022) storage meeting have any impact on the `JoinHandles` here?

jmpesp:
> Do the proposed async changes from todays (10/3/2022) storage meeting have any impact on the JoinHandles here?

I don't think so, no

upstairs/src/lib.rs
```diff
@@ -893,7 +858,47 @@ where
                             ).await;
                             bail!("expected version 1, got {}", version);
                         }
+
                         negotiated = 1;
+
+                        up.ds_repair_address(up_coms.client_id, repair_addr).await;
+
+                        // If this downstairs is returning, then attempt repair
+                        {
```
leftwo, line 867: I'm curious why you moved this inside here and away from the top of the function.  I don't know what the impact of moving the `ds.re_new()` call to this phase of negotiation will have, and by moving it down we require getting to this phase of negotiation before we re_new the work on the work queue.  There is also some coming work that will handle that panic case (the return of a failed downstairs).

If it's just because we want to print the repair address and we don't have the repair address yet, then we don't have to print it.  I was going to remove that anyway, but if we do want to print, we can just add a message inside this match to print it.

Also, nit:  > 80 characters.  Moving it back would fix that :)

upstairs/src/lib.rs
```diff
@@ -2248,20 +2254,10 @@ struct Downstairs {
...
-            let port = addr.port() + REPAIR_PORT_OFFSET;
-            let repair_addr = SocketAddr::new(addr.ip(), port);
-            ds_repair.insert(i as u8, repair_addr);
-        }
-
+    fn new(log: Logger) -> Self {
         Self {
             ds_uuid: HashMap::new(),
-            ds_repair,
+            ds_repair: HashMap::new(),
```
leftwo, line 2260: Not for this PR, but man this seems ripe for a refactor.
Every element in this is a tuple or hash map of the three downstairs.  It seems like there could be a better way to do this.  Some are index based, and some by a hash key, it's.. it's not my proudest moment.

upstairs/src/lib.rs
```diff
@@ -5567,6 +5561,11 @@ impl Upstairs {
 
         Ok(notify_guest)
     }
+
+    async fn ds_repair_address(&self, client_id: u8, addr: SocketAddr) {
+        let mut ds = self.downstairs.lock().await;
+        ds.ds_repair.insert(client_id, addr);
```
leftwo, line 5567: I take it there is no issue with replacement here?  Downstairs will come and go and the repair address could be different each time, we would need to handle that.  Maybe we want to consider removing it as part of the re-attachment process?

upstairs/src/lib.rs
```diff
@@ -5567,6 +5561,11 @@ impl Upstairs {
 
         Ok(notify_guest)
     }
+
+    async fn ds_repair_address(&self, client_id: u8, addr: SocketAddr) {
```
leftwo, line 5565: This name, maybe call it `ds_set_repair`? Or something to indicate this is setting the repair address?

It is a little confusing that there is a `Downstairs` `repair_addr` method that will return the repair address for a given client ID, and an `Upstairs` method (this here) which will set the repair address.

leftwo (commented):
Some thoughts, but in general this is great.  Looking forward to having tests pass again.

downstairs/src/lib.rs
```diff
@@ -2277,29 +2287,42 @@ pub async fn start_downstairs(
...
-    /*
-     * Establish a listen server on the port.
-     */
-    println!("Using address: {:?}", listen_on);
-    let listener = TcpListener::bind(&listen_on).await?;
+    {
+        let mut ds = d.lock().await;
+        ds.repair_address = Some(repair_listener);
+        println!("Using repair address: {:?}", repair_listener);
+    }
```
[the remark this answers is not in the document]
jmpesp: :+1:

integration_tests/src/lib.rs
```diff
@@ -12,44 +13,25 @@ mod test {
...
-            {
-                let mut ports = PORTS.lock().await;
-                if !ports.insert(port) {
-                    panic!("port {} reused!", port);
-                }
-            }
-
             let tempdir = tempfile::Builder::new()
-                .prefix(&format!("downstairs-{}", port))
+                .prefix(&"downstairs-")
```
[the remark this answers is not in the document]
jmpesp: :-D

upstairs/src/lib.rs
jmpesp: yeah, I think I see what you're getting at here. I'm thinking that I should remove the previous address in `re_new`, does that sound right?

downstairs/src/repair.rs
```diff
@@ -40,10 +40,11 @@ fn build_api() -> ApiDescription<FileServerContext> {
     api
 }
 
+/// Returns Ok if everything launched ok, Err otherwise
```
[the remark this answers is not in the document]
jmpesp: yep, corrected in 05f0248

upstairs/src/lib.rs
jmpesp: I was under the impression that `ds.re_new()` performed some part of the repair but that's wrong. Undid the move in 0c041ea

upstairs/src/lib.rs
jmpesp: done in 2585e35

upstairs/src/lib.rs
jmpesp: sounds good, done in a97e6cd

PR #469, opened 2022-10-04
Crash consistent removal of block contexts

jmpesp:
Previous to this commit, `truncate_encryption_contexts_and_hashes` was removing all but the last entered encryption context and hash rows for any given block, and this is incorrect - **any** of the encryption context and hash rows could be valid in the presense of a crash. The only guarantee that the OS gives us is that after a flush, the last write to an extent will be on the physical disk, but any of the extent writes could be on the disk between flushes.

This commit adds an on-disk hash to a block's context that the Downstairs uses to remove rows that do not match what is on disk after a flush. It also combines the encryption context and hash tables into one called block_context, and similarly combines the encryption context and hash that are fields in ReadResponse and Write into a BlockContext.

I originally thought this commit would degrade performance, but using crudd to write and then read 1GiB shows an average write bandwidth of 56.5MiB/s and an average read bandwidth of 68.6MiB/s! I'm sure that there is more potential for improvement, namely in how truncate_encryption_contexts_and_hashes is called.

### DIFF
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -38,6 +38,15 @@ pub struct Extent {
     inner: Option<Mutex<Inner>>,
 }
 
+/// BlockContext, with the addition of block index and on_disk_hash
+#[derive(Clone)]
+pub struct DownstairsBlockContext {
+    pub block_context: BlockContext,
+
+    pub block: u64,
+    pub on_disk_hash: u64,
+}
+
 #[derive(Debug)]
 pub struct Inner {
     file: File,
@@ -129,61 +138,31 @@ impl Inner {
         Ok(())
     }
 
-    /// For a given block range, return all encryption contexts since the last
-    /// flush. `get_hashes` returns a `Vec<Vec<u64>>` of length equal to
-    /// `count`. Each `Vec<u64>` inside this parent Vec contains all
-    /// contexts for a single block, ordered so the latest context is last.
-    /// If the region is not using encryption, the inner Vecs will all be
-    /// empty, but the outer Vec will still be of length `count`.
-    fn get_encryption_contexts(
+    /// For a given block range, return all context rows since the last flush.
+    /// `get_block_contexts` returns a `Vec<Vec<DownstairsBlockContext>>` of
+    /// length equal to `count`. Each `Vec<DownstairsBlockContext>` inside this
+    /// parent Vec contains all contexts for a single block.
+    fn get_block_contexts(
         &self,
         block: u64,
         count: u64,
-    ) -> Result<Vec<Vec<EncryptionContext>>> {
+    ) -> Result<Vec<Vec<DownstairsBlockContext>>> {
         // NOTE: "ORDER BY RANDOM()" would be a good --lossy addition here
-        let stmt = "SELECT block, nonce, tag FROM encryption_context \
+        let stmt =
+            "SELECT block, hash, nonce, tag, on_disk_hash FROM block_context \
              WHERE block BETWEEN ?1 AND ?2 \
-             ORDER BY counter ASC";
+             ORDER BY ROWID ASC";
         let mut stmt = self.metadb.prepare_cached(stmt)?;
 
         let stmt_iter =
             stmt.query_map(params![block, block + count - 1], |row| {
-                let block: u64 = row.get(0)?;
-                let nonce: Vec<u8> = row.get(1)?;
-                let tag: Vec<u8> = row.get(2)?;
-                Ok((block, nonce, tag))
-            })?;
-
-        let mut results = Vec::with_capacity(count as usize);
-        for _i in 0..count {
-            results.push(Vec::new());
-        }
-
-        for row in stmt_iter {
-            let (row_block, nonce, tag) = row?;
-            results[(row_block - block) as usize]
-                .push(EncryptionContext { nonce, tag });
-        }
-
-        Ok(results)
-    }
-
-    /// For a given block range, return all hashes since the last flush.
-    /// `get_hashes` returns a `Vec<Vec<u64>>` of length equal to `count`. Each
-    /// `Vec<u64>` inside this parent Vec contains all hashes for a single
-    /// block, ordered so the latest hash is last.
-    pub fn get_hashes(&self, block: u64, count: u64) -> Result<Vec<Vec<u64>>> {
-        // NOTE: "ORDER BY RANDOM()" would be a good --lossy addition here
-        let stmt = "SELECT block, hash FROM integrity_hashes \
-             WHERE block BETWEEN ?1 AND ?2 \
-             ORDER BY counter ASC";
-        let mut stmt = self.metadb.prepare_cached(stmt)?;
-
-        let stmt_iter =
-            stmt.query_map(params![block, block + count - 1], |row| {
-                let block: u64 = row.get(0)?;
+                let block_index: u64 = row.get(0)?;
                 let hash: Vec<u8> = row.get(1)?;
-                Ok((block, hash))
+                let nonce: Option<Vec<u8>> = row.get(2)?;
+                let tag: Option<Vec<u8>> = row.get(3)?;
+                let on_disk_hash: Vec<u8> = row.get(4)?;
+
+                Ok((block_index, hash, nonce, tag, on_disk_hash))
             })?;
 
         let mut results = Vec::with_capacity(count as usize);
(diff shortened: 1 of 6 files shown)


leftwo:
Could you share some outputs from `crucible-downstairs dump` with the new code?  I'd like to see how it looks with these changes.

jmpesp:
Sure thing:

```
     Running `target/debug/crucible-downstairs dump -d disks/d1 -d disks/d2 -d disks/d3`
EXT          BLOCKS GEN0 GEN1 GEN2  FL0 FL1 FL2  D0 D1 D2
  0 0000000-0040959    1    1    1    1   1   1   F  F  F
  1 0040960-0081919    1    1    1    1   1   1   F  F  F
  2 0081920-0122879    1    1    1    1   1   1   F  F  F
  3 0122880-0163839    1    1    1    1   1   1   F  F  F
  4 0163840-0204799    1    1    1    1   1   1   F  F  F
  5 0204800-0245759    1    1    1    1   1   1   F  F  F
  6 0245760-0286719    1    1    1    1   1   1   F  F  F
  7 0286720-0327679    1    1    1    1   1   1   F  F  F
  8 0327680-0368639    1    1    1    2   2   2   F  F  F
  9 0368640-0409599    1    1    1    2   2   2   F  F  F
 10 0409600-0450559    1    1    1    2   2   2   F  F  F
 11 0450560-0491519    1    1    1    2   2   2   F  F  F
 12 0491520-0532479    1    1    1    2   2   2   F  F  F
 13 0532480-0573439    1    1    1    2   2   2   F  F  F
 14 0573440-0614399    1    1    1    2   2   2   F  F  F
 15 0614400-0655359    1    1    1    2   2   2   F  F  F
 16 0655360-0696319    1    1    1    2   2   2   F  F  F
 17 0696320-0737279    1    1    1    2   2   2   F  F  F
 18 0737280-0778239    1    1    1    2   2   2   F  F  F
 19 0778240-0819199    1    1    1    2   2   2   F  F  F
 20 0819200-0860159    1    1    1    2   2   2   F  F  F
 21 0860160-0901119    1    1    1    2   2   2   F  F  F
 22 0901120-0942079    1    1    1    2   2   2   F  F  F
 23 0942080-0983039    1    1    1    2   2   2   F  F  F
 24 0983040-1023999    1    1    1    2   2   2   F  F  F
 25 1024000-1064959    1    1    1    2   2   2   F  F  F
 26 1064960-1105919    1    1    1    2   2   2   F  F  F
 27 1105920-1146879    1    1    1    2   2   2   F  F  F
 28 1146880-1187839    1    1    1    3   3   3   F  F  F
 29 1187840-1228799    1    1    1    3   3   3   F  F  F
 30 1228800-1269759    1    1    1    3   3   3   F  F  F
 31 1269760-1310719    1    1    1    3   3   3   F  F  F
 32 1310720-1351679    1    1    1    3   3   3   F  F  F
 33 1351680-1392639    1    1    1    3   3   3   F  F  F
 34 1392640-1433599    1    1    1    3   3   3   F  F  F
 35 1433600-1474559    1    1    1    3   3   3   F  F  F
 36 1474560-1515519    1    1    1    3   3   3   F  F  F
 37 1515520-1556479    1    1    1    3   3   3   F  F  F
 38 1556480-1597439    1    1    1    3   3   3   F  F  F
 39 1597440-1638399    1    1    1    3   3   3   F  F  F
 40 1638400-1679359    1    1    1    3   3   3   F  F  F
 41 1679360-1720319    1    1    1    3   3   3   F  F  F
 42 1720320-1761279    1    1    1    3   3   3   F  F  F
 43 1761280-1802239    1    1    1    3   3   3   F  F  F
 44 1802240-1843199    1    1    1    3   3   3   F  F  F
 45 1843200-1884159    1    1    1    4   4   4   F  F  F
 46 1884160-1925119    1    1    1    4   4   4   F  F  F
 47 1925120-1966079    1    1    1    4   4   4   F  F  F
 48 1966080-2007039    1    1    1    4   4   4   F  F  F
 49 2007040-2047999    1    1    1    4   4   4   F  F  F
 50 2048000-2088959    1    1    1    4   4   4   F  F  F
Max gen: 1,  Max flush: 4

     Running `target/debug/crucible-downstairs dump -d disks/d1 -d disks/d2 -d disks/d3 -e 0`
           Extent 0
GEN             1        1        1 
FLUSH_ID        1        1        1 
DIRTY                               

BLOCK  D0 D1 D2  C0 C1 C2 DIFF 
    0   A  A  A   A  A  A     
    1   A  A  A   A  A  A     
    2   A  A  A   A  A  A     
    3   A  A  A   A  A  A     
    4   A  A  A   A  A  A     
    5   A  A  A   A  A  A     
    6   A  A  A   A  A  A     
    7   A  A  A   A  A  A     
    8   A  A  A   A  A  A     
    9   A  A  A   A  A  A     
   10   A  A  A   A  A  A     
...

     Running `target/debug/crucible-downstairs dump -d disks/d1 -d disks/d2 -d disks/d3 -b 128`
Extent 0 Block in extent 128  Actual block 128

  DATA  SHA256                                                            VER
------  ----------------------------------------------------------------  ---
     0  978dd917d1a8e628f764764c25e0e38f1a92ee43cde90ca1e0761c2382acda24  A
     1  978dd917d1a8e628f764764c25e0e38f1a92ee43cde90ca1e0761c2382acda24  A
     2  978dd917d1a8e628f764764c25e0e38f1a92ee43cde90ca1e0761c2382acda24  A

NONCES             0                        1                        2              DIFF 
------  ------------------------ ------------------------ ------------------------ ----- 
     0  6515a59872dbcdca600e3c7b 6515a59872dbcdca600e3c7b 6515a59872dbcdca600e3c7b 

  TAGS                 0                                1                                2                  DIFF 
------  -------------------------------- -------------------------------- -------------------------------- ----- 
     0  906364dbec6c51a9d55da31d7eb94728 906364dbec6c51a9d55da31d7eb94728 906364dbec6c51a9d55da31d7eb94728 

HASHES         0                1                2          DIFF 
------  ---------------- ---------------- ---------------- ----- 
     0  605528d0643778aa 605528d0643778aa 605528d0643778aa 

```

jmpesp:
There's one difference in the extent specific output because of the combination of encryption context and hash into block context, so there is only one column per downstairs now.

leftwo:
I think it gives an error if there is no data yet:
```
final:jmpespfork$ ./target/release/crucible-downstairs dump -d /var/tmp/dsc/region/8810 -d /var/tmp/dsc/region/8820 -d /var/tmp/dsc/region/8830/ -e 0
           Extent 0
FLUSH_ID        0        0        0
DIRTY

BLOCK  D0 D1 D2  C0 C1 C2 DIFF
Error: Error: no such table: block_context

Caused by:
    Error code 1: SQL error or missing database
```

But the original dump does not:
```
final:crucible$ ./target/release/crucible-downstairs dump -d /var/tmp/dsc/region/8810 -d /var/tmp/dsc/region/8820 -d /var/tmp/dsc/region/8830/ -e 0
           Extent 0
FLUSH_ID        0        0        0
DIRTY

BLOCK  D0 D1 D2  E0 E1 E2  H0 H1 H2 DIFF
    0   A  A  A   A  A  A   A  A  A
    1   A  A  A   A  A  A   A  A  A
    2   A  A  A   A  A  A   A  A  A
    3   A  A  A   A  A  A   A  A  A
    4   A  A  A   A  A  A   A  A  A
```

Same is true for `-b` option, it fails and exits when it can't read the database.

leftwo:
Deleted my comment where I thought an "empty" volume was returning errors.
It turns out the region was created with the older format and the database lookup failed because of different column names.

upstairs/src/lib.rs
```diff
@@ -2873,107 +2887,121 @@ impl Downstairs {
...
+
+            // Validate integrity hash before decryption
+            let computed_hash = integrity_hash(&[
+                &block_encryption_ctx.nonce[..],
+                &block_encryption_ctx.tag[..],
+                &response.data[..],
+            ]);
+
+            if computed_hash == ctx.hash {
+                successful_hash = true;
```
leftwo, line 2936: No matter what happens below this, we are exiting the `for` loop here.

Could there be a situation where the `computed_hash == ctx.hansh`, but the decryption failed, and there were other `block_contexts` that we did not get to yet that did pass?

downstairs/src/region.rs
```diff
@@ -160,71 +171,59 @@ impl Inner {
...
-
         let stmt =
-            "INSERT INTO encryption_context (counter, block, nonce, tag) \
-             VALUES ( \
-                 (SELECT IFNULL(MAX(counter), 0) + 1 FROM encryption_context WHERE block=?1), \
-                 ?1, ?2, ?3 \
-             )";
+            "INSERT INTO block_context (block, hash, nonce, tag, on_disk_hash) \
+             VALUES (?1, ?2, ?3, ?4, ?5)";
+
```
leftwo, line 207: Why not the `SELECT IFNULL ..` stuff?  Is that part of what we doing differently here with the database commands?

leftwo (approved):
Just some questions, but yeah looks good.

upstairs/src/lib.rs
jmpesp: I think so, yes, if there are two block contexts with difference nonces and tags, but that both return the same `integrity_hash`:

```rust
let hash1 = integrity_hash(&[&ctx1.nonce[..], &ctx1.tag[..], &response.data[..]]);
let hash2 = integrity_hash(&[&ctx2.nonce[..], &ctx2.tag[..], &response.data[..]]);

hash1 == hash2
```

Given that xxHash is a non-cryptographic hash and that the hash value is a `u64`, collisions are possible, though unlikely (I've been running something to try and find such a collision for several hours). The fix for this is to **not** break if decryption is unsuccessful, and try the other block contexts. I've done this in 8a8bb8a

leftwo: Ah, I had not considered the collision case, but yeah, that could be possible.

Issue #461, opened 2022-09-28
Does Guest need the active field?


jmpesp:
The Guest has its own idea of being active:

```rust
    pub async fn is_active(&self) -> bool {
        // A Guest is active if it's seen the Upstairs return that it's active
        let active = self.active.lock().await;
        *active
    }
```

Does it need this? What if the Guest idea of active and the Upstairs / BlockIO idea of active go out of sync?

leftwo:
I do not think the guest needs it.
The guest should just pass the request down to the upstairs where it has the most up to date information.

I figured we would rip it out when we do the proper generation number support, but no reason this piece can't 
be fixed independently.

PR #475, opened 2022-10-04
Remove Guest's idea of active and locking in len()

jmpesp:
Guest previously had its own idea of active, and this is slightly confusing. This commit removes that, and instead lets the Upstairs tell consumers if it is active or not.

Additionally, getting the length of a Buffer shouldn't take a lock. Make `len()` and `is_empty()` synchronous by storing the length (it shouldn't change!).

Closes https://github.com/oxidecomputer/crucible/issues/461

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -82,6 +82,7 @@ pub trait BlockIO: Sync {
     // Total bytes of Volume
     async fn total_size(&self) -> Result<u64, CrucibleError>;
 
+    /// Return the block size - this should never change at runtime!
     async fn get_block_size(&self) -> Result<u64, CrucibleError>;
 
     async fn get_uuid(&self) -> Result<Uuid, CrucibleError>;
@@ -4494,7 +4495,7 @@ impl Upstairs {
         let nwo = extent_from_offset(
             *ddef,
             offset,
-            Block::from_bytes(data.len().await, &ddef),
+            Block::from_bytes(data.len(), &ddef),
         );
 
         /*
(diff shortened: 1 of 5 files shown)


upstairs/src/lib.rs
```diff
@@ -6080,12 +6085,12 @@ impl Buffer {
...
-    pub async fn len(&self) -> usize {
-        self.data.lock().await.len()
+    pub fn len(&self) -> usize {
+        self.len
     }
 
-    pub async fn is_empty(&self) -> bool {
-        self.len().await == 0
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
```
leftwo, line 6093: Is this getting the length of `Buffer`?
Should this be. `self.len == 0` ?

upstairs/src/lib.rs
```diff
@@ -7015,12 +6961,6 @@ impl BlockIO for Guest {
     }
 
     async fn show_work(&self) -> Result<WQCounts, CrucibleError> {
```
leftwo, line 6963: Could you leave behind a comment that this job will be sent even if the upstairs is not active?  Either here, or over in  `process_new_io()`?  There may come a time or reason to turn that off (or at least check)

leftwo (approved):
Do we have a test that verifies work is refused if the upstairs is not active?
It's very important that a not active upstairs do any work at all, and knowing that we had done some sanity here might prevent us debugging pain in the future.

upstairs/src/lib.rs
jmpesp: yeah, changed in a8d32c0

upstairs/src/lib.rs
jmpesp: :+1:, done in 675f812

leftwo:
Looks good, :shipit:

PR #481, opened 2022-10-05
Synchronous until send in Guest BlockIO impl

jmpesp:
Insertion order into Guest's queue is not preserved if there's an await before enqueuing to reqs. Split `get_block_size` into a synchronous BlockIO sub-trait, and store block size statically in implementations of it (it shouldn't ever change at runtime!).

This also affected the *_from_byte_offset functions, which were checking if the BlockIO impl was active befroe calling read or write. Remove this check too, so that it's synchronous code until enqueuing to reqs.


jmpesp:
I've added a few tests to ensure that "synchronous until send" means that insertion order for jobs is preserved. They currently fail :)

PR #482, opened 2022-10-06
Add validation and flush tests

jmpesp:
This commit introduces a bunch of tests for validating unencrypted and encrypted read responses ~~, and in writing those tests uncovered an issue: if a write to a blank block (where there were previously no block contexts) is not flushed, then an Upstairs with encryption will reject the block. This is fixed by detecting this case, and returning Ok(None).~~

### DIFF
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -3243,8 +3243,14 @@ mod test {
     #[test]
     fn test_flush_removes_partial_writes() -> Result<()> {
         // Validate that incorrect context rows are removed by a flush so that a
-        // repair will get rid of partial writes' block context rows (this is
-        // necessary for write_unwritten to work after a crash).
+        // repair will get rid of partial writes' block context rows. This is
+        // necessary for write_unwritten to work after a crash.
+        //
+        // Specifically, this test checks for the case where we had a brand new
+        // block and a write to that blocks that failed such that only the
+        // write's block context was persisted, leaving the data all zeros. In
+        // this case, there is no data, so we should remove the invalid block
+        // context row.
 
         let dir = tempdir()?;
         let mut region = Region::create(&dir, new_region_options())?;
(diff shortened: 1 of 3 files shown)


jmpesp:
I reverted the changes to the Upstairs' validation functions because the case I was worried about cannot exist: if there is a partial write to a blank block, then a crash (so the data isn't durable), activation will run reconciliation, which will call flush, which will remove block context rows that do not match what is on disk. This means that blank blocks will always have no context rows. It also means that write_unwritten will still work in this case, which is validated in 525cd6c678d7e0e1e6407b05393b9a51e1beef0a

upstairs/src/test.rs
```diff
@@ -448,6 +448,353 @@ mod up_test {
...
+        let successful_hash = Downstairs::validate_encrypted_read_response(
+            &mut read_response,
+            &Arc::new(context),
+            &csl(),
+        )?;
+
+        assert_eq!(successful_hash, Some(read_response_hash));
+
+        // `Downstairs::validate_encrypted_read_response` will mutate the read
+        // response's data value, make sure it decrypted
```
leftwo, line 500: Is this comment still true?

leftwo (approved):
Let's change the name of the PR here so looking back on it won't confuse us as to what this change did.

upstairs/src/test.rs
jmpesp: It is, that function calls

```rust
                let decryption_result = encryption_context.decrypt_in_place(
                    &mut response.data[..],
                    Nonce::from_slice(&block_encryption_ctx.nonce[..]),
                    Tag::from_slice(&block_encryption_ctx.tag[..]),
                );
```

downstairs/src/region.rs
```diff
@@ -3238,6 +3240,75 @@ mod test {
         Ok(())
     }
 
+    #[test]
+    fn test_flush_removes_partial_writes() -> Result<()> {
+        // Validate that incorrect context rows are removed by a flush so that a
+        // repair will get rid of partial writes' block context rows (this is
+        // necessary for write_unwritten to work after a crash).
```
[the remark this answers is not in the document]
jmpesp: agree, done in 8101e55

PR #483, opened 2022-10-13
Add encrypted option to dsc

leftwo:
This adds an `--encrypted` option to the start/create paths of `dsc`.

Updated the `tools/test_up.sh` script to use `dsc` instead of hand crafted downstairs and bash process management.

### DIFF
--- a/tools/test_up.sh
+++ b/tools/test_up.sh
@@ -67,9 +67,9 @@ case ${1} in
     "unencrypted")
         ;;
     "encrypted")
-        ssl_key=$(openssl rand -base64 32)
-        echo "Upstairs using SSL key: $ssl_key"
-        args+=( --key "$ssl_key" )
+        upstairs_key=$(openssl rand -base64 32)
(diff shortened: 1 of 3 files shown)


tools/test_up.sh
```diff
@@ -28,31 +28,49 @@ fi
...
 args=()
+dsc_args=()
 
 case ${1} in
     "unencrypted")
         ;;
     "encrypted")
-        args+=( --key "$(openssl rand -base64 32)" )
+        ssl_key=$(openssl rand -base64 32)
+        echo "Upstairs using SSL key: $ssl_key"
```
jmpesp, line 71: nit: this isn't an SSL key, it's for AES (but more generically for on-disk encryption)

jmpesp (approved):
:ship:

tools/test_up.sh
leftwo: Yeah, I guess there are other types of keys...

PR #490, opened 2022-10-24
dsc no auto restart

leftwo:
**Updates to the `dsc` test program**

Update dsc to not restart exiting downstairs by default. 
This could have hid some test issues where a downstairs would restart and the test would continue (thanks to replay) and never actually know that things had restarted.  

Update dsc to know the difference between a downstairs that exited after being killed by dsc, and a downstairs that exited because it panicked.  When a downstairs panics, the auto-restarter will not restart it. We now have a new downstairs state in `dsc` to reflect this.

Updated the dsc test script

Updated test_perf.sh and test_up.sh to be explicit to disable auto downstairs restart. If a downstairs fails, these test should also fail.

Updated test_reconnect.sh to specifically turn on auto restart as it is no longer the default.

### DIFF
--- a/dsc/src/main.rs
+++ b/dsc/src/main.rs
@@ -931,6 +931,7 @@ enum DownstairsState {
     Running,
     Exit,
     Error,
+    Failed,
 }
 
 /// Downstairs actions.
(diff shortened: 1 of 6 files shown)


jmpesp (approved):
Needs a `cargo fmt` but otherwise :shipit:

PR #497, opened 2022-11-01
Some minor updates while plumbing the scrubber in Propolis.

leftwo:
Added a method on the volume object Propolis can use to determine if a scrub is needed.

The scrubber now takes two additional options.
1. A number of seconds to delay before starting.
2. A intern scrub "pause" after ever operation.

These will help us tune the scrubber as well as let the scrubber behave differently when it's attached to propolis vs. when it's attached to the pantry.

Count needs to be an Arc so the scrubber can clone it in Propolis.

Added a log message or two when a snapshot is taken.

### DIFF
--- a/upstairs/src/volume.rs
+++ b/upstairs/src/volume.rs
@@ -8,7 +8,7 @@ use std::sync::atomic::{AtomicU32, AtomicU64, Ordering};
 
 use crucible_client_types::VolumeConstructionRequest;
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct Volume {
     uuid: Uuid,
 
@@ -24,9 +24,10 @@ pub struct Volume {
      * Each sub volume should be the same block size (unit is bytes)
      */
     block_size: u64,
-    count: AtomicU32,
+    count: Arc<AtomicU32>,
 }
 
+#[derive(Clone)]
 pub struct SubVolume {
     lba_range: Range<u64>,
     block_io: Arc<dyn BlockIO + Send + Sync>,
@@ -48,7 +49,7 @@ impl Volume {
             read_only_parent: None,
             scrub_point: Arc::new(AtomicU64::new(0)),
             block_size,
-            count: AtomicU32::new(0),
+            count: Arc::new(AtomicU32::new(0)),
         }
     }
 
@@ -83,7 +84,7 @@ impl Volume {
             read_only_parent: None,
             scrub_point: Arc::new(AtomicU64::new(0)),
             block_size,
-            count: AtomicU32::new(0),
+            count: Arc::new(AtomicU32::new(0)),
         })
     }
 
@@ -185,6 +186,11 @@ impl Volume {
         Ok(())
     }
 
+    // Check to see if this volume has a read only parent
+    pub fn has_read_only_parent(&self) -> bool {
+        self.read_only_parent.is_some()
+    }
+
     // Imagine three sub volumes:
     //
     //  0 -> Range(0, a)
@@ -264,6 +270,7 @@ impl Volume {
             info!(log, "Scrub for {} begins", self.uuid);
             info!(log, "Scrub with total_size:{:?} block_size:{:?}", ts, bs);
             let scrub_start = Instant::now();
+
             let start = read_only_parent.lba_range.start;
             let end = read_only_parent.lba_range.end;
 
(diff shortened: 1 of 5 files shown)


upstairs/src/volume.rs
```diff
@@ -351,6 +358,9 @@ impl Volume {
                     );
                     showat += showstep;
                 }
+                // Pause just a bit between IOs so we don't starve the guest
+                // TODO: More benchmarking to find a good value here.
+                tokio::time::sleep(Duration::from_millis(250)).await;
```
jmpesp, line 363: If there's a 10 GB volume (10737418240 bytes), divided by 262144 byte chunks, that's 40960 chunks. Multiply that by 0.250 and that's 10240 seconds, or 2.84 hours!

We shouldn't pause here. With the better job dependencies, I don't think guest IO will be blocked unless they happen to be interacting with the same blocks at the same time, and even then only for a moment.

jmpesp (changes requested):

upstairs/src/volume.rs
leftwo: I'll test again with the dependencies work.  I did my initial testing before that work was done, and found that boot time was 4x as long if the scrubber is running while the system is booting.

I'm sure there is also more tuning to do with regards to the size of scrubber IOs.

And, a third note, the scrubber is not actually turned on yet in propolis, so we have a little time to play around with different values before committing to something (but not too much time)

leftwo:
The TL;DR summary is, both the size of each IO the scrubber does, and
adding a delay between the scrubber's IOs can have an impact on both
IO performance and how long the scrubber takes to scrub the volume.

In general, I would feel safe saying:
A larger IO size per "scrub loop" results in a faster overall scrub.
As the delay between each "scrub loop" goes up, the system boots faster.

Really, more accurate measurements are needed to determine what the proper
numbers here are, but we have enough data to show an impact.

I used the `systemd-analyze critical-chain` on ubuntu (focal) image.
This was useful in showing there were differences between different
settings, but the run to run variation with the same settings suggests
a better test is needed to get more specific numbers.

As a Baseline, booting takes 0 min 57.022 seconds without the scrubber active.
Volume scrub size was 2361393152 with a block_size of 512.

Below is a table showing the variables tested and the resulting boot time
with those settings..

Scrub IO size: How much data the scrubber reads/writes per IO.
Pause MS: How long we pause in the scrubber between IOs.
Scrub Time: How long in seconds the full scrub too.
Boot Time: How long the system took to boot (systemd-analyze critical-chain).

Some scrub times are blank because I did not want to wait for it to finish.
```
 scrub |pause| scrub| boot time
IO size|   MS|  time| min:sec
       |     |   sec|
==================================
      -     -      -  0:57.022 <-- baseline

 262144     0    253  1:40.832
 262144     1    268  1:42.222
 262144     2    275  1:39.939
 262144     5    298  1:34.009
 262144    15    391  1:35.082
 262144    25    479  1:34.074
 262144    50    ???  1:33.782
 262144   250    ???  1:32.418

 scrub |pause| scrub| boot time
IO size|   MS|  time| min:sec
       |     |   sec|
================================== 
 131072     0    367  1:43.789
 131072     0    335  1:37.074
 131072     0    330  1:37.749
 131072     1    385  1:36.271
 131072     1    391  1:38.247
 131072     2    409  1:39.352
 131072     2    418  1:40.038
 131072     5    460  1:37.317
 131072     5    453  1:31.946
 131072    15    637  1:32.697
 131072    15    646  1:33.790
 131072    25    814  1:30.148
 131072    25    805  1:28.380
 131072    25         1:30.188
 131072    50   1269  1:33.929
 131072    50         1:37.214
 131072    50         1:37.756
 131072    50         1:34.926
 
 scrub |pause| scrub| boot time
IO size|   MS|  time| min:sec
       |     |   sec|
==================================
  65536     0    590  1:52.494
  65536     0    557  1:45.007
  65536     1         1:37.845
  65536     2         1:31.639
  65536     5    705  1:26.139
  65536    15         1:30.488
  65536    25         1:29.773
  65536    25         1:39.403
  
  32768     0   1095  1:49.099
  32768     0    944  1:46.347
  32768     0    989  1:47.438
  32768     0         1:37.446
  32768     0         1:40.993
  32768     1   1155  1:42.562
  32768     1         1:33.293
  32768     2         1:32.034
  32768     5         1:29.096
  32768    15         1:32.117
  32768    25         1:30.900
  32768    50  ~4800  1:35.600
  ```

leftwo:
I'm going to propose a starting value of  131072 bytes for IO side and 15ms between IOs
That seems, given the current testing granularity and accuracy, to give us a balance between
slowing down performance and the scrubber taking forever.  More tuning will be needed.

I'm also going to have a 120 second pause after starting an instance before the scrubber kicks off.
This should allow most OSs to boot.  Again, further tuning needed.

leftwo:
I've updated the `scrub` method to take as args start delay as well as delay between IOs.
This means propolis and/or the pantry can pick the numbers they want.

PR #498, opened 2022-11-01
[smf] Allow Crucible Zones to be self-assembling

smklein:
Make the "crucible agent" self-assembling. 

This intends to replace the following parts of the sled agent in Omicron:

- [Setting up IP addresses and routes after the zone has booted](https://github.com/oxidecomputer/omicron/blob/daf316d50ada51fe54a9c723074bc4d80f13104b/sled-agent/src/storage_manager.rs#L497-L502) -> The Crucible agent can do this setup itself
- [Setting SMF properties and enabling the service after the zone has booted](https://github.com/oxidecomputer/omicron/blob/daf316d50ada51fe54a9c723074bc4d80f13104b/sled-agent/src/storage_manager.rs#L395-L450) -> SMF properties can be injected via a profile

Part of https://github.com/oxidecomputer/omicron/issues/1898


smklein:
https://github.com/oxidecomputer/omicron/pull/1902 shows what the updates look like on the sled agent side

leftwo:
Does this change also need https://github.com/oxidecomputer/omicron/pull/1902 to go in before it will work?

smklein:
> Does this change also need [oxidecomputer/omicron#1902](https://github.com/oxidecomputer/omicron/pull/1902) to go in before it will work?

Yeah, the two need to land together to work. If both are approved, I would merge this, then update the crucible rev while merging 1902 in omicron.

ahl (approved):
looks good to me

agent/agent_method_script.sh
```diff
@@ -0,0 +1,28 @@
...
+DATASET="$(svcprop -c -p config/dataset "${SMF_FMRI}")"
+LISTEN_ADDR="$(svcprop -c -p config/listen_addr "${SMF_FMRI}")"
+LISTEN_PORT="$(svcprop -c -p config/listen_port "${SMF_FMRI}")"
+PORTBASE="$(svcprop -c -p config/portbase "${SMF_FMRI}")"
+DOWNSTAIRS_PREFIX="$(svcprop -c -p config/downstairs_prefix "${SMF_FMRI}")"
+SNAPSHOT_PREFIX="$(svcprop -c -p config/snapshot_prefix "${SMF_FMRI}")"
+
+ipadm show-addr "$DATALINK/ll" || ipadm create-addr -t -T addrconf "$DATALINK/ll"
+ipadm show-addr "$DATALINK/omicron6"  || ipadm create-addr -t -T static -a "$LISTEN_ADDR" "$DATALINK/omicron6"
+route get -inet6 default -inet6 "$GATEWAY" || route add -inet6 default -inet6 "$GATEWAY"
```
jclulow, line 17: We should really move this network initialisation into a separate service that's uniform in all of the zones, I think.  If we're happy with the shape of this (i.e., this sequence of commands with this set of input variables) we could even include it in the _brand_ as effectively a stable interface.  You'd be able to just turn it on and specify the datalink/address/gateway in the site profile and that'd be it.  We could make `svc:/milestone/network` depend on it, and all our services could just depend on _that_.

We can also do that later, of course, if you want to get moving now.

pantry/smf/manifest.xml
jclulow, line 1: Can you rename this file from `manifest.xml` to `pantry.xml`?

We should also take this opportunity to rename this service from `system/illumos/crucible_pantry` to `oxide/crucible/pantry` -- which will require a change in `sled-agent/src/services.rs` in the other PR as well.

agent/agent_method_script.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+set -o errexit
+set -o pipefail
```
jclulow, line 4: Let's add `set -o xtrace` here, at least for now?

pantry/pantry_method_script.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+set -o errexit
+set -o pipefail
```
jclulow, line 4: Let's add `set -o xtrace` here, at least for now?

package-manifest.toml
```diff
@@ -15,6 +16,7 @@ source.type = "local"
 source.rust.binary_names = ["crucible-pantry"]
 source.rust.release = true
 source.paths = [
-  { from = "pantry/smf", to = "/var/svc/manifest/site/crucible_pantry" }
+  { from = "pantry/smf", to = "/var/svc/manifest/site/crucible_pantry" },
```
jclulow, line 19: Let's just make this path `/var/svc/manifest/site/crucible` like the other one, and renaming `manifest.xml` to `pantry.xml` will mean the full path will end up being `/var/svc/manifest/site/crucible/pantry.xml`.

pantry/pantry_method_script.sh
```diff
@@ -0,0 +1,19 @@
...
+
+DATALINK="$(svcprop -c -p config/datalink "${SMF_FMRI}")"
+GATEWAY="$(svcprop -c -p config/gateway "${SMF_FMRI}")"
```
jclulow, line 7: I think we should probably explicitly check that the `DATALINK` and `GATEWAY` values here are not set to the `unknown` default value?

After L3-4, but before L6, we could insert:

```sh
. /lib/svc/share/smf_include.sh
```

Then, if we detect missing configuration, we can exit `$SMF_EXIT_ERR_CONFIG` to have SMF flag a permanent configuration error; e.g.,

```sh
if [[ $DATALINK == unknown ]] || [[ $GATEWAY == unknown ]]; then
        printf 'ERROR: missing datalink or gateway\n' >&2
        exit "$SMF_EXIT_ERR_CONFIG"
fi
```

agent/agent_method_script.sh
```diff
@@ -0,0 +1,28 @@
...
+
+DATALINK="$(svcprop -c -p config/datalink "${SMF_FMRI}")"
+GATEWAY="$(svcprop -c -p config/gateway "${SMF_FMRI}")"
```
jclulow, line 7: Same comment about `unknown` checks applies here as for the other method script.

jclulow (changes requested):
Broadly looks good!  Thanks for doing this.  A few nits and small changes, but otherwise I think pretty much ready to go!  I looked at this in tandem with the other PR: https://github.com/oxidecomputer/omicron/pull/1902

agent/agent_method_script.sh
smklein: Done

agent/agent_method_script.sh
smklein: Done

agent/agent_method_script.sh
smklein: I'm totally on-board with doing this , and agree that it would be better served as a service. Is https://github.com/oxidecomputer/helios-omicron-brand a better spot for this? Where should it live in there?

(I do plan on moving forward with this implemented manually, but I'd be happy to switch over to another service if I could get some guidance on setting up the service within the brand)

pantry/pantry_method_script.sh
smklein: Done

pantry/pantry_method_script.sh
smklein: Sure, updated.

Couple things:
- Is there a standard for checking "empty values" of properties in illumos? I noticed that I could either...
  - (1) totally delete the property, which obfuscates the value by removing it from the manifest.xml file
  - (2) set it to `''`, but then it appears to be set in the shell to `\"\"`, rather than an actual empty string, which is kinda odd.
  - (3) use this quirky naming scheme, at least for `astring` values, of setting it to some reserved sentinel value (like `default`).

I ask because if we generalize some of this stuff to an extra service, I don't *think* all these values actually need to be set, so some could be only supplied optionally. For example, I think the underlay address we're supplying might not always be available when the zone starts (such as within the switch zone).

pantry/smf/manifest.xml
smklein: Done

package-manifest.toml
smklein: Done

PR #499, opened 2022-11-02
Correct generation number requirement for activation.

leftwo:
This enables and enforces the proper generation number during activation of an upstairs.  Now all upstairs trying to activate with a downstairs region will be required to have provided a >= generation number to what the downstairs report as their highest found generation number.

Added better logic in the downstairs on how to handle generation number comparisons and who should be active when faced with various situations.

Added a bunch of tests to cover the various generation number situations to verify their correct behavior.

Added proper generation number accounting to all the tests I could find that needed it.  Minor improvements to logging in test scripts.

Fixed hammer_loop.sh test, it now passes.

Before this can go in (or before crucible can be used by propolis/omicron) the[ generation number changes work in omicron](https://github.com/oxidecomputer/omicron/pull/1897) will need to be approved and merged.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -3775,7 +3775,7 @@ mod test {
     async fn test_promote_to_active_multi_read_write_different_uuid_lower_gen(
     ) -> Result<()> {
         // Attempting to activate multiple read-write (where it's different
-        // Upstairs) but with the same gen should be blocked
+        // Upstairs) but with a lower gen should be blocked.
         let ads = build_test_downstairs(false)?;
 
         let upstairs_connection_1 = UpstairsConnection {
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -7,6 +7,7 @@
 
 use futures::executor;
 use futures::lock::{Mutex, MutexGuard};
+use std::cmp::Ordering;
 use std::collections::HashMap;
 use std::fmt;
 use std::fs::File;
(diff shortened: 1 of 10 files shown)


downstairs/src/lib.rs
```diff
@@ -995,17 +998,17 @@ where
...
-                             * At this point, should we just return error?
-                             * XXX
-                             */
                         } else {
-                            // matches_self above should include a check for
-                            // gen, but gen number may change between
-                            // negotiation and activation (see `XXX because the
-                            // generation number` upstairs). update generation
-                            // number here.
                             if upstairs_connection.gen != gen {
```
jmpesp, line 1012: do we still need this?

downstairs/src/lib.rs
```diff
@@ -3696,23 +3752,84 @@ mod test {
...
 
         Ok(())
     }
 
     #[tokio::test]
-    async fn test_promote_to_active_multi_read_write_same_uuid() -> Result<()> {
+    async fn test_promote_to_active_multi_read_write_different_uuid_lower_gen(
+    ) -> Result<()> {
+        // Attempting to activate multiple read-write (where it's different
+        // Upstairs) but with the same gen should be blocked
```
jmpesp, line 3778: s/same gen/lower gen/

jmpesp (approved):

downstairs/src/lib.rs
leftwo: Hmm..  Maybe.
If we only have `activation` and not `activation_with_gen`, then I would say yeah.
But, since I think were are heading that way, having the warning still printed would alert us that we were in this situation.

downstairs/src/lib.rs
leftwo: fixed

leftwo:
@faithanalog , If you could look at the changes I just made to [tools/crudd-speed-battery.sh](https://github.com/oxidecomputer/crucible/pull/499/files#diff-5290a7e0f3b74119495340eee62b1974450f566593109f062e5054c8c2455736)
I would appreciate it.  James already reviewed the rest of the code, and you are welcome to add any other
comments you see fit, but don't spend time there unless you want to.  It's just the updates I made tonight after
merging with your crudd-speed-battery that really need your attention.

faithanalog (approved):

PR #515, opened 2022-11-08
Add timeout test to perf test

leftwo:
test_perf.sh is currently taking waaaaaay to long to finish, and while
performance issues that this test is tripping over are worked out, the 
script has been both reduced in what it does, and limited to 40 minutes 
of runtime.

Added a new dtrace script to log performance for this test, as well
as collect a bunch of other debug information that could be useful
when a test does hit a timeout.

Minor cleanup of the test_perf.sh script, send logs and error to the
proper places for collection.


leftwo (commented):
This should give us more information when the test fails, and not have to wait
for the default timeout to trigger when the test takes forever.

Issue #500, opened 2022-11-03
Volume layer needs better generation number handling


leftwo:
The volume layer activation does not need generation number argument, nor does
really any `fn activation` for things that impl `BlockIO`.

Volume has sub-volume and read-only-parents of type `VolumeConstructionRequest`

The only thing that actually needs a generation number is the `VolumeConstructionRequest` of type `Region`.
Type `Region` means a `Guest` will be created.

The generation number for `VCR::Region` is passed to up_main when that volume is "created".
The `BlockIO for Guest` will send a `BlockOpGoActive` to the crucible upstairs.  Since up_main already has the
generation number, the upstairs can just use what gen it has internally instead of taking it from the guest.

This does mean that the generation number for an upstairs must be known at the time up_main is called, and
that an upstairs can only connect and reconnect with the same generation number that it was started with.
I can't currently think of a place where this would be a problem.

PR #516, opened 2022-11-09
Take the generation out of activation

leftwo:
This removes the generation number requirement from the BlockIO activate.  When the upstairs is started, it should receive a generation number from the caller (most likely from a VolumeConstructionRequest) and that value should be used when the volume (or subvolume, or guest) is activated.

A guest specific activate_with_gen method still exists for now that will enable a different generation number to be used.
This allows tests and possibly migration to still have the ability to deactivate and then re-activate a single upstairs 
instance without having to start over.

### DIFF
--- a/crudd/src/main.rs
+++ b/crudd/src/main.rs
@@ -565,7 +565,7 @@ async fn main() -> Result<()> {
     eprintln!("Crucible runtime is spawned");
 
     // IO time
-    guest.activate(opt.gen).await?;
+    guest.activate().await?;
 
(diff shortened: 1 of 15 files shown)


leftwo:
This was 90% done by @jmpesp before leave started, so credit goes to him.
I took the WIP code and added the final steps.

leftwo:
https://github.com/oxidecomputer/crucible/issues/500

faithanalog (approved):
looks good to me, tests run locally just fine too.

PR #517, opened 2022-11-10
Updates for test_up.sh script and buildomat

leftwo:
Updated test_up.sh to start dsc in two steps instead of all at once. 
This allows slower systems to take the time they need to create the regions.

Plumbed the no color option through to the dump command so buildomat 
does not have to print color codes.

Fixed the archive file list for the test_up tests, as the location had changed and 
buildomat was not longer getting the files.


leftwo:
This should be helpful in debugging an issue seen in other active PRs

PR #524, opened 2022-11-16
Short mode for crudd in CI

leftwo:
Make the default mode for the `crudd-speed-battery.sh` test only run one permutation.

If you pass the "full" arg, it will run all the permutations.


leftwo:
This puts a full CI run at just over 30 minutes, which is a better balance for tests that run on
every commit.  The soon to be coming nightly test should pick up the slack for longer running
tests.

faithanalog:
looks good to me

faithanalog (approved):

PR #528, opened 2022-11-17
DTrace probes for upstairs work queue

leftwo:
Added internal counters for job states in the upstairs and hooked those into the dtrace probes we already had.

Updated the dtrace info script to print out these counters.

### DIFF
--- a/tools/dtrace/README.md
+++ b/tools/dtrace/README.md
@@ -271,7 +271,7 @@ pfexec dtrace -s upstairs_info.d
 ```
 
 You start crucible, then run the above script.  Output should start appearing
-with a few seconds.
+within a few seconds.
 
 The output has several columns.  The first three will list the state of each
 of the three downstairs.  Following that the remaining columns all indicate
--- a/tools/dtrace/README.md
+++ b/tools/dtrace/README.md
@@ -263,7 +263,7 @@ Trace a downstairs IO and measure time for in in the following three parts:
 * 3rd report is OS done to downstairs sending the ACK back to upstairs
 
 ## upstairs_info.d
-This is a dtrace script for printing some simple upstairs state info.
+This is a dtrace script for printing upstairs state and work queue info.
 If the upstairs is not yet running, add the -Z flag to dtrace so it will
 wait to find the matching probe.
 ```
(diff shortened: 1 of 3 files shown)


faithanalog:
Could you put an example invocation of using `tools/dtrace/upstairs_info.d` with dtrace on a running upstairs, maybe as a comment in the `upstairs_info.d` file itself or wherever is convenient? For the benefit of us less familiar with using dtrace.

faithanalog:
oh i didnt even know that README file existed, dont i feel like a fool now!

leftwo:
> oh i didnt even know that README file existed

It needed updating, so thanks for reminding me about it.
I'm actually going to add a bit more to it to explain better what
the numbers mean.

leftwo:
Okay, I'm done with polishing.
Hopefully these updates help address any concerns.

tools/dtrace/README.md
```diff
@@ -263,24 +263,53 @@ Trace a downstairs IO and measure time for in in the following three parts:
...
 If the upstairs is not yet running, add the -Z flag to dtrace so it will
 wait to find the matching probe.
 ```
 pfexec dtrace -s upstairs_info.d
 ```
 
 You start crucible, then run the above script.  Output should start appearing
-right away with the state of the three downstairs and a count of active
-jobs for upstairs and downstairs.
+with a few seconds.
```
faithanalog, line 274: extreme nit moment: probably "within"

tools/dtrace/README.md
```diff
@@ -263,24 +263,53 @@ Trace a downstairs IO and measure time for in in the following three parts:
...
+
+Now, back to the columns.
+Columns four and five show the count of active jobs for the guest side (UPW)
+and for the downstairs side (DSW).
+
+
+The remaining columns 6-20 show the count of job states on each downstairs.
+Jobs for the downstairs can be in one of five possible states:
+ `New`, `In Progress`, `Done`, `Skipped`, `Error`.
+For each state/downstairs, we print a count of jobs in that state.
```
faithanalog, line 300: this is very helpful thanks!!

faithanalog (approved):
thanks for the readme updates! i left a nit on something that might be a typo, otherwise, ship it imo

tools/dtrace/README.md
leftwo: Neither extreme nor a nit, that's a straight up error!

PR #541, opened 2022-11-29
Update agent worker loop to not miss notifications

leftwo:
Update the agent worker loop so it has the lock while pulling work off the list of regions, and if no work is on the list then waiting on the condvar.  This avoids a situation where new work (and the notification) could arrive on the list after the worker thread had checked it, but before it was listening on the convar.


leftwo:
I've done 4422 delete/create loops and not seen any hang like I did before.
This same test could at best get two or three loops but usually would hang right away.

PR #550, opened 2022-12-02
Move Rust to 2021 edition

leftwo:
Move to 2021 Rust edition.

Also fixed a clippy nit that snuck in.


faithanalog (approved):
the future is now

PR #552, opened 2022-12-02
always specify branch in git dependencies

sunshowers:
This is PR 1 of 3 to make sure that omicron doesn't get two separate
dependencies, one for progenitor with main and one for progenitor
without main.

We choose this strategy over not doing so to maintain uniformity in
cases where a non-default branch must be used.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -415,12 +415,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "convert_case"
-version = "0.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


sunshowers:
There are no functional changes here so I'm just going to land this to unblock omicron.

leftwo:
The one tricky non obvious dependency here is being sure that both Omicron and Propolis can build with
new versions of Crucible.  We sometimes change the API, and those changes have to all go together with each
other.  The crucible `crucible-client-types` is the place where things typically break.

In this case, nothing has changed recently, so it should be fine.  And usually when we are in these intermediate 
states, everyone knows about it.

sunshowers:
@leftwo yeah! I did run into this in https://github.com/oxidecomputer/propolis/pull/262.

PR #554, opened 2022-12-02
[dsc] use async tokio::sync::Mutex

sunshowers:
I'm trying to upgrade to Rust 1.65 (see #553), and clippy warns about
mutexes being held across async points. Switch over to an async mutex.

I had to switch a bunch of functions over to being async. The
alternative,
[`blocking_lock`](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#method.blocking_lock),
is unfortunately not available if we're already in an async context
(which we are in most of these cases). It's easier to just make
everything async.

### DIFF
--- a/dsc/src/control.rs
+++ b/dsc/src/control.rs
@@ -116,8 +116,8 @@ pub struct Max {
     max: u64,
 }
 
-fn cid_bad(dsci: &DscInfo, cid: usize) -> bool {
-    let rs = dsci.rs.lock().unwrap();
+async fn cid_bad(dsci: &DscInfo, cid: usize) -> bool {
+    let rs = dsci.rs.lock().await;
     rs.ds_state.len() <= cid
 }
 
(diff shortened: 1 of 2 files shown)


leftwo:
Are these changes similar to what James has in https://github.com/oxidecomputer/crucible/pull/463 ?

I was looking at #552 and #553 and it seems like this might need to go first?

sunshowers:
> Are these changes similar to what James has in https://github.com/oxidecomputer/crucible/pull/463 ?

Yeah, it's basically similar.

This needs to go before #553. #552 is independent of this though.

Looking at the build breakages.

leftwo:
Ah, finally looking at what James had, it's in the downstairs, but the same idea, yeah.
I had kicked Jame's PR down the road because I was working on other things and it was
a lower priority (at the moment) and I wanted to be sure we were not breaking something.
These changes do look pretty simple though.

There is a `tools/test_dsc.sh` that we have (that this code passes on).
Also, many of the tests in `tools/test_up.sh` use dsc as well, so that should hopefully shake out any issues.

Let me know if you want a 2nd set of eyes on the build breakage issues.

sunshowers:
Thanks for the pointers! I *think* the build breakage has been addressed, and both `tools/test_dsc.sh` and `tools/test_up.sh` pass for me now.

leftwo (approved):
Assuming this passes the CI test, it's good to go.

I ran some of the other non CI tests in `tools` directory that make use of dsc and
they also ran without issue.

PR #556, opened 2022-12-05
region IO dtrace probes

faithanalog:
this adds a bunch of dtrace probes to region IO to get some numbers on how long each part of the IO is taking

in the dtrace script, i accumulate the entire time of flushes, but for reads/writes i divide by the number of blocks to get a per-block measurement.

I also threaded job_id in to read/write so i can report that in the dtrace probe.

I'm not differentiating between PID for the accumulations right now since I just want more data to average, and im only running one set of 3 downstairs anyway. We could change it to differentiate the PIDs if we want to.


leftwo (approved):
More probes less questions!

You could probably make the dtrace script a little smaller with more wildcard usage and making use of the probename (`tools/dtrace/perf-downstairs.d`), but I've done it both ways and sometimes I just can reason better about what I'm looking at when I break
it down into specific probe names like you have done here.  

This is not a recommendation, just passing on info so you know about it in the future. 
That is assuming you don't already know it, which you may, and in that case I'll shut up
now.

PR #571, opened 2022-12-16
Remove Boolean specifier from crucible-agent to comply with clap update

gjcolombo:
The upgrade to clap v4 (#566) changed the behavior of the crucible-downstairs `--encrypted` flag such that it no longer takes an explicit Boolean value. Instead, the presence or absence of the flag determines the value produced by parsing, and explicitly passing "true" or "false" causes an error. Make the Crucible agent play by these new rules so that Omicron can pick up Crucible packages including the clap v4 change. Without this, creating a new disk fails during region creation because crucible-downstairs rejects crucible-agent's `--encrypted true` arguments.

Tested by building the Crucible package, deploying the private package into an Omicron build, and observing that instance creation succeeds.

### DIFF
--- a/agent/src/main.rs
+++ b/agent/src/main.rs
@@ -910,7 +910,7 @@ fn worker_region_create(
         .arg(region.extent_size.to_string())
         .arg("--extent-count")
         .arg(region.extent_count.to_string())
-        .arg(format!("--encrypted={}", region.encrypted))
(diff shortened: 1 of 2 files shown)


leftwo (approved):
This should have been caught with the clap update, thanks for fixing.

leftwo:
I went ahead and merged this one, because I wanted to! :)

PR #579, opened 2022-12-19
Move to a stable Rust toolchain

bnaecker:


leftwo (approved):
Thanks for doing this

bnaecker:
Thanks Alan -- I think there's still an update to Dropshot I need to get plumbed through to actually support macOS stable builds. Will try to get that landed today, then merge this.

bnaecker:
@leftwo Ready for another quick review, Alan!

downstairs/build.rs
```diff
@@ -2,17 +2,19 @@
...
-// Copyright 2022 Oxide Computer Company
+// Copyright 2023 Oxide Computer Company
 
 fn main() {
     println!("cargo:rerun-if-changed=build.rs");
 
-    if version_check::is_min_version("1.59").unwrap_or(false) {
-        println!("cargo:rustc-cfg=usdt_stable_asm");
+    if !version_check::is_min_version("1.59").unwrap_or(false) {
+        println!("cargo:rustc-cfg=usdt_need_asm");
```
leftwo, line 11: Is this print going to fire if the version is < 1.59?

bnaecker: That's correct. We flipped the logic everywhere, including the meaning of the thing we print. The goal is to positively detect when we need to include `#![feature(asm)]`, which is for any version strictly before 1.59.

leftwo (approved):
Changing state from: `Approved` to: `Approved`

bnaecker:
Well [_that is sad_](https://github.com/oxidecomputer/crucible/actions/runs/3899303148). The PR passed checks but the merge to main failed. Will investigate.

bnaecker:
Hmm, looks like the merge resulted in two entries for `base64@0.20.0` in the lockfile. It looks like a dependabot update in 8257315f85dbc922306252b7362eae75937c536e added the same version, and I did not rebase before merging this. Odd that this resulted in a duplicate, I'd have expected the merge to correctly resolve that. Apparently not. See https://github.com/oxidecomputer/crucible/pull/602 for the fix.

PR #580, opened 2022-12-20
New OnlineRepair state, new Faulted state.

leftwo:
Created the OnlineRepair downstairs state, and renamed failed to faulted.  This is the first part (of several) to enable support of bringing back a downstairs after we have given up on it.

The OnlineRepair state currently does nothing other than ignore all work submitted to it.

The Faulted state is entered when a downstairs fails a write or flush (and can no longer be trusted).  Once a downstairs enters this state, all New/InProgress jobs are pushed to Skipped, and future jobs will automatically go to Skipped.

The decision to skip jobs was moved from in_progress() to enqueue() To enable this move, we need to pass the ds_done_rx channel to the enqueue function on the task that takes work from the guest and sends it all to the downstairs tasks.

More tests added for making sure jobs proceed through the work queue when one or more downstairs has faulted.

A few state change tests were fixed that were panicing for the wrong reason.

Added some more downstairs state transitions and checking around them. Put the dtrace __start points before enqueue, as enqueue can now finish some operations and we should avoid ending a trace point before we have started it.


leftwo:
I'm going to hold this till after #576 goes in, as that should decrease the memory load to an acceptable level.

upstairs/src/lib.rs
```diff
@@ -2659,12 +2861,47 @@ impl Downstairs {
...
+            }
+        }
+
+        // Puts the IO onto the downstairs work queue.
+        let ds_id = io.ds_id;
+        self.ds_active.insert(ds_id, io);
+
+        // If we skipped all three jobs, then it's possible no downstairs
+        // task is around to tell ds_up_listen() that this job is done, so
+        // do that work here.
```
jmpesp, line 2894: This occurs when all three downstairs are Faulted, right, because the tasks will be killed?

upstairs/src/test.rs
```diff
@@ -1560,40 +1576,12 @@ mod up_test {
         assert!(ds.downstairs_errors.get(&0).is_some());
         assert!(ds.downstairs_errors.get(&1).is_some());
         assert!(ds.downstairs_errors.get(&2).is_none());
-
-        // another read. Make sure only client 2 returns data.
```
jmpesp, line 1564: why was this removed?

jmpesp (changes requested):
A few things:

upstairs/src/lib.rs
```diff
@@ -1348,12 +1465,33 @@ where
                 {
                     warn!(
                         up_c.log,
-                        "[{}] Error processing message: {}",
+                        "[{}] ZZZZ Error processing message: {}",
```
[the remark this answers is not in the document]
leftwo: I'm asleep at the debug wheel..

upstairs/src/lib.rs
leftwo: Correct, if no downstairs connection exists, then there is no way for IO to make progress through crucible.  This would be true even before this PR, just taking place in a different location.

upstairs/src/test.rs
leftwo: With the new fault behavior, this test does not have enough framework to move IOs forward.
The marking of a downstairs as faulted now happens in `process_ds_operation` instead of
`process_ds_completion`.  New tests added now cover what was covered in this test before.

upstairs/src/lib.rs
```diff
@@ -1305,9 +1365,66 @@ where
...
-        do_reconcile_work(up, &mut fr, &mut fw, up_coms).await?;
+    // Match on DS state to decide what to do.
+    // Either:
+    // New: do_reconcile_work
+    // Replay: Set more work, move state to Active
+    // OnlineRepair: XXX write more code, all jobs should stay skipped.
+
+    let mut more_work = false;
+    let up_state = {
+        let active = up.active.lock().await;
```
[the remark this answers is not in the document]
leftwo: Yeah, the active state is a little tricky here, and I'm not super happy with having multiple locks.
But, I've not come up with a better way to do it yet either.

upstairs/src/lib.rs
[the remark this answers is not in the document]
leftwo: Fixed in https://github.com/oxidecomputer/crucible/pull/580/commits/a2ddb2991a24f472dfb36b64f51d828dfe80cdbd

jmpesp (approved):

PR #583, opened 2022-12-27
Updates for pantry package

leftwo:


leftwo:
Updating the pantry package format for testing with the WIP Omicron pantry code.

PR #584, opened 2022-12-30
Update internal bookkeeping for crutest write counter

leftwo:
This fixes two old bug with the crutest write verification that would crop up in a few situations.
1) Some situations where the write counter crossed a u8 boundary.
2) When we repaired downstairs 2 and 3 using 1 as the source when after running the`repair_workload` test and none of the 30 operations were a flush.
 
Updated the way the crutest write counter works, added more tests to be sure the counts increment as expected when dealing with ranges.

Added a little more debug information to the `test_repair.sh` script.
Added and improved the debug messages that show up with crutest detects a mismatch.

Update the crutest `verify_workload` test to read the whole region and not just quit on the first error it finds.

### DIFF
--- a/crutest/src/main.rs
+++ b/crutest/src/main.rs
@@ -420,13 +420,8 @@ impl WriteLog {
                 self.count_cur[index],
             );
 
-            if s_value >= self.count_min[index]
-                && s_value <= self.count_cur[index]
-            {
-                res = true
-            } else {
-                res = false
-            };
+            res = s_value >= self.count_min[index]
+                && s_value <= self.count_cur[index];
 
             // Only update if requested and the range was valid.
             if update && res && self.count_cur[index] != s_value {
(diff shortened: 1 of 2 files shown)


leftwo:
This is a work in progress, don't review!!

leftwo:
> This is a work in progress, don't review!!

Now ready for review.

This updated test ran some 10,000 loops, where as before we never got to 2,000

faithanalog:
> Some situations where the write counter crossed a u8 boundary.

the code + extra tests have me convinced this is fixed, looks good

> When we repaired downstairs 2 and 3 using 1 as the source when after running the `repair_workload` test and none of the 30 operations were a flush.

I think I don't really understand how these changes affect this one

crutest/src/main.rs
```diff
@@ -1946,9 +2007,6 @@ async fn repair_workload(
     // TODO: Allow the user to specify a seed here.
     let mut rng = rand_chacha::ChaCha8Rng::from_entropy();
 
-    // Any state coming in should have been verified, so we can
-    // consider the current write log to be the minimum possible values.
-    ri.write_log.commit();
```
leftwo, line 1951: Removal of this line is the fix for number 2 in the description.

leftwo:
> > When we repaired downstairs 2 and 3 using 1 as the source when after running
>>  the repair_workload test and none of the 30 operations were a flush.
>
> I think I don't really understand how these changes affect this one

Ah, yeah, that is my bad for not explaining it.

The removal of the `ri.write_log.commit();` in the beginning of the `repair_workload()` was the 
fix for that.

The write log works by counting the number of writes to a block, and writing that value in
the block as the data.  When we write, we add one to the write counter.
We keep a copy of the write counter in a different file, and we use that to verify that each
block has a number that we expect.  It crude, but does add some value.

For the repair test, we store two values in the write log, a "current value" and the "minimum"
possible value.  When we issue a flush, we replace the minimum with the current and
continue.  This gives us a "window" of possible values for our write counter.  When
dealing with this test, we intentionally try to issue writes without a flush and then let
the repair happen, but we can only verify a "range" of possible values, as we don't actually
know what write finished last, and with repair, we only know if a block is dirty, we don't
know which block has the "newest" dirty data, so it's possible that during repair we
actually go back to a previous write (up to the most recent flush).

The repair test does 30 IOs, but will issue a "commit" of the internal write log if it issues 
(and gets and ACK) back from a flush.  This commit of the write means we know the write
counter on disk is "at least" as high as the current (and after the commit, the minimum) in
our write log.

The problem would be that the test would do a bunch of IO, but never issue a flush, then the
test would quit and restart, and then the next test would run, but the first thing the test
does is "commit" the current write log.  This initial commit is not correct, as we never had
a flush, so we can't yet update the minimum.

This was also a rare occurrence as usually their would be 1 flush in 30 operations, and even
if there was no a flush, we also have to have a block having data older than the write logs
current value at the time of the new test starting and issuing the commit (meaning a repair
happened, and it did pick the extent with "older" data).

faithanalog (approved):

crutest/src/main.rs
```diff
@@ -1997,6 +1992,33 @@ async fn burst_workload(
 /*
  * issue some random number of IOs, then wait for an ACK for all.
  * We try to exit this test and leave jobs outstanding.
+ * The write log works by counting the number of writes to a block, and
```
leftwo, line 1995: Oh dammit,  I did not want to commit these comments here.

PR #591, opened 2023-01-05
Switch ImpactedBlocks from bitvec to a range

faithanalog:
Switch ImpactedBlocks from bitvec to a range
    
The bitvec implementation of ImpactedBlocks was really slow, allocating kilobytes of bitvecs at a time, and doing kilobytes of bit operatoins to check for conflicts. The replacement I've opted for after talking with Alan is just a simple range; none of our usecases need anything more complex. This eliminates one of the largest sources of Upstairs CPU utilization.
    
If in the future we want a single ImpactedBlocks so support multiple ranges, we should probably just store that as a Vec or binary tree of ranges, but we'll burn that bridge when we come to it. As it is, Upstairs' API doesn't create a scenario where that would be relevant.
    
Also, this commit brings in Proptest to test ImpactedBlocks extremely thorougly! Give that a look, it's pretty neat. As part of that I did also modify crucible_region to expose MIN_SHIFT and MAX_SHIFT so I wouldn't need to re-derive it from MIN_BLOCK_SIZE and MAX_BLOCK_SIZE.

### DIFF
--- a/upstairs/src/impacted_blocks.rs
+++ b/upstairs/src/impacted_blocks.rs
@@ -127,11 +127,12 @@ impl ImpactedBlocks {
         first_impacted: ImpactedAddr,
         last_impacted: ImpactedAddr,
     ) -> Self {
-        if first_impacted <= last_impacted {
-            ImpactedBlocks::InclusiveRange(first_impacted, last_impacted)
-        } else {
-            ImpactedBlocks::Empty
-        }
+        assert!(last_impacted >= first_impacted);
+        ImpactedBlocks::InclusiveRange(first_impacted, last_impacted)
+    }
+
+    pub fn empty() -> Self {
+        ImpactedBlocks::Empty
     }
 
     /// Create a new ImpactedBlocks range starting at a given offset, and
@@ -142,13 +143,15 @@ impl ImpactedBlocks {
         first_impacted: ImpactedAddr,
         n_blocks: u64,
     ) -> Self {
-        debug_assert!(extent_size > 0);
+        assert!(extent_size > 0);
 
         if n_blocks == 0 {
             return ImpactedBlocks::Empty;
         }
 
-        // So because we're inclusive, if we have 1 block then the first impacted should be the same as the last impacted. If we have 2 blocks, it's +1. etc.
+        // So because we're inclusive, if we have 1 block then the first
+        // impacted should be the same as the last impacted. If we have
+        // 2 blocks, it's +1. etc.
         let ending_block = n_blocks + first_impacted.block - 1;
 
         let last_impacted = ImpactedAddr {
@@ -265,38 +268,29 @@ pub fn extent_from_offset(
     num_blocks: Block,
 ) -> ImpactedBlocks {
     let extent_size = ddef.extent_size().value;
-
-    // I don't really think these assertions belong here, but we have a test
-    // for it I guess
     let shift = ddef.block_size().trailing_zeros();
-    debug_assert_eq!(offset.shift, shift);
-    debug_assert_eq!(num_blocks.shift, shift);
+
+    // If we get invalid data at this point, something has gone terribly
+    // wrong further up the stack. So here's the assumptions we're making:
+    // First, our blocks are actually the same size as this region's blocks
+    assert_eq!(offset.shift, shift);
+    assert_eq!(num_blocks.shift, shift);
+
+    // Second, the offset is actually within the region
+    assert!(offset.value < ddef.extent_count() as u64 * extent_size);
+
+    // Third, the last block is also within the region
+    assert!(
+        offset.value + num_blocks.value
+            <= ddef.extent_count() as u64 * extent_size
+    );
 
     let fst = ImpactedAddr {
         extent_id: offset.value / extent_size,
         block: offset.value % extent_size,
     };
 
-    // If offset is outside the region, Empty
-    if fst.extent_id >= ddef.extent_count() as u64 {
-        return ImpactedBlocks::Empty;
-    }
-
-    let mut impact = ImpactedBlocks::from_offset(
-        ddef.extent_size().value,
-        fst,
-        num_blocks.value,
-    );
-
-    // Clamp range to last block in region
-    if let ImpactedBlocks::InclusiveRange(_, lst) = &mut impact {
-        if lst.extent_id >= ddef.extent_count() as u64 {
-            lst.extent_id = ddef.extent_count() as u64 - 1;
-            lst.block = extent_size - 1;
-        }
-    }
-
-    impact
+    ImpactedBlocks::from_offset(extent_size, fst, num_blocks.value)
 }
 
 #[cfg(test)]
@@ -304,7 +298,7 @@ mod test {
     use super::*;
     use proptest::prelude::*;
     use std::panic;
-    use std::panic::AssertUnwindSafe;
+    use std::panic::UnwindSafe;
     use test_strategy::{proptest, Arbitrary};
 
     fn extent_tuple(eid: u64, offset: u64) -> (u64, Block) {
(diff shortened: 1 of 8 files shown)


faithanalog:
review request for rain is for the proptest stuff in particular, to make sure I'm using it right

faithanalog:
tests what do you want from me, works on my machine

upstairs/src/impacted_blocks.rs
```diff
@@ -284,8 +439,731 @@ mod test {
...
+    fn iblocks_blocks_iterates_over_all_blocks(
+        // Keep these reasonably sized so we don't OOM running this test
+        #[strategy(1..=128u32)] extent_count: u32,
+        #[strategy(1..=128u64)] extent_size: u64,
+
+        #[strategy(0..#extent_count as u64)] start_eid: u64,
+        #[strategy(0..#extent_size)] start_block: u64,
+
+        #[strategy(#start_eid..#extent_count as u64)] end_eid: u64,
+        #[strategy(#start_block..#extent_size)] end_block: u64,
```
sunshowers, line 903: Same here.

upstairs/src/impacted_blocks.rs
```diff
@@ -2,96 +2,237 @@
...
+            Some((fst, lst)) => {
+                let extents = lst.extent_id - fst.extent_id;
+                (extents * self.extent_size + lst.block - fst.block + 1)
+                    as usize
             }
         }
+    }
+}
+
+impl DoubleEndedIterator for ImpactedBlockIter {
```
leftwo, line 95: Is this impl used anywhere?  I was trying to figure out what it does and
why we need it, but I could not.  Is is part of the new test stuff?

upstairs/src/impacted_blocks.rs
```diff
@@ -2,96 +2,237 @@
...
-            for i in block_offsets.iter().sorted() {
-                result
-                    .push((*eid, Block::new_with_ddef(*i as u64, &self.ddef)));
-            }
+        if n_blocks == 0 {
+            return ImpactedBlocks::Empty;
         }
 
-        result
+        // So because we're inclusive, if we have 1 block then the first impacted should be the same as the last impacted. If we have 2 blocks, it's +1. etc.
```
leftwo, line 151: nit, this comment goes over 80 columns

upstairs/src/impacted_blocks.rs
```diff
@@ -2,96 +2,237 @@
...
+        };
+
+        ImpactedBlocks::InclusiveRange(first_impacted, last_impacted)
     }
 
-    /// Returns the number of impacted blocks
-    pub fn len(&self) -> usize {
-        let mut len = 0;
+    pub fn intersection(&self, other: &ImpactedBlocks) -> ImpactedBlocks {
+        let ImpactedBlocks::InclusiveRange(fst_self, lst_self) = self else {
```
leftwo, line 163: TIL about the `let else`. Nice.

upstairs/src/impacted_blocks.rs
```diff
@@ -119,83 +260,95 @@ impl ImpactedBlocks {
...
-        (offset.value + num_blocks.value)
-            <= (ddef.extent_size().value * ddef.extent_count() as u64)
-    );
-    assert_eq!(offset.block_size_in_bytes() as u64, ddef.block_size());
+    let extent_size = ddef.extent_size().value;
 
-    let mut result = ImpactedBlocks::new(ddef);
-    let mut o: u64 = offset.value;
-    let mut blocks_left: u64 = num_blocks.value;
+    // I don't really think these assertions belong here, but we have a test
```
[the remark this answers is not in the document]
leftwo: Is this maybe need for the proptest stuff?  Like it might as part of the tests it does, send 
invalid offsets into `extent_from_offset()` and it needs that function to not panic?

upstairs/src/test.rs
```diff
@@ -230,10 +231,12 @@ mod up_test {
     }
 
     #[tokio::test]
-    #[should_panic]
```
leftwo, line 233: I think we should put back the assertions if invalid input is given to `extent_from_offset`,
however, if we don't, then maybe change this test to "off_to_extent_ignore_length_to_big"?

upstairs/src/test.rs
```diff
@@ -907,7 +909,16 @@ mod up_test {
             vec![],
             10,
             vec![request.clone()],
-            ImpactedBlocks::default(),
+            ImpactedBlocks::new(
```
leftwo, line 912: small nit: I see this same pattern repeated quite a few times.
Is there any way to abstract it out to a single place that everyone uses?

leftwo (commented):
I don't really understand exactly how the proptest stuff works, but I'm comforted by
@sunshowers ' review so I don't feel I need to :)

My only real issue is about the assertions and if handling invalid input is
something that `extent_from_offset` needs to handle, see my comments inline.

upstairs/src/impacted_blocks.rs
faithanalog: This isn't used anywhere now, but it comes nearly free out of the implementation, and if we ever wanted to iterate the blocks in reverse order this makes that O(1) to call `.rev()` on the iterator instead of O(n) + having to allocate memory for the whole thing

upstairs/src/test.rs
faithanalog: So it's worth noting that the previous use of `::default()` was just incorrect. it was issuing these commands and stating that they impacted zero blocks, which was just wrong.

Looking back on it I actually found another instance where I replaced a default() with an Empty incorrectly.

Maybe we should have a function that takes a ReadRequest or WriteRequest and returns the ImpactedBlocks that covers it

This wouldn't be particularly helpful to have outside of the test code, because those requests are always one block long. Creating a bunch of ImpactedBlocks from that in release code would be horribly inefficient. But in the test code which is all doing one-block things anyway, it would help a bit.

This raises some questions about how create_read_eob and create_write_eob are called in general in my mind, seeing how easy it is to accidentally write code that passes in uncorrelated requests and impactedblocks into it. But i think that further question is outside the scope of this PR.

upstairs/src/impacted_blocks.rs
```diff
@@ -2,96 +2,237 @@
...
-        let mut result = Vec::with_capacity(self.len());
+    /// Create a new ImpactedBlocks range starting at a given offset, and
+    /// stretching n_blocks further into the extent. Panics an error if
+    /// `extent_size` is 0. Returns ImpactedBlocks::Empty if n_blocks is 0.
+    pub fn from_offset(
+        extent_size: u64,
+        first_impacted: ImpactedAddr,
+        n_blocks: u64,
+    ) -> Self {
+        debug_assert!(extent_size > 0);
```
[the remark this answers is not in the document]
faithanalog: I was thinking of actually making it be a Result but I could change it back to an assert. I don't really like release-mode assertions because it brings back the ol java fun times of random functions throwing undeclared exceptions, instead of the code actually knowing where all the errors come from and pushing them all through the correct error handling facilities (if only to log them properly before suddenly dying)

upstairs/src/impacted_blocks.rs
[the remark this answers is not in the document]
faithanalog: It is not needed for proptest stuff. We can do whatever we want with proptest stuff. I'm just not sure what we gain from randomly dying inside the ImpactedBlocks creation, so I opted for coercing input into something that makes sense for the region. But I can put it back to sudden-death mode if we prefer that.

upstairs/src/impacted_blocks.rs
```diff
@@ -2,96 +2,237 @@
...
+                }
+
+                Some(result)
+            }
         }
     }
 
-    pub fn blocks_in_extent(&self) -> usize {
-        self.ddef.extent_size().value as usize
+    fn size_hint(&self) -> (usize, Option<usize>) {
```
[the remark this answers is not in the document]
faithanalog: size_hint is part of the Iter trait

upstairs/src/impacted_blocks.rs
faithanalog: yeah it got added into rust stable recently and it is suuuper helpful for a lot of common patterns

upstairs/src/test.rs
leftwo: You bring up good points here, but I was thinking something simple like this:

```
    fn generic_impacted_blocks() -> ImpactedBlocks {
        ImpactedBlocks::new(
            ImpactedAddr { 
                extent_id: 0,
                block: 7,
            },  
            ImpactedAddr { 
                extent_id: 0,
                block: 7,
            },
        )
    }
```

Then call `create_read_eob()` like this:
```
        let op = create_read_eob(
            next_id,
            vec![],
            10,
            vec![request.clone()],
            generic_impacted_blocks(),
        );  
```

Just as a way of having a little more code density.
It was a "small nit", so not something I was really concerned about :)

leftwo (approved):
Thanks for the updates.

Sorry to leave you with some conflicts to merge, let me know if you have any
trouble with them.  Looking forward to building on this PR.

PR #610, opened 2023-01-13
Check incoming IOs are inside the region.

leftwo:
Added a method to the RegionDefinition that allows verification of an IO offset+length to be sure it will be inside the region.

Added tests at both the common/region layer and the integration test layer.


faithanalog:
rest of this change looks good to me.

PR #616, opened 2023-01-19
Added nightly build script, removed tools/downstairs_daemon.sh

leftwo:
Removed the now unused tools/downstairs_daemon.sh script.

Updated the test_restart_repairs.sh to find only test related downstairs process.

Added a script to build and tar up the scripts and binaries required to run the test_nightly.sh script.

Updated the tools/README.md


tools/make-nightly.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+set -eux
+
+cargo build --release --all-features
+
+rm -f out/crucible-nightly.tar.gz 2> /dev/null
+
+mkdir -p out
+
```
faithanalog, line 9: will this `rm` command error (halting the script) if `out/crucible-nightly.tar.gz` doesn't exist already, or does `-f` prevent that?

leftwo: Yeah, the `-f` command will tell `rm` to not report error.

```
alan@atrium$ rm this_does_not_exist
rm: this_does_not_exist: No such file or directory
alan@atrium:crucible$ echo $?
2
alan@atrium$ rm -f this_does_not_exist
alan@atrium$ echo $?
0
```

faithanalog (approved):

tools/make-nightly.sh
faithanalog: neat!

PR #619, opened 2023-01-24
Adapting to dropshot 0.9.x changes

lifning:
Such as TypedBody being last parameter, and various small type changes. https://github.com/oxidecomputer/dropshot/blob/9f3c27091e8c3bbdb577f39f573a4a29b16f9b02/CHANGELOG.adoc#breaking-changes

This will help things get compatible with latest omicron main.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -74,24 +74,15 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


leftwo (approved):
Thanks for making these changes!

PR #620, opened 2023-01-24
Return info when closing an extent.

leftwo:
Another piece of the live repair puzzle.

Updated the extent close command to return the flush, gen, and the dirty bit for the extent just closed.
This behavior will be used by future code to do live repair of a downstairs.

Added some tests for close to verify correct behavior.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -599,12 +599,13 @@ where
                     flush_number,
                     gen_number
                 );
+
                 match d
                     .region
                     .region_flush_extent(
                         *extent_id,
-                        *flush_number,
                         *gen_number,
+                        *flush_number,
                         *repair_id,
                     )
                     .await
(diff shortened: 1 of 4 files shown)


downstairs/src/lib.rs
```diff
@@ -599,12 +599,13 @@ where
                     flush_number,
                     gen_number
                 );
+
                 match d
                     .region
                     .region_flush_extent(
                         *extent_id,
-                        *flush_number,
```
jmpesp, line 606: :scream:

We might want to add a separate type for either gen number of flush number, to prevent this from happening!

jmpesp (approved):

downstairs/src/lib.rs
leftwo: This was not actually a bug, I was just making function signatures the same
where I did flush/gen/dirty always in that order.  But, yeah, an official type
for flush and/or gen might not be a bad idea to avoid someone flipping them.

PR #621, opened 2023-01-26
An activate failure due to generation number should say so

leftwo:
If the reason for activation failure is the supplied generation number, then the error should indicate that.

Cleaned up the connect_region_set decision path, got rid of unnecessary variables.  Made collate_downstairs method return a result and true/false if we need to repair.  If its an error, then pass that error back.

Now, git is freaking out about this diff, but it's really not as bad as it looks at first glance.

The diff in `collate_downstairs()` is pretty normal.

The diff in `connect_region_set()` is a mess, but there is actually very little change from
the original.  I will explain inline below.

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -5394,7 +5394,10 @@ impl Upstairs {
      * Verify the guest given gen number is highest.
      * Decide if we need repair, and if so create the repair list
      */
-    async fn collate_downstairs(&self, ds: &mut Downstairs) -> Result<bool> {
+    async fn collate_downstairs(
+        &self,
+        ds: &mut Downstairs,
+    ) -> Result<bool, CrucibleError> {
         /*
          * Show some (or all if small) of the info from each region.
          *
@@ -5452,7 +5455,7 @@ impl Upstairs {
         let requested_gen = self.get_generation().await;
         if requested_gen == 0 {
             error!(self.log, "generation number should be at least 1");
-            bail!("Generation number should be at least 1");
+            crucible_bail!(GenerationNumberTooLow, "Generation 0 illegal");
         } else if requested_gen < max_gen {
             /*
              * We refuse to connect. The provided generation number is not
@@ -5460,12 +5463,13 @@ impl Upstairs {
              */
             error!(
                 self.log,
-                "found/using gen number {}, larger than requested: {}",
+                "found generation number {}, larger than requested: {}",
                 max_gen,
                 requested_gen,
             );
-            bail!(
-                "found/using gen number {}, larger than requested: {}",
+            crucible_bail!(
+                GenerationNumberTooLow,
+                "found generation number {}, larger than requested: {}",
                 max_gen,
                 requested_gen,
             );
@@ -5685,10 +5689,7 @@ impl Upstairs {
          * Determine the highest flush number and make sure our generation
          * is high enough.
          */
-        let need_repair;
-        let repair_commands;
-        let failed_collate;
-        {
+        let collate_status = {
             let active = self.active.lock().await;
             if active.up_state != UpState::Initializing {
                 return Ok(());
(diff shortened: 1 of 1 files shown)


upstairs/src/lib.rs
```diff
@@ -5685,10 +5689,7 @@ impl Upstairs {
          * Determine the highest flush number and make sure our generation
          * is high enough.
          */
-        let need_repair;
-        let repair_commands;
-        let failed_collate;
-        {
+        let collate_status = {
```
leftwo, line 5692: This part is the meat of the change, instead of having these three booleans, we just
have one `collate_status` now, that is a `Result<bool, CrucibleError>`.  This was what the
three variables were trying to express, and now it's all wrapped up nicely in a single place.

upstairs/src/lib.rs
```diff
@@ -5721,164 +5722,161 @@ impl Upstairs {
...
-                    need_repair = res;
-                    failed_collate = false;
-                }
-                Err(e) => {
-                    need_repair = false;
-                    failed_collate = true;
-                    error!(self.log, "Failed collate with {}", e);
-                }
-            }
-            repair_commands = ds.reconcile_task_list.len();
```
leftwo, line 5735: Yes, this is no longer here, it's moved down below to the one place where we use it.

upstairs/src/lib.rs
```diff
@@ -5721,164 +5722,161 @@ impl Upstairs {
...
-             * expect them to be.  Any change means we abort and require
-             * all downstairs to reconnect, even if repair work is finished
-             * as a disconnected DS does not yet know it is all done with
-             * work and could have its state reset to New.
-             */
-
-            assert_eq!(ds.ds_active.len(), 0);
-            assert_eq!(ds.reconcile_task_list.len(), 0);
+            self.collate_downstairs(&mut ds).await
+        };
```
leftwo, line 5726: Here is where the diff just falls apart, and I would suggest not trying to make any sense of it at all.

upstairs/src/lib.rs
```diff
@@ -5721,164 +5722,161 @@ impl Upstairs {
...
 
-            let ready = ds
-                .ds_state
-                .iter()
-                .filter(|s| **s == DsState::Repair)
-                .count();
+        match collate_status {
```
leftwo, line 5728: Instead of the three "if"'s the old code had: 
`if failed_collate`
`else if need_repair`
`else `
We now have a lovely `match` on the `collate_status` which has been updated to incorporate the
three different choices.

Nothing else (other than moving `let repair_commands = ...` down inside where it is used) was
changed for what actually happens for the three choices.
I was required to indent everything over, and from that, re-align some comments that had
crossed the 80 character boundary, but despite the git freak out, nothing else is different.

leftwo (commented):
Here is a sample of what the errors look like with these changes:
```
final:stock$ ./target/release/crutest cli
cli connecting to 0.0.0.0:5050
connected to 0.0.0.0:5050
>> activate
ERROR: GenerationNumberTooLow("found generation number 11, larger than requested: 1")
>> activate -g 0
ERROR: GenerationNumberTooLow("Generation 0 illegal")
>> activate -g 9
ERROR: GenerationNumberTooLow("found generation number 11, larger than requested: 9")
>> activate -g 11
Ok
```

upstairs/src/lib.rs
```diff
@@ -5721,164 +5722,161 @@ impl Upstairs {
...
-            let ready = ds
-                .ds_state
-                .iter()
-                .filter(|s| **s == DsState::WaitQuorum)
-                .count();
+                let ready = ds
+                    .ds_state
+                    .iter()
+                    .filter(|s| **s == DsState::WaitQuorum)
+                    .count();
```
jordanhendricks, line 5857: nit: looks like some extra spaces made it in here

upstairs/src/lib.rs
jordanhendricks: agreed that this is much nicer!

jordanhendricks (approved):
thanks! I appreciate you outlining where the diff is different too

PR #622, opened 2023-01-26
change crucible-pantry to crucible_pantry

jmpesp:
PR #583 changed how the pantry service name was constructed, but these changes are required to match 1) where the tar extracts to and 2) the service name the sled agent tries to start.


leftwo (approved):
I take it you have tested these with latest Omicron and all works as expected?

I'm not sure why it was all working for me, though I've not used the pantry
branch with latest Omicron, so your tests would be newer than mine.

PR #634, opened 2023-02-02
WIP, ideas for refactor of Target

leftwo:
We need some of the `Target` struct to be in a place where we can update it, and access it
after the initial connection is made.

Many tasks want access to the channels to talk with other tasks, and rather than passing those
channels around, we can just hang them off the downstairs struct and then anyone can find
them.


leftwo:
Abandon!

leftwo:
Better idea was implemented.  Bye bye.

PR #635, opened 2023-02-03
Refactor for future repair needs

leftwo:
Moved the downstairs IP:Port to inside the downstairs struct (and protected behind the downstairs mutex).  This will allow for future changes of that IP:Port.  Changed the way the looper method works to make use of the mutex value instead of requiring it to start.

Removed the downstairs IP:Port from the structures we pass around.

Changed the ds_work channel to be a MPSC.  This will allow for a future repair task to also put work on the downstairs queue apart from the task that is answering work requests from the guest.

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -301,7 +301,8 @@ pub fn deadline_secs(secs: u64) -> Instant {
 async fn process_message(
     u: &Arc<Upstairs>,
     m: &Message,
-    up_coms: UpComs,
+    client_id: u8,
+    ds_done_tx: &mpsc::Sender<u64>,
 ) -> Result<()> {
     let (upstairs_id, session_id, ds_id, result) = match m {
         Message::Imok => return Ok(()),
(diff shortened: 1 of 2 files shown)


upstairs/src/lib.rs
```diff
@@ -2264,6 +2254,10 @@ async fn looper(
         } else {
             tokio::time::sleep(Duration::from_secs(1)).await;
         }
+        // Get the specific information for the downstairs we will operate on.
```
leftwo, line 2257: After every disconnect from a downstairs, the looper loop will reacquire the IP:Port for the downstairs and try to connect again.   We will probably put this somewhere else when
support for changing a downstairs comes in, but having it where it is now is a step in the
right direction.

upstairs/src/lib.rs
```diff
@@ -8689,56 +8691,48 @@ pub async fn up_main(
...
-            let (ds_work_tx, ds_work_rx) = watch::channel(1);
-            /*
-             * Create the channel used to submit reconcile work to each
-             * downstairs (when work is required).
-             */
-            let (ds_reconcile_work_tx, ds_reconcile_work_rx) =
-                watch::channel(1);
+    let mut dst = Vec::new();
+    for client_id in 0..3 {
+        /*
```
leftwo, line 8700: This loop is basically the same as before with the changes being:

Always do three client_ids.
The `ds_work_` is now a `mpsc::channel`
We don't store the IP:Port in the `Target` struct any longer, it's now part of the downstairs struct.

jmpesp (approved):

PR #636, opened 2023-02-03
Add extent to impacted blocks

leftwo:
Create a function that, if given an extent number, will return an ImpactedBlocks for all the blocks in that extent.

There is not currently a user for this, but there will be with the coming downstairs live repair work, and given
the size that work is becoming, I'm trying to break off manageable chunks.


faithanalog (approved):
looks good, ship it!

PR #648, opened 2023-02-15
Writes have to depend on all in-flight jobs

### LINKED ISSUE #595: Read after flush means hash mismatch
Running the test_repair.sh script, I encountered this panic on hash mismatch:
```
02/30 Write block 385  len 32768  data:179  14 234 182 247 205 165  93 
03/30 Read  block  28  len 32768
04/30 Read  block 459  len 16384
05/30 Write block 172  len 40960  data:135 230 223 200 244  30 254 210 102  80 
06/30 Read  block 192  len 32768
07/30 Read  block  91  len 12288
08/30 Flush
09/30 Write block  85  len 32768  data:133 142 217 174 187 157 110 100 
10/30 Read  block 310  len 36864
11/30 Flush
12/30 Write block 171  len  8192  data:165 136 
thread 'tokio-runtime-worker' panicked at '[0] read hash mismatch on id 1006
Expected [Some(fc14853c232d8295), Some(9845e4de2549c015), Some(9aee68ebcf51cfb)]
Computed [Some(f1f24e1091a86eac), Some(f55039b4811a7962), Some(9aee68ebcf51cfb)]
guest_id:7 request:[ReadRequest { eid: 4, offset: Block { value: 11, shift: 12 } }, ReadRequest { eid: 4, offset: Block { value: 12, shift: 12 } }, ReadRequest { eid: 4, offset: Block { value: 13, shift: 12 } }]
job state:{1: Done, 0: Done, 2: Done}', upstairs/src/lib.rs:3265:29
Exit on repair fail, loop: 53575, choice: 0
```

I suspect what happened is the following:

I've taken the output from the test and replaced the job number/total jobs with the downstairs side job ID.
Job IDs always start with 1001 and (just discovered as I'm writing this) there is a bug in crutest where it
does not print anything for the first job.
```
    ID    IO  BLOCK  SIZE  DATA WRITTEN (If a write)
  1001  Write   385 32768  179  14 234 182 247 205 165  93
  1002   Read    28 32768
  1003   Read   459 16384
  1004  Write   172 40960  135 230 223 200 244  30 254 210 102  80
  1005   Read   192 32768
  1006   Read    91 12288
  1007  Flush
  1008  Write    85 32768  133 142 217 174 187 157 110 100
```
For this analysis, we can ignore jobs before 1006.
At 1006, A Read of block 91 went out,
One downstairs (not 0) responded (which records the hash of the read result) ack send to the guest.
At 1007 A flush was issued, at least 2 downstairs ack'd.
At 1008 A write was issued to block 85 of length 8 (so also covering block 91).
The write, job 1007 finished on at least downstairs 0.
Downstairs 0 now handles the read request at 1006, but the write for 1008 has already
been serviced, this read now returns different data than the first read did, and we get our panic.

So, I think there may be a dependency issue with allowing a read issued before a flush then write
to be serviced both before and after that flush then write.

If I check block 91 now, all three downstairs agree on what it has, which does fit within
this analysis.

jmpesp:
`test_repair.sh` found a problem where a read job executed after a write! The current job dependency rules for writes stops searching for jobs that have overlapping blocks if a flush is seen, but this is obviously wrong: jobs may arrive at different Downstairs in different orders, or an aggressively parallel Downstairs could re-order jobs however it wants. We have to construct job dependencies so that even when this happens jobs still complete in the correct order.

When searching for write dependencies, search all in-flight jobs - do not stop looking when a flush is seen. This commit also adds a unit test for the problem that `test_repair.sh` identified.

Also fix a clippy problem in downstairs/src/region.rs.

Fixes #595

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -5077,22 +5077,6 @@ impl Upstairs {
          * incorrect read. It's important to search for write dependencies in
          * the list of all active jobs.
          *
-         * Take another example. Again say that a flush was a barrier to
-         * searching for write dependencies:
-         *
-         *       block
-         * op# | 0 1 2 | deps
-         * ----|-------------
-         *   0 | W     |
-         *   1 |   W   |
-         *   2 | FFFFF | 0,1
-         *   3 | W     | 2
-         *   4 |   W   | 2
-         *
-         * Without dependencies between ops 0 and 3, and 1 and 4, there's
-         * nothing that prevents an aggressively parallel downstairs from
-         * executing these writes in the wrong order.
-         *
          * TODO: any overlap of impacted blocks will create a dependency.
          * take this an example (this shows three writes, all to the
          * same block, along with the dependency list for each
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -5056,9 +5056,42 @@ impl Upstairs {
          * Construct a list of dependencies for this write based on the
          * following rules:
          *
-         * - ignore everything that happened before the last flush
          * - writes have to depend on the last flush completing
          * - any overlap of impacted blocks requires a dependency
+         *
+         * It's important to remember that jobs may arrive at different
+         * Downstairs in different orders (they should still complete in job
+         * dependency order!). For example, say that searching for the
+         * dependency of a write stopped at the last flush. Then say that the
+         * following set of jobs were submitted:
+         *
+         *       block
+         * op# | 0 1 2 | deps
+         * ----|-------------
+         *   0 | R R   |
+         *   1 | F F F |
+         *   2 |   W W | 1
+         *
+         * Without any dependencies, a downstairs could choose to perform op 0
+         * at any time, including after the write! This would result in an
+         * incorrect read. It's important to search for write dependencies in
+         * the list of all active jobs.
+         *
+         * Take another example. Again say that a flush was a barrier to
+         * searching for write dependencies:
+         *
+         *       block
+         * op# | 0 1 2 | deps
+         * ----|-------------
+         *   0 | W     |
+         *   1 |   W   |
+         *   2 | FFFFF | 0,1
+         *   3 | W     | 2
+         *   4 |   W   | 2
+         *
+         * Without dependencies between ops 0 and 3, and 1 and 4, there's
+         * nothing that prevents an aggressively parallel downstairs from
+         * executing these writes in the wrong order.
          *
          * TODO: any overlap of impacted blocks will create a dependency.
          * take this an example (this shows three writes, all to the
@@ -5090,11 +5123,10 @@ impl Upstairs {
         {
             let job = &downstairs.ds_active[job_id];
 
-            // Depend on the last flush, then break - flushes are a barrier for
+            // Depend on the last flush - flushes are a barrier for
             // all writes.
             if job.work.is_flush() {
                 dep.push(**job_id);
-                break;
             }
 
             // If this job impacts the same blocks as something already active,
--- a/upstairs/src/test.rs
+++ b/upstairs/src/test.rs
@@ -5994,7 +5994,7 @@ mod up_test {
         // ----|-------|-----
         //   0 | W     |
         //   1 | FFFFF | 0
-        //   2 | W     | 1
+        //   2 | W     | 0,1
 
         let upstairs = make_upstairs();
         let (ds_done_tx, _ds_done_rx) = mpsc::channel(500);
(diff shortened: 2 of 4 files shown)


upstairs/src/lib.rs
```diff
@@ -5056,10 +5056,43 @@ impl Upstairs {
...
+         * ----|-------------
+         *   0 | W     |
+         *   1 |   W   |
+         *   2 | FFFFF | 0,1
+         *   3 | W     | 2
+         *   4 |   W   | 2
+         *
+         * Without dependencies between ops 0 and 3, and 1 and 4, there's
+         * nothing that prevents an aggressively parallel downstairs from
+         * executing these writes in the wrong order.
```
leftwo, line 5094: I think this case is actually okay. 
Job 3 depends on 2,
Job 2 depends on 0,1

So, job 3 can't go untill 0 and 1 (and then 2) are all done.

upstairs/src/test.rs
```diff
@@ -5994,7 +5994,7 @@ mod up_test {
         // ----|-------|-----
         //   0 | W     |
         //   1 | FFFFF | 0
-        //   2 | W     | 1
+        //   2 | W     | 0,1
```
leftwo, line 5997: This, like the example above, I think was still going to be okay.

0 can go when it's ready, as it has no dependencies.
1 can't run until 0 is finished.
2 can't go until 1 has finished, so also requiring 0 here does not change anything.

leftwo (commented):
I wonder if you might consider an alternative approach.

What if we really do make the flush a barrier, and the flush depends on all outstanding IO, read or write?

This would be simpler to reason about, there would be no chance for things to cross the boundary.

It would probably be slower for reads, as they would now have to wait for the flush to finish, though we should be able to run some perf tests to see that impact.

Or dependency list would only have to extend to the most recent flush, not continue searching and comparing `ImpactedBlocks` for all outstanding IOs. This is probably a
small perf win, but I don't expect to see that show up in measurements.

leftwo:
> I wonder if you might consider an alternative approach.
> 
> What if we really do make the flush a barrier, and the flush depends on all outstanding IO, read or write?
> 
> This would be simpler to reason about, there would be no chance for things to cross the boundary.
> 
> It would probably be slower for reads, as they would now have to wait for the flush to finish, though we should be able to run some perf tests to see that impact.
> 
> Or dependency list would only have to extend to the most recent flush, not continue searching and comparing `ImpactedBlocks` for all outstanding IOs. This is probably a small perf win, but I don't expect to see that show up in measurements.

As a side point, I'm working on the OnlineRepair code path, and having the flush really be the barrier helps reason about repair behavior as well.

upstairs/src/test.rs
jmpesp: Right, but it's only like this because we don't elide dependencies yet :)

upstairs/src/lib.rs
jmpesp: Yeah, this part of the comment is wrong. Removed in a0b0b2d

jmpesp:
> What if we really do make the flush a barrier, and the flush depends on all outstanding IO, read or write?
>
> This would be simpler to reason about, there would be no chance for things to cross the boundary.
>
> It would probably be slower for reads, as they would now have to wait for the flush to finish, though we should be able to run some perf tests to see that impact.

I hesitate to make reads depend on flushes - it would extend our flush latency bubble to reads too, and it would violate the idea of Crucible acting like an NVMe disk.

> Or dependency list would only have to extend to the most recent flush, not continue searching and comparing ImpactedBlocks for all outstanding IOs. This is probably a small perf win, but I don't expect to see that show up in measurements.

This is what we're doing now! :) and it's wrong unless we do what's in this PR or make flushes a universal barrier.

jmpesp:
> it would violate the idea of Crucible acting like an NVMe disk.

I should expand on this - I think NVMe disks *don't* do this. Looking at the NVMe 1.1b spec:

> The Flush command shall commit data and metadata associated with the specified namespace(s) to non-volatile media. The flush applies to all commands completed prior to the submission of the Flush command. The controller may also flush additional data and/or metadata from any namespace.

I interpret "commit data" here as the flush command only applying to writes.

leftwo:
> > it would violate the idea of Crucible acting like an NVMe disk.
> 
> I should expand on this - I think NVMe disks _don't_ do this. Looking at the NVMe 1.1b spec:
> 
> > The Flush command shall commit data and metadata associated with the specified namespace(s) to non-volatile media. The flush applies to all commands completed prior to the submission of the Flush command. The controller may also flush additional data and/or metadata from any namespace.
> 
> I interpret "commit data" here as the flush command only applying to writes.

I don't think the NVMe spec dictates how in flight reads are required to perform with an in flight flush, the same as
what it says for writes.  My understanding of the situation is we, as implementers, are not required to include outstanding reads or writes before handling an outstanding flush, only that, once a flush has completed and acknowledged to the guest, we need to return data from writes that completed before the flush (at minimum).  The "may also flush" sentence that you quoted above I think covers us flushing more if we so desire, but are not required to.

So, by requiring in flight reads and/or writes to finish before in flight flush, we are not violating the NVMe spec, just 
adding our own requirements on top of it, but still within what the spec allows.

All this being said, I think I can work with just using `ImpactedBlocks` to stack our dependencies so all three downstairs
will layer on the data in the same (eventual) order and should work with what I'm planning for OnlineRepair.

leftwo:
> > What if we really do make the flush a barrier, and the flush depends on all outstanding IO, read or write?
...
> 
> > Or dependency list would only have to extend to the most recent flush, not continue searching and comparing ImpactedBlocks for all outstanding IOs. This is probably a small perf win, but I don't expect to see that show up in measurements.
> 
> This is what we're doing now! :) and it's wrong unless we do what's in this PR or make flushes a universal barrier.

My "Or" here was a load bearing typo, it should have been "Our", and the statement should have been included in the initial proposal of "What if we really do make the flush a barrier, and the flush depends on all outstanding IO, read or write?".   At any rate, I think I'm not going to convince you of my idea, and the solution you have here should fix the original issue, so we should move forward and get this in, and if I want to circle back later I'll do so :)

leftwo (approved):
I added some comments, but let's move forward with checking this in.

PR #651, opened 2023-02-21
Add an extent limit to the flush command

leftwo:
Make an optional extent limit field to the flush command that will tell the downstairs to only flush a subset of extents.

What this option represents is a boundary where extent flushes will go no further.  Any extent less than this value will be issued a flush, any extent at or beyond this value will not be flushed.

The intent here is for use by the online repair process, which is still under development, so there are no actual users of this option as of yet.  When online repair is ready, the repair task will keep track of the "extent limit" and update it
as each extent on a downstairs under repair is repaired (or deemed not to need repair).

If an extent is not "dirty" then no action will be taken (just like a regular flush).

### DIFF
--- a/protocol/src/lib.rs
+++ b/protocol/src/lib.rs
@@ -370,6 +370,11 @@ pub enum Message {
         flush_number: u64,
         gen_number: u64,
         snapshot_details: Option<SnapshotDetails>,
+        /*
+         * The ending extent where a flush should stop.
+         * This value is unique per downstairs.
+         */
+        extent_limit: Option<usize>,
     },
     FlushAck {
         upstairs_id: Uuid,
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -481,6 +481,11 @@ where
             continue;
         }
 
+        // ZZZ if this is a flush, then we have to determine if there is
+        // an extent_limit value for this downstairs.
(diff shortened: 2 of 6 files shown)


protocol/src/lib.rs
```diff
@@ -370,6 +370,11 @@ pub enum Message {
         flush_number: u64,
         gen_number: u64,
         snapshot_details: Option<SnapshotDetails>,
+        /*
+         * The ending extent where a flush should stop.
+         * This value is unique per downstairs.
```
jmpesp, line 375: This is only true during repair, right?

upstairs/src/lib.rs
```diff
@@ -481,6 +481,11 @@ where
             continue;
         }
 
+        // ZZZ if this is a flush, then we have to determine if there is
```
jmpesp, line 484: ZZZ

protocol/src/lib.rs
jmpesp: the comment [here](https://github.com/oxidecomputer/crucible/pull/651/files#diff-6d9d1c0cbb1fdf3f309bac3c91b38560a952bf9e17b4e64b6d1e04a13f160b1bR2490) answers my question

jmpesp (approved):
:rocket:

PR #653, opened 2023-02-22
Update perf test, add memory test

leftwo:
Updated the perf test to run a few more combinations, and allow for more user selectable input.

Added a first pass at a downstairs memory usage test along with buildomat script to call it.  Added this test to the list of scripts the nightly test will include (but not call it yet).


leftwo:
Lets see what I forgot to do to get this to work in CI..

faithanalog (approved):
ran this on my machine, gave a thorough readthru of the new mem test. all looks great to me!

PR #658, opened 2023-03-06
Flush currently has to be a barrier

### LINKED ISSUE #655: You can't depend on dependencies
The downstairs currently assumes any job ID issued before the most recent flush has been completed:

https://github.com/oxidecomputer/crucible/blob/75519cb2dc91be78ee5792718575e5348ddbdc1c/downstairs/src/lib.rs#L2585

This breaks the dependency model where a read does not depend on a flush and a flush does not depend
on a previous read.  This can possibly introduce reads landing on different downstairs in a different order, 
resulting in different data being returned.

Consider this example.

Read block 0
Flush
Write block 0

The read has no dependencies, so it can go whenever.
The flush has no dependencies, so it can go whenever.
The write requires the flush.

On downstairs 1, the flush could go first, then the write, then the read.  The read would get data from the write.
On downstairs 2, the read could go first (and get original block 0 data), then the flush, then the write.

Even if the write depends on the read, because the downstairs stops checking dependencies with job IDs
lower than the most recent flush, it can't tell the difference between the read not having yet arrived, and it
having arrived before the flush.

jmpesp:
The Downstairs currently assumes that all jobs previous to the last flush have completed, but work was done Upstairs to relax certain dependencies. This created the potential for the case where a job could be created Upstairs with a dependency but that dependency would be ignored by the Downstairs. To correct this, flushes must be a barrier for all jobs. There's future work that needs to be done to once again undo this but let's get back to being correct.

Fixes #655


leftwo (approved):
I depend on this to work.

leftwo:
Oh wait, what about this:

https://github.com/oxidecomputer/crucible/blob/e4b54426b95904a09638baa2632b202a820085c2/upstairs/src/lib.rs#L3974

The `retire_check()` method has to change

jmpesp:
You're right - have a look at d0ca965 for the rollback

upstairs/src/lib.rs
```diff
@@ -3964,27 +3964,15 @@ impl Downstairs {
             kvec.sort_unstable();
 
             for id in kvec.iter() {
-                // Remove everything before this flush
+                // Remove everything before this flush (only valid if flushes
```
leftwo, line 3967: Minor comment, I would word this as "Remove everything before this flush because flushes depend on everything and everything depends on flushes" Instead of the "only valid if" part.  If we change the code, then we should change the comment, the comment does not need the conditional.

leftwo (approved):

upstairs/src/lib.rs
jmpesp: :+1:, done in f243621

PR #664, opened 2023-03-15
Remember a summary of completed jobs

leftwo:
The upstairs now keeps track of a summary of the last eight jobs it has removed from its downstairs work queue. These jobs are ones where all three downstairs have returned a result and we have ACK'd back to the guest and are removing the job from the active list.

The control "jobs" endpoint will also display the summary info for the last 8 jobs.

The code to summarize a DownstairsIO has been moved into the objects that do the summarization.
By doing this, we can use it to record a summary of a completed job without having to include large
things, like the data buffers, but still give us enough information to be useful.

In addition, I snuck in a few fields to the Downstairs struct
```
    extents_repaired: Vec<usize>,
    extents_confirmed: Vec<usize>,
    extent_limit: Vec<Option<usize>>,
```

These are not currently used for anything, but will be with the coming OnlineRepair work.

### DIFF
--- a/upstairs/src/control.rs
+++ b/upstairs/src/control.rs
@@ -112,6 +112,9 @@ struct UpstairsStats {
     ds_jobs: usize,
     repair_done: usize,
     repair_needed: usize,
+    extents_repaired: Vec<usize>,
+    extents_confirmed: Vec<usize>,
+    extent_limit: Vec<Option<usize>>,
 }
 
 /**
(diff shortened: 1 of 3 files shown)


leftwo:
This is all stuff I have/had in the OnlineRepair work, but I figured breaking it out would make that PR not quite
as mammoth as it's looking to be.

upstairs/src/control.rs
```diff
@@ -190,126 +171,8 @@ async fn build_downstairs_job_list(up: &Arc<Upstairs>) -> Vec<WorkSummary> {
     let mut jobs = Vec::new();
     for id in kvec.iter() {
         let job = ds.ds_active.get(id).unwrap();
-
-        let (job_type, num_blocks, deps): (String, usize, Vec<u64>) =
```
leftwo, line 194: This all just moved to the io_summarize() and ioop_summary() methods.
The only thing that changed really is converting IOState to a string instead of
making up a whole new struct.

upstairs/src/lib.rs
```diff
@@ -2484,6 +2485,16 @@ struct Downstairs {
      */
     io_state_count: IOStateCount,
 
+    /**
```
leftwo, line 2488: These two are for keeping track of online repair stats, and are not used yet.

jmpesp (approved):

PR #696, opened 2023-04-07
Add clap feature so dsc --help alignment setting is honored

leftwo:
clap needed this feature so it would respect the dsc `--help` alignment config that was requested.


faithanalog (approved):
seems fine to me

PR #697, opened 2023-04-07
Add picture for downstairs states, remove Verifying

leftwo:
Added a pretty picture of downstairs states.

Removed the `Verifying` state, as we don't use that and there are no plans to do so.

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -6549,9 +6549,76 @@ impl FlushInfo {
 }
 
 /*
- * States a downstairs can be in.
- * XXX This very much still under development. Most of these are place
- * holders and the final set of states will change.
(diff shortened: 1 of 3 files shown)


upstairs/src/lib.rs
```diff
@@ -6549,9 +6549,76 @@ impl FlushInfo {
 }
 
 /*
- * States a downstairs can be in.
- * XXX This very much still under development. Most of these are place
- * holders and the final set of states will change.
+ * States of a downstairs (from the upstairs point of view).
+ *
+ * This shows the different states a downstairs can be in (from the point of
+ * view of the upstairs.
```
jmpesp, line 6555: This is a little redundant.

jmpesp (approved):
Nice!

PR #698, opened 2023-04-10
Add support for SMF refresh

bnaecker:


bnaecker:
It's used in https://github.com/oxidecomputer/dendrite/pull/212, where the process itself updates the SMF properties and then refreshes them.

jmpesp:
:+1:

PR #699, opened 2023-04-16
crutest use SIGUSR1 to stop tests

leftwo:
Added a simple signal handler for crutest that will receive a signal and, depending on that signal will send a message over a channel.

Currently there are two signals supported , `SIGUSR1` and `SIGUSR2` and only the test command `generic` supports
them.  `SIGUSR1` will send the message over the channel to stop running, and `SIGUSR2` will send the message to verify the volume.

Added a test mode, "continuous", that will run the test indefinitely.

This is inspired by the signal code in `crudd`.  By inspired I mean copied.

All of this allows us to run a test that just sends IO.
We can interrupt it to have it verify the volume (then go back to sending IO)
We can interrupt it and have it stop gracefully.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -970,6 +970,8 @@ dependencies = [
  "ringbuffer",
  "serde",
  "serde_json",
+ "signal-hook",
+ "signal-hook-tokio",
  "statistical",
  "tokio",
  "tokio-util 0.7.3",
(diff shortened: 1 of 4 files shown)


faithanalog (approved):
sorry for the delay, but this all looks right to me.

PR #700, opened 2023-04-17
Hold the downstairs lock a little longer

leftwo:
Hold the downstairs lock when transitioning a downstairs from `WaitActive` to `WaitQuorum` 
to prevent the up_listen task from jumping into repair before we have added all the info it needs.

Previously, we would get the lock, change the downstairs to WQ, then release the lock.
Next, we get the lock agan, add the region_metadata to the downstairs struct, then release the
lock.  This small window allowed the up_listen task to sneak in and try a reconciliation
and then panic when it did not find the region_metadata.
 
This should the a panic seen in the `LiveRepair` CI, though these changes are unrelated to LiveRepair
and could happen even without the `LiveRepair` work.

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -1291,22 +1291,28 @@ where
                         if negotiated != 4 {
                             bail!("Received ExtentVersions out of order!");
                         }
+                        let active = up.active.lock().await;
+                        let up_state = active.up_state;
+                        let mut ds = up.downstairs.lock().await;
+                        drop(active);
 
-                        let my_state = {
-                            let state = &up.downstairs.lock().await.ds_state;
-                            state[up_coms.client_id as usize]
-                        };
+                        let my_state = ds.ds_state[up_coms.client_id as usize];
                         match my_state {
                             DsState::WaitActive => {
-                                up.ds_transition(
-                                    up_coms.client_id, DsState::WaitQuorum
-                                ).await;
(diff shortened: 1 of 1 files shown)


upstairs/src/lib.rs
```diff
@@ -1291,22 +1291,28 @@ where
                         if negotiated != 4 {
                             bail!("Received ExtentVersions out of order!");
                         }
+                        let active = up.active.lock().await;
+                        let up_state = active.up_state;
+                        let mut ds = up.downstairs.lock().await;
+                        drop(active);
```
jmpesp, line 1297: Is it required that the `up.active` lock be taken before the `up.downstairs` lock?, or that `active` be dropped before taking the `up.downstairs` lock?

jmpesp (approved):
:ship: except for one comment

upstairs/src/lib.rs
leftwo: Yes, it is required that we have the upstairs state when we change the downstairs state.
To make sure we don't deadlock, we always get the upstairs state first.
 We need to know what the current upstairs state is (active or not) when we transition
downstairs states to know if that transition is valid.

We must get the downstairs lock while holding the upstairs active lock so we know we are
comparing the states of both at the "same time", and we don't release the downstairs lock
until we have made (or not made) any transition for the downstairs.

PR #703, opened 2023-04-20
Fix paths in method scripts

smklein:


leftwo (approved):
Thanks for the offline description of what this is for.

Feel free to merge this when ready. I'm assuming this will land along with the self
assembling zone stuff (or just before/after).

PR #705, opened 2023-04-24
Slightly better versioning between upstairs and downstairs

leftwo:
This adds a slight improvement on message version checking between the upstairs and the downstairs.  We now do a little better version checking, and add a field for future use when more versions are supported.

This also makes the first official change of the message version, as several updates since the initial version have taken place, and its time to begin changing the version going forward whenever there is a change in the Message enum.

Added some basic tests to check for the version.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -1193,12 +1193,27 @@ where
                         let mut ds = ads.lock().await;
 
                         if let Some(upstairs_connection) = upstairs_connection {
-                            info!(
-                                log,
-                                "upstairs {:?} disconnected, {} jobs left",
-                                upstairs_connection,
-                                ds.jobs(upstairs_connection).await?,
-                            );
+
+                            // If our upstairs never completed activation,
+                            // or some other upstairs activated, we won't
+                            // be able to report how many jobs.
+                            match ds.jobs(upstairs_connection).await {
+                                Ok(jobs) => {
+                                    info!(
+                                        log,
+                                        "upstairs {:?} disconnected, {} jobs left",
+                                        upstairs_connection,
+                                        jobs,
+                                    );
+                                }
+                                Err(e) => {
+                                    info!(
+                                        log,
+                                        "upstairs {:?} disconnected, {}",
+                                        upstairs_connection, e
+                                    );
+                                }
+                            }
 
                             if ds.is_active(upstairs_connection) {
                                 info!(
@@ -1226,14 +1241,46 @@ where
                         gen,
                         read_only,
                         encrypted,
+                        supported_versions,
                     }) => {
                         if negotiated != 0 {
                             bail!("Received connect out of order {}",
                                 negotiated);
                         }
+                        info!(log, "Connection request from {} with version {}",
+                            upstairs_id, version);
 
-                        if version != 1 {
-                            bail!("expected version 1, got {}", version);
+                        // Verify we can communicate with the upstairs.  First
+                        // check our message version.  If that fails,  check
+                        // to see if our version is one of the supported
+                        // versions the upstairs has told us it can support.
+                        if version != CRUCIBLE_MESSAGE_VERSION {
+                            if supported_versions
+                                .contains(&CRUCIBLE_MESSAGE_VERSION)
+                            {
+                                warn!(
+                                    log,
+                                    "downstairs requires down rev version {}",
+                                    CRUCIBLE_MESSAGE_VERSION
+                                );
+                            } else {
+                                let m = Message::VersionMismatch {
+                                    version: CRUCIBLE_MESSAGE_VERSION,
+                                };
+                                let mut fw = fw.lock().await;
+                                if let Err(e) = fw.send(m).await {
+                                    warn!(
+                                        log,
+                                        "Failed to send VersionMismatch: {}",
+                                        e
+                                    );
+                                }
+                                bail!(
+                                    "Required version {}, got {}",
+                                    CRUCIBLE_MESSAGE_VERSION,
+                                    version,
+                                );
+                            }
                         }
 
                         // Reject an Upstairs negotiation if there is a mismatch
@@ -1275,12 +1322,16 @@ where
                             gen,
                         });
                         info!(
-                            log, "upstairs {:?} connected",
-                            upstairs_connection.unwrap());
+                            log, "upstairs {:?} connected, version {}",
+                            upstairs_connection.unwrap(),
+                            CRUCIBLE_MESSAGE_VERSION);
 
                         let mut fw = fw.lock().await;
                         if let Err(e) = fw.send(
-                            Message::YesItsMe { version: 1, repair_addr }
+                            Message::YesItsMe {
+                                version: CRUCIBLE_MESSAGE_VERSION,
+                                repair_addr
+                            }
                         ).await {
                             bail!("Failed sending YesItsMe: {}", e);
                         }
@@ -3059,6 +3110,7 @@ mod test {
     use rand_chacha::ChaCha20Rng;
     use std::net::Ipv4Addr;
     use tempfile::{tempdir, TempDir};
+    use tokio::net::TcpSocket;
     use tokio::sync::mpsc::error::TryRecvError;
 
     // Create a simple logger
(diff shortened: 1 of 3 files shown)


leftwo:
I'm trying to walk the line here and not get pulled too deep into the solution set 
of what we could do, and how much time we have with an incremental 
improvement over what we have.

leftwo:
Going forward, the messages between the upstairs and the downstairs should be
clear that there is a mismatch in version and the connection will be dropped.

A version difference between this PR and things before will not be so clear, but in
my testing with the current head of crucible main and this change, the upstairs and
downstairs will not stay connected, which at least is "better" than staying up for
some number of IOs, then resetting.

downstairs/src/lib.rs
```diff
@@ -1226,14 +1241,46 @@ where
...
+                                let mut fw = fw.lock().await;
+                                if let Err(e) = fw.send(m).await {
+                                    warn!(
+                                        log,
+                                        "Failed to send VersionMismatch: {}",
+                                        e
+                                    );
+                                }
+                                bail!(
+                                    "Required version {}, got {}",
```
jmpesp, line 1279: this should list the supported versions that the Upstairs sent too

jmpesp (approved):

downstairs/src/lib.rs
```diff
@@ -1226,14 +1241,46 @@ where
...
+                        // check our message version.  If that fails,  check
+                        // to see if our version is one of the supported
+                        // versions the upstairs has told us it can support.
+                        if version != CRUCIBLE_MESSAGE_VERSION {
+                            if supported_versions
+                                .contains(&CRUCIBLE_MESSAGE_VERSION)
+                            {
+                                warn!(
+                                    log,
+                                    "downstairs requires down rev version {}",
```
jordanhendricks, line 1263: with the way this is written now, it doesn't seem like this will always be a down rev? (assuming that means a less recent version)

downstairs/src/lib.rs
jordanhendricks: +1

downstairs/src/lib.rs
```diff
@@ -5774,4 +5826,230 @@ mod test {
...
+                panic!("Invalid answer from downstairs: {:?}", x);
+            }
+        }
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_version_downrev() -> Result<()> {
+        // Test that a newer crucible version will result in a message
+        // indicating their is a version mismatch.
```
jordanhendricks, line 5912: nit: "there"

jordanhendricks: I know this is the pattern in all the tests, but it seems like the downstairs version is implicit right? (implicitly `CRUCIBLE_MESSAGE_VERSION`?)

downstairs/src/lib.rs
```diff
@@ -5774,4 +5826,230 @@ mod test {
...
+                panic!("Invalid answer from downstairs: {:?}", x);
+            }
+        }
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_version_uprev_only() -> Result<()> {
+        // Test sending only the +1 version to the DS, verify it rejects
+        // this version as supported.
```
jordanhendricks, line 5945: "unsupported"?

protocol/src/lib.rs
```diff
@@ -132,32 +132,68 @@ pub struct SnapshotDetails {
     pub snapshot_name: String,
 }
 
+/**
+ * Crucible Upstairs Downstairs message protocol version.
+ * Crude, but its something.  This should be changed whenever there is
+ * any change to the enum Message below.
+ */
+pub const CRUCIBLE_MESSAGE_VERSION: u32 = 2;
```
jordanhendricks, line 140: A pattern we have in propolis for managing the (crude) versioning with bhyve uses a u32 like this to represent the current version, but its type is an `IntoPrimitive` enum: https://github.com/oxidecomputer/propolis/blob/master/crates/bhyve-api/src/lib.rs#L256-L287. I like this approach because it allows us to embed in the code some notes about what changed during versions. It might be worth doing something like this here, since I imagine crucible protocol changes might be even more gnarly than bhyve ones, and it could be worthwhile to have notes like this.

upstairs/src/lib.rs
```diff
@@ -941,22 +942,40 @@ where
                             up.encrypted(),
                         );
                     }
+                    Some(Message::VersionMismatch { version }) => {
+                        // Upstairs cannot communicate with the downstairs.
+                        up.ds_transition(
+                            up_coms.client_id, DsState::BadVersion
+                        ).await;
+                        bail!(
+                            "downstairs version is {}, ours is {}!",
```
jordanhendricks, line 951: probably also worth printing supported versions here

upstairs/src/lib.rs
```diff
@@ -941,22 +942,40 @@ where
...
+                            CRUCIBLE_MESSAGE_VERSION => {}
+                            x => {
+                                up.ds_transition(
+                                    up_coms.client_id,
+                                    DsState::BadVersion
+                                ).await;
+                                bail!("expected version {}, got {}",
+                                    CRUCIBLE_MESSAGE_VERSION,
+                                    x
+                                );
```
jordanhendricks, line 977: could this be an if/else? the empty match block for `CRUCIBLE_MESSAGE_VERSION` throws me a bit

jordanhendricks (commented):
Thanks for prioritizing this! Some questions and review comments.

I also think it would be very nice if there was a CLI command to ask the downstairs what version it is. (I'm not imagining querying a *running* downstairs, just being able to tell what version of the binary I have built.)

downstairs/src/lib.rs
leftwo: Technically, no, the versions in the "supported_versions" Vec are not required to be lower
than the `CRUCIBLE_MESSAGE_VERSION`.

I thought of this like `CRUCIBLE_MESSAGE_VERSION` is what the `Message` enum would
be versioned as.  If the upstairs in some future decides there were other versions of the
`Message` enum it could support, it would include them in that `supported_versions` Vec.
Like, `CRUCIBLE_MESSAGE_VERSION` is 6, but we can talk to a downstairs with 5 or 4.

None of that code is written yet though.  And, sure, we might send a version that is
higher than `CRUCIBLE_MESSAGE_VERSION`.

Would this read better as `downstairs requires alternate message version {}`?

downstairs/src/lib.rs
leftwo: Yes, the downstairs that is created in `start_ds_and_connect` will be using whatever
the `CRUCIBLE_MESSAGE_VERSION` is.  I'll add a comment to make that clear.

downstairs/src/lib.rs
jordanhendricks: Hm, I don't think that quite feels right. IIUC, the downstairs is reporting that it and the upstairs are at different (but compatible) versions, and that we are carrying on as usual. (I think the "requires" phrasing is throwing me off.) If that's the case, then maybe something like: "downstairs and upstairs using different but compatible versions"? And then I would probably put the version of each in the log message, as well as even the supported upstairs versions.

leftwo: Yes, I'll update to that suggestion, thanks.

leftwo:
> Thanks for prioritizing this! Some questions and review comments.
> 
> I also think it would be very nice if there was a CLI command to ask the downstairs what version it is. (I'm not imagining querying a _running_ downstairs, just being able to tell what version of the binary I have built.)

I like this idea, but I want to put it into a follow up PR, where we not only print the message version, but git info as well
It's a little more work.

downstairs/src/lib.rs
leftwo: See if recent changes are to your liking.

downstairs/src/lib.rs
leftwo: Comments added

protocol/src/lib.rs
leftwo: See if the new changes are what you were thinking of.

jordanhendricks: No problem, definitely don't want to block on that idea.

upstairs/src/lib.rs
```diff
@@ -769,7 +774,7 @@ where
         gen: up.get_generation().await,
         read_only: up.read_only,
         encrypted: up.encrypted(),
-        supported_versions: vec![CRUCIBLE_MESSAGE_VERSION],
+        supported_versions: alternate_message_versions.clone(),
```
jordanhendricks, line 777: maybe worth renaming this field since it means something slightly different now?

jordanhendricks (approved):
one small suggestion but otherwise LGTM. thanks!

PR #733, opened 2023-05-15
`measure-iops` should submit different blocks

jmpesp:
Submit different block data as part of the IO loop. Add an option to submit all zeroes instead if that workload is still desired.

### DIFF
--- a/measure_iops/src/main.rs
+++ b/measure_iops/src/main.rs
@@ -48,6 +48,10 @@ pub struct Opt {
 
     #[clap(long, action)]
     bw_limit_in_bytes: Option<usize>,
+
+    /// Submit all zeroes instead of random data
+    #[clap(long, action)]
+    all_zeroes: bool,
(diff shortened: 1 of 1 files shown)


measure_iops/src/main.rs
```diff
@@ -191,7 +199,7 @@ async fn main() -> Result<()> {
 
             io_operations_sent = 0;
             bw_consumed = 0;
-            io_operation_time = Instant::now();
+            measurement_time = Instant::now();
```
bcantrill, line 202: Note that you also need to clear `total_io_time` here to make sure that `fractional_seconds` doesn't measure the I/O time since the start of the run. e.g.:

```suggestion
            measurement_time = Instant::now();
            total_io_time = Duration::ZERO;
```

leftwo:
Did this get rolled into #738 ?  If so, we can just close this.

Issue #735, opened 2023-05-17
Crucible upstairs should take a `Logger` field as part of startup


leftwo:
When working with Propolis or others, crucible should take a `Logger` from them and use that for all log
messages rather than create its own logger.

leftwo:
Possible a new `Option<Logger>` for `up_main()`
This seems to fit with how Propolis will be passing us the log, through volume create, which calls `add_subvolume_create_guest` which then calls `up_main`

PR #736, opened 2023-05-17
up_main take a log option, Block on logging by default

leftwo:
Update up_main to take an option Logger arg and use that instead of creating its own logger.

Made a generic build_logger function (stolen from propolis) that everyone can use.  This build_logger will not drop logging messages if too many are received and instead will force back pressure on whatever is generating the logs.

Part of what we want for: https://github.com/oxidecomputer/crucible/issues/735

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -676,12 +676,18 @@ name = "crucible-common"
 version = "0.0.1"
 dependencies = [
  "anyhow",
+ "atty",
  "nix",
  "rusqlite",
  "rustls-pemfile",
  "schemars",
  "serde",
  "serde_json",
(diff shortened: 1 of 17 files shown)


leftwo:
> Will there be a separate commit for adding this to `Volume::construct`? That's what [propolis uses](https://github.com/oxidecomputer/propolis/blob/181ffc03beab1b24176cd794b7f3fedc09a4b338/lib/propolis/src/block/crucible.rs#L46)

Yeah
The Propolis side changes need to happen also, so I thought I would roll that piece when propolis is ready.

jmpesp (approved):

Issue #731, opened 2023-05-14
upstairs should avoid scans/sorts of ds_active


bcantrill:
After talking to @leftwo, @jclulow, @askfongjojo and @faithanalog to get a lay of the land, I explored Crucible performance a bit, by running `measure-iops` on a local Gimlet (Rev B) with each downstairs dedicated to a single U.2 SSD.  I ran with the defaults for region sizing (512 byte blocks, 100 blocks per extent, 15 extents -- 768K total).

Running `measure-iops` on stock bits yields a five second sawtooth, with a mean of ~560 IOPS (but sawtoothing from ~305 IOPS to ~1300 IOPS).  Looking at CPU profiling when running this yielded us spending time in scans of the `ds_active` HashMap -- specifically in `total_live_work`:

![flame graph of CPU utilization](https://github.com/oxidecomputer/crucible/assets/328614/dc421fc2-d24d-47ea-a385-83706f499eda)

This is consistent with what @faithanalog has seen as well in that scans of `ds_active` serve to make us quadratic with respect to load (we have thousands of jobs in the `ds_active` hash -- ~2800 in this run).

Looking at the code for `total_live_work` it appears that it is trying to get counts by state -- but those counts *appear* to be already maintained by `client_id`.  Draft change for this:

```diff
diff --git a/upstairs/src/lib.rs b/upstairs/src/lib.rs
@@ -3025,14 +3014,11 @@ impl Downstairs {
      * work we have for a downstairs.
      */
     fn total_live_work(&self, client_id: u8) -> usize {
-        self.ds_active
-            .values()
-            .filter(|job| {
-                Some(&IOState::InProgress) == job.state.get(&client_id)
-                    || Some(&IOState::New) == job.state.get(&client_id)
-            })
-            .count()
+        (self.io_state_count.new[client_id as usize]
+            + self.io_state_count.in_progress[client_id as usize])
+            as usize
     }
+
     /**
      * Build a list of jobs that are ready to be acked.
      */
```

Fixing this improved performance, but we immediately ran into the next scan in `submitted_work` -- which can be fixed similarly:

```diff
diff --git a/upstairs/src/lib.rs b/upstairs/src/lib.rs
index ad2f615..9d02971 100644
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -3012,12 +3006,7 @@ impl Downstairs {
      * for this client, but don't yet have a response.
      */
[...]

PR #738, opened 2023-05-18
Eliminate some scans + store jobs in BTree

### LINKED ISSUE #757: default flush timeout should be much lower, tunable to less than 1 second
When creating a Crucible upstairs, the `flush_timeout` is optionally specified in `CrucibleOpts` to be the number of seconds between flushes.  Long flush times increase both space used by Crucible and time (as the number of outstanding jobs grows larger), but short flush times may result in excess CPU usage as fixed work is amortized over less time.  When running `measure-iops` with the fixes described in #731 (implemented in #738), here is the effect of varying flush time:

2023-05-28-21-22-42_2109x1634

The default of 5 seconds is too long -- it leaves quite a bit of performance on the table.  Moreover, it is clear that there is a case to be made for a sub-second flush timeout.  To avoid changing the units or semantics of `flush_timeout`, we propose changing it from an unsigned integer to a float -- and to changing the default from 5.0 to 0.5 (that is, 500ms).

bcantrill:
In addition to fixing #731, this also fixed #757 -- and pulls in #733.

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -4,7 +4,7 @@
 #![allow(clippy::mutex_atomic)]
 
 use std::clone::Clone;
-use std::collections::{HashMap, VecDeque};
+use std::collections::{BTreeMap, HashMap, VecDeque};
 use std::convert::TryFrom;
 use std::fmt;
 use std::fmt::{Debug, Formatter};
@@ -21,7 +21,6 @@ pub use crucible_protocol::*;
 use anyhow::{anyhow, bail, Result};
 pub use bytes::{Bytes, BytesMut};
 use futures::{SinkExt, StreamExt};
-use itertools::Itertools;
 use oximeter::types::ProducerRegistry;
 use rand::prelude::*;
 use ringbuffer::{AllocRingBuffer, RingBufferExt, RingBufferWrite};
@@ -2474,7 +2473,7 @@ struct Downstairs {
      */
     ds_last_flush: Vec<u64>,
     downstairs_errors: HashMap<u8, u64>, // client id -> errors
-    ds_active: HashMap<u64, DownstairsIO>,
+    ds_active: BTreeMap<u64, DownstairsIO>,
     ds_new: Vec<Vec<u64>>,
 
     next_id: u64,
@@ -2552,7 +2551,7 @@ impl Downstairs {
             ds_state: vec![DsState::New; 3],
             ds_last_flush: vec![0; 3],
             downstairs_errors: HashMap::new(),
-            ds_active: HashMap::new(),
+            ds_active: BTreeMap::new(),
             ds_new: vec![Vec::new(); 3],
             completed: AllocRingBuffer::with_capacity(2048),
             completed_jobs: AllocRingBuffer::with_capacity(8),
@@ -2815,19 +2814,14 @@ impl Downstairs {
      * when there is work in the queue.
      */
     fn ds_deactivate_offline(&mut self, client_id: u8) {
-        let mut kvec: Vec<u64> =
-            self.ds_active.keys().cloned().collect::<Vec<u64>>();
-        kvec.sort_unstable();
-
         info!(
             self.log,
             "[{}] client skip all {} jobs for deactivate",
             client_id,
-            kvec.len(),
+            self.ds_active.len(),
         );
-        for ds_id in kvec.iter() {
-            let job = self.ds_active.get_mut(ds_id).unwrap();
 
+        for (ds_id, job) in self.ds_active.iter_mut() {
             let state = job.state.get(&client_id).unwrap();
 
             if *state == IOState::InProgress || *state == IOState::New {
@@ -2858,23 +2852,17 @@ impl Downstairs {
      */
     fn re_new(&mut self, client_id: u8) {
         let lf = self.ds_last_flush[client_id as usize];
-        let mut kvec: Vec<u64> =
-            self.ds_active.keys().cloned().collect::<Vec<u64>>();
-        kvec.sort_unstable();
 
         info!(
             self.log,
-            "[{}] client re-new {} jobs since flush {}",
-            client_id,
-            kvec.len(),
-            lf
+            "[{client_id}] client re-new {} jobs since flush {lf}",
+            self.ds_active.len(),
         );
-        for ds_id in kvec.iter() {
-            let is_read = self.is_read(*ds_id).unwrap();
-            let wc = self.state_count(*ds_id).unwrap();
-            let jobs_completed_ok = wc.completed_ok();
 
-            let job = self.ds_active.get_mut(ds_id).unwrap();
+        for (ds_id, job) in self.ds_active.iter_mut() {
+            let is_read = job.work.is_read();
+            let wc = job.state_count();
+            let jobs_completed_ok = wc.completed_ok();
 
             // We don't need to send anything before our last good flush
             if *ds_id <= lf {
@@ -2939,20 +2927,15 @@ impl Downstairs {
     // notify the correct upstairs task that all downstairs related work
     // for a skipped job has completed.
     fn ds_set_faulted(&mut self, client_id: u8) -> bool {
-        let mut kvec: Vec<u64> =
-            self.ds_active.keys().cloned().collect::<Vec<u64>>();
-        kvec.sort_unstable();
-
         info!(
             self.log,
-            "[{}] client skip {} in process jobs because fault",
-            client_id,
-            kvec.len(),
+            "[{client_id}] client skip {} in process jobs because fault",
+            self.ds_active.len(),
         );
         let mut notify_guest = false;
-        for ds_id in kvec.iter() {
-            let job = self.ds_active.get_mut(ds_id).unwrap();
+        let mut retire_check = vec![];
 
+        for (ds_id, job) in self.ds_active.iter_mut() {
             let state = job.state.get(&client_id).unwrap();
 
             if *state == IOState::InProgress || *state == IOState::New {
@@ -2968,7 +2951,9 @@ impl Downstairs {
                 // Check to see if this being skipped means we can ACK
                 // the job back to the guest.
                 if job.ack_status == AckStatus::Acked {
-                    self.retire_check(*ds_id);
+                    // Push this onto a queue to do the retire check when
+                    // we aren't doing a mutable iteration.
+                    retire_check.push(*ds_id);
                 } else if job.ack_status == AckStatus::NotAcked {
                     let wc = job.state_count();
                     if (wc.error + wc.skipped + wc.done) == 3 {
@@ -2990,6 +2975,11 @@ impl Downstairs {
                 }
             }
         }
+
+        for ds_id in retire_check {
+            self.retire_check(ds_id);
+        }
+
         notify_guest
     }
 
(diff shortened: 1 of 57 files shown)


leftwo:
Love the improvements, a marked difference.

I did hit a new panic running the `tools/test_reconnect.sh` test, but I'm not sure yet if this
is the fault of these changes, though they are my first suspect. 

The panic happens when the upstairs has decided that a downstairs has been "gone too long"
and has marked that downstairs as faulted.  This should result in all new IO for this downstairs, and 
any current IO that has not completed being marked as "skipped".  Without the `LIveRepair`
changes in place yet, the downstairs will just stay as faulted.   And, indeed on stock bits, this
is what happens and we don't see a panic.  Even when that downstairs "comes back", it remains
in purgatory (OnlineRepairReady) and is not allowed to re-join the region set.

However, when this same situation happens on the perf bits, I'm hitting this panic:
```
Dec 28 17:25:28.389 WARN Full(5398) Failed to notify client 2 of work 5398
05398/15000 Write block 4853  len 24576  data:   4   3   2   2   3   3
Dec 28 17:25:28.389 WARN Full(5399) Failed to notify client 2 of work 5399
05399/15000 Write block 3288  len  8192  data:   3   3
Dec 28 17:25:28.390 WARN Full(5400) Failed to notify client 2 of work 5400
Dec 28 17:25:28.390 INFO [2] a1bcf246-0610-4b3a-bd97-22b1d6a593e1 (f84b9698-7b00-4e87-97e5-bbd78792b8d4) Active Active Faulted ds_transition to OnlineRepairReady
Dec 28 17:25:28.390 INFO [2] Transition from Faulted to OnlineRepairReady
Dec 28 17:25:28.390 WARN [2] new RM replaced this: Some(RegionMetadata { generation: [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1], flush_numbers: [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1], dirty: [false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false] })
Dec 28 17:25:28.390 INFO [2] a1bcf246-0610-4b3a-bd97-22b1d6a593e1 Enter Ready for Online Repair mode
Dec 28 17:25:28.390 INFO [2] Starts cmd_loop
Dec 28 17:25:28.390 INFO [2] 127.0.0.1:8830 task reports connection:true
thread 'tokio-runtime-worker' panicked at 'called `Option::unwrap()` on a `None` value', upstairs/src/lib.rs:2601:19
Dec 28 17:25:28.446 INFO accepted connection, remote_addr: 127.0.0.1:45964, local_addr: 127.0.0.1:9998
```

The panic is in `in_progress()` while attempting to pull a job off the `ds_active` list:
```
    fn in_progress(&mut self, ds_id: u64, client_id: u8) -> Option<IOop> {
        let job = self.ds_active.get_mut(&ds_id).unwrap();
```

Now, I don't know what `job_id` that was yet, and I don't know which `client_id` it was either, trying to see
if I can figure that out, and how I also might improve the panic message here.

I can re-create this panic pretty easy, so I'll just dig a little further and try to understand where exactly we
hit it.

leftwo:
Ah, I think what is missing is having something like this:
```
self.ds_new[client_id as usize] = Vec::new();
```
In `ds_set_faulted()`.

Adding that and I no longer get the panic when a faulted downstairs attempts to rejoin the region set.

bcantrill:
> Here is the diff that will fix the panic I found, plus an improvement on the panic message.
> 
> ```
> diff --git a/upstairs/src/lib.rs b/upstairs/src/lib.rs
> index beb8c00..e3edc7b 100644
> --- a/upstairs/src/lib.rs
> +++ b/upstairs/src/lib.rs
> @@ -2598,7 +2598,12 @@ impl Downstairs {
>       * has no work to do, so return None.
>       */
>      fn in_progress(&mut self, ds_id: u64, client_id: u8) -> Option<IOop> {
> -        let job = self.ds_active.get_mut(&ds_id).unwrap();
> +        let job = match self.ds_active.get_mut(&ds_id) {
> +            Some(job) => job,
> +            None => {
> +                panic!("[{client_id}] in_progress can't find job {ds_id}");
> +            }
> +        };
>  
>          // If current state is Skipped, then we have nothing to do here.
>          if job.state[&client_id] == IOState::Skipped {
> @@ -2948,6 +2953,8 @@ impl Downstairs {
>          let mut notify_guest = false;
>          let mut retire_check = vec![];
>  
> +        // There should be no new jobs for this downstairs
> +        self.ds_new[client_id as usize] = Vec::new();
>          for (ds_id, job) in self.ds_active.iter_mut() {
>              let state = job.state.get(&client_id).unwrap();
> ```

I was able to reproduce this by pstop'ing a downstairs while running `measure-iops`.  Doing this revealed several problems: there were three other places that `ds_new` could become out of sync with `ds_active`; those have now all been fixed.

upstairs/src/lib.rs
```diff
@@ -3324,45 +3342,36 @@ impl Downstairs {
...
-            .filter_map(|job| {
-                if let Some(IOState::New) = job.state.get(&client_id) {
-                    Some(job.ds_id)
-                } else {
-                    None
-                }
-            })
-            .collect()
+    fn new_work(&mut self, client_id: u8) -> Vec<u64> {
+        self.ds_new[client_id as usize].drain(..).collect()
```
jmpesp, line 3346:
```patch
index 99df635..80cf58f 100644
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -3343,7 +3345,22 @@ impl Downstairs {
      * requests for this client.
      */
     fn new_work(&mut self, client_id: u8) -> Vec<u64> {
-        self.ds_new[client_id as usize].drain(..).collect()
+        let left: Vec<u64> = self.ds_active
+            .values()
+            .filter_map(|job| {
+                if let Some(IOState::New) = job.state.get(&client_id) {
+                    Some(job.ds_id)
+                } else {
+                    None
+                }
+            })
+            .collect();
+
+        let right: Vec<u64> = self.ds_new[client_id as usize].drain(..).collect();
+
+        debug_assert_eq!(left.len(), right.len());
+
+        right
     }
 
     /**
```

upstairs/src/lib.rs
```diff
@@ -3324,45 +3342,36 @@ impl Downstairs {
...
+    }
+
+    /**
+     * Called to requeue work that was previously found by calling
+     * [`new_work`], presumably due to flow control.
+     */
+    fn requeue_work(&mut self, client_id: u8, work: &[u64]) {
+        self.ds_new[client_id as usize].extend_from_slice(work);
```
jmpesp, line 3354:
```patch
index 99df635..80cf58f 100644
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -3351,6 +3368,10 @@ impl Downstairs {
      * [`new_work`], presumably due to flow control.
      */
     fn requeue_work(&mut self, client_id: u8, work: &[u64]) {
+        for job_id in work {
+            let job = self.ds_active.get_mut(&job_id).unwrap();
+            debug_assert_eq!(*job.state.get(&client_id).unwrap(), IOState::New);
+        }
         self.ds_new[client_id as usize].extend_from_slice(work);
     }
 
```

upstairs/src/lib.rs
```diff
@@ -3324,45 +3342,36 @@ impl Downstairs {
...
      */
     fn total_live_work(&self, client_id: u8) -> usize {
-        self.ds_active
-            .values()
-            .filter(|job| {
-                Some(&IOState::InProgress) == job.state.get(&client_id)
-                    || Some(&IOState::New) == job.state.get(&client_id)
-            })
-            .count()
+        (self.io_state_count.new[client_id as usize]
```
jmpesp, line 3370:
```patch
index 99df635..136ba02 100644
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -3367,6 +3388,11 @@ impl Downstairs {
      * work we have for a downstairs.
      */
     fn total_live_work(&self, client_id: u8) -> usize {
+        debug_assert_eq!(
+            self.io_state_count.new[client_id as usize] as usize,
+            self.ds_new[client_id as usize].len(),
+        );
+
         (self.io_state_count.new[client_id as usize]
             + self.io_state_count.in_progress[client_id as usize])
             as usize
```

jmpesp (approved):
Overall I'm good with this PR (IOPS GO BRRRR), except:

> Doing this revealed several problems: there were three other places that `ds_new` could become out of sync with `ds_active`; those have now all been fixed.

I'm anxious about the validity of our current tests, and about future cases of drift here. All tests pass on this PR, but 68 tests fail if I add an assertion that `ds_new` **doesn't** contain the job we're trying to mark InProgress:

```patch
index 99df635..124946e 100644
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -2884,6 +2885,7 @@ impl Downstairs {
         let new_state = IOState::InProgress;
         let old_state = job.state.insert(client_id, new_state.clone()).unwrap();
         assert_eq!(old_state, IOState::New);
+        assert!(!self.ds_new[client_id as usize].contains(&ds_id));
         self.io_state_count.decr(&old_state, client_id);
         self.io_state_count.incr(&new_state, client_id);

```

The tests that fail call `ds.in_progress` directly instead of getting `io_send` to do it, which properly takes care of keeping `ds_new` correct. If I instead assert that the ds_id **is** contained in `ds_new`, then I instead see the two new dummy downstairs tests fail (where `io_send` has removed the job id from `ds_new` before calling `in_progress`). I don't think that assertion is correct or should be added in, but I do think more checks should be added in. I've suggested a few in the comments, keeping them as `debug_assert` so that they are (I assume!) compiled out to preserve the performance gains seen here.

(not blocking this PR) I wonder if it's worth it to separate ds_active, ds_new, io_state_count, and possibly more into one new struct, where correctness can be maintained through the functions of that struct.

upstairs/src/lib.rs
```diff
@@ -4475,26 +4487,23 @@ impl Downstairs {
...
-                .cloned()
-                .filter(|&x| x <= ds_id)
-                .collect::<Vec<u64>>();
+            // Retire all the jobs that happened before and including this
+            // flush.
+            let mut retired = Vec::new();
 
-            kvec.sort_unstable();
+            loop {
+                let id = match self.ds_active.keys().next() {
```
leftwo, line 4495: I guess this will still be sorted by key?  I'm not sure how we are making sure that
the first key we get that is above our flush allows us to break unless we are sure
here that `self.ds_active.keys().next()` will iterate from lowest to highest key value.

It may just be that, by the nature of our insertion/creation of keys that they always
end up in order, but if so, perhaps we need to comment that somewhere so we
don't accidentally lose that order that we now depend on here.

upstairs/src/live_repair.rs
```diff
@@ -3064,9 +3055,7 @@ pub mod repair_test {
 
         // Verify that the future repair jobs were added to our IOs
         // dependency list.
-        let keys: Vec<&u64> = ds.ds_active.keys().sorted().collect();
-        let jobs: Vec<&DownstairsIO> =
-            keys.iter().map(|k| ds.ds_active.get(k).unwrap()).collect();
+        let jobs: Vec<&DownstairsIO> = ds.ds_active.values().collect();
```
leftwo, line 3058: This, not needing to sort the job list, probably makes the test run faster too!

leftwo (commented):
I just have the one concern about guaranteeing that our `ds_active`, when iterating,
will indeed always return in key_id lowest to highest.  I know we always allocate in
order, and add to the end of the list in order.  I'm just a little nervous that there is a
path back in where we could put something out of order.  The `LiveRepair` stuff
does do some reservations for job IDs before they actually exist, and this might be
a place where it could happen.

I'll dig a little further to see if I can find a situation like I'm concerned about.

leftwo:
> 

Ah, yes, it's a b-tree now.  Ignore my fears about this.

upstairs/src/lib.rs
```diff
@@ -600,24 +600,36 @@ where
...
             continue;
         }
 
         /*
          * If in_progress returns None, it means that this job on this
          * client should be skipped.
          */
-        let job = u.downstairs.lock().await.in_progress(*new_id, client_id);
+        let job = u.downstairs.lock().await.in_progress(new_id, client_id);
         if job.is_none() {
```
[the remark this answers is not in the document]
leftwo: https://github.com/oxidecomputer/crucible/issues/760 for this

leftwo (approved):
Ok, BTree, I'm all good to go here.

upstairs/src/lib.rs
bcantrill: Filed as #765.

PR #740, opened 2023-05-19
Add dummy downstairs tests

jmpesp:
As a precursor to changing the Upstairs to deal with a Downstairs **not** considering a job complete when there's an error, implement a test framework that allows Upstairs tests to be written where specific messages can be sent to the Upstairs from three dummy downstairs.

Two tests were written to show what the framework can do:

- verify flow control behaviour
- verify replay of messages when a downstairs disconnects and reconnects

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,3 +23,6 @@ members = [
     "smf",
     "upstairs",
 ]
+
+[profile.dev]
+panic = 'abort'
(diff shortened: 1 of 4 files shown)


leftwo (approved):
Approve,  ...but..

> As a precursor to changing the Upstairs to deal with a Downstairs not considering a job complete when there's an error

I still think we have to discuss this idea.  Don't go off and implement it before we have decided if it's the best way forward :)

I don't want you to confuse my approval of the code changes here with the potential future plan, as that idea has details we need to flush out.

I do like this test idea, and we can grow this to checking the various error cases that could come up during negotiation as well.

PR #747, opened 2023-05-26
Add a clippy run to Rust workflows

jmpesp:
Also address what clippy complained about.


jmpesp:
> Does this mean the build will fail if clippy fails?

I don't think so, I think for that to happen it has to be added (in the repo settings) to the list of checks that need to pass?

leftwo (approved):

PR #751, opened 2023-05-26
Add the first part of downstairs replacement

jmpesp:
Adds `replace_downstairs` to the BlockIO trait, a ReplaceDownstairs BlockOp, and an integration test that should pass once the functionality is in place.

This commit came from a pairing session with Alan, so Github please consider him a co-author!

### DIFF
--- a/integration_tests/src/lib.rs
+++ b/integration_tests/src/lib.rs
@@ -235,6 +235,21 @@ mod test {
 
             Ok(())
         }
+
+        pub async fn new_downstairs(&self) -> Result<TestDownstairs> {
+            TestDownstairs::new(
+                "127.0.0.1".parse()?,
+                true,
+                false,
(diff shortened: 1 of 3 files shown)


integration_tests/src/lib.rs
```diff
@@ -2130,6 +2145,62 @@ mod test {
...
+
+        // XXX this should work!
+        volume
+            .replace_downstairs(
+                test_downstairs_set.opts().id,
+                test_downstairs_set.downstairs1_address().await,
+                new_downstairs.address().await,
+            )
+            .await
+            .err()
```
leftwo, line 2198: Is the `.err()` here why this test passes now?

jmpesp: It is yeah

leftwo (approved):

PR #766, opened 2023-05-31
remove some &mut self for Downstairs

jmpesp:


jmpesp:
Note, this is the first PR from breaking up #659

leftwo (approved):
The build-and-test job was failing, but it looks like it was some transient
issue with some pantry jobs.  I forced a re-run of the job and it worked the
second time.  I'm sure it has nothing to do with these changes though.

Issue #762, opened 2023-05-30
Fix accounting in DTrace new job


leftwo:
when a downstairs is Offline, the dtrace counters for new jobs accumulate.

```
       DS STATE 0        DS STATE 1        DS STATE 2   UPW  DSW  NEW0 NEW1 NEW2   IP0  IP1  IP2    D0   D1   D2    S0   S1   S2  E0 E1 E2
           active            active            active     0    0     0    0    0     0    0    0     0    0    0     0    0    0   0  0  0
           active            active            active     0    0     0    0    0     0    0    0     0    0    0     0    0    0   0  0  0
           active            active            active     0    0     0    0    0     0    0    0     0    0    0     0    0    0   0  0  0
           active           offline            active     0    0     0    0    0     0    0    0     0    0    0     0    0    0   0  0  0
           active           offline            active     1    2     1    2    1     0    0    0     1    0    1     0    0    0   0  0  0
           active           offline            active     0    2     0    2    0     0    0    0     2    0    2     0    0    0   0  0  0
           active           offline            active     1    5     1    5    1     0    0    0     4    0    4     0    0    0   0  0  0
       DS STATE 0        DS STATE 1        DS STATE 2   UPW  DSW  NEW0 NEW1 NEW2   IP0  IP1  IP2    D0   D1   D2    S0   S1   S2  E0 E1 E2
           active           offline            active     1   37     1   37    1     0    0    0    36    0   36     0    0    0   0  0  0
           active           offline            active     1   53     1   53    1     0    0    0    52    0   52     0    0    0   0  0  0
           active           offline            active     0   53     0   53    0     0    0    0    53    0   53     0    0    0   0  0  0
```

These should reflect going straight to skipped, which is what the jobs actually do.

leftwo:
When the Live Repair happens, we also have "New" jobs showing up that don't get cleared.
In this example, the LR happened too fast for DTrace to see it, but we are left with a bunch of New
jobs that don't go away.
```
       DS STATE 0        DS STATE 1        DS STATE 2   UPW  DSW  NEW0 NEW1 NEW2   IP0  IP1  IP2    D0   D1   D2    S0   S1   S2  E0 E1 E2
           active           faulted            active     0    0     0    0    0     0    0    0     0    0    0     0    0    0   0  0  0
           active           faulted            active     0    0     0    0    0     0    0    0     0    0    0     0    0    0   0  0  0
[...]

PR #767, opened 2023-05-31
Fix double counting of repair jobs

leftwo:
Found and removed the double stat counting for DTrace #762

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4673,7 +4673,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "97fee6b57c6a41524a810daee9286c02d7752c4253064d0b05472833a438f675"
 dependencies = [
  "cfg-if",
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -4673,7 +4673,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "97fee6b57c6a41524a810daee9286c02d7752c4253064d0b05472833a438f675"
 dependencies = [
  "cfg-if",
- "rand 0.8.5",
```
jmpesp, line 4676: this is a strange downgrade to be present on this PR?

jmpesp (approved):

Cargo.lock
leftwo: Yeah, weird, your haunted cargo problems are contagious.
I've reverted that change.

PR #769, opened 2023-06-01
Update all Omicron packages in Cargo.lock

jmpesp:
Turns out the Omicron revs in Cargo.lock were OLD!

Here's what I did:

```
for dep in $(
    (for cargo_file in $(find . -name Cargo.toml ! -type d | xargs rg omicron -l);
    do
        grep omicron ${cargo_file} | grep git;
    done;) | sort -u | cut -d' ' -f1);
do
    cargo update -p $dep;
done;
```

This updated Omicron from bac9b299 (Feb 7, 2023) to bd6c6280 (main as of this commit).

This pulled in a new dropshot and oximeter, which necessitated some other changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -56,10 +56,21 @@ dependencies = [
 ]
 
 [[package]]
-name = "aho-corasick"
-version = "0.7.18"
+name = "ahash"
+version = "0.8.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


leftwo (approved):
What has dependabot been doing (or not doing) all this time..

If it passes CI, then yeah, go for it.

leftwo:
We should probably do the "workspace" thing that Omicron and Propolis have done, so all workspaces share the same dependencies.

PR #772, opened 2023-06-01
Allow dsc more than three downstairs.

leftwo:
Updated the dsc tool to be able to create and start more than just three downstairs.  
Updated the dsc test to check and allow for more than three.

This will pave the way for test support for replacement of a downstairs.  By allowing dsc
to create and start additional downstairs processes, we can automate replacing one
downstairs with another.

### DIFF
--- a/dsc/src/main.rs
+++ b/dsc/src/main.rs
@@ -92,6 +92,10 @@ enum Action {
             action
         )]
         region_dir: Vec<PathBuf>,
+
+        /// The number of region directories to create.
+        #[clap(long, default_value = "3", action)]
+        region_count: usize,
     },
     /// Test creation of downstairs regions
     RegionPerf {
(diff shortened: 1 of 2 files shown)


leftwo:
I broke this little bit out to make the downstairs replacement PR have less in it.

dsc/src/main.rs
```diff
@@ -341,56 +351,38 @@ impl DscInfo {
...
+        let mut ds_state = Vec::new();
+        let mut ds_pid = Vec::new();
+
+        for _ in 0..region_count {
+            ds_state.push(DownstairsState::Stopped);
+            ds_pid.push(None);
+        }
 
         // If we only have one region directory, then use that for all
         // three downstairs.  If we have three, then each downstairs can
```
jmpesp, line 363: `If we have N` here?

jmpesp (approved):
:package:

dsc/src/main.rs
leftwo: Ah yeah, I'll update the comment before merging.

Issue #771, opened 2023-06-01
downstairs: switch to unix-excl VFS


faithanalog:
The SQLite VFS is at the lowest level of how SQLite stack. You should read the article here https://www.sqlite.org/vfs.html if you want to know more. We'll be talking about this, and also the WAL Index ( https://www.sqlite.org/walformat.html#shm )

For the purposes of this issue, here's what you need to know:

By default SQLite uses the `unix` VFS. To allow multiple processes to access the database at the same time, SQLite memory-maps a `shm` file. If you're curious, SQLite doesn't actually care about this data being persisted at all:

>  Since the content of the shm file does not need to be preserved across a crash, the shm file is never `fsync()`-ed to disk. In fact, if there were a mechanism by which SQLite could tell the operating system to never persist the shm file to disk but always hold it in cache memory, SQLite would use that mechanism to avoid any unnecessary disk I/O associated with the shm file. However, no such mechanism exists in standard posix.

The `unix` VFS also does not acquire any locks on the database file itself, so each connection has to assume some other process may be changing it at the same time.

Anyways, all this allows multiple processes to access the database simultaneously, but introduces some overhead. For crucible, we don't care about multiple processes accessing our extent DBs simultaneously, and in fact, we'd really prefer they didn't.

So, we can switch to the `unix-excl` VFS. Under this, a process acquires an exclusive lock on the DB file when it connects to the database, and the shm file is eliminated because it's no longer needed. This results in a nice bump in perf, and as a bonus adds a barrier to multiple processes opening the same extents at the same time (perhaps we'd like more such barriers in future?)

(sorry for messing up your graph bryan, but i could not see that top line on my monitor because it is from 2003)

![2023-05-28-15-20-30_2106x1630-adjusted](https://github.com/oxidecomputer/crucible/assets/1389549/d6a74f65-affe-4b7b-a64f-9ebeae5ec0c7)

## What this **DOESNT** do

This doesn't forbid multiple connections _within the same process_ from accessing the DB. Instead, synchronization between mulitple in-process sqlite connections is just moved to the heap. If we wanted to squeeze out a bit more perf under the single-connection model (and maybe we do, right now) we could ALSO set the db to exclusive locking mode. https://www.sqlite.org/pragma.html#pragma_locking_mode

[...]

PR #773, opened 2023-06-02
downstairs: switch to unix-excl VFS

faithanalog:
see https://github.com/oxidecomputer/crucible/issues/771 for the details on this. This is the implementation.

This is the code that actually matters:

```
    let metadb: Connection = Connection::open_with_flags_and_vfs(
        path,
        OpenFlags::default(),
        "unix-excl",
    )?;
```

The rest of this PR is stripping out the `shm` from the rest of the codebase, and updating a few test scripts to shut down running downstairs before running dumps.

### DIFF
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -474,15 +474,15 @@ pub fn validate_repair_files(eid: usize, files: &[String]) -> bool {
 /// Always open sqlite with journaling, and synchronous.
 /// Note: these pragma_updates are not durable
 fn open_sqlite_connection<P: AsRef<Path>>(path: &P) -> Result<Connection> {
-    // By default, sqlite uses these shm files for synchronization to support
-    // DB access from multiple processes. We don't need that at all in crucible;
-    // we are only going to access the DB from the single process. So we change
-    // it to unix-excl which prevents other processes from accessing the DB, and
+    // By default, sqlite uses these shm files for synchronization to support DB
+    // access from multiple processes. We don't need that at all in crucible; we
+    // are only going to access the DB from the single process. So we change it
+    // to unix-excl which prevents other processes from accessing the DB, and
     // keeps everything in the heap, but still allows multiple connections from
     // in-process. If we ultimately decide we don't want to hold multiple
     // connections per DB (and thus, we don't want any parallelism of SQLite
-    // operations within an extent), we could change this to unix-none to forego
-    // locking altogether.
+    // operations within an extent), we could additionally set the
+    // `locking_mode` pragma to EXCLUSIVE.
     //
     // See https://www.sqlite.org/vfs.html#standard_unix_vfses for more info.
     let metadb: Connection = Connection::open_with_flags_and_vfs(
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -485,14 +485,11 @@ fn open_sqlite_connection<P: AsRef<Path>>(path: &P) -> Result<Connection> {
     // locking altogether.
     //
     // See https://www.sqlite.org/vfs.html#standard_unix_vfses for more info.
-    // metadb.pragma_update(None, "", "")?;
     let metadb: Connection = Connection::open_with_flags_and_vfs(
         path,
         OpenFlags::default(),
         "unix-excl",
     )?;
-
-    // let metadb = Connection::open(path)?;
 
     assert!(metadb.is_autocommit());
     metadb.pragma_update(None, "journal_mode", "WAL")?;
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -1848,34 +1848,29 @@ impl Region {
         };
         save_stream_to_file(extent_db, repair_stream.into_inner()).await?;
 
-        // These next two are optional.
-        // YYY
-        for opt_file in &[ExtentType::DbWal] {
-            let filename = extent_file_name(eid as u32, opt_file.clone());
+        // the WAL log isn't always there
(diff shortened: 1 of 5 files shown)


downstairs/src/region.rs
```diff
@@ -2943,6 +2914,10 @@ mod test {
 
     #[test]
     fn validate_repair_files_quad_duplicate() {
+        // XXX I don't know why this test exists. we definitely shouldnt be
+        // involving shm in anything anymore. but what was this supposed to even
+        // be doing? idk.
```
jmpesp, line 2919: I think it's a test for the case where there are 4 files (previously expected to be 2 or 4), but one is a duplicate of the other, instead of 4 unique files.

tools/test_repair.sh
```diff
@@ -140,17 +140,14 @@ for (( i = 0; i < 100; i += 1 )); do
...
-        wait "$ds0_pid" || true
-    elif [[ $choice -eq 1 ]]; then
-        kill "$ds1_pid"
-        wait "$ds1_pid" || true
-    else
-        kill "$ds2_pid"
-        wait "$ds2_pid" || true
-    fi
+
+    # Stop everything before dump so dump can access the extent metadata
```
jmpesp, line 144: The previous comment's mention of not allowing IOs to complete is good context here, that should be kept.

jmpesp (changes requested):
Looks good to me, but unfortunately this doesn't work: any snapshot of a disk / volume we take means the ZFS snapshot will be taken when the database is open, and trying to launch a read-only downstairs from this snapshot doesn't work:

```
[ Jun  2 09:31:47 Executing start method ("/opt/oxide/lib/svc/manifest/crucible/downstairs.sh"). ]
{"msg":"Opened existing region file \"/data/regions/eeb75f44-c435-4cff-8c93-3d6d4dc5ff58/.zfs/snapshot/580a7b89-f74d-4d2d-87f9-78f83b58e946/region.json\"","v":0,"name":"crucible","level":30,"time":"2023-06-02T09:31:47.376914861-07:00","hostname":"oxz_crucible_oxp_ed9a1cbd-55d2-ce10-eed0-f6f4aed4bcf7","pid":15654}
{"msg":"Error: Open of db file \"/data/regions/eeb75f44-c435-4cff-8c93-3d6d4dc5ff58/.zfs/snapshot/580a7b89-f74d-4d2d-87f9-78f83b58e946/00/000/000.db\" for extent#0 returned: unable to open database file","v":0,"name":"crucible","level":50,"time":"2023-06-02T09:31:47.377533137-07:00","hostname":"oxz_crucible_oxp_ed9a1cbd-55d2-ce10-eed0-f6f4aed4bcf7","pid":15654}
Error: Open of db file "/data/regions/eeb75f44-c435-4cff-8c93-3d6d4dc5ff58/.zfs/snapshot/580a7b89-f74d-4d2d-87f9-78f83b58e946/00/000/000.db" for extent#0 returned: unable to open database file
[ Jun  2 09:31:47 Stopping because service exited with an error. ]
[ Jun  2 09:31:47 Failing too quickly, throttling. ]
[ Jun  2 09:31:48 Executing start method ("/opt/oxide/lib/svc/manifest/crucible/downstairs.sh"). ]
```

faithanalog:
> any snapshot of a disk / volume we take means the ZFS snapshot will be taken when the database is open, and trying to launch a read-only downstairs from this snapshot doesn't work:

ah bullshit, ok. I'll ponder/poke at that

downstairs/src/region.rs
leftwo: Yeah, exactly, it was checking to be sure that the count was 4 and that we had 4 unique file names.

tools/test_repair.sh
leftwo: For a "--lossy" downstairs, the intent was that that downstairs was doing IO slower
than the other three downstairs.  Slow enough that indeed it had not completed all
the IOs that the other two downstairs had completed.  This functionality is important
for the test as we are testing a repair here, so to perform a repair, there must be some
difference between the three downstairs :)

faithanalog: i thought the point of lossy was that it would straight up drop the IOs rather than just do them slower?

leftwo: Perhaps poorly named, yeah.
No, we don't drop any IOs, but the downstairs will "skip over" some as it walks the work list
It will also add a tiny delay into completing them.  So, eventually the IOs will finish, but it
will take longer than the other downstairs.

faithanalog:
the change just pushed does work for reading from RO filesystems. but i have some misgivings (see the comments). I really don't like that it would allow you to run `crucible-downstairs dump` without error against a currently-open region. feedback requested

jclulow:
I would use flock, not fcntl.  It's difficult to guarantee that fcntl locks are being used correctly in library routines and in the face of multiple threads.

jclulow:
I'm also not sure if you'd be able to lock files in a ZFS snapshot, so it may first be worth investigating a way to tell if what you're looking at is a snapshot or a live file system, and only doing the immutable open in the snapshot case?

faithanalog:
> I'm also not sure if you'd be able to lock files in a ZFS snapshot, so it may first be worth investigating a way to tell if what you're looking at is a snapshot or a live file system, and only doing the immutable open in the snapshot case?

yeah i'd prefer that I think. I could scan all the snapshots' mount points and check if its a descendant of any of those. Does that seem reasonable or is there something better I ought to do?

faithanalog:
read through that today and it all makes sense to me. ill look at implementing it into crucible

Issue #778, opened 2023-06-03
retire_check loop leaves no job behind when it sometimes should


leftwo:
Retire check will loop until either there are no jobs left on the `ds_active` list, or
it comes to one that is greater than the flush.  

```
            loop {
                let id = match self.ds_active.keys().next() {
                    Some(id) if *id > ds_id => break,
                    None => break,
                    Some(id) => *id,
                };
```

However, there exists two ways that a job < the flush ID is not ready to retire
yet.  First that the ACK for the flush from the downstairs came back ahead of
the ACK for a job that did depend on the flush, and did finish before the flush, but
lost the race back to the upstairs.

```
                if wc.active != 0 {
                    warn!(
                        self.log,
                        "Leave job {} on the work queue for now", id
                    );
                    continue;
                }
```

There is also the possibility that a job with an ID below the flush
might not have acked back to the guest yet, and for those jobs they should
remain on the active till the upstairs has sent the ack back to the guest:
 
```
                let oj = self.ds_active.get(&id).unwrap();
                if oj.ack_status != AckStatus::Acked {
                    warn!(
                        self.log,
                        "[rc] leave job {} on the queue when removing {}",
                        oj.ds_id,
                        ds_id,
                    );
                    continue;
                }
```

These cases need to be handled.  I believe the fix is just to iterate over the `ds_active` list once
and pull out the ones to retire, leaving behind the ones that still have things to do.

PR #779, opened 2023-06-05
Allow some jobs to remain behind

leftwo:
The retire_check loop could get stuck when it encountered a job that was bellow the 
flush ID, but was not ready to retire.  Update the loop to allow retire_check to finish, but 
add back jobs that were not ready to retire.

I suspect this may have been causing the `test_live_repair.sh` test to run forever.


leftwo:
https://github.com/oxidecomputer/crucible/issues/778

jmpesp (approved):
ssssssssssssssssssssssssssssssssssssship it

PR #781, opened 2023-06-06
use workspace dependencies

jmpesp:


leftwo (approved):
YES!
I've been wanting this.  I thought I had created an issue for this, but apparently I did not 
even do that much.  If this passes CI, then check it in.

jmpesp:
> I thought I had created an issue for this, but apparently I did not...

You said https://github.com/oxidecomputer/crucible/pull/769#issuecomment-1572869897, which was good enough for me :)

PR #782, opened 2023-06-06
Add downstairs database version.

leftwo:
Added two fields to the region.json file for identifying the read and write versions for database files.

It's a bit rudimentary, and if there is a mismatch the downstairs will refuse to start, but it at least gives
us a starting point to work forward from.


faithanalog (approved):
looks good. I like that the version check requires an exact match right now. We can relax that to a `>=` = ok check later once we actually implement format migrations.

PR #783, opened 2023-06-06
Accept a length for pantry validation requests

jmpesp:
When the user is uploading a file to a disk, they may specify that the disk's size is larger than the file size. In that case, an unbounded validation request's digest would not match the file's digest because the pantry would be hashing all the blocks of zeros after the user's uploaded bits.

Change the validation request to accept a size to validate in bytes.

### DIFF
--- a/integration_tests/src/lib.rs
+++ b/integration_tests/src/lib.rs
@@ -3768,96 +3768,9 @@ mod test {
 
         let tds = TestDownstairsSet::small(false).await.unwrap();
 
-        // Start a pantry, get the client for it, then use it to bulk_write in
-        // data
+        // Start a pantry, get the client for it
         let (_pantry, volume_id, client) =
             get_pantry_and_client_for_tds(&tds).await;
-
-        // parallel bulk write in a bunch of random data in a random order
-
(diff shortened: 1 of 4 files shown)


integration_tests/src/lib.rs
```diff
@@ -3624,4 +3758,135 @@ mod test {
 
         client.detach(&volume_id.to_string()).await.unwrap();
     }
+
+    // Test validating a non-block size amount fails
+    #[tokio::test]
+    async fn test_pantry_validate_fail() {
+        const BLOCK_SIZE: usize = 512;
```
leftwo, line 3765: This seems like a bunch of code for a test that might not need all this?
Could this be simpler in that it has a volume created but without any import,
then just a verify with invalid size?  Like we know it will never get to the
validate part, so why bother actually doing it when the check itself will
stop first?

leftwo (approved):
Just one thought on a test simplification.

integration_tests/src/lib.rs
jmpesp: You're right, I simplified this test in 293c712

PR #784, opened 2023-06-06
Relax base64 dependency

jmpesp:


jmpesp:
I'm going to bump the crucible rev in propolis and omicron again, and this is needed because omicron specifies `0.21.2`. Unfortunately even Propolis refused to build without this. I'm going to make this change here, then also relax omicron's dep to `0.21`.

jmpesp:
I managed to get propolis to build by running `cargo update -p base64` there. I don't think this PR is required anymore...

PR #789, opened 2023-06-09
Add a quick fill for crutest, fix clippy warnings

leftwo:
Added a quick fill for crutest that will do one write to every extent in a region.  This can be used to be sure that all extents are dirty/changed without having to write to every single block in a volume.

Fixed clippy warnings for region default

### DIFF
--- a/crutest/src/main.rs
+++ b/crutest/src/main.rs
@@ -1452,11 +1452,12 @@ async fn fill_sparse_workload(
 
     // Figure out how many extents we have
     let extents = ri.total_blocks / (ri.extent_size.value as usize);
+    let extent_size = ri.extent_size.value as usize;
 
     // Do one write to each extent.
-    for extent in 0..extents {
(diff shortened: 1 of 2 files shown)


crutest/src/main.rs
```diff
@@ -1432,6 +1440,41 @@ async fn fill_workload(
...
+) -> Result<()> {
+    let mut rng = rand_chacha::ChaCha8Rng::from_entropy();
+
+    // Figure out how many extents we have
+    let extents = ri.total_blocks / (ri.extent_size.value as usize);
+
+    // Do one write to each extent.
+    for extent in 0..extents {
+        let mut block_index = extent * (ri.extent_size.value as usize);
+        let random_offset = rng.gen_range(0..100);
```
jmpesp, line 1459: `0..ri.extent_size.value` instead?

jmpesp (approved):

PR #791, opened 2023-06-12
Update dumb_downstairs test to be smarter

leftwo:
Updated parts of the dummy_downstairs to allow for ping packets as well as different results from channels that might be either closed or empty.

### DIFF
--- a/.github/buildomat/jobs/build.sh
+++ b/.github/buildomat/jobs/build.sh
@@ -20,8 +20,6 @@ rustc --version
 banner build
 ptime -m cargo build --verbose
 
-ptime -m cargo test test_successful_live_repair --verbose
-
 banner output
 mkdir -p /work/bins
 for t in crucible-downstairs crucible-hammer crutest dsc; do
--- a/.github/buildomat/jobs/build.sh
+++ b/.github/buildomat/jobs/build.sh
@@ -20,6 +20,8 @@ rustc --version
 banner build
 ptime -m cargo build --verbose
 
+ptime -m cargo test test_successful_live_repair --verbose
+
 banner output
 mkdir -p /work/bins
 for t in crucible-downstairs crucible-hammer crutest dsc; do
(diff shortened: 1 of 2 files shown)


.github/buildomat/jobs/build.sh
```diff
@@ -20,6 +20,8 @@ rustc --version
 banner build
 ptime -m cargo build --verbose
 
+ptime -m cargo test test_successful_live_repair --verbose
```
jmpesp, line 23: *remove before flight*

upstairs/src/dummy_downstairs_tests.rs
```diff
@@ -156,102 +156,123 @@ pub(crate) mod protocol_test {
...
-                    version,
-                    upstairs_id: _,
-                    session_id: _,
-                    gen: _,
-                    read_only: _,
-                    encrypted: _,
-                    alternate_versions: _,
-                } => {
-                    info!(self.inner.log, "negotiate packet {:?}", packet);
+            loop {
```
jmpesp, line 159: currently, a ping can't happen before `HereIAm`, it only happens in the select loop after that packet has been sent.

upstairs/src/dummy_downstairs_tests.rs
```diff
@@ -156,102 +156,123 @@ pub(crate) mod protocol_test {
...
-                .unwrap()
-                .unwrap();
-            match &packet {
-                Message::PromoteToActive {
-                    upstairs_id,
-                    session_id,
-                    gen,
-                } => {
-                    assert!(*gen == 1);
+            loop {
```
jmpesp, line 201: There should be a comment explaining the ping behaviour up here maybe, instead of in the match?

jmpesp (approved):
Sorry about the test fails!

upstairs/src/dummy_downstairs_tests.rs
leftwo: Done

upstairs/src/dummy_downstairs_tests.rs
leftwo: Done

leftwo:
This makes it better (found some issues) but there is still some out there that prevents it from working
every time.  I'm going to merge this change but we should continue to iterate on whatever is ailing this test.

PR #799, opened 2023-06-14
Update values for flow control and gone_too_long

leftwo:
New settings for the constants that control how much work we send to a downstairs, and how many IOs we build up before we disconnect from a downstairs.
`IO_OUTSTANDING_MAX: usize = 5000;`
`MAX_ACTIVE_COUNT: usize = 600;`

Moved the max active count to a common place so the downstairs could use it.

Update a few places that just had hard coded message queue depths that would not work with the new constants.

Other things not related to the above changes.
Updated in_progress() to not assume all job IDs it receives will be valid, and return None if it encounters an ID that is no longer valid.

Added per downstairs counters for when we enabled flow control. 
Put the live repair counters with the rest of the DTrace counters.

Fixed and renamed the fast_fill option, It looping over wrong value.

I tested this a bunch on an actual gimlet, but no propolis/omicron in the loop.
I used a disk for each region:

```
/target/release/dsc start --create --cleanup --extent-size 16384 --extent-count 1600 --block-size=4096 \
  --region-dir /oxp_0ca797a6-f467-4296-bc27-e7590c8330c2/dsc/region/  \
  --region-dir /oxp_627cda87-085b-44af-a70e-d067599c3fe2/dsc/region/ \
  --region-dir /oxp_9f5a50c7-08ce-41f7-8efd-5d1323a1f070/dsc/region/ \
  --region-dir /oxp_d2e09ee0-b7ff-479c-ab50-0cdb5ec76503/dsc/region --region-count 4
```

With the above changes, the upstairs can now send `--io-depth 64` workloads using both `crutest` and
`measure-iops` and still keep up.  Previous runs would pretty quickly hit flow-control and then soon
after, send a downstairs to faulted.

Here is some example dtrace output showing the work queues for each downstairs taking
jobs and moving them through.  Each line is at a 1 second interval:
```
DS STATE 0        DS STATE 1        DS STATE 2   UPW  DSW  NEW0 NEW1 NEW2   IP0  IP1  IP2    D0   D1   D2    S0   S1   S2  E0 E1 E2
    active            active            active   100  256     0    0    0   256  100   99     0  156  157     0    0    0   0  0  0
    active            active            active    36  128     0    0    0    11   92   36   117   36   92     0    0    0   0  0  0
    active            active            active    45 1025     0    0    0   257   45   44   768  980  981     0    0    0   0  0  0
    active            active            active   108  256     0   43    0   128   65   99   128  148  157     0    0    0   0  0  0
    active            active            active    12  256     0    0    0     8   12   47   248  244  209     0    0    0   0  0  0
    active            active            active    54  186     0   10    8   115   44   46    71  132  132     0    0    0   0  0  0
    active            active            active   129 1021     0    0    0     0  129  129  1021  892  892     0    0    0   0  0  0
    active            active            active     1 1153     0    0    0   129    1    0  1024 1152 1153     0    0    0   0  0  0
    active            active            active   112  256    47    0    0    65  114  104   144  142  152     0    0    0   0  0  0
    active            active            active   106  384     0    0    0   106  106  107   278  278  277     0    0    0   0  0  0
    active            active            active    22  256     0    0    0    58   22   16   198  234  240     0    0    0   0  0  0
    active            active            active   129 1025     0    0    0     0  129  129  1025  896  896     0    0    0   0  0  0
    active            active            active   117 1281     0   45   59   257   66   58  1024 1170 1164     0    0    0   0  0  0
    active            active            active     1  213     0    0    0    12    1    1   201  212  212     0    0    0   0  0  0
    active            active            active     3  256     0    0    0    93    3    0   163  253  256     0    0    0   0  0  0
    active            active            active    53  256     0    0    0   151   53   51   105  203  205     0    0    0   0  0  0
    active            active            active   121  128     0    0    0     0  121  128   128    7    0     0    0    0   0  0  0
    active            active            active    94 1153     0    0    0   257   93   94   896 1060 1059     0    0    0   0  0  0
```

The old threshold of 100 outstanding IOs would never stand up to this workload.

I believe we can get further tuning to possibly improve things more, but this is moving in the
right direction.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -24,6 +24,10 @@ pub mod x509;
 
 pub const REPAIR_PORT_OFFSET: u16 = 4000;
 
+// Max number of submitted IOs between the upstairs and the downstairs, above
+// which flow control kicks in.
+pub const MAX_ACTIVE_COUNT: usize = 600;
+
 #[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(
     thiserror::Error,
--- a/crutest/src/main.rs
+++ b/crutest/src/main.rs
@@ -103,7 +103,7 @@ enum Workload {
         /// Before each replacement, do a fill of the region so
         /// the replace will have to copy the entire region..
         #[clap(long, action)]
-        fill: bool,
+        fast_fill: bool,
 
         /// The address:port of a running downstairs for replacement
         #[clap(long, action)]
@@ -607,7 +607,7 @@ async fn main() -> Result<()> {
     // If we are running the replace workload, we need to know the
     // current list of targets the upstairs will be started with.
     let full_target = if let Workload::Replace {
-        fill: _,
+        fast_fill: _,
         replacement,
         work: _,
     } = opt.workload
@@ -882,6 +882,9 @@ async fn main() -> Result<()> {
             read_loops,
             write_loops,
         } => {
+            // Pathetic.  This tiny wait is just so all my output from the
+            // test will be after all the upstairs messages have finised.
+            tokio::time::sleep(tokio::time::Duration::from_secs(2)).await;
             println!("Perf test");
             let ops = {
                 if opt.count == 0 {
@@ -991,7 +994,7 @@ async fn main() -> Result<()> {
                 .await?;
         }
         Workload::Replace {
-            fill,
+            fast_fill,
             replacement: _,
             work,
         } => {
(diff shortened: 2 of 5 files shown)


jmpesp (approved):

leftwo:
I do want to record your (and my) chat discussion where we agree that this is a bandaid and the real
solution here needs to come with job flow control back pressure to slow down incoming work when
we are not quite keeping up and do have healthy downstairs.

PR #800, opened 2023-06-14
debug CI failures

leftwo:
damn you test_successful_live_repair, why can't you behave.

### DIFF
--- a/.github/buildomat/jobs/build-release.sh
+++ b/.github/buildomat/jobs/build-release.sh
@@ -51,6 +51,9 @@ rustc --version
 banner rbuild
 ptime -m cargo build --verbose --release --all-features
 
+banner LRtest
+ptime -m cargo test -p crucible --lib test_successful_live_repair --verbose -- --nocapture
+
 banner rtest
 ptime -m cargo test --verbose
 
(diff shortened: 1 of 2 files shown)


upstairs/src/dummy_downstairs_tests.rs
```diff
@@ -511,7 +511,7 @@ pub(crate) mod protocol_test {
     /// downstairs responds with a read response for a certain job, then more
     /// work is sent.
     #[tokio::test]
-    async fn test_flow_control() {
+    async fn _test_flow_control() {
```
ahl, line 514: What's the leading `_` do? Did you mean to disable this test like `_test_successful_live_repair` below?

leftwo: I did mean to disable this test initially, and confused myself later when I saw it still failing.  

I'm really using this PR as a way to trigger the CI to run (the only place I can get it to fail) 
as I debug the failures.

ahl: Can I be helpful in debugging it? Happy to pitch in if I can!

PR #801, opened 2023-06-14
Issue extent flushes in parallel

jmpesp:
First, move the concept of whether an Extent is opened or closed *out* of Extent by removing the Option around Mutex<Inner>:

    -    inner: Option<Mutex<Inner>>,
    +    pub inner: Mutex<Inner>,

This moves up to Region as a new enum:

    -    pub extents: Vec<Extent>,
    +    pub extents: Vec<Arc<Mutex<ExtentState>>>,

where ExtentState uses the variant to store if the extent is opened or closed:

    +pub enum ExtentState {
    +    Opened(Arc<Extent>),
    +    Closed,
    +}

This is all in service of `fn region_flush` issuing a `tokio::spawn` for each `extent.flush` call, and waiting on each join handle that is returned. This is possible when an `Arc<Extent>` is used because a clone can be passed to each task.

Using an enum in this way is also a little safer: `Region::close_extent` actually consumes the old extent while closing it, meaning that the old Extent no longer hangs around. It does however mean that the downstairs will panic if the ExtentState is not expected: both `get_opened_extent` and `close_extent` panic if the extent in question is closed. `close_extent` also panics if you're trying to close an extent where there's an Arc clone somewhere as part of doing some other work.

Note this work required updating to Rust 1.70 (for `Arc::into_inner`). There's a few unrelated changes due to this.

### DIFF
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -32,6 +32,10 @@ pub struct Extent {
     block_size: u64,
     extent_size: Block,
     iov_max: usize,
+
+    /// Inner contains information about the actual extent file that holds the
+    /// data, the metadata (stored in the database) about that extent, and the
+    /// set of dirty blocks that have been written to since last flush.
     pub inner: Mutex<Inner>,
 }
 
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -6,7 +6,8 @@ use std::fs::{rename, File, OpenOptions};
 use std::io::{BufReader, IoSlice, IoSliceMut, Read, Seek, SeekFrom, Write};
 use std::net::SocketAddr;
 use std::path::{Path, PathBuf};
-use tokio::sync::{Mutex, MutexGuard};
+use tokio::sync::Mutex;
+use tokio::task::JoinHandle;
 
 use anyhow::{anyhow, bail, Result};
 use futures::TryStreamExt;
@@ -31,15 +32,7 @@ pub struct Extent {
     block_size: u64,
     extent_size: Block,
     iov_max: usize,
-    /// Inner contains information about the actual extent file that holds
-    /// the data, and the metadata (stored in the database) about that
-    /// extent.
-    ///
-    /// If Some(), it means the extent file and database metadata for
-    /// it are opened.
-    /// If None, it means the extent is currently
-    /// closed (and possibly being updated out of band).
-    inner: Option<Mutex<Inner>>,
+    pub inner: Mutex<Inner>,
 }
 
 /// BlockContext, with the addition of block index and on_disk_hash
@@ -58,6 +51,14 @@ pub struct Inner {
 
     /// Set of blocks that have been written since last flush.
     dirty_blocks: HashSet<usize>,
+}
+
+/// An extent can be Opened or Closed. If Closed, it is probably being updated
+/// out of band. If Opened, then this extent is accepting operations.
+#[derive(Debug)]
+pub enum ExtentState {
+    Opened(Arc<Extent>),
+    Closed,
 }
 
 impl Inner {
@@ -623,11 +624,11 @@ impl Extent {
             block_size: def.block_size(),
             extent_size: def.extent_size(),
             iov_max: Extent::get_iov_max()?,
-            inner: Some(Mutex::new(Inner {
+            inner: Mutex::new(Inner {
                 file,
                 metadb,
                 dirty_blocks: HashSet::new(),
-            })),
+            }),
         };
 
         // Clean out any irrelevant block contexts, which may be present if
@@ -639,16 +640,21 @@ impl Extent {
         Ok(extent)
     }
 
+    #[cfg(test)]
+    pub async fn dirty(&self) -> bool {
+        self.inner.lock().await.dirty().unwrap()
+    }
+
     /**
      * Close an extent and the metadata db files for it.
      */
-    pub async fn close(&mut self) -> Result<(u64, u64, bool), CrucibleError> {
-        let inner = self.inner.as_ref().unwrap().lock().await;
+    pub async fn close(self) -> Result<(u64, u64, bool), CrucibleError> {
+        let inner = self.inner.lock().await;
+
         let gen = inner.gen_number().unwrap();
         let flush = inner.flush_number().unwrap();
         let dirty = inner.dirty().unwrap();
-        drop(inner);
-        self.inner = None;
+
         Ok((gen, flush, dirty))
     }
 
@@ -812,11 +818,11 @@ impl Extent {
             block_size: def.block_size(),
             extent_size: def.extent_size(),
             iov_max: Extent::get_iov_max()?,
-            inner: Some(Mutex::new(Inner {
+            inner: Mutex::new(Inner {
                 file,
                 metadb,
                 dirty_blocks: HashSet::new(),
-            })),
+            }),
         })
     }
 
@@ -824,10 +830,10 @@ impl Extent {
      * Create the copy directory for this extent.
      */
     fn create_copy_dir<P: AsRef<Path>>(
-        &self,
         dir: P,
+        eid: usize,
     ) -> Result<PathBuf, CrucibleError> {
-        let cp = copy_dir(dir, self.number);
+        let cp = copy_dir(dir, eid as u32);
 
         /*
          * Verify the copy directory does not exist
@@ -845,12 +851,12 @@ impl Extent {
      * remote downstairs.
      */
     fn create_copy_file(
-        &self,
         mut copy_dir: PathBuf,
+        eid: usize,
         extension: Option<ExtentType>,
     ) -> Result<File> {
         // Get the base extent name before we consider the actual Type
-        let name = extent_file_name(self.number, ExtentType::Data);
+        let name = extent_file_name(eid as u32, ExtentType::Data);
         copy_dir.push(name);
         if let Some(extension) = extension {
             let ext = format!("{}", extension);
@@ -870,10 +876,6 @@ impl Extent {
         Ok(file)
     }
 
-    pub async fn inner(&self) -> MutexGuard<'_, Inner> {
-        self.inner.as_ref().unwrap().lock().await
-    }
-
     pub fn number(&self) -> u32 {
         self.number
     }
@@ -891,7 +893,8 @@ impl Extent {
         cdt::extent__read__start!(|| {
             (job_id, self.number, requests.len() as u64)
         });
-        let inner = self.inner().await;
+
+        let inner = self.inner.lock().await;
 
         // This code batches up operations for contiguous regions of
         // ReadRequests, so we can perform larger read syscalls and sqlite
@@ -1042,12 +1045,12 @@ impl Extent {
             (job_id, self.number, writes.len() as u64)
         });
 
-        let mut inner = self.inner().await;
+        let mut inner_guard = self.inner.lock().await;
         // I realize this looks like some nonsense but what this is doing is
         // borrowing the inner up-front from the MutexGuard, which will allow
         // us to later disjointly borrow fields. Basically, we're helping the
         // borrow-checker do its job.
-        let inner = &mut *inner;
+        let inner = &mut *inner_guard;
 
         for write in writes {
             self.check_input(write.offset, &write.data)?;
@@ -1260,7 +1263,7 @@ impl Extent {
         job_id: u64,
         log: &Logger,
     ) -> Result<(), CrucibleError> {
-        let mut inner = self.inner().await;
+        let mut inner = self.inner.lock().await;
 
         if !inner.dirty()? {
             /*
@@ -1443,7 +1446,7 @@ impl Extent {
         &self,
         force_override_dirty: bool,
     ) -> Result<(), CrucibleError> {
-        let mut inner = self.inner().await;
+        let mut inner = self.inner.lock().await;
 
         if !force_override_dirty && !inner.dirty()? {
             return Ok(());
@@ -1511,7 +1514,7 @@ impl Extent {
 pub struct Region {
     pub dir: PathBuf,
     def: RegionDefinition,
-    pub extents: Vec<Extent>,
+    pub extents: Vec<Arc<Mutex<ExtentState>>>,
     read_only: bool,
     log: Logger,
 }
@@ -1628,6 +1631,15 @@ impl Region {
         self.def.get_encrypted()
     }
 
+    async fn get_opened_extent(&self, eid: usize) -> Arc<Extent> {
+        match &*self.extents[eid].lock().await {
+            ExtentState::Opened(extent) => extent.clone(),
+            ExtentState::Closed => {
+                panic!("attempting to get closed extent {}", eid)
+            }
+        }
+    }
+
     /**
      * If our extent_count is higher than the number of populated entries
      * we have in our extents Vec, then open all the new extent files and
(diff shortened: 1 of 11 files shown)


downstairs/src/region.rs
```diff
@@ -1754,14 +1792,15 @@ impl Region {
...
-        &mut self,
+        &self,
         eid: usize,
         repair_addr: SocketAddr,
     ) -> Result<(), CrucibleError> {
         // Make sure the extent:
         // is currently closed, matches our eid, is not read-only
-        assert!(self.extents[eid].inner.is_none());
-        assert_eq!(self.extents[eid].number, eid as u32);
+        let mg = self.extents[eid].lock().await;
```
leftwo, line 1801: What happens if we are repairing extents and another downstairs takes over?

I'm not sure if have any tests for this, but.  hmm.  What would we need to to to
create a test of this sort of situation to be sure we remain crash consistent here?
I just wonder about the lock and preventing two upstairs connections from making
a mess of things.

downstairs/src/repair.rs
```diff
@@ -77,14 +77,7 @@ pub async fn repair_main(
...
 
-    tokio::spawn(async move {
-        /*
-         * Wait for the server to stop.  Note that there's not any code to
-         * shut down this server, so we should never get past this
-         * point.
-         */
-        server.await
-    });
+    tokio::spawn(server);
```
leftwo, line 80: I think we will be back here shortly for all the cancel safe stuff.

integration_tests/src/lib.rs
```diff
@@ -2589,7 +2589,6 @@ mod test {
     async fn integration_test_guest_replace_many_downstairs() -> Result<()> {
         // Test using the guest layer to verify we can replace one
         // downstairs, but not another while the replace is active.
-        const BLOCK_SIZE: usize = 512;
```
leftwo, line 2592: These are no longer needed?  Did new rust clippy find this?  Not sure how otherwise we would have been leaving them in.

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.66"
+channel = "1.70"
```
leftwo, line 2: If we bump this, do we need it bumped anywhere else in Omicron/Propolis?
I don't think so, as the downstairs itself is standalone, but not sure about any
other consumers of the upstairs

leftwo (commented):
We should run the `./tools/test_repair.sh` and `./tools/test_restart_repair.sh` tests, and possibly bump their internal loop counters and let them run for longer.

downstairs/src/region.rs
jmpesp: I wrote one in 4955d17 - it seems like the old upstairs that's in the middle of repairing gets deactivated, the new upstairs tries to activate, then repairs the downstairs during activation due to the flush numbers mismatch.

downstairs/src/region.rs
```diff
@@ -31,15 +32,7 @@ pub struct Extent {
...
-    /// Inner contains information about the actual extent file that holds
-    /// the data, and the metadata (stored in the database) about that
-    /// extent.
-    ///
-    /// If Some(), it means the extent file and database metadata for
-    /// it are opened.
-    /// If None, it means the extent is currently
-    /// closed (and possibly being updated out of band).
-    inner: Option<Mutex<Inner>>,
+    pub inner: Mutex<Inner>,
```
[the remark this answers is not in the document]
jmpesp: You're right, I added part of that comment back in b3484b0.

The Mutex covers all state changes, yeah, and it's exclusive access for either reads or writes. Not sure if I've answered your question or not...

downstairs/src/repair.rs
jmpesp: :grimacing:

integration_tests/src/lib.rs
jmpesp: Yeah, (what I assume is) the new clippy found this.

rust-toolchain.toml
jmpesp: I'm not sure, leaning towards yes - I can't see how (for example) Propolis would be able to compile this in as a library without support for `Arc::into_inner`. We'll see when that rev gets bumped I guess :)

leftwo: We might want to try to bump propolis and see if it compiles.  If there is some entanglement
and it turns out to be a huge amount of work, we might consider holding this fix till we can
get propolis ready for us.  I'm sure we will have other changes we want in Crucible and I want
to be able to update omicron/propolis frequently over the next few weeks.

downstairs/src/region.rs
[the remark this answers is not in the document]
leftwo: You answered it by not answering it :)
I think it means this mutex only protects the change, and not access to the extent itself.

At some point in the past, I thought we had a lock around the extent, but that does not
appear to be the case (and perhaps never was).

downstairs/src/region.rs
leftwo: That's the behavior I would want, but what I really want is to run that test XX times and make sure
we don't have any race conditions or other cases that we are not seeing here.

I'm just not sure we have protected ourselves for this case.  Granted it's not likely to come
up, but I just don't know what we do during repair if another upstairs takes over.  Maybe it
is all okay and the takeover upstairs does the right thing by waiting for the other tasks for
the other upstairs (that is now no longer active) to be finished with it's work.

I believe there is some logic in the work queue for the downstairs that check for this
situation and self destructs if it finds it.?

integration_tests/src/lib.rs
```diff
@@ -2422,6 +2426,111 @@ mod test {
...
+                test_downstairs_set.downstairs1_address().await,
+                new_downstairs.address().await,
+            )
+            .await
+            .unwrap();
+
+        assert_eq!(res, ReplaceResult::Started);
+
+        eprintln!("Wait for some repair work to proceed");
+        tokio::time::sleep(tokio::time::Duration::from_secs(4)).await;
```
leftwo, line 2489: I've been pulling out hair tracking down failing CI tests, and the sleep here makes me
nervous.  Is there any way to make this deterministic?  Like checking for some job progress but
not completed repair yet?  On a busy system or a fast system, could this sleep be too little
or too much?

rust-toolchain.toml
faithanalog: i was compiling latest propolis against this commit during my perf tests fwiw

downstairs/src/region.rs
jmpesp: > Maybe it is all okay and the takeover upstairs does the right thing by waiting for the other tasks for the other upstairs (that is now no longer active) to be finished with it's work.

I don't think this is the case: the old Upstairs receives the YouAreNoLongerActive and terminates itself in the middle of the repair. I tested this with 

```diff
diff --git a/upstairs/src/dummy_downstairs_tests.rs b/upstairs/src/dummy_downstairs_tests.rs
index 8962ad3..b58d23f 100644
--- a/upstairs/src/dummy_downstairs_tests.rs
+++ b/upstairs/src/dummy_downstairs_tests.rs
@@ -1069,6 +1069,20 @@ pub(crate) mod protocol_test {
                 }
             };
 
+            if eid == 5 {
+                // Send a YouAreNoLongerActive to nuke the Upstairs
+                harness.ds2.fw
+                    .lock()
+                    .await
+                    .send(Message::YouAreNoLongerActive {
+                        new_upstairs_id: Uuid::new_v4(),
+                        new_session_id: Uuid::new_v4(),
+                        new_gen: 9876,
+                    })
+                    .await
+                    .unwrap();
+            }
+
             // Extent limit is Some(eid), where eid is the current loop
             // iteration. It marks the extent at and below are clear to receive
             // IO. Issue some single extent reads and writes to make sure that
@@ -1081,6 +1095,7 @@ pub(crate) mod protocol_test {
 
             let mut responses = vec![Vec::new(); 3];
 
+            /*
             for io_eid in 0usize..10 {
                 // read
 
@@ -1280,6 +1295,7 @@ pub(crate) mod protocol_test {
                     _ => panic!("saw {:?}", m3),
                 }
             }
+            */
 
             // The repair task then waits for the close responses.
 

```

in the middle of `dummy_downstairs_tests::protocol_test::test_successful_live_repair` and progress stopped:

```
Jun 19 15:22:11.323 INFO received ExtentLiveClose { upstairs_id: c4752b14-c3e0-49a8-89b5-317ed16cd02d, session_id: 29d863af-b9d3-4b97-9a36-bea3c630d997, job_id: 2221, dependencies: [], extent_id: 5 }, downstairs: 1
Jun 19 15:22:11.323 INFO received ExtentLiveFlushClose { upstairs_id: c4752b14-c3e0-49a8-89b5-317ed16cd02d, session_id: 29d863af-b9d3-4b97-9a36-bea3c630d997, job_id: 2221, dependencies: [], extent_id: 5, flush_number: 7, gen_number: 1 }, downstairs: 2
Jun 19 15:22:11.323 INFO received ExtentLiveFlushClose { upstairs_id: c4752b14-c3e0-49a8-89b5-317ed16cd02d, session_id: 29d863af-b9d3-4b97-9a36-bea3c630d997, job_id: 2221, dependencies: [], extent_id: 5, flush_number: 7, gen_number: 1 }, downstairs: 3
Jun 19 15:22:11.366 INFO received ExtentLiveReopen { upstairs_id: c4752b14-c3e0-49a8-89b5-317ed16cd02d, session_id: 29d863af-b9d3-4b97-9a36-bea3c630d997, job_id: 2224, dependencies: [2221, 2222, 2223], extent_id: 5 }, downstairs: 3
Jun 19 15:22:11.366 INFO received ExtentLiveReopen { upstairs_id: c4752b14-c3e0-49a8-89b5-317ed16cd02d, session_id: 29d863af-b9d3-4b97-9a36-bea3c630d997, job_id: 2224, dependencies: [2221, 2222, 2223], extent_id: 5 }, downstairs: 2
Jun 19 15:22:11.366 INFO received ExtentLiveReopen { upstairs_id: c4752b14-c3e0-49a8-89b5-317ed16cd02d, session_id: 29d863af-b9d3-4b97-9a36-bea3c630d997, job_id: 2224, dependencies: [2221, 2222, 2223], extent_id: 5 }, downstairs: 1
Jun 19 15:22:11.367 ERRO [1] c4752b14-c3e0-49a8-89b5-317ed16cd02d (29d863af-b9d3-4b97-9a36-bea3c630d997) cmd_loop saw YouAreNoLongerActive 82a545bb-2df4-4ba4-a62a-c16fb6947f16 adcbfbbb-72f3-426b-af21-c8bb8ae2b729 9876, upstairs: 1
Jun 19 15:22:11.367 INFO [1] c4752b14-c3e0-49a8-89b5-317ed16cd02d (29d863af-b9d3-4b97-9a36-bea3c630d997) LiveRepair Active Active ds_transition to Disabled, upstairs: 1
Jun 19 15:22:11.367 INFO [1] Transition from Active to Disabled, upstairs: 1
Jun 19 15:22:11.367 INFO c4752b14-c3e0-49a8-89b5-317ed16cd02d set inactive, session 29d863af-b9d3-4b97-9a36-bea3c630d997, upstairs: 1
Jun 19 15:22:11.367 ERRO 127.0.0.1:40203: proc: Generation number is too low: saw YouAreNoLongerActive with larger gen 9876 than ours 1, looper: 1, upstairs: 1
Jun 19 15:22:11.367 INFO [1] c4752b14-c3e0-49a8-89b5-317ed16cd02d Gone missing, transition from Disabled to Disconnected, upstairs: 1
Jun 19 15:22:11.367 INFO [1] c4752b14-c3e0-49a8-89b5-317ed16cd02d connection to 127.0.0.1:40203 closed, looper: 1, upstairs: 1
Jun 19 15:22:11.367 WARN [1] pm_task rx.recv() is None, upstairs: 1
Jun 19 15:22:11.367 INFO [1] 127.0.0.1:40203 task reports connection:false, upstairs: 1
Jun 19 15:22:11.367 INFO c4752b14-c3e0-49a8-89b5-317ed16cd02d LiveRepair Disconnected Active, upstairs: 1
Jun 19 15:22:11.367 INFO [1] 127.0.0.1:40203 task reports offline, upstairs: 1
Jun 19 15:22:12.368 INFO [1] c4752b14-c3e0-49a8-89b5-317ed16cd02d looper connected, looper: 1, upstairs: 1
Jun 19 15:22:12.368 INFO [1] Proc runs for 127.0.0.1:40203 in state Disconnected, upstairs: 1
Jun 19 15:23:02.369 ERRO 127.0.0.1:40203: proc: timed out during negotiation, looper: 1, upstairs: 1
Jun 19 15:23:02.369 INFO [1] c4752b14-c3e0-49a8-89b5-317ed16cd02d Gone missing, transition from Disconnected to Disconnected, upstairs: 1
Jun 19 15:23:02.369 INFO [1] c4752b14-c3e0-49a8-89b5-317ed16cd02d connection to 127.0.0.1:40203 closed, looper: 1, upstairs: 1
Jun 19 15:23:02.369 INFO [1] 127.0.0.1:40203 task reports connection:false, upstairs: 1
Jun 19 15:23:02.369 INFO c4752b14-c3e0-49a8-89b5-317ed16cd02d LiveRepair Disconnected Active, upstairs: 1
Jun 19 15:23:02.369 INFO [1] 127.0.0.1:40203 task reports offline, upstairs: 1
Jun 19 15:23:03.370 INFO [1] c4752b14-c3e0-49a8-89b5-317ed16cd02d looper connected, looper: 1, upstairs: 1
Jun 19 15:23:03.370 INFO [1] Proc runs for 127.0.0.1:40203 in state Disconnected, upstairs: 1
test dummy_downstairs_tests::protocol_test::test_successful_live_repair has been running for over 60 seconds
```

> I believe there is some logic in the work queue for the downstairs that check for this situation and self destructs if it finds it.?

You mean this?

https://github.com/oxidecomputer/crucible/blob/18d2fa49d716f62742d42fa9140cccbd0d5b894d/downstairs/src/lib.rs#L1988-L2001

integration_tests/src/lib.rs
jmpesp: Good point, I changed this to check for the `StartedAlready` and panic if the repair completed in 9994e79

leftwo:
I ran 
`tools/test_reconnect.sh` <- this test is broken now, and I'll be turning it off soon.
`tools/test_restart_repair.sh`
`tools/test_repaly.sh` (new, coming soon replacement for test_reconnect)
`tools/test_repair.sh`
`tools/test_live_repair.sh`
`tools/hammer_loop.sh`

All passed.  I upped the loop count for each to get some extended runtime.

downstairs/src/region.rs
```diff
@@ -2180,12 +2236,27 @@ impl Region {
...
+                extent.flush(flush_number, gen_number, job_id, &log).await
+            });
+            join_handles.push(jh);
         }
+
+        // Wait for all flushes to finish
+        for join_handle in join_handles {
+            join_handle
+                .await
+                .map_err(|e| CrucibleError::GenericError(e.to_string()))??;
```
leftwo, line 2257: If we error here, where does this bubble out to?

leftwo (approved):
Just that one question

downstairs/src/region.rs
jmpesp: if any of the flushes fails, then it will get returned as a failure here for `region_flush`. Though, see 53c45fbef40f410eabe218d9df2502a840e8e6f2 for the fix for an issue you brought up during a chat: we should wait for all results before returning, instead of (potentially) cancelling these flush tasks in the middle of execution if one of them errors.

PR #802, opened 2023-06-14
Perform relative check in gone_too_long

jmpesp:
Instead of an absolute check, perform a relative check in `gone_too_long`: look for a downstairs that is *relatively* slower than the others. This function should tolerate a **lot** of work being dispatched to the downstairs all at once.


jmpesp:
> If all three downstairs are offline, we won't ever kick them out as they will all have the same amount of outstanding work, right? Would this lead to the upstairs holding incoming IO forever and potentially growing without bounds?

Yes it would. Hmm...

leftwo:
Gone too long is now gone, so we probably don't need a relative check :)

PR #804, opened 2023-06-15
Make consume_req cancel safe

jmpesp:
John helpfully audited our Crucible code for cancel safety of futures used in tokio::select! arms, and found that consume_req was not safe to cancel. This commit makes consume_req cancel safe by splitting out the work into a non-async function: either consume_req is cancelled trying to acquire all the necessary locks, or it runs to completion.

There were a few async functions that were not required to be async, so this commit also changes those.

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -8605,7 +8605,7 @@ impl BlockOp {
      * limits.  Though, if too many volumes are created with scrubbers
      * running, we may have to revisit that.
      */
-    pub async fn iops(&self, iop_sz: usize) -> Option<usize> {
+    pub fn iops(&self, iop_sz: usize) -> Option<usize> {
         match self {
             BlockOp::Read { offset: _, data } => {
                 Some(ceiling_div!(data.len(), iop_sz))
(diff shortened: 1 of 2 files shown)


upstairs/src/lib.rs
```diff
@@ -9097,49 +9115,39 @@ impl Guest {
         let mut bw_check_ok = true;
         let mut iop_check_ok = true;
 
-        // XXX if recv ever is called from multiple threads, token locks must be
-        // taken for the whole of the procedure, not multiple times in the below
-        // if blocks!
```
jgallagher, line 9102: This PR looks correct from an async-cancel-safety perspective. I'm less sure about the semantic (or performance, maybe?) change affected by acquiring all three locks up front, where the code previously only acquired the later two locks if and when needed (and kept them locked for a shorter duration). Will defer to you and Alan.

jmpesp: Looking back, I don't think the previous code was correct if consume_req was called from multiple tasks / threads (it's currently not, hence removing that comment). There definitely seems to have been a TOCTOU problem.

leftwo (approved):
This looks good.
If there is an perf impact to the locking, its probably not in a place that matters
as we are attempting to throttle IOs back to the guest.

At any rate, correctness trumps performance :)

PR #805, opened 2023-06-15
Debug CI PR

leftwo:
This is just my "debug CI failures" PR.

I throw various logging and debug things in here and see how it does on buildomat to see if I can figure out
why `cargo test` fails... sometimes...


leftwo:
This PR has served its purpose, and things taken from here have already gone back to main.

PR #808, opened 2023-06-16
Update to latest dropshot

leftwo:
Picks up the change to add a default_handler_task_mode to detach spawned tasks and avoid cancellation woes.

Copied directly from John and omicron #3362


leftwo:
Inspired by: https://github.com/oxidecomputer/omicron/pull/3362

PR #812, opened 2023-06-19
Widen assert values to u128 to deal with u64::MAX

jmpesp:
Cast to u128 in extent_from_offset when performing the "last block is also within the region" assert, as there was a proptest regression where first_block + n_blocks == u64::MAX.

Also committing upstairs/proptest-regressions/impacted_blocks.txt like it suggests :)

### DIFF
--- a/upstairs/proptest-regressions/impacted_blocks.txt
+++ b/upstairs/proptest-regressions/impacted_blocks.txt
@@ -1,0 +1,7 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
+# It is recommended to check this file in to source control so that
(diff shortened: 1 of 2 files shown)


faithanalog:
I'm ok with this if it doesn't impact performance, but it's possible that it would be better to change the test (which I can do). if first_block + n_blocks == u64::MAX then that means we're dealing with a region that is over 18000 petabytes in size. which uh... yeah good luck with that ever actually happening

faithanalog (approved):
i should say i dont really expect it would impact perf meaningfully because afaik ImpactedBlocks isnt really used in a way where a few integer ops will matter right now. so if you want to just merge this now, seems fine to me. not really worth the mental thinking to deliberate on it when changing it again wouldnt be hard anyway

PR #814, opened 2023-06-20
Change size_to_validate from usize to u64

jmpesp:
Enforce that size_to_validate should be 64 bits wide, to allow validating disks that are larger than 2**32 bytes.


jmpesp:
Note I don't know if this was a problem, given that usize is enough to store 64 bits on our systems... but it's something to ensure.

leftwo (approved):

PR #817, opened 2023-06-23
Retry jobs until they succeed

### LINKED ISSUE #816: The Downstairs should not consider jobs that errored as completed
https://github.com/oxidecomputer/crucible/pull/694#issuecomment-1526021298 showed a scenario where a downstairs running with `--write-errors` could (with a few contrived sleeps) cause a read hash mismatch.

It's important that the Downstairs should **not** consider jobs that errored as completed, and should not notify the Upstairs that is has completed a job until that job successfully completes. The Downstairs should keep retrying until it is successful. If it is `gone_too_long` as determined by the Upstairs, then live repair will kick in, but importantly the flush numbers will not be inaccurate.

jmpesp:
The Downstairs should not consider jobs that errored to be completed, and should not notify the Upstairs until a job has completed successfully. It should retry until that job succeeds.

Fixes #816


protocol/src/lib.rs
```diff
@@ -485,6 +485,57 @@ pub enum Message {
...
+            Message::ExtentLiveRepairAckId { result, .. } => result.is_err(),
+            Message::ExtentLiveAckId { result, .. } => result.is_err(),
+            Message::WriteAck { result, .. } => result.is_err(),
+            Message::FlushAck { result, .. } => result.is_err(),
+            Message::ReadResponse { responses, .. } => responses.is_err(),
+            Message::WriteUnwrittenAck { result, .. } => result.is_err(),
+        }
+    }
+}
+
```
leftwo, line 538: The comment below this I think needs to go back up and be before the `impl`, or even moved above it.

downstairs/src/lib.rs
```diff
@@ -1039,20 +1039,35 @@ where
...
-                    fw.send(&m).await?;
-                    drop(fw);
-
-                    ads.lock()
-                        .await
-                        .complete_work(upstairs_connection, job_id, m)
-                        .await?;
-                    cdt::work__done!(|| job_id);
+
+                    if m.err() {
```
leftwo, line 1043: If we reset to new here, it won't retry right away, but put the job back on
the list as "new", much like what a lossy would do sometimes, right?
And it will retry it, after having walked the entire list in `new_work`?

If we have just one job, and that one job fails, I don't think we will go back through
the loop, right?  Because I believe we need to see `job_channel_rx.recv().await.is_some()`
to take another lap through the work queue.

leftwo (commented):
We should have some tests for this condition as well.
Like, have a job fail and verify that it moves back to new and is
retried, while other jobs that are not dependent are allowed forward, and
jobs that are dependent are prevented from moving forward.

protocol/src/lib.rs
jmpesp: done in d487192

downstairs/src/lib.rs
jmpesp: I changed the logic to loop instead of setting the state back to New (0970a29b90ea7723d8eec169bc32134367bac17c) This should retry jobs until they pass without needing another `job_channel_rx.recv().await.is_some()`

upstairs/src/lib.rs
```diff
@@ -5143,8 +5157,8 @@ impl Upstairs {
         let ds_state = self.ds_state_copy().await;
         let ds = self.downstairs.lock().await;
         let ds_io_count = ds.io_state_count;
-        let ds_reconciled = ds.reconcile_repaired.clone();
-        let ds_reconcile_needed = ds.reconcile_repair_needed.clone();
+        let ds_reconciled = ds.reconcile_repaired;
```
leftwo, line 5160: Thanks for fixing

leftwo (approved):

downstairs/src/lib.rs
```diff
@@ -2763,12 +2798,14 @@ impl Work {
...
-     * return the job ID and the upstairs UUID, moving the state of the
-     * job as InProgress.
+     * return the job ID and the upstairs UUID, moving the state of the job as
+     * InProgress. If the dependencies are not met, move the state to DepWait.
      *
-     * If this job is not new, then just return none.  This can be okay as
-     * we build or work list with the new_work fn above, but we drop and
-     * re-aquire the Work mutex and things can change.
+     * If this job is not new, then just return none. This can be okay as we
+     * build or work list with the new_work fn above, but we drop and re-aquire
```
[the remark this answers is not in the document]
jmpesp: bef2b39

upstairs/src/lib.rs
jmpesp: :+1:

Issue #759, opened 2023-05-30
skipped IO should not count toward success of Write/Flush


leftwo:
If an IO is skipped, we should not count that toward success with a Write/Flush

Right now, under failure situations, a single successful write/flush with two other `Skipped` will
result in success for the IO back to the guest.  While a write may have the data they want, the
result should not be success, as we can only guarantee that one downstairs has or has flushed
the data.

PR #819, opened 2023-06-27
Treat a skipped IO like an error IO for ACK results.

leftwo:
When an IO is skipped, treat that like an error when we are deciding what response to send back to the guest.  Success on one IO alone is not enough on a write or flush, we need at least two before we can pass that success back to the guest.

New tests added to cover this case as well.

This fixes [this issue](https://github.com/oxidecomputer/crucible/issues/759)

### DIFF
--- a/upstairs/src/test.rs
+++ b/upstairs/src/test.rs
@@ -1710,7 +1710,7 @@ pub(crate) mod up_test {
     async fn write_single_skip() {
         // Verfy that a write with one skipped job will still result in
         // success being sent back to the guest.
-        w_io_single_skip(true).await;
+        w_io_single_skip(false).await;
     }
 
     async fn w_io_single_skip(is_write_unwritten: bool) {
@@ -1804,7 +1804,7 @@ pub(crate) mod up_test {
     #[tokio::test]
     async fn write_double_skip() {
         // Verify that write IO errors are counted.
-        w_io_double_skip(true).await;
+        w_io_double_skip(false).await;
     }
 
     async fn w_io_double_skip(is_write_unwritten: bool) {
@@ -1887,7 +1887,7 @@ pub(crate) mod up_test {
     async fn write_fail_and_skip() {
         // Verify that a write error + a skip results in error back to the
         // guest
-        w_io_fail_and_skip(true).await;
+        w_io_fail_and_skip(false).await;
     }
 
     async fn w_io_fail_and_skip(is_write_unwritten: bool) {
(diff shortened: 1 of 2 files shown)


upstairs/src/test.rs
```diff
@@ -1700,6 +1700,529 @@ pub(crate) mod up_test {
...
+    async fn write_unwritten_single_skip() {
+        // Verfy that a write_unwritten with one skipped job will still
+        // result in success sent back to the guest.
+        w_io_single_skip(true).await;
+    }
+    #[tokio::test]
+    async fn write_single_skip() {
+        // Verfy that a write with one skipped job will still result in
+        // success being sent back to the guest.
+        w_io_single_skip(true).await;
```
jmpesp, line 1713: s/true/false/

upstairs/src/test.rs
```diff
@@ -1700,6 +1700,529 @@ pub(crate) mod up_test {
...
+
+    #[tokio::test]
+    async fn write_unwritten_double_skip() {
+        // Verify that write IO errors are counted.
+        w_io_double_skip(true).await;
+    }
+    #[tokio::test]
+    async fn write_double_skip() {
+        // Verify that write IO errors are counted.
+        w_io_double_skip(true).await;
```
jmpesp, line 1807: s/true/false/

upstairs/src/test.rs
```diff
@@ -1700,6 +1700,529 @@ pub(crate) mod up_test {
...
+    async fn write_unwritten_fail_and_skip() {
+        // Verify that an write_unwritten error + a skip results in error
+        // back to the guest
+        w_io_fail_and_skip(true).await;
+    }
+    #[tokio::test]
+    async fn write_fail_and_skip() {
+        // Verify that a write error + a skip results in error back to the
+        // guest
+        w_io_fail_and_skip(true).await;
```
jmpesp, line 1890: s/true/false/

upstairs/src/test.rs
```diff
@@ -1700,6 +1700,529 @@ pub(crate) mod up_test {
...
+            assert_eq!(done.guest_id, gw_id);
+            ds.ack(*ds_id_done);
+
+            assert!(ds.result(*ds_id_done).is_ok());
+        }
+    }
+
+    #[tokio::test]
+    async fn flush_io_double_skip() {
+        // up_ds_listen test, a double skip on a write or write_unwritten
```
jmpesp, line 2066: write -> flush

upstairs/src/test.rs
```diff
@@ -1700,6 +1700,529 @@ pub(crate) mod up_test {
...
+            assert_eq!(done.guest_id, gw_id);
+            ds.ack(*ds_id_done);
+
+            assert!(ds.result(*ds_id_done).is_err());
+        }
+    }
+
+    #[tokio::test]
+    async fn flush_io_fail_and_skip() {
+        // up_ds_listen test, a fail plus a skip on a write or write_unwritten
```
jmpesp, line 2138: write -> flush

jmpesp (approved):
Looks good, some comments:

PR #821, opened 2023-06-29
Remove panics and asserts in dummy tests

jmpesp:
Remove all the panics and asserts in the dummy downstairs protocol tests, and replace them with bails. Hopefully this helps with getting to the bottom of why the CI tests fail.


leftwo:
Lol.. that this passes CI is probably proof enough of it's value.

PR #823, opened 2023-06-30
Reorder select arms so pings can't be starved out

jmpesp:
If the Guest submits a boatload of work, then the current `tokio::select` biased ordering will cause pings to not be sent. If a Downstairs is *not* responding at all (like the dummy downstairs in `test_successful_live_repair`) while this is going on, then the ping and timeout deadlines are not being pushed. This results in the Upstairs kicking out the Downstairs, causing this test to fail.

Reorder the select branches so that this situation can't occur.

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -1903,13 +1903,28 @@ where
         tokio::select! {
             /*
              * We set biased so the loop will:
-             * First make sure the pm task is still running.
-             * Second, get and look at messages received from the downstairs.
-             *   Some messages we can handle right here, but ACKs from
-             *   messages we sent are passed on to the pm task.
              *
-             * By handling messages from the downstairs before sending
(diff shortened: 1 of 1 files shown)


upstairs/src/lib.rs
```diff
@@ -3436,6 +3435,11 @@ impl Downstairs {
             }
         }
 
+        info!(
+            self.log,
+            "[{}] changed {} to fault skipped", client_id, number_jobs_skipped
```
leftwo, line 3440: Maybe: `Changed {} jobs to fault skipped`
Otherwise, I'll think {} is the job id :)

leftwo (approved):

upstairs/src/lib.rs
jmpesp: agreed, a7eaf8f

PR #825, opened 2023-07-04
Add Logger Option to Volume construct method

leftwo:
Added a Option<Logger> to the Volume construct and from there, pass it down to add_subvolume_create_guest().  

up_main already took an optional Logger, so this just plumbs it up into the Volume layer.

Once this goes in, we can update propolis and then have logs that don't clobber each other.


jmpesp (approved):
Sssssssssssssssssssssssssssssship it

Issue #822, opened 2023-06-29
tools/test_reconnect.sh may not be able to finish


leftwo:
the tools/test_reconnect.sh test can get itself stuck by restarting too fast, which can result in
the test never finishing.  If we send enough IO when a downstairs is off, the upstairs will kick it
out and require LiveRepair.  This results in things taking longer before that downstairs is back
to Active.  If another downstairs goes away during this, then that means it will take even longer
before everything is back to normal.

This test was a test for the replay functionality, and in its current form it may not be able to
test that.  

There is now a `crutest replay` option that does what we need, so this test (test_reconnect.sh)
should probably be abandoned and a new test developed.

PR #826, opened 2023-07-06
New replay test for buildomat

leftwo:
New replay test for buildomat. This uses crutest and dsc together
to create a better reconnect test than the previous bash crimes that
were how this was tested before.

Remove and delete test_reconnect.sh from usage, it was broken anyway.
Updated the live repair test, less bash more rust.

Fixed a few DTrace scripts that were broken.

Updated the nightly test on to use the new replay test.
Updated several tests to take a "number of loops" options so regular
callers can have the default (shorter) length tests, and the nightly
test can supply a larger number of loops.


leftwo:
Fixes: https://github.com/oxidecomputer/crucible/issues/822

tools/test_live_repair.sh
```diff
@@ -1,75 +1,71 @@
 #!/bin/bash
 
-# Create and start up the downstairs.
-# In a loop:
-#  Send IO through crutest
-#  Fault a downstairs
-#  Let the upstairs repair.
-
+# Create regions and start up four downstairs.
+# We create four and use the forth downstairs as the replacement
```
jmpesp, line 4: typo: fourth

tools/test_live_repair.sh
```diff
@@ -1,75 +1,71 @@
...
 
 loop_log=/tmp/test_live_repair_summary.log
 test_log=/tmp/test_live_repair.log
-dsc_test_log=/tmp/test_live_repair_dsc.log
-verify_file=/tmp/test_live_verify
 
 ROOT=$(cd "$(dirname "$0")/.." && pwd)
-export BINDIR=${BINDIR:-$ROOT/target/release}
-cds="$BINDIR/crucible-downstairs"
+export BINDIR=${BINDIR:-$ROOT/target/debug}
```
jmpesp, line 27: changing to debug here?

tools/test_replay.sh
```diff
@@ -0,0 +1,114 @@
...
+trap ctrl_c INT
+function ctrl_c() {
+    echo "Stopping at your request"
+    ${dsc} cmd shutdown
+}
+
+test_log=/tmp/test_replay.log
+
+ROOT=$(cd "$(dirname "$0")/.." && pwd)
+export BINDIR=${BINDIR:-$ROOT/target/debug}
```
jmpesp, line 22: debug here?

jmpesp (approved):
LGTM, but: any way to remove the sleeps?

PR #829, opened 2023-07-08
Update to latest dropshot

leftwo:
Get the latest dropshot and watch the panics flee


leftwo:
Ha, now with the dropshot fix, we get no panics at all.  
Let's see if that happens twice (re-running the tests now)

PR #830, opened 2023-07-13
Switch to building on heliosv2

citrus-it:
This is a pre-requisite for moving the rack software on top of heliosv2. Once this is integrated, we can point the omicron pins at the CI artefacts built with the newer helios version. It's going to take a few days to get everything over and build machines updated.

### DIFF
--- a/.github/buildomat/jobs/build-release.sh
+++ b/.github/buildomat/jobs/build-release.sh
@@ -2,7 +2,7 @@
 #:
 #: name = "rbuild"
 #: variety = "basic"
-#: target = "helios"
+#: target = "helios-2.0"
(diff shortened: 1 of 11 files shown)


leftwo (approved):
Once this change happens, will we need to update both propolis and omicron with the
new pointers to v2 artifacts.  Both Omicron and Propolis have V1 artifacts for crucible
now and when either of them switch to V2, it should probably include V2 crucible, right?

citrus-it:
Yes. omicron will be last (after 5 pre-requisite PRs are merged) but there's going to have to be a second propolis one to pull the new crucible bits.

PR #832, opened 2023-07-14
Minor clippy cleanup

leftwo:
This is just a shell PR so I can test CI


leftwo:
This seems to perform better with sending test output to a file.

Issue #605, opened 2023-01-12
Determine how much storage to "reserve" for overhead


leftwo:
Figure out how much physical storage crucible needs for overhead to avoid filling up 100% of space on the disk.

askfongjojo:
Per storage huddle, the minimum for FCS is to set a reasonable quota.

leftwo:
I think we can do a first pass at this with both a quota and a reservation when we create
the dataset for a region.

I'm not sure yet how snapshots will factor into this, as a snapshot might require a 2x reservation
while the snapshot exists. 
Also TBD, if we have multiple snapshots, how much do we need to reserve for those?

leftwo:
The fix for this has been merged to main in the crucible repo.
Propolis PR is out,
once that PR is merged, then Omicron will need to get a new pointer to latest crucible.

leftwo:
This has landed in Omicron.

PR #835, opened 2023-07-15
Add quota to agent created datasets

leftwo:
This adds a quota and a reservation to crucible datasets created when a region
is created in the crucible agent.  A somewhat arbitrary .25 overhead is reserved
for each region.  17% is the largest possible space the metadata will take, and this
gives us a little room in the unlikely event that all regions are full.

This does not consider snapshot overhead.
 
For issue #605

### DIFF
--- a/agent/src/main.rs
+++ b/agent/src/main.rs
@@ -14,19 +14,25 @@ use std::sync::Arc;
 const PROG: &str = "crucible-agent";
 const SERVICE: &str = "oxide/crucible/downstairs";
 /*
- * As a safety mechanism to prevent the agent from creating more regions
- * than we have physical space, we using a zfs reservation on each dataset
- * we create.  While not a complete solution, it does help in some
- * situations to avoid allocation of a region that we don't have the
- * space for if that region was to fill up.
- * As for how much we reserve, it's the region size itself, plus the
- * amount we need for metadata (currently around 17%) then an additional
- * 8% that, if things are behaving, we will have for snapshots and
- * to prevent us from using all the data in the pool.
+ * As a safety mechanism to prevent the agent from allocating more space to
+ * regions than we have physical space on disk, we use a zfs reservation on
+ * each dataset we create. While it does prevent us from allocating more space
+ * to regions than the disk could accommodate, it is not a complete solution in
+ * that we ideally would like Nexus to track allocations (for crucible and
+ * other significant datasets), an adequate capacity margin to ensure that disk
+ * performance doesn't become pathological, and monitors any over-use for
+ * appropriate mitigation (migration, termination, etc).
  *
- * In addition, we throw a quota of 3x the region size.  If the region
- * has grown that big, then something is wrong and we should prevent it
- * from growing any larger and impacting other regions in the dataset.
+ * As for how much we reserve, it's the region size itself, plus the amount
+ * we need for metadata (currently around 17%) then an additional 8% that, if
+ * things are behaving, we will have for snapshots and to prevent us from
+ * using all the data in the pool.
+ *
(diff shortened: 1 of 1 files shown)


leftwo:
After a discussion with @ahl , I've updated the comments a bit, and changed the quota to be larger
than the reservation.

agent/src/main.rs
```diff
@@ -13,6 +13,23 @@ use std::sync::Arc;
...
+ * situations to avoid allocation of a region that we don't have the
+ * space for if that region was to fill up.
+ * As for how much we reserve, it's the region size itself, plus the
+ * amount we need for metadata (currently around 17%) then an additional
+ * 8% that, if things are behaving, we will have for snapshots and
+ * to prevent us from using all the data in the pool.
+ *
+ * In addition, we throw a quota of 3x the region size.  If the region
+ * has grown that big, then something is wrong and we should prevent it
+ * from growing any larger and impacting other regions in the dataset.
```
ahl, line 29:
```suggestion
 * In addition, we throw a quota of 3x the region size. The failure modes of exhausting the quota are potentially severe and not (as yet) well-tested so we want some buffer between the reservation and the quota. If the region
 * has grown that big however, something is potentially very wrong and we should prevent it
 * from growing any larger and impacting other regions on the disk.
```

agent/src/main.rs
```diff
@@ -13,6 +13,23 @@ use std::sync::Arc;
...
+ */
+const METADATA_BUFFER: f64 = 1.25;
+const REGION_QUOTA: u64 = 3;
```
ahl, line 32: maybe: `RESERVATION_FACTOR` and `QUOTA_FACTOR`? I leave it to you... just a thought

agent/src/main.rs
```diff
@@ -117,6 +140,63 @@ impl ZFSDataset {
             bail!("zfs create failed! out:{} err:{}", out, err);
         }
 
+        // If requested, set a quota and reservation here.
+        if let Some(reservation) = reservation {
```
ahl, line 144: I'd suggest we might want something like this:

```rust
        let mut cmd = std::process::Command::new("zfs")
            .arg("create");

    if let Some(reservation) = reservation {
        cmd = cmd.arg("-o").arg(format!("reservation={}", reservation));
    }

        cmd = cmd
            .arg(&dataset)
            .output()?;

```

i.e. I think you can create the dataset with the properties preset.

ahl (approved):

agent/src/main.rs
leftwo: Done

agent/src/main.rs
leftwo: Yeah, I like it.  Fewer lines of code and easier cleanup.

agent/src/main.rs
```diff
@@ -13,6 +13,23 @@ use std::sync::Arc;
 
 const PROG: &str = "crucible-agent";
 const SERVICE: &str = "oxide/crucible/downstairs";
+/*
+ * As a safety mechanism to prevent the agent from creating more regions
+ * than we have physical space, we using a zfs reservation on each dataset
```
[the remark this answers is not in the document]
leftwo: Done

agent/src/main.rs
```diff
@@ -13,6 +13,23 @@ use std::sync::Arc;
...
+ * we create.  While not a complete solution, it does help in some
+ * situations to avoid allocation of a region that we don't have the
+ * space for if that region was to fill up.
```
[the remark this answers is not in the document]
leftwo: Done

agent/src/main.rs
leftwo: Done

PR #838, opened 2023-07-19
Wait for ZFS snapshot directory + agent fixes

### LINKED ISSUE #827: Snapshots don't instantly mount in the dataset's `.zfs/snapshot` directory
When creating a snapshot during the finalizing an image upload, the pantry had a requested snapshot:

	18:07:26.653Z INFO crucible-pantry (dropshot): accepted connection
	    local_addr = [fd00:1122:3344:102::5]:17000
	    remote_addr = [fd00:1122:3344:101::4]:35744
	18:07:26.654Z INFO crucible: flush with snap requested
	18:07:26.716Z INFO crucible-pantry (dropshot): request completed
	    local_addr = [fd00:1122:3344:102::5]:17000
	    method = POST
	    remote_addr = [fd00:1122:3344:101::4]:35744
	    req_id = 51bbc8b7-5c84-4319-a9ec-6fbb4208efa4
	    response_code = 204
	    uri = /crucible/pantry/0/volume/c0a6042a-6c60-4e69-872c-951a06778f44/snapshot

The next step of the snapshot create saga is to validate that those snapshots exist, and create running snapshots out of those. However, here's the agent saying that the snapshot doesn't exist:

	18:07:26.888Z INFO crucible-agent (datafile): checking if path "/data/regions/f7d172bf-3c56-4aa5-a0ff-8057c720ca76/.zfs/snapshot" exists
	18:07:26.888Z INFO crucible-agent (datafile): checking path "/data/regions/f7d172bf-3c56-4aa5-a0ff-8057c720ca76/.zfs/snapshot"
	18:07:26.888Z INFO crucible-agent (dropshot): request completed
	    error_message_external = Not Found
	    error_message_internal = region RegionId("f7d172bf-3c56-4aa5-a0ff-8057c720ca76") snapshot "4d8bc95c-a82d-4942-b7ff-d21a7518afa8" not found
	    local_addr = [fd00:1122:3344:102::7]:32345
	    method = GET
	    remote_addr = [fd00:1122:3344:101::4]:53379
	    req_id = 23fcdea4-dfc8-44e5-a115-33d30f2c03c9
	    response_code = 404
	    uri = /crucible/0/regions/f7d172bf-3c56-4aa5-a0ff-8057c720ca76/snapshots/4d8bc95c-a82d-4942-b7ff-d21a7518afa8

Checking later, It does though:

	root@oxz_crucible_oxp_4ac23275-d8fc-e984-b72c-a5d148ffde7c:~# ls /data/regions/f7d172bf-3c56-4aa5-a0ff-8057c720ca76/.zfs/snapshot
4d8bc95c-a82d-4942-b7ff-d21a7518afa8

This caused the snapshot-create saga to fail, and the unwind fails too because it can't clean up a snapshot that isn't there. It seems like 172ms isn't long enough to wait for a snapshot to be mounted in the dataset's directory.

jmpesp:
When `zfs snapshot dataset@name` is called, the ZFS snapshot is not instantaneously mounted to the appropriate .../.zfs/snapshot/<name> path (as long as 172 ms was measured for a zpool with file backed vdevs). There's two fixes that result from this:

1. Use `zfs list -t snapshot ...` to tell if a snapshot exists, not the presence of the mounted snapshot directory

2. Wait for the ZFS snapshot directory to get mounted before starting a read-only downstairs service that points to it. Wait a maximum of 5 seconds before returning an error.

Also bundled in this commit is a fix for an agent bug: it's possible for a snapshot to have a running read-only downstairs that is in the destroyed state (but not deleted yet), and this will prevent the deletion of said snapshot, as the deletion code simply checked for the presence of a running read-only snapshot and not its state. Fix this by matching on the state and allowing deletion.

Also (!) bundled in this commit is another fix for another agent bug: when getting snapshots for a region, if the region wasn't in state Created then the endpoint would return 500. This is wrong, and we can handle more states than !Created. The only state that should return 500 is Failed.

Fixes #827.

### DIFF
--- a/agent/src/datafile.rs
+++ b/agent/src/datafile.rs
@@ -795,13 +795,23 @@ impl DataFile {
             .arg(dataset.dataset())
             .output()?;
 
+        let snapshots_stdout = String::from_utf8_lossy(&cmd.stdout);
         if !cmd.status.success() {
+            let snapshots_stderr = String::from_utf8_lossy(&cmd.stderr);
+
+            error!(
+                self.log,
+                "zfs list snapshot for dataset {:?} list failed: out {:?} err {:?}",
+                dataset.dataset(),
+                snapshots_stdout,
+                snapshots_stderr,
+            );
+
             bail!("zfs list snapshots failed!");
         }
 
         let mut results = Vec::new();
 
-        let snapshots_stdout = String::from_utf8_lossy(&cmd.stdout);
         let snapshots_list: Vec<&str> = snapshots_stdout
             .trim_end()
             .split('\n')
@@ -838,6 +848,20 @@ impl DataFile {
 
                 cmd_stdout
             };
+
+            if !cmd.status.success() {
+                let err = String::from_utf8_lossy(&cmd.stderr);
+
+                error!(
+                    self.log,
+                    "zfs get for snapshot {} failed: out {:?} err {:?}",
+                    snapshot,
+                    cmd_stdout,
+                    err,
+                );
+
+                bail!("zfs get failed!");
+            }
 
             results.push(Snapshot {
                 name: snapshot_name.to_string(),
--- a/agent/src/datafile.rs
+++ b/agent/src/datafile.rs
@@ -827,7 +827,10 @@ impl DataFile {
             let parts: Vec<&str> = snapshot.split('@').collect();
             info!(self.log, "parts is {:?}", &parts);
             if parts.len() != 2 {
-                panic!("bad snapshot name");
+                // If some non-snapshot-name output snuck in here, then don't
+                // panic! Continue to the next one in the list.
+                error!(self.log, "bad snapshot name {}!", &snapshot);
+                continue;
             }
             let snapshot_name = parts[1];
 
--- a/agent/src/datafile.rs
+++ b/agent/src/datafile.rs
@@ -251,6 +251,39 @@ impl DataFile {
         }
 
         /*
+         * Wait for ZFS snapshot directory to get mounted before
+         * starting a read-only downstairs that points to it. Note
+         * `create_running_snapshot_request` is only entered if the
+         * snapshot exists for the region so this should eventually
+         * be a directory.
+         */
+        {
+            let mut snapshot_path = self.base_path.to_path_buf();
+            snapshot_path.push("regions");
+            snapshot_path.push(request.id.0.clone());
+            snapshot_path.push(".zfs");
+            snapshot_path.push("snapshot");
+            snapshot_path.push(request.name.clone());
+
+            // Wait a maximum of 5 seconds for the
+            // <region>/.zfs/snapshot/<snapshot> directory to appear
+            let mut appeared = false;
+            for _ in 0..50 {
+                if snapshot_path.is_dir() {
+                    appeared = true;
+                    break;
+                }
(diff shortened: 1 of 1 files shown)


agent/src/datafile.rs
```diff
@@ -341,17 +374,26 @@ impl DataFile {
...
+                match running_snapshot.state {
+                    State::Requested | State::Created | State::Tombstoned => {
+                        bail!(
+                            "read-only downstairs running for region {} snapshot {}",
+                            request.id.0,
+                            request.name
+                        );
+                    }
+
+                    State::Destroyed | State::Failed => {
```
leftwo, line 393: Which state here are we looking at?  The state of a downstairs somewhere?

If the state is failed, could it be failed and then come back later and not be failed and then want this snapshot?

agent/src/datafile.rs
```diff
@@ -714,106 +756,93 @@ impl DataFile {
...
+            .arg("-o")
+            .arg("name")
+            .arg("-r")
+            .arg(dataset.dataset())
+            .output()?;
 
-                    // Remove newline
-                    let cmd_stdout = cmd_stdout.trim_end().to_string();
+        if !cmd.status.success() {
+            bail!("zfs list snapshots failed!");
```
leftwo, line 799: Do we want to print out the error here if we saw one?

agent/src/datafile.rs
```diff
@@ -714,106 +756,93 @@ impl DataFile {
...
+        for snapshot in snapshots_list {
+            if snapshot == dataset.dataset() {
+                info!(self.log, "skipping {} matches dataset name", snapshot);
+                continue;
+            }
+            info!(self.log, "snapshot is {}", &snapshot);
+            let parts: Vec<&str> = snapshot.split('@').collect();
+            info!(self.log, "parts is {:?}", &parts);
+            if parts.len() != 2 {
+                panic!("bad snapshot name");
```
leftwo, line 820: When would we hit this panic?   I worry that some output from the zfs command we don't expect would cause us to panic here, and that would be sad :(

agent/src/datafile.rs
```diff
@@ -714,106 +756,93 @@ impl DataFile {
...
+
+            let cmd = Command::new("zfs")
+                .arg("get")
+                .arg("-pH")
+                .arg("-o")
+                .arg("value")
+                .arg("creation")
+                .arg(snapshot)
+                .output()?;
+
```
leftwo, line 832: Do we need a `if !cmd.status.success() {` check here?  Or if we fail, does cmd_stdout just end up
with nothing in it?

leftwo (commented):
Some questions for you

agent/src/datafile.rs
jmpesp: This is the stored state for a "running snapshot" (aka read only downstairs booted at the snapshot's mount dir: I like and don't like the "running snapshot" name but here we are).

As far as I can tell, there's no path in the `state` state machine to transition out of `Failed` for any of the resources that the agent can interact with.

agent/src/datafile.rs
jmpesp: yes definitely, done in dfb1358

agent/src/datafile.rs
jmpesp: Error check added in dfb1358

Issue #837, opened 2023-07-18
crucible upstairs stuck in live repair loop resulted in continuous I/O errors in guest


jordanhendricks:
On the dogfood cluster in rack2, some ubuntu 22.04 instances started complaining about I/O errors immediately after booting and never stopped. They generally were so sad that one could not log in on the serial console, either.

Some sample errors from such a guest:
```
42083.744118] EXT4-fs error (device nvme0n1p1): __ext4_find_entry:1663: inode #80932: comm cron: reading directory lblock 0
[42143.743759] EXT4-fs error (device nvme0n1p1): __ext4_get_inode_loc_noinmem:4433: inode #33445: block 2379: comm cron: unable to read itable block
[42143.746472] EXT4-fs error (device nvme0n1p1): __ext4_find_entry:1663: inode #80932: comm cron: reading directory lblock 0
[42203.746123] EXT4-fs error (device nvme0n1p1): __ext4_get_inode_loc_noinmem:4433: inode #33445: block 2379: comm cron: unable to read itable block
```

I drilled into one such instance and saw that one of the downstairs came and went a couple of times:
```
[2023-07-16T22:49:57.858213302Z]  INFO: propolis-server/25180 on oxz_propolis-server_da88ee76-b2a5-4c9c-840c-97130ae73896: Creating storage device sysbench-mysql-1 of kind Nvme
[...]

PR #840, opened 2023-07-21
Do not eat ErrorReport messages related to repairs

jmpesp:
If an ErrorReport is received for a live repair, this should cause the Upstairs to kick out the associated Downstairs, and attempt to reconnect. The Downstairs that sends such an ErrorReport will abort its tasks and wait for the Upstairs to reconnect, so if the Upstairs does not kick out the associated Downstairs, then it will get stuck in a repair task waiting for the repair ack response, and no IO can flow due to how repair related dependencies work.

### DIFF
--- a/upstairs/src/dummy_downstairs_tests.rs
+++ b/upstairs/src/dummy_downstairs_tests.rs
@@ -12,6 +12,7 @@ pub(crate) mod protocol_test {
     use crate::BlockContext;
     use crate::BlockIO;
     use crate::Buffer;
+    use crate::CrucibleError;
     use crate::Guest;
     use crate::IO_OUTSTANDING_MAX;
     use crate::MAX_ACTIVE_COUNT;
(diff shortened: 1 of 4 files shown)


jmpesp:
I believe this fixes https://github.com/oxidecomputer/crucible/issues/837 but I haven't tested it yet

upstairs/src/dummy_downstairs_tests.rs
```diff
@@ -1880,4 +1881,710 @@ pub(crate) mod protocol_test {
...
+                        .read(Block::new_512(0), buffer)
+                        .await
+                        .unwrap();
+                });
+            }
+
+            if i < MAX_ACTIVE_COUNT {
+                // Before flow control kicks in, assert we're seeing the read
+                // requests
+                bail_assert!(matches!(
```
leftwo, line 1925: Do we need to ignore ping messages too?

upstairs/src/dummy_downstairs_tests.rs
```diff
@@ -1880,4 +1881,710 @@ pub(crate) mod protocol_test {
...
+                    ds1_messages.recv().await.unwrap(),
+                    Message::ReadRequest { .. },
+                ));
+            } else {
+                // After flow control kicks in, we shouldn't see any more
+                // messages
+                match ds1_messages.try_recv() {
+                    Err(TryRecvError::Empty) => {}
+                    Err(TryRecvError::Disconnected) => {}
+                    x => {
```
leftwo, line 1935: Do we need to ignore ping messages here too?

For all these cases really, for ds2 and ds3, they could all receive a ping, right?

upstairs/src/lib.rs
```diff
@@ -498,18 +498,53 @@ async fn process_message(
...
+            session_id,
             job_id,
             error,
         } => {
-            // XXX currently, this error report goes nowhere except to the log.
-            // The Upstairs should track this for each Downstairs.
+            // The Upstairs should not consider a job completed until it has
+            // returned an Ok result, and should therefore log and eat all
+            // ErrorReport messages here. This will change in the future when
+            // the Upstairs tracks the number of errors per Downstairs, and acts
```
leftwo, line 509: Let's file an issue for this lest we forget it.

upstairs/src/lib.rs
```diff
@@ -498,18 +498,53 @@ async fn process_message(
...
+            // However, there is one case (see `check_message_for_abort` in
+            // downstairs/src/lib.rs) where the Upstairs **does** need to
+            // act: when a repair job in the Downstairs fails, that Downstairs
+            // aborts itself and reconnects.
+            let _active = u.active.lock().await;
+            let ds = u.downstairs.lock().await;
+
+            if let Some(job) = ds.ds_active.get(job_id) {
+                if job.work.is_repair() {
+                    // Return the is error and let the previously written error
```
leftwo, line 525: Extra `is` here

upstairs/src/lib.rs
```diff
@@ -498,18 +498,53 @@ async fn process_message(
...
+                    // processing code work.
+                    cdt::ds__repair__done!(|| (job_id, client_id as u64));
+
+                    // XXX uncomment this to see the upstairs disconnect this
```
leftwo, line 529: We should take out the XXX comments here, I think once we know this does what we expect
we don't need to comments.  I don't know why (other than testing) we would want to get the
upstairs stuck in live repair.

upstairs/src/lib.rs
```diff
@@ -4193,6 +4228,8 @@ impl Downstairs {
                             self.downstairs_errors
                                 .insert(client_id, errors + 1);
                         }
+
+                        // If an repair job errors, mark that downstairs as bad
```
leftwo, line 4232: Nit: If `a` repair job fails..

upstairs/src/lib.rs
```diff
@@ -4231,6 +4268,7 @@ impl Downstairs {
                             self.downstairs_errors
                                 .insert(client_id, errors + 1);
                         }
+
```
leftwo, line 4271: Nit: looks like an extra blank line slipped in here..

leftwo (approved):
I think this is good, it sends the IO back to the repair task directly which means
we don't need a whole new set of plumbing for notifying the repair task that
the task should exit.

I _think_ that's the right thing here... I think...

upstairs/src/lib.rs
jmpesp: I'm allowed a few XXX, this PR's still in draft! :P

upstairs/src/dummy_downstairs_tests.rs
jmpesp: Luckily the task in `spawn_message_receiver` eats pings (and responds to them!) for us.

upstairs/src/dummy_downstairs_tests.rs
jmpesp: See other comment :)

upstairs/src/lib.rs
jmpesp: I filed https://github.com/oxidecomputer/crucible/issues/857, which is a general version of the work required (as the previous error handling code did count the number of errors per Downstairs!).

upstairs/src/lib.rs
jmpesp: fixed in 20def81

upstairs/src/lib.rs
jmpesp: fixed in 5f10b2c

upstairs/src/lib.rs
jmpesp: That was an intention separation, but I added a comment in 940cacb

PR #859, opened 2023-07-26
A bounced agent must reinitialize all SMF config

jmpesp:
Inspired by Angela's testing where she disabled and enabled the Pantry to see what would happen, I tried this with the Agent, and saw that the agent would **not** reconfigure running snapshots when bounced. This is due to object state transitioning from Requested to Created, but `apply_smf` only taking action for running snapshots in state Requested. If it was successfully created once, then it will remain in state Created, and when the agent stops and starts again it will 1) read the datafile, and 2) initialize all running snapshot SMF for objects with state Requested.

The fix for this is to take action in `apply_smf` on both Requested and Created states. This is different from how the region/downstairs objects are handled, but importantly the agent doesn't create snapshots like it does for region requests.

There's another bug fix here where it was previously possible to delete a snapshot if the running read-only downstairs corresponding to that snapshot was in state Failed. It's now not possible so we can investigate.

Testing the fix involved creating two general interfaces, one for interacting with SMF and one for interacting with ZFS snapshots. Two implementations of both were created: a real one for use, and a mock one for testing. The test that confirmed the running snapshot bug that this commit fixes is `test_smf_running_snapshot`.

### DIFF
--- a/agent/src/main.rs
+++ b/agent/src/main.rs
@@ -1467,7 +1467,9 @@ fn worker(
 
                 if let Err(e) = result {
                     error!(log, "SMF application failure: {:?}", e);
-                    // XXX no fail_rs?! RS is still in requested
+
+                    // There's no fail_rs here: a future `apply_smf` should
+                    // attempt to start the service again.
                 } else {
                     info!(log, "SMF ok!");
 
--- a/agent/src/main.rs
+++ b/agent/src/main.rs
@@ -1,5 +1,5 @@
 // Copyright 2021 Oxide Computer Company
-#![allow(clippy::needless_collect)]
+
 use anyhow::{anyhow, bail, Result};
 use clap::Parser;
 use dropshot::{ConfigLogging, ConfigLoggingIfExists, ConfigLoggingLevel};
@@ -13,6 +13,7 @@ use std::sync::Arc;
 
 const PROG: &str = "crucible-agent";
 const SERVICE: &str = "oxide/crucible/downstairs";
+
 /*
  * As a safety mechanism to prevent the agent from allocating more space to
  * regions than we have physical space on disk, we use a zfs reservation on
@@ -40,9 +41,12 @@ const QUOTA_FACTOR: u64 = 3;
 mod datafile;
 mod model;
 mod server;
+mod smf_interface;
+mod snapshot_interface;
 
 use model::Resource;
 use model::State;
+use smf_interface::*;
 
 #[derive(Debug, Parser)]
 #[clap(name = PROG, about = "Crucible zone management agent")]
@@ -279,6 +283,11 @@ async fn main() -> Result<()> {
                 listen,
                 lowport,
                 lowport + 999, // TODO high port as an argument?
+                Arc::new(snapshot_interface::ZfsSnapshotInterface::new(
+                    log.new(
+                        o!("component" => String::from("ZfsSnapshotInterface")),
+                    ),
+                )),
             )?);
 
             let regions_dataset = dataset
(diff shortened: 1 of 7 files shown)


agent/src/main.rs
```diff
@@ -580,7 +613,19 @@ fn apply_smf(
...
+            // original request to start a running snapshot is made, the object
+            // begins in state Requested, and once apply_smf returns
+            // successfully has its state changed to Created (this is used by
+            // the caller to query if the service has been created yet).
+            //
+            // If the agent zone is bounced, we want it to read the datafile
+            // and restore all the previously created SMF instances. This means
+            // that we have to take action on both Requested and Created for
+            // running snapshots. This is in contrast to Region, which has
+            // different actions for Requested and Created.
```
leftwo, line 627: Is the agent single threaded here?  Like could two different tasks/threads both call apply_smf_actual()?  I just want to be sure that the startup case, where we have a bunch of restart work to do won't clash with some external entity trying to act on the same snapshot at the same time, and being confused for the startup path.

jmpesp: As far as I can tell it runs two threads: one for the worker and one for the dropshot server:

```rust
            let log0 = log.new(o!("component" => "worker"));
            let df0 = Arc::clone(&df);
            std::thread::spawn(|| {
                worker(
                    log0,
                    df0,
                    regions_dataset,
                    downstairs_program,
                    downstairs_prefix,
                    snapshot_prefix,
                )
            });

            server::run_server(&log, listen, df).await
```

datafile uses interior mutability to make changes to the datafile, and both those threads share an `Arc<Datafile>`. It's possible for state changes to interleave because neither the server nor the worker hold a persistent guard to the inner part of the datafile. I don't think there's any guarding against multiple threads calling `apply_smf`, but only the worker function calls it: the server thread only makes datafile changes then signals for the worker via the Condvar.

jmpesp: Note that the agent does call `apply_smf` when starting up, before spawning the worker and starting the dropshot server.

agent/src/main.rs
```diff
@@ -984,6 +1467,7 @@ fn worker(
 
                 if let Err(e) = result {
                     error!(log, "SMF application failure: {:?}", e);
+                    // XXX no fail_rs?! RS is still in requested
```
leftwo, line 1470: Should we do something about the XXX here?

leftwo (approved):
Just a few help me understand it questions.

agent/src/smf_interface.rs
```diff
@@ -0,0 +1,762 @@
...
+
+        Ok(self.get_instance(name)?.unwrap())
+    }
+
+    /// When testing to see if running `apply_smf` based on the on-disk
+    /// datafile, disabled stuff remains in the current SMF interface and will
+    /// cause PartialEq comparison to fail. Prune that here so that comparison
+    /// can be done.
+    #[cfg(test)]
+    fn prune(&self) {
```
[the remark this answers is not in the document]
jmpesp: Yes but that'd be a different type of pruning: disabled services don't go away when a zone is shutdown and boot again. This is purely for comparison purposes during these tests.

agent/src/snapshot_interface.rs
```diff
@@ -0,0 +1,283 @@
...
+        ZfsSnapshotInterface { log }
+    }
+}
+
+impl SnapshotInterface for ZfsSnapshotInterface {
+    fn get_snapshots_for_dataset(
+        &self,
+        dataset: String,
+    ) -> Result<Vec<Snapshot>> {
+        let cmd = Command::new("zfs")
```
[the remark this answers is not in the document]
jmpesp: Yep, unless I messed that up somehow :)

PR #863, opened 2023-08-08
Replace a downstairs via a new VCR 

leftwo:
This adds support for replacing a crucible downstairs from the Volume layer by submitting a current and
new `VolumeConstructionRequest` (VCR).  If the new VCR matches the current with only a few allowed differences
 (gen, a single new target) then the guest layer `replace_downstairs` is called with the old and new downstairs targets.

To make this code a little more readable, new structure types were created for each of the enums in the `VolumeConstructionRequest`, and the `VolumeConstructionRequest` now uses the new types.
These changes can be seen in `crucible-client-types/src/lib.rs`, and that's probably a good place to start.

A bunch of new tests were created to verify the old/new VCRs only differ in the required ways, and any other
differences are blocked.

The meat of the changes here are the new `compare_vcr_for_replacement()` call that compares two VCRs and, if it
finds them to match enough, returns the old and new target downstairs.  This can then be passed on to the
existing `replace_downstairs()` call that will tell the upstairs to perform the replacement.

There is a fair chunk of code that is just testing all the possible differences that two VCRs might have, which is sadly
like 85% the same for each test.  I'm happy to take suggestions on how to turn that into less code if possible, but
trying to keep a balance such that it is pretty clear as to what the test was doing.

### DIFF
--- a/upstairs/src/volume.rs
+++ b/upstairs/src/volume.rs
@@ -1396,25 +1396,28 @@ impl Volume {
         replacement: VolumeConstructionRequest,
         log: &Logger,
     ) -> Result<(), CrucibleError> {
-        let (original_target, new_target) =
-            match Self::compare_vcr_for_update(original, replacement, log)
-                .await
-            {
-                Ok(Some((o, n))) => (o, n),
-                Ok(None) => {
-                    crucible_bail!(
-                        ReplaceRequestInvalid,
-                        "VCR replacement does not differ on targets",
-                    )
-                }
-                Err(e) => {
-                    crucible_bail!(
-                        ReplaceRequestInvalid,
-                        "VCR replacement invalid: {}",
-                        e
-                    )
-                }
-            };
+        let (original_target, new_target) = match Self::compare_vcr_for_update(
+            original,
+            replacement,
+            log,
+        )
+        .await
+        {
+            Ok(Some((o, n))) => (o, n),
+            Ok(None) => {
+                crucible_bail!(
+                    ReplaceRequestInvalid,
+                    "VCR replacement does not differ on targets",
+                )
+            }
+            Err(e) => {
+                crucible_bail!(
+                    ReplaceRequestInvalid,
+                    "VCR replacement invalid: {}",
+                    e
+                )
+            }
+        };
 
         info!(
             log,
@@ -3317,10 +3320,9 @@ mod test {
 
         let log = csl();
         // This should return OK, but not have any targets.
-        let res =
-            Volume::compare_vcr_for_update(original, replacement, &log)
-                .await
-                .unwrap();
+        let res = Volume::compare_vcr_for_update(original, replacement, &log)
+            .await
+            .unwrap();
 
         assert!(res.is_none());
     }
@@ -3367,13 +3369,9 @@ mod test {
             });
 
         let log = csl();
-        assert!(Volume::compare_vcr_for_update(
-            original,
-            replacement,
-            &log
-        )
-        .await
-        .is_err());
+        assert!(Volume::compare_vcr_for_update(original, replacement, &log)
+            .await
+            .is_err());
     }
 
     #[tokio::test]
@@ -3418,13 +3416,9 @@ mod test {
             });
 
         let log = csl();
-        assert!(Volume::compare_vcr_for_update(
-            original,
-            replacement,
-            &log
-        )
-        .await
-        .is_err());
+        assert!(Volume::compare_vcr_for_update(original, replacement, &log)
+            .await
+            .is_err());
     }
 
     #[tokio::test]
@@ -3479,13 +3473,9 @@ mod test {
             });
 
         let log = csl();
-        assert!(Volume::compare_vcr_for_update(
-            original,
-            replacement,
-            &log
-        )
-        .await
-        .is_err());
+        assert!(Volume::compare_vcr_for_update(original, replacement, &log)
+            .await
+            .is_err());
     }
 
     #[tokio::test]
@@ -3532,13 +3522,9 @@ mod test {
             });
 
         let log = csl();
-        assert!(Volume::compare_vcr_for_update(
-            original,
-            replacement,
-            &log
-        )
-        .await
-        .is_err());
+        assert!(Volume::compare_vcr_for_update(original, replacement, &log)
+            .await
+            .is_err());
     }
 
     #[tokio::test]
@@ -3585,13 +3571,9 @@ mod test {
             });
 
         let log = csl();
-        assert!(Volume::compare_vcr_for_update(
-            original,
-            replacement,
-            &log
-        )
-        .await
-        .is_err());
+        assert!(Volume::compare_vcr_for_update(original, replacement, &log)
+            .await
+            .is_err());
     }
 
     #[tokio::test]
@@ -3638,13 +3620,9 @@ mod test {
             });
 
         let log = csl();
-        assert!(Volume::compare_vcr_for_update(
-            original,
-            replacement,
-            &log
-        )
-        .await
-        .is_err());
+        assert!(Volume::compare_vcr_for_update(original, replacement, &log)
+            .await
+            .is_err());
     }
 
     // This is a wrapper function to test changing CrucibleOpts structures.
--- a/upstairs/src/volume.rs
+++ b/upstairs/src/volume.rs
@@ -1092,6 +1092,280 @@ impl Volume {
             }
         }
     }
+
+    pub async fn compare_vcr_for_replacement(
+        original: VolumeConstructionRequest,
+        replacement: VolumeConstructionRequest,
+        log: &Logger,
+    ) -> Result<(), CrucibleError> {
+        let (o_id, o_block_size, o_sub_volumes, o_read_only_parent) =
+            match original {
+                VolumeConstructionRequest::Volume {
+                    id,
+                    block_size,
+                    sub_volumes,
+                    read_only_parent,
+                } => {
+                    (id, block_size, sub_volumes, read_only_parent)
+                },
+                VolumeConstructionRequest::Url { .. } => {
+                    crucible_bail!(Unsupported, "Cannot replace URL VCR")
+                }
+
+                VolumeConstructionRequest::Region { .. } => {
+                    crucible_bail!(Unsupported, "Cannot replace Region VCR")
+                }
+
+                VolumeConstructionRequest::File { .. } => {
+                    crucible_bail!(Unsupported, "Cannot replace File VCR")
+                }
+            };
+
+        info!(log, "Original {o_id}");
+        let (id, block_size, sub_volumes, read_only_parent) =
+            match replacement {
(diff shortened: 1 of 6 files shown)


leftwo:
Notes to self:
Do we need to update the volume`self`?  Just the target (and the gen) or the whole thing?
What does it mean to have a new generation number in the Volume, but still be active
with the old one?  Okay I think.

upstairs/src/volume.rs
```diff
@@ -1092,15 +1094,355 @@ impl Volume {
...
+    //    A. The new generation number must be greater than the old.
+    //    B. Only one CrucibleOpts::target[] is different.
+    //    C. The new Volume can have None for read only parent if the
+    //       original had Some(), or it must match the Some().
+    //
+    // Any other difference between the two volumes is considered a failure.
+    //
+    // On success, we record the old and new downstairs target and return
+    // those to the caller.
+    pub async fn compare_vcr_for_replacement(
```
gjcolombo, line 1113: A high-level thought: would it make sense to use this during live migration to ensure that two Crucible backends have been configured compatibly? (You'd have the old VCR on the source and the new VCR on the target; the source would send its VCR to the target; the target could then verify that its VCR is plausibly compatible with the one on the source.) We'd have a little work to do in Propolis to enable this but it's nothing complicated.

The reason I ask is that if this *could* be used this way, it may be nicer to phrase the routine name and error messages throughout in terms of "compatibility" instead of "replaceability."

upstairs/src/volume.rs
```diff
@@ -1092,15 +1094,355 @@ impl Volume {
...
+            crucible_bail!(
+                ReplaceRequestInvalid,
+                "subvolume len mismatch {} vs. {}",
+                o_volume.sub_volumes.len(),
+                n_volume.sub_volumes.len(),
+            )
+        }
+
+        // Presently, we only support one sub_volume for replacement.
+        if n_volume.sub_volumes.len() != 1 {
```
gjcolombo, line 1208: This check doesn't compare `n_volume` to `o_volume` like the others nearby. Is there (or should there be) a separate routine that checks the validity of an individual VCR? (I could also be misunderstanding--is having more than one `sub_volume` OK in general but invalid for replacement? If that's the case I might elaborate on why that is in the comment here.)

upstairs/src/volume.rs
```diff
@@ -1092,15 +1094,355 @@ impl Volume {
...
+                "sub_volume opts target unchanged {:?} vs. {:?}",
+                o_opts.target,
+                n_opts.target
+            )
+        }
+
+        // Walk the targets, we can only have one different
+        let mut new_target_cid = None;
+        for cid in 0..3 {
+            if o_opts.target[cid] != n_opts.target[cid] {
```
gjcolombo, line 1353: Does target ordering matter in other contexts? (I'm fretting about a hypothetical future bug where Nexus replaces a target by calling `remove()` to take one out and `push()` to put one in...)

upstairs/src/volume.rs
```diff
@@ -1092,15 +1094,355 @@ impl Volume {
...
+
+    // Given two VolumeConstructionRequests, compare them to verify they
+    // only have the proper differences and if the VCRs are valid, submit
+    // the targets for replacement.  A success here means the upstairs has
+    // accepted the replacement and the process has started.
+    pub async fn target_replace(
+        &self,
+        original: VolumeConstructionRequest,
+        replacement: VolumeConstructionRequest,
+        log: &Logger,
```
gjcolombo, line 1388: Is this the interface Propolis is going to call? It looks great, except that we don't have a logger handy in the `CrucibleBackend`, because the backend doesn't hang onto the logger that it passed down to the upstairs in `Volume::construct`. Can the top-level Volume hold onto that logger in addition to passing it down to the subvolumes so that the caller doesn't have to construct a new one here?

upstairs/src/volume.rs
leftwo: Yes and no.  The initial order for targets does not matter, but once we have started the upstairs, whomever is at index 0 must remain their for the life of that upstairs process.  We can't do `remove()` and `push()` for that reason inside Crucible upstairs.

The upstairs is also pretty dependent on three targets always existing (even if it can't connect to one of them).

If you were to stop the upstairs and create a new upstairs, then the order can be different and the upstairs will be just fine.  There is no dependency on a specific downstairs being at a specific target index when starting cold.

You are correct to fret about Nexus trying to `remove()` and `push()`, and when we write that side of the downstairs replacement code, it will have to do a replacement of the specific target at the index it currently resides in.

upstairs/src/volume.rs
leftwo: The original idea was that Volumes could have multiple SubVolumes for different VA ranges, so most of the Volume code just treats the `sub_volume` as a `Vec` and walks all the elements acting on each one.  
However, a few other dependencies have crept in (like activation handling) that now prevent us from supporting more than one `sub_voume` for a `Volume`.  
I wanted to be explicit here and check for a Volume with more than one `sub_volume` and fail it, as this code as written will not handle it.  If we decide in the future to come back and allow it, then we can write the replacement code to handle it then (along with the other places that need updating to support it).

I can leave a comment/bread-crumb here though to better describe that decision to make it easier for future us to find.

upstairs/src/volume.rs
leftwo: I think this could be a more generic check, with a little more polish.

For a migration, the generation on the new side should be > than the old, which we check here.  For a migration, the read only parent could be `None` on the new side, which we check here.  I'm assuming we don't want to migrate and replace a downstairs at the same time (at least initailly).

To make this generic, it would have to differentiate between the "okay for migration" VCR and "ok for downstairs replacement" VCR.

I think a way to do that would be to return `Ok(None)` if the VCRs are the same (given the above constraints) and 'Ok(Some(old_target, new_target))` if the VCRs are the same with the exception of one target being different.

How does that sound?

upstairs/src/volume.rs
```diff
@@ -1092,15 +1094,355 @@ impl Volume {
...
+                crucible_bail!(ReplaceRequestInvalid, "Cannot replace File VCR")
+            }
+        };
+
+        let n_volume = match replacement {
+            VolumeConstructionRequest::Volume(vcr_v) => vcr_v,
+            VolumeConstructionRequest::Url { .. } => {
+                crucible_bail!(
+                    ReplaceRequestInvalid,
+                    "Replacement URL VCR invalid "
```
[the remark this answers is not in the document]
leftwo: Will fix

upstairs/src/volume.rs
leftwo: Yeah, this interface was what I was hoping Propolis could use.

I've wanted a few times to add a `Logger` field to the volume layer, but so far have not done so.  Maybe this is the time to do that.  Let me see if I can figure out how much work it is to do that.  I think since we have updated Propolis to supply a logger at `Volume::construct`, crucible can attach it there.  It's really all the test code I'll have to change :)

leftwo (commented):
Thanks for the comments!

I've tried to answer them, and I'm going to work on some suggestions you had and I'll push up changes when I have that work done.

upstairs/src/volume.rs
gjcolombo: Makes sense to me--thanks for the context!

leftwo (commented):
I believe I've addressed all comments, this is ready for a re-review.

upstairs/src/volume.rs
```diff
@@ -1096,25 +1101,35 @@ impl Volume {
...
+    //  Specifically for the downstairs target replacement case we also require
+    //  this difference:
+    //    C. Only one CrucibleOpts::target[] is different.
+    //
     // Any other difference between the two volumes is considered a failure.
     //
-    // On success, we record the old and new downstairs target and return
-    // those to the caller.
-    pub async fn compare_vcr_for_replacement(
+    // We return Ok(None) if requirements 1, 2, 3A, 3B are met.  This would
```
gjcolombo, line 1122: This makes sense to me, since we only need to return a socket address pair if we're replacing, but it feels a little weird to me to have the caller check for the flavor of `Ok` it got back when it could have specified to the callee which checks to do. That is, I think the downstairs-replacement path is going to end up with something like this:

```rust
match compare_vcr_for_update(original, replacement, log)? {
    Ok(Some(_)) => todo!("okay to proceed"),
    Ok(None) => bail!("target replacement check failed"),
}
```

Might it make sense to have two functions? `compare_vcr_for_update` checks 1, 2, 3A, and 3B and returns a `Result<()>`; migration calls that one. `compare_vcr_for_replacement` calls `compare_vcr_for_update` and, if that succeeds, does check 3C, and if that succeeds it returns `Result<(SocketAddr, SocketAddr)>`. Then the downstairs replacement code can just `let (old_addr, new_addr) = compare_vcr_for_replacement()?;`.

WDYT?

gjcolombo (commented):
Oof, sorry for inciting all those logging changes... One more suggestion about the comparison routines, but otherwise this is looking pretty good.

upstairs/src/volume.rs
leftwo: The challenge here is that, and perhaps I need to make it clear in the comment as well, is:
* For migration, I don't want 3C, the targets must be the same.  
* For replacement, I do want 3C, the targets must be different.  

The `compare_vcr_for_replacement` does not build on top of `compare_vcr_for_migration`

I think I can do it with a wrapper in front of `compare_vcr_for_update`, where I just hide the `Ok` check like you have shown above and have a `migration_check` and a `replacement_check`, 
`migration_check` just looks for `Ok(None)`, and `replacement_check` wants the `Ok(Some(_))`
Then I'm at least sweeping the `Ok` check under the rug and the callers don't have to match.

I could also make a "check everything except the `sub_volume` targets" function (what is common between the replace/migrate cases). If I did that I would have to make another two functions that then did the translation/selection from `VolumeConstructionRequest` -> `VcrVolume` -> `SubVolume` -> Target Vec, and returned the Ok/Err depending on what they wanted, which would then look something like this:

```
fn compare_vcr_for_migration(..) -> Result<()> {
   compare_vcr_common(..)?
   compare_vcr_subtarget_the_same()
}
```
and
```
fn compare_vcr_for_migration(..) -> Result<(SocketAddr, SocketAddr)> {
   compare_vcr_common(..)?
   compare_vcr_subtarget_one_different()
}
```

Or, maybe I make a `give_me_the_subvolume_target_from_this_vcr` function, then compare the two targets after I have done the `compare_vcr_common`, but.. I don't know what ends up being the better solution here.   I kind of like just hiding the Ok check. 

Let me know if you feel strongly about either of these two solutions (or some third idea).  I do think it's worth updating this to be a little better and having two functions to call depending on what the caller wants, but I need a little direction on which path you think might be better here.

leftwo: I went ahead and tried option A (the wrapper around `compare_vcr_for_update`)
Let me know what you think.

PR #868, opened 2023-08-14
Undo the last VolumeConstructionRequest changes

leftwo:
To change the VCR, we have to support transitioning from the current version to the new version, as the VCR is a string field in the database.  The recent VCR changes to support downstairs replacement were not required for functionality.  As we don't want to make the changes on Omicron side to support VCR changes yet, we can transition the VCR back to what it was but still support the new functionality for downstairs replacement.


jmpesp (approved):
:package:

PR #874, opened 2023-08-17
ACK the write after adding it to the work queue

leftwo:
Ack writes after submission to downstairs work queues                               
                                                                                    
Write jobs are marked for ACK after we submit them to the downstairs work           
queue.  This changes the previous behavior where we waited for 2/3 writes           
to be ACK'd back from the downstairs.                                               
                                                                                    
Updated some replay tests to not change the ACK status for writes when a            
replay happens.                                                                     
                                                                                    
As we can absorb much more outstanding IOs with this change, the queue              
depth for downstairs outstanding work as well as the number of active               
jobs we hold in the upstairs before deciding there is a problem has                 
been increased.                                                                     
                                                                                    
The jobs outstanding to a downstairs, MAX_ACTIVE_COUNT, goes from 600 to 2600.      
The total jobs outstanding, IO_OUTSTANDING_MAX, goes from 5000 to 57000.            
Both these numbers probably need more tuning.                                       
                                                                                    
dsc now takes a port_base instead of always starting at 8810.                       
This enables multiple dscs to be running at the same time.                          
                                                                                    
Took some of the log messages out of dummy_downstairs_tests in a hope               
that they would run faster in CI.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -26,7 +26,7 @@ pub const REPAIR_PORT_OFFSET: u16 = 4000;
 
 // Max number of submitted IOs between the upstairs and the downstairs, above
 // which flow control kicks in.
-pub const MAX_ACTIVE_COUNT: usize = 600;
+pub const MAX_ACTIVE_COUNT: usize = 2600;
 
 #[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(
--- a/dsc/src/main.rs
+++ b/dsc/src/main.rs
@@ -80,12 +80,19 @@ enum Action {
         #[clap(long, global = true, default_value = "/tmp/dsc", action)]
         output_dir: PathBuf,
 
+        /// The port_base where the downstairs will listen.
+        /// Note: is used as part of the expected region directory path.
+        #[clap(long, default_value = "8810", action)]
+        port_base: u32,
+
         /// The directory where the downstairs regions will be created.
         /// Either provide once, or "N" times.  One means all downstairs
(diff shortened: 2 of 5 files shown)


andrewjstone:
It'll go faster if you don't even queue it!

leftwo:
With this set, I quickly hit flow control on my downstairs IO.

The flow control is set by MAX_ACTIVE_COUNT

I updated MAX_ACTIVE_COUNT from 600 to 2600.
However this resulted in the upstairs thinking there were too many outstanding jobs, which caused the upstairs to kick out a downstairs.
That max job count is set by IO_OUTSTANDING_MAX

So, it looks like IO_OUTSTANDING_MAX needs to be increased as well.

My bench gimlet is having some trouble at the moment, so I've not been able to try running with both
a higher MAX_ACTIVE_COUNT and IO_OUTSTANDING_MAX.

leftwo:
The current flow control + IO max outstanding values need further tuning.
With this change, we can overwhelm the upstairs and kick out a downstairs, which then
reults in the upstairs breaking with this issue:
```
Error: IO Error: 2 out of 3 downstairs failed to complete this IO
```
This code path needs more work to handle the new instant ack for Writes.

leftwo:
I think further tuning will be needed to both:

* Make sure we don't kick downstairs out because they can't quite keep up.
* Make sure we don't take too much memory if a downstairs goes away before we decide to kick it out.

jmpesp (approved):

PR #875, opened 2023-08-18
Only flush extents that might actually be dirty.

### LINKED ISSUE #872: dispatch less flush tasks
we don't need to flush extents that aren't dirty. Right now we spawn one flush task for *every* extent during a flush command. We should be able to cut down on this a lot by keeping a set in the region struct of which extents actually need flushes. 

So what extents _do_ need flushes? Any of them that are marked dirty.

- when opening an extent, check if it is dirty, if so, add to the list. will get flushed during the first flush command received from upstairs
- when writing to an extent, add it to the list
- when flushing, issue flush jobs for everything on the list. clear the list

Question: what about if a flush fails? Well maybe we have bigger problems, and i think we failkill the downstairs at that point anyway.

faithanalog:
This change keeps track at the region level which extents might be dirty. That way, during a flush, it only locks/sends flush commands to those extents.

This makes a big difference for regions with a lot of extents. Right now our write perf goes down the more extents you have, and it's pretty gnarly.

When opening extents, we check the dirty flag. If it's dirty, we add it to the region's dirty flag cache.

When we send a write command to an extent, we unconditionally set the dirty flag cache for that extent. If the extent decides the write is a NOP and doesn't actually mark itself dirty in its metadata, it'll still get flushed. That's fine- it's technically a waste to do, but it's not going to happen a lot and it's minor if it does. So it's not worth the complexity we'd need to perfectly track the dirty state of the extents.

Basically, an incorrect negative is unacceptable, but an incorrect positive is fine.

fixes #872 

I am fairly confident in the logic of this but I'd like an extra confirmation from the other crucible-knowers that I'm not going to end up with a stale cache somehow, so please really make sure the logic of this seems sound. Also feel free to suggest tests if you think our existing test suite doesn't cover some aspect of this.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -923,7 +923,7 @@ where
             extent_id,
         } => {
             let msg = {
-                let d = ad.lock().await;
+                let mut d = ad.lock().await;
                 debug!(d.log, "{} Reopen extent {}", repair_id, extent_id);
                 match d.region.reopen_extent(*extent_id).await {
                     Ok(()) => Message::RepairAckId {
@@ -1797,7 +1797,7 @@ pub struct ActiveUpstairs {
  */
 #[derive(Debug)]
 pub struct Downstairs {
-    pub region: Arc<Region>,
+    pub region: Region,
     lossy: bool,        // Test flag, enables pauses and skipped jobs
     read_errors: bool,  // Test flag
     write_errors: bool, // Test flag
@@ -1829,7 +1829,7 @@ impl Downstairs {
             ))),
         };
         Downstairs {
-            region: Arc::new(region),
+            region,
             lossy,
             read_errors,
             write_errors,
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -1766,21 +1766,23 @@ impl Region {
 
         for eid in eid_range {
             let extent = if create {
-                Extent::create(&self.dir, &self.def, eid)
+                Extent::create(&self.dir, &self.def, eid)?
             } else {
-                Extent::open(
+                let extent = Extent::open(
                     &self.dir,
                     &self.def,
                     eid,
                     self.read_only,
                     &self.log,
                 )
-                .await
-            }?;
+                .await?;
 
-            if extent.dirty().await {
-                self.dirty_extents.insert(eid as usize);
-            }
+                if extent.dirty().await {
+                    self.dirty_extents.insert(eid as usize);
+                }
+
+                extent
+            };
 
             these_extents.push(Arc::new(Mutex::new(ExtentState::Opened(
                 Arc::new(extent),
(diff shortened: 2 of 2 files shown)


faithanalog:
for context, on a 256 gig disk with 32MiB extents:

```
[root@nixos:~]# pgbench -j 8 -c 50 -f insert.sql -T 30 -s 50
pgbench (14.8, server 15.3)
starting vacuum...end.
transaction type: insert.sql
scaling factor: 50
query mode: simple
number of clients: 50
number of threads: 8
duration: 30 s
number of transactions actually processed: 487386
latency average = 3.076 ms
initial connection time = 22.624 ms
tps = 16252.809476 (without initial connection time)
```

this is good, i was getting 6600 tps on a 32 gig disk, and had to drop down to like a 512MiB disk to see numbers like this. now i can make the disk huge and it still works about the same regardless

some other stuff im copying here from chat:

> fio randwrite-16k on this is still slower than the 256MiB-extents run from last night was, but it matches the randwrite-16k speed of the 32MiB-extents run with a disk 1/8 the size, so that's nifty. makes a degree of sense, you can probably hit quite a number of extents in the flush interval with a fully random write pattern and get a lot of extents queued for a flush at a time
> meanwhile, seq writes, and large random writes outperform all the tests i was running last night. so i guess there's a sweet spot where if you write to a handful of extents before a flush, the parallelism from that gives you a speedup compared to writing to just a single extent. but if you write to a LOT of extents before a flush, you start getting slower than if you just had a single extent. interesting

> the important thing is that this no longer scales linearly with the size of the disk, it scales with the number of extents written to since the last flush. which is much more manageable

> and it also might affect what the ideal flush interval is, may need to re-tune that

downstairs/src/lib.rs
```diff
@@ -1797,7 +1797,7 @@ pub struct ActiveUpstairs {
  */
 #[derive(Debug)]
 pub struct Downstairs {
-    pub region: Arc<Region>,
```
jmpesp, line 1800: Curious, why this change too?

downstairs/src/region.rs
```diff
@@ -2317,23 +2344,41 @@ impl Region {
...
+                let dirty_extents = self
+                    .dirty_extents
+                    .iter()
+                    .copied()
+                    .filter(|x| *x <= el)
+                    .collect();
+
+                for eid in &dirty_extents {
+                    self.dirty_extents.remove(eid);
+                }
```
jmpesp, line 2368: The Downstairs will re-execute jobs until they succeed if errors are seen, so these should only be removed after all potentially erroneous operations have been done, else a rerun of this will not flush anything.

jmpesp (changes requested):

downstairs/src/lib.rs
jmpesp: Nevermind this question, I get it now - otherwise `dirty_extents` would have to be guarded with a Mutex or something.

faithanalog: Yeah basically, its because i couldnt make some func take an `&mut self` and still call it where i wanted. and nothing actually was using the Arc behavior, like no code ever cloned the reference, so i just removed the arc and everything was still happy

downstairs/src/region.rs
faithanalog: ah- i thought downstairs failed out if errors were seen and that a repair would then be needed. i wrote this under that assumption.

downstairs/src/region.rs
```diff
@@ -1765,9 +1776,14 @@ impl Region {
                     &self.log,
                 )
                 .await
-            };
+            }?;
+
+            if extent.dirty().await {
```
leftwo, line 1781: Could this be inside the else?
Like, if we are creating an extent, it would seem strange if it was already dirty :)

It's probably not going to make a noticeable difference in startup time though.

leftwo (approved):
Looks good.

I'm going to fire off a LiveRepair test and let it run over the weekend using these bits, just
to verify there is not some edge case I can't remember, but I think we are all good here.

downstairs/src/region.rs
faithanalog: it could be, yeah

downstairs/src/lib.rs
jmpesp: Not blocking this review, but I'm wondering what it would take to keep this. I have a branch that changes the downstairs to operate on requests on parallel, which will clone this reference.

jmpesp (approved):
:ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship: :ship:

leftwo:
I ran several hours of LiveRepair tests on these bits and found no problems.

PR #878, opened 2023-08-23
Use libc::getrandom instead of ChaCha20Rng

jmpesp:
illumos' getrandom uses ChaCha20 anyway, and this avoids either reconstructing the ChaCha20Rng object from entropy each time, or storing it in a Mutex and locking it when required (as otherwise `&mut self` would be required and this bubbles up the call chain).

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -4840,7 +4840,9 @@ impl EncryptionContext {
 
     #[cfg(target_os = "illumos")]
     fn get_random_nonce(&self) -> Nonce {
-        let mut random_iv: [u8; 12] = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+        let mut random_iv: Nonce = aes_gcm_siv::aead::generic_array::arr![u8;
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
+        ];
 
         // illumos' libc contains this
         extern "C" {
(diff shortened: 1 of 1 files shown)


jmpesp:
It tried that first! Unfortunately https://rust-lang.github.io/libc/x86_64-unknown-illumos/doc/libc/ doesn't include `arc4random`, and apparently neither does https://rust-lang.github.io/libc/x86_64-unknown-linux-gnu/doc/libc/.

jclulow:
You can just add an extern C declaration for the function you want inside this file, it doesn't have to come from libc specifically

faithanalog (approved):
we can actually avoid an extra copy like this, since `Nonce` is just a `GenericArray` is just a fancy wrapper around an array:

```diff
diff --git a/upstairs/src/lib.rs b/upstairs/src/lib.rs
index 4c4d432..233bb5a 100644
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -4840,7 +4840,9 @@ impl EncryptionContext {
 
     #[cfg(target_os = "illumos")]
     fn get_random_nonce(&self) -> Nonce {
-        let mut random_iv: [u8; 12] = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+        let mut random_iv: Nonce = aes_gcm_siv::aead::generic_array::arr![u8;
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
+        ];
 
         // illumos' libc contains this
         extern "C" {
@@ -4851,12 +4853,14 @@ impl EncryptionContext {
             arc4random_buf(random_iv.as_mut_ptr() as *mut libc::c_void, 12)
         }
 
-        Nonce::clone_from_slice(&random_iv)
+        random_iv
     }
 
     #[cfg(not(target_os = "illumos"))]
     fn get_random_nonce(&self) -> Nonce {
-        let mut random_iv: [u8; 12] = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+        let mut random_iv: Nonce = aes_gcm_siv::aead::generic_array::arr![u8;
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
+        ];
 
         let filled = unsafe {
             libc::getrandom(
@@ -4868,7 +4872,7 @@ impl EncryptionContext {
 
         assert_eq!(filled, 12);
 
-        Nonce::clone_from_slice(&random_iv)
+        random_iv
     }
 
     pub fn encrypt_in_place(
```

other than that, lgtm

PR #882, opened 2023-08-24
Do not scrub blank blocks

jmpesp:
During scrub, if the block is blank, don't worry about writing it.

### DIFF
--- a/upstairs/src/volume.rs
+++ b/upstairs/src/volume.rs
@@ -415,7 +415,10 @@ impl Volume {
                     )
                     .await?;
                 } else {
-                    info!(self.log, "Scrub eliding blank write at offset {}", offset);
+                    info!(
+                        self.log,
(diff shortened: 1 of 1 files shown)


upstairs/src/volume.rs
```diff
@@ -406,12 +406,20 @@ impl Volume {
...
+                if !buffer.as_vec().await.iter().all(|&x| x == 0) {
+                    // TODO: Nexus needs to know about this failure.
+                    self.write_unwritten(
+                        Block::new(offset, bs.trailing_zeros()),
+                        Bytes::from(buffer.as_vec().await.clone()),
+                    )
+                    .await?;
+                } else {
+                    info!(
+                        self.log,
```
leftwo, line 419: This might result in many log messages if the volume has many blank blocks.

leftwo (approved):
How does this play into the encryption/non-

Issue #883, opened 2023-08-24
MacOS build failure


leftwo:
Current main fails to build on macos:

```
error[E0425]: cannot find function `getrandom` in crate `libc`
    --> upstairs/src/lib.rs:4879:19
     |
4879 |             libc::getrandom(
     |                   ^^^^^^^^^ not found in `libc`

error[E0425]: cannot find value `GRND_NONBLOCK` in crate `libc`
    --> upstairs/src/lib.rs:4882:23
     |
4882 |                 libc::GRND_NONBLOCK,
     |                       ^^^^^^^^^^^^^ help: a constant with a similar name exists: `MNT_NOBLOCK`
     |
    ::: /Users/alan/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libc-0.2.147/src/unix/bsd/apple/mod.rs:4686:1
     |
4686 | pub const MNT_NOBLOCK: ::c_int = 0x00020000;
     | ------------------------------ similarly named constant `MNT_NOBLOCK` defined here
```

leftwo:
Fixed

PR #884, opened 2023-08-25
Make macos rand like illumos

leftwo:
It turns out `macos` has `arc4random_buf` but not `libc::getrandom`

Make `macos` compile like we do for `illumos` builds.


leftwo:
Fixes https://github.com/oxidecomputer/crucible/issues/883

PR #899, opened 2023-08-29
Only rehash if a write may have failed

mkeeter:
Previously, we always rehashed data during a `flush` operation, which required reading it back from disk and computing an integrity hash.

However, if a block was successfully written, then we already know what its hash value will be (since we computed it earlier)!  We _only_ need to read + rehash if we failed to write the data, because if we failed to write the data, then the on-disk bytes are in an unknown state.

This PR tracks dirty blocks as a map from `block -> Option<u64>`, so that if the hash is known, it won't be recomputed during a flush operation.

It shows a small but noticeable (1-3%) performance increase when running `pgbench`:

Screenshot 2023-08-29 at 11 48 39 AM

### DIFF
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -4840,8 +4840,6 @@ mod test {
             *a_buf = 9;
         }
 
-        region.region_flush(1, 0, &None, 0, None).await?;
-
         // read data into File, compare what was written to buffer
         let mut read_from_files: Vec<u8> = Vec::with_capacity(total_size);
 
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -1232,7 +1232,7 @@ impl Extent {
             (job_id, self.number, writes.len() as u64)
         });
 
-        let mut hashes_to_write = vec![];
+        let mut hashes_to_write = Vec::with_capacity(writes.len());
         for write in writes {
             if writes_to_skip.contains(&write.offset.value) {
                 hashes_to_write.push(None);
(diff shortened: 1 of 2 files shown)


downstairs/src/region.rs
```diff
@@ -1285,6 +1316,20 @@ impl Extent {
         // Write any remaining data
         batched_pwritev.perform_writes()?;
 
+        // At this point, we know that the on-disk values for any written blocks
+        // must match the integrity hashes calculated above (and stored to
+        // SQLite).  We can therefore store hash values for many dirty blocks,
```
jmpesp, line 1321: Unfortunately, we don't know that until `flush` returns Ok - `write` returning ok is insufficient to know that bytes have landed in durable storage.

jmpesp (changes requested):

downstairs/src/region.rs
jmpesp: After a few re-reads of this PR, this comment needs to change but at this point I think the PR itself is sound.

mkeeter: Yeah, I was struggling for the right way to phrase "written-but-not-flushed" data, but "on-disk" is definitely misleading.

Changed in 485307332c91e99af6f35abf169475c6fb997954, let me know what you think of the new working.

jmpesp: :+1:

faithanalog (approved):

downstairs/src/region.rs
```diff
@@ -1203,9 +1232,10 @@ impl Extent {
             (job_id, self.number, writes.len() as u64)
         });
 
+        let mut hashes_to_write = vec![];
```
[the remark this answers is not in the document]
mkeeter: yeah, seems like best practice (although it probably won't be noticeable in the performance test)

Issue #893, opened 2023-08-25
Crucible fails the live repair test


leftwo:
This dependabot PR, chore(deps): update rust crate tokio to 1.32 #890, failed the live_repair test:

https://github.com/oxidecomputer/crucible/pull/890/checks?check_run_id=16220798170

We need to figure out why.  Nothing in that PR should have any impact on live repair, which suggests
there is an underlying issue with live repair that we have not seen yet.  Either that, or something in
Tokio 1.32 broke in a strange way to introduce this.

leftwo:
Output printed for the failing test:
[test-repair-fail-testout.log](https://github.com/oxidecomputer/crucible/files/12444133/test-repair-fail-testout.log)

The crutest log (not including stedrr, an oversight that will get fixed, stderr is in the above log)
[crutest-replace.log](https://github.com/oxidecomputer/crucible/files/12444135/crutest-replace.log)

The log from DSC (the downtstairs)
[dsc.log](https://github.com/oxidecomputer/crucible/files/12444136/dsc.log)

leftwo:
The test was running this command:
```
ptime -m /var/tmp/bins/crutest replace -t 127.0.0.1:8810 -t 127.0.0.1:8820 -t 127.0.0.1:8830 --replacement 127.0.0.1:8840 -g 1 -c 10 --stable
```

The panic/mismatch is:
```
2023-08-25T17:44:24.943Z thread 'tokio-runtime-worker' panicked at '[2] read hash mismatch on 4993 upstairs/src/lib.rs:4438:29                                                                                                                  
2023-08-25T17:44:24.950Z Expected [Some(c05cf20bce896a22), Some(c69f8cfabdfa715f), Some(14b1d79450301735)]                                                                                                                                      
2023-08-25T17:44:24.959Z Computed [Some(4b685ce4967184af), Some(c69f8cfabdfa715f), Some(14b1d79450301735)]
``` 
By default (which this test uses), `dsc` will create 4k blocks, 100 blocks per extent, and 15 extents.
The downstairs are not running with encryption.

The Job that had the mismatch was coming  from downstairs 2 (from the panic message)
Here is the job itself:  Note that sometimes the hash is printed in hex, sometimes decimal, but the
decimal values are the same as the hex.
```
job: DownstairsIO {
    ds_id: 4993,
    guest_id: 3994,
    work: Read {
        dependencies: [],
        requests: [
[...]

PR #908, opened 2023-09-07
Move active jobs into a separate data structure and optimize `ackable_work`

mkeeter:
This PR does two things:

First, it moves `ds_active` from a `BTreeMap<u64, DownstairsIO>` into an opaque `ActiveJobs` data structure that exposes a very similar API to the user.  This data structure exposes functions like `deps_for_{read,write,flush}`, which gives us room for future optimization.

Dependency tracking remains identical to before, albeit in a new location (this means that #893 remains an issue).

Second, ackable jobs are now tracked in a `BTreeSet<u64>`, inside the new `struct ActiveJobs`.

Here's the motivation: the longest-running tests (`test_successful_live_repair` and `test_error_during_live_repair_no_halt`) are spending a huge amount of time in `ackable_work`, because each call iterates over every single active job:

Screenshot 2023-09-07 at 10 08 36 AM

However, tracking ackable jobs in a separate set is tricky, because existing code expects to modify a `&mut DownstairsIO` directly (and can therefore change its `ack_status` willy-nilly).

Instead of relying on users to do this right, I changed `ActiveJobs` to never hand out an uncontrolled `&mut DownstairsIO`.  Instead, it provides a new `struct DownstairsIOHandle`.  This handle wraps a `&mut DownstairsIO`, but also guarantees (on drop) that the list of ackable jobs is correctly updated; as such, it borrows from the parent `ActiveJobs`, so this is all checked by the borrow checker.

The result is that `test_successful_live_repair` goes from 33 seconds down to about 21 (in `cargo test --release`).

I've got more advanced optimizations planned, but this seems like a reasonable point to open a PR.

### DIFF
--- a/upstairs/src/active_jobs.rs
+++ b/upstairs/src/active_jobs.rs
@@ -2,8 +2,8 @@
 
 use crucible_common::RegionDefinition;
 
-use crate::{DownstairsIO, ImpactedBlocks};
-use std::collections::BTreeMap;
+use crate::{AckStatus, DownstairsIO, ImpactedBlocks};
+use std::collections::{BTreeMap, BTreeSet};
 
 /// `ActiveJobs` tracks active jobs by ID
 ///
@@ -19,12 +19,14 @@ use std::collections::BTreeMap;
 #[derive(Debug)]
 pub(crate) struct ActiveJobs {
     jobs: BTreeMap<u64, DownstairsIO>,
+    ackable: BTreeSet<u64>,
 }
 
 impl ActiveJobs {
     pub fn new() -> Self {
         Self {
             jobs: BTreeMap::new(),
+            ackable: BTreeSet::new(),
         }
     }
 
@@ -39,7 +41,7 @@ impl ActiveJobs {
     pub fn get_mut(&mut self, job_id: &u64) -> Option<DownstairsIOHandle> {
         self.jobs
             .get_mut(job_id)
-            .map(|job| DownstairsIOHandle { job })
+            .map(|job| DownstairsIOHandle::new(job, &mut self.ackable))
     }
 
     /// Returns the total number of active jobs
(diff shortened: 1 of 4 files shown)


upstairs/src/lib.rs
```diff
@@ -3301,7 +3305,7 @@ impl Downstairs {
                 self.io_state_count.incr(&IOState::Skipped, client_id);
                 self.ds_skipped_jobs[client_id as usize].insert(*ds_id);
```
jmpesp, line 3306: This initially surprised me: it's capturing `&mut self`, which is `Downstairs` here, right?

upstairs/src/active_jobs.rs
```diff
@@ -0,0 +1,349 @@
...
+        }
+    }
+
+    pub fn job(&mut self) -> &mut DownstairsIO {
+        self.job
+    }
+}
+
+impl<'a> std::ops::Drop for DownstairsIOHandle<'a> {
+    fn drop(&mut self) {
```
jmpesp, line 323: I love this, nice

jmpesp (approved):

upstairs/src/lib.rs
mkeeter: Oh, maybe you're confused by how we can call a `&mut self` function on `self.ds_active`, while also mutating other members of `self`?

That's allowed by the Rust 2021 [disjoint capture](https://doc.rust-lang.org/edition-guide/rust-2021/disjoint-capture-in-closures.html) behavior.

jmpesp: My confusion stems from seeing `self` in a closure :)

leftwo (approved):
I'm going to read this a few time so I understand it, but I don't have any comments and I'm all good with it going in now.

PR #911, opened 2023-09-08
Crutest updates, rename tests, new options

leftwo:
Renamed the Rand test to WFR, as it actually was doing a write then a flush then a read and that better describes what it does.

Renamed verify option to verify-at-end

Created a new verify-at-start and made that option require the verify-in option through clap instead of a specific check.

Updated the amount of work the replace test does as I was seeing that the IO part of the test could sometimes finish before the replacement would happen.

Removed a final volume verify from a few tests now that we can optionally select to do that final verify and sometimes we don't want to do it.  this allows more flexibility

### DIFF
--- a/crutest/src/main.rs
+++ b/crutest/src/main.rs
@@ -2095,11 +2095,9 @@ async fn perf_workload(
     let es = ri.extent_size.value;
     let ec = ri.total_blocks as u64 / es;
 
-    // To make a random block offset modulus, we take the total
-    // block number and subtract the IO size in blocks.
-    let offset_mod =
-        rng.gen::<u64>() % (ri.total_blocks - blocks_per_io) as u64;
-
+    // To make a random block offset, we take the total block count and subtract
+    // the IO size in blocks (so that we don't overspill the region)
+    let offset_mod = (ri.total_blocks - blocks_per_io) as u64;
     for _ in 0..write_loop {
         let mut wtime = Vec::with_capacity(count);
         let big_start = Instant::now();
--- a/crutest/src/main.rs
+++ b/crutest/src/main.rs
@@ -32,7 +32,7 @@ use dsc_client::{types::DownstairsState, Client};
  * The various tests this program supports.
  */
 /// Client: A Crucible Upstairs test program
-#[allow(clippy::derive_partial_eq_without_eq)]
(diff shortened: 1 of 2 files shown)


crutest/src/main.rs
```diff
@@ -1068,6 +1062,23 @@ async fn main() -> Result<()> {
         Workload::Version => {
             panic!("This case handled above");
         }
+        Workload::WFR => {
+            println!("Run Write-Flush-Read random IO test");
+            let count = {
+                if opt.count == 0 {
```
mkeeter, line 1068: Nit: it would be cleaner to store `count: Option<usize>`, then do `opt.count.unwrap_or(10)`

crutest/src/main.rs
```diff
@@ -1068,6 +1062,23 @@ async fn main() -> Result<()> {
...
+                    10
+                } else {
+                    opt.count
+                }
+            };
+            write_flush_read_workload(&guest, count, &mut region_info).await?;
+        }
+    }
+
+    if opt.verify_at_end {
```
mkeeter, line 1078: Would it make sense to have `verify_at_end` default to `true`? 
 
For example, `balloon_workload` used to unconditionally call `verify_volume`; that call has been removed, so we'll now only do the verification here if the flag is set.  Making the flag default to true would preserve our current behavior, but still let you turn it off with `--verify-at-end=false`

leftwo: I don't want all the tests to verify-at-end though, as that can add a bunch of run time to, like the perf tests or other larger tests.  

How about I go add the `--verify-at-end` for the one place in CI that used balloon to preserve that behavior and let any other test that wants it can add it.

mkeeter: Okay, that's fine by me.  It looks like `perf` is also used in CI (in `test_perf.sh`); up to you whether you want to keep previous behavior there as well.

mkeeter (approved):

crutest/src/main.rs
leftwo: Yeah, if perf was doing a final verify then I think that's actually the wrong behavior.  The perf test
should just be testing perf (and it's already rather long for CI).

I did just update the count -> Option suggestion you made.

PR #917, opened 2023-09-11
Don't pass `MutexGuard` into functions

mkeeter:
Small cleanup: `MutexGuard` implements `deref_mut` into a plain mutable reference, so we can use that instead.

Note that this could _theoretically_ change behavior in `submit_flush_internal`, which took the `MutexGuard` by value (i.e. dropping it at the end of the function).

Now, the `MutexGuard` is now dropped _outside_ of `submit_flush_internal`.  However, the call to `submit_flush_internal` is the last call in all parent function, so there's no actual behavior change.

All other functions changed in this PR were taking a `&mut MutexGuard`, which is harmless to change to a `&mut` reference.

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -5346,7 +5346,7 @@ impl Upstairs {
          * guest and downstairs lock at the same time.
          */
         let mut active = self.active.lock().await;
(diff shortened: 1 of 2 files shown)


leftwo (approved):
Yeah, that looks good.

PR #919, opened 2023-09-12
Use a newtype for `JobId`

mkeeter:
This is a huge amount of changed code, but the vast majority is boilerplate.

There's only one actual logic change: the notification queues are changed from sending a `u64` (job ID) to sending a `()`.  The value was never used; it's just used to wake the task.  While doing this refactoring, I realized that there were a bunch of cases that didn't actually _have_ a valid job ID, so we were just making one up (i.e. sending 0 or 1).

### DIFF
--- a/protocol/src/lib.rs
+++ b/protocol/src/lib.rs
@@ -17,7 +17,6 @@ use crucible_common::{Block, CrucibleError, RegionDefinition};
 ///
 /// A job ID is used to identify a specific job to the downstairs.  It is used
 /// in resolving dependencies.
-// TODO(matt) do we need to manually implement JsonSchema?
 #[derive(
     Copy,
     Clone,
@@ -296,8 +295,6 @@ pub enum Message {
      * this is reconciliation work and not actual IO.  The reconciliation work
      * uses a different work queue and each reconciliation job must finish on
      * all three downstairs before the next one can be sent.
-     *
-     * TODO(matt) is this comment still correct?
      */
     /// Send a close the given extent ID on the downstairs.
     ExtentClose {
--- a/protocol/src/lib.rs
+++ b/protocol/src/lib.rs
@@ -13,6 +13,36 @@ const MAX_FRM_LEN: usize = 100 * 1024 * 1024; // 100M
 
 use crucible_common::{Block, CrucibleError, RegionDefinition};
 
+/// Wrapper type for a job ID
+///
+/// A job ID is used to identify a specific job to the downstairs.  It is used
+/// in resolving dependencies.
+// TODO(matt) do we need to manually implement JsonSchema?
+#[derive(
+    Copy,
+    Clone,
+    Debug,
+    Eq,
+    Hash,
+    PartialEq,
+    Ord,
+    PartialOrd,
+    Serialize,
+    Deserialize,
+    schemars::JsonSchema,
+)]
+#[serde(transparent)]
+pub struct JobId(pub u64);
+
+impl std::fmt::Display for JobId {
+    fn fmt(
+        &self,
+        f: &mut std::fmt::Formatter<'_>,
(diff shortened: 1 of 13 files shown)


protocol/src/lib.rs
```diff
@@ -266,6 +296,8 @@ pub enum Message {
      * this is reconciliation work and not actual IO.  The reconciliation work
      * uses a different work queue and each reconciliation job must finish on
      * all three downstairs before the next one can be sent.
+     *
+     * TODO(matt) is this comment still correct?
```
leftwo, line 300: Yeah, it is still true.  The live repair is different than the initial reconciliation process that happens when we
first start up, though both make use of the same mechanism under the covers.

upstairs/src/lib.rs
```diff
@@ -9414,10 +9427,10 @@ struct Repair {
  * Send a message there is repair work to do.
  */
 pub struct Target {
-    ds_work_tx: mpsc::Sender<u64>,
-    ds_done_tx: mpsc::Sender<u64>,
-    ds_active_tx: watch::Sender<u64>,
-    ds_reconcile_work_tx: watch::Sender<u64>,
+    ds_work_tx: mpsc::Sender<u64>, // guest work ID?  lastcast?
```
leftwo, line 9430: The comment above this struct is out of date.
`ds_work_tx`: Used to indicate that new work has arrived for a downstairs.  
The value for the u64 here is not used, and this could probably be replaced with a doorbell like mechanism.  At some point in the past we may have actually been sending the `JobId` over the channel, but that has long since passed.

`ds_done_tx`:  This is used when the upstairs has decided that we can send a response to the guest that a JobID has completed.  
As you determined, the actual JobId is not used here, as the other side of the channel will figure it out.

`ds_active_tx`: Used to indicate to each of the downstairs that the upstairs has received an activation request, and each downstairs task should take action on that request.  
The value sent is the generation number, but, again, not used.

`ds_reconcile_work_tx`: Used to send notification to downstairs tasks that there is reconciliation work to do.  The
same basic idea as `ds_work_tx`.

leftwo (commented):
So, I wonder if this means we need to update the `CRUCIBLE_MESSAGE_VERSION`?

I think what is actually going over the wire between upstairs and downstiars is not
going to change (it's still a u64), right?  If that is different, then for sure we need
to bump the version here, but if not, should we?

This message version is used to detect a difference between upstairs and
downstairs messages and avoid a bad decoding of messages.

mkeeter:
> I think what is actually going over the wire between upstairs and downstiars is not
> going to change (it's still a u64), right? If that is different, then for sure we need
> to bump the version here, but if not, should we?

The intention is to make this transparent to the wire format, and that's why I used `#[serde(transparent)]` on the new type.  Without that annotation, this fails the OpenAPI checks, because the (JSON) encoding changes.

I have tested with old upstairs + new downstairs, and everything still works.

upstairs/src/lib.rs
mkeeter: Thanks, updated in 8fe2ad92e34d9c27126ffa17da5322a71456766e

At some point, I may go through and turn them all into `notify` (or `watch::Sender<()>`), but this PR touches enough stuff already.

protocol/src/lib.rs
mkeeter: Thanks, updated the comment.  At some point, I may add a `ReconciliationId` to make this _also_ strongly typed, but one thing at a time.

upstairs/src/lib.rs
leftwo: Yeah, feel free to create some issues and the rest of us can pick of stuff like this too.  You don't have to fix
everything by yourself :)
 But, agreed, too many unrelated changes can make future bisection difficult.

leftwo (approved):

PR #922, opened 2023-09-14
A few Crucible Agent fixes

jmpesp:
All these issues were found running the snapshot antagonist as part of testing for oxidecomputer/omicron#3866.

It is possible to call `delete_running_snapshot_request` for a snapshot that was in any state, and that function would Tombstone the running snapshot, even if it was already Destroyed! This would result in a cycle of Tombstoned -> Destroyed -> Tombstoned -> Destroyed ...  for as long as delete requests were being made. This commit changes that function to do the right thing: only Tombstone running snapshots that are in Requested or Created. Unit tests were added to validate some transitions.

The Crucible agent is very noisy: choose to use `debug` for messages that repeat a lot, and hide some messages behind conditionals.

Add more information to some log messages, and correctly identify region and running snapshot IDs.

### DIFF
--- a/agent/src/datafile.rs
+++ b/agent/src/datafile.rs
@@ -349,22 +349,49 @@ impl DataFile {
                 return Ok(());
             }
 
-            Some(request) => {
-                info!(
-                    self.log,
-                    "removing snapshot {}-{}", request.id.0, request.name
-                );
+            Some(existing) => {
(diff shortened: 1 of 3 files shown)


leftwo (approved):
concerns discussed during a live chat, all good here.

PR #926, opened 2023-09-15
Fast return for zero length IOs

leftwo:
If a read or write with length zero is submitted to either the Volume layer or via a Guest, we just return success without having to involve any of the layers below us.

Updated crutest to allow a zero length IO.
Added tests to verify a zero length IO via Volume or Guest does not return error or panic.

### DIFF
--- a/crutest/src/main.rs
+++ b/crutest/src/main.rs
@@ -1164,8 +1164,11 @@ fn fill_vec(
     wl: &WriteLog,
     bs: u64,
 ) -> Vec<u8> {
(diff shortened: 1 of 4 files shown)


leftwo:
Thinking about this a bit, a zero length IO might actually be an interesting performance
test case that executes almost everything except the final IO.  Almost like a loopback test
of sorts.

I think for a customer, we would not want to send a zero length IO all the way through the
system, but it might be an interesting test of some sort... maybe...

jmpesp (approved):

PR #928, opened 2023-09-18
Use a strong type for repair ID

mkeeter:
In the same vein as #925 and #919, this switches from a `u64` to a strong type for repair IDs.  Luckily, there's less to change here!

I'm ambivalent about the `RepairId` name, and wonder if we should use `ReconciliationId` instead (to more strongly distinguish from live repair, which also has the word "repair").  Opinions are welcome!

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -811,9 +811,7 @@ where
                         extent_id,
                         gen_number,
                         flush_number,
-                        // TODO(matt) this is a repair ID, not a job ID, but is
-                        // only used for logging, so ¯\_(ツ)_/¯
-                        JobId(repair_id.0),
+                        repair_id,
                     )
                     .await
                 {
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -813,7 +813,7 @@ where
                         flush_number,
                         // TODO(matt) this is a repair ID, not a job ID, but is
                         // only used for logging, so ¯\_(ツ)_/¯
-                        JobId(repair_id),
+                        JobId(repair_id.0),
                     )
                     .await
                 {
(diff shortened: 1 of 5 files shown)


leftwo:
`ReconciliationId` > `RepairId` for sure.  This "RepairID" came before we had LR, but changing it is what we should do.

leftwo:
> `ReconciliationId` > `RepairId` for sure. This "RepairID" came before we had LR, but changing it is what we should do.

To be clear,  `ReconciliationId` is better than, greater than, or, at least in my opinion, the preferred name. :)

mkeeter:
Okay, `s/RepairId/ReconciliationId/g` in ee016151af3d3ada055fcfbcd7fff324ff2aa521

downstairs/src/lib.rs
```diff
@@ -813,7 +813,7 @@ where
                         flush_number,
                         // TODO(matt) this is a repair ID, not a job ID, but is
                         // only used for logging, so ¯\_(ツ)_/¯
```
leftwo, line 815: Should we update this comment, and, or, why does only this one  location gets the update 
for`JobId(repair_id)` ?  There are `repair_id`'s for `Message::ExtentClose`, `Message::ExtentRepair`,
and `Message::ExtentReopen` but none of them need this translation?

jmpesp (approved):

downstairs/src/lib.rs
mkeeter: `Region::reopen_extent`, `Region::repair_extent`, and `Region::close_extent` all don't take a `JobId` argument; `Region::region_flush_extent` is unique in that respect!

(`Region::region_flush_extent` then calls into `Extent::flush`, passing that `JobId`, which is used as an ID in DTrace probes)

mkeeter: Alright, What do you think about 73b141ab7b7a04fdbb496ee89b13d0ed2df58fd7 ?  It's still strongly typed, but no longer relies on punning the inner `u64`.

PR #931, opened 2023-09-19
Add a "dynamometer" option to crucible-downstairs

jmpesp:
Before measuring the performance impact of a new on-disk format, add a dynamometer option: measure the performance of sending writes to a Region in isolation from the rest of Crucible.

### DIFF
--- a/downstairs/src/dynamometer.rs
+++ b/downstairs/src/dynamometer.rs
@@ -28,18 +28,25 @@ pub async fn dynamometer(
     let mut blocks_since_last_flush = 0;
 
     let ddef = region.def();
+    eprintln!("{:?}", ddef);
 
     // Fill test: write bytes in whole region
-    let block = vec![0x1; ddef.block_size() as usize];
-    let nonce =
-        vec![0x0, 0x1, 0x2, 0x3, 0x4, 0x5, 0x6, 0x7, 0x8, 0x9, 0xa, 0xb];
-    let tag = vec![
-        0x0, 0x1, 0x2, 0x3, 0x4, 0x5, 0x6, 0x7, 0x8, 0x9, 0xa, 0xb, 0xc, 0xd,
-        0xe, 0xf,
-    ];
-    let hash = integrity_hash(&[&nonce, &tag, &block]);
+
+    let mut rng = SmallRng::from_entropy();
 
     'outer: loop {
+        let block = (0..ddef.block_size() as usize)
+            .map(|_| rng.sample(rand::distributions::Standard))
+            .collect::<Vec<u8>>();
+        let nonce = (0..12)
+            .map(|_| rng.sample(rand::distributions::Standard))
+            .collect::<Vec<u8>>();
+        let tag = (0..16)
+            .map(|_| rng.sample(rand::distributions::Standard))
+            .collect::<Vec<u8>>();
+
+        let hash = integrity_hash(&[&nonce, &tag, &block]);
+
         for eid in 0..ddef.extent_count() {
             let mut block_offset = 0;
             loop {
--- a/downstairs/src/dynamometer.rs
+++ b/downstairs/src/dynamometer.rs
@@ -47,6 +47,8 @@ pub async fn dynamometer(
 
         let hash = integrity_hash(&[&nonce, &tag, &block]);
 
+        let block_bytes = bytes::Bytes::from(block.clone());
+
         for eid in 0..ddef.extent_count() {
             let mut block_offset = 0;
             loop {
@@ -54,7 +56,6 @@ pub async fn dynamometer(
                     break;
                 }
 
-                let block = block.clone();
                 let nonce = nonce.clone();
                 let tag = tag.clone();
 
@@ -65,13 +66,13 @@ pub async fn dynamometer(
                             i as u64 + block_offset,
                             &ddef,
                         ),
-                        data: bytes::Bytes::from(block.clone()),
+                        data: block_bytes.clone(),
                         block_context: BlockContext {
                             hash,
                             encryption_context: Some(
                                 crucible_protocol::EncryptionContext {
-                                    nonce: nonce.clone(),
-                                    tag: tag.clone(),
+                                    nonce: nonce.as_slice().try_into().unwrap(),
+                                    tag: tag.as_slice().try_into().unwrap(),
                                 },
                             ),
                         },
@@ -82,8 +83,8 @@ pub async fn dynamometer(
                 region.region_write(&writes, JobId(1000), false).await?;
 
                 total_io_time += io_operation_time.elapsed();
-                io_operations_sent += 1;
-                iops_since_last_flush += 1;
+                io_operations_sent += num_writes;
+                iops_since_last_flush += num_writes;
                 blocks_since_last_flush += num_writes;
                 bw_consumed += num_writes * ddef.block_size() as usize;
 
--- a/downstairs/src/dynamometer.rs
+++ b/downstairs/src/dynamometer.rs
@@ -1,10 +1,17 @@
 // Copyright 2023 Oxide Computer Company
 use super::*;
 
+pub enum DynoFlushConfig {
+    FlushPerIops(usize),
+    FlushPerBlocks(usize),
+    FlushPerMs(usize),
+}
+
 pub async fn dynamometer(
     mut region: Region,
     num_writes: usize,
     samples: usize,
+    flush_config: DynoFlushConfig,
 ) -> Result<()> {
     // TODO: pull into another crate? this is copied from measure-iops tool
     let mut io_operations_sent = 0;
@@ -13,6 +20,12 @@ pub async fn dynamometer(
     let mut total_io_time = Duration::ZERO;
     let mut iops: Vec<f32> = vec![];
     let mut bws: Vec<f32> = vec![];
+
+    let mut flush_number = 0;
+    let mut gen_number = 0;
+    let mut flush_time = Instant::now();
+    let mut iops_since_last_flush = 0;
+    let mut blocks_since_last_flush = 0;
 
     let ddef = region.def();
 
@@ -62,7 +75,9 @@ pub async fn dynamometer(
                 region.region_write(&writes, JobId(1000), false).await?;
 
                 total_io_time += io_operation_time.elapsed();
-                io_operations_sent += num_writes;
+                io_operations_sent += 1;
+                iops_since_last_flush += 1;
+                blocks_since_last_flush += num_writes;
                 bw_consumed += num_writes * ddef.block_size() as usize;
 
                 if measurement_time.elapsed() > Duration::from_secs(1) {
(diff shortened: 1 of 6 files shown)


jmpesp:
This is in draft because it doesn't include a random write test yet

downstairs/src/dynamometer.rs
```diff
@@ -0,0 +1,204 @@
...
+                let tag = tag.clone();
+
+                let writes: Vec<_> = (0..num_writes)
+                    .map(|i| crucible_protocol::Write {
+                        eid: eid as u64,
+                        offset: Block::new_with_ddef(
+                            i as u64 + block_offset,
+                            &ddef,
+                        ),
+                        data: bytes::Bytes::from(block.clone()),
```
mkeeter, line 68: This adds an extra clone of the full data; I think it would be faster if we constructed the `Bytes` object outside the loop (since it's reference-counted and cheaper to clone here).

downstairs/src/dynamometer.rs
```diff
@@ -0,0 +1,204 @@
...
+                                    None, // extent_limit
+                                )
+                                .await?;
+
+                            flush_number += 1;
+                            gen_number += 1;
+                            flush_time = Instant::now();
+                        }
+                    }
+                }
```
mkeeter, line 164: How about factoring out the actual flush, e.g.
```suggestion
                let needs_flush = match flush_config {
                    DynoFlushConfig::FlushPerIops(value)
                        if iops_since_last_flush > value =>
                    {
                        iops_since_last_flush = 0;
                        true
                    }

                    DynoFlushConfig::FlushPerBlocks(value)
                        if blocks_since_last_flush > value =>
                    {
                        blocks_since_last_flush = 0;
                        true
                    }

                    DynoFlushConfig::FlushPerMs(value)
                        if flush_time.elapsed()
                            > Duration::from_millis(value as u64) =>
                    {
                        flush_time = Instant::now();
                        true
                    }
                    _ => false,
                };
                if needs_flush {
                    region
                        .region_flush(
                            flush_number,
                            gen_number,
                            &None, // snapshot_details
                            JobId(1000),
                            None, // extent_limit
                        )
                        .await?;

                    flush_number += 1;
                    gen_number += 1;
                }
```

downstairs/src/main.rs
```diff
@@ -182,6 +182,45 @@ enum Args {
...
+        #[clap(long, action, default_value = None)]
+        flush_per_iops: Option<usize>,
+
+        // Flush per blocks written
+        #[clap(long, action, default_value = None)]
+        flush_per_blocks: Option<usize>,
+
+        // Flush per ms
+        #[clap(long, action, default_value = None)]
+        flush_per_ms: Option<usize>,
```
mkeeter, line 222: It would be neat to store a `Duration` here, so that it's strongly typed by the time it's parsed.  This requires a custom value parser, see [this discussion](https://stackoverflow.com/a/72314001).

downstairs/src/main.rs
```diff
@@ -386,5 +425,64 @@ async fn main() -> Result<()> {
...
+                log.clone(),
+            )
+            .await?;
+
+            eprintln!(
+                "{:?} {:?} {:?}",
+                flush_per_iops, flush_per_blocks, flush_per_ms
+            );
+
+            let mut flush_types_set = 0;
```
mkeeter, line 458: You could use `conflicts_with` / `conflicts_with_all` to enforce this check in clap, e.g. 
```rust
        #[clap(long, action, default_value = None,
               conflicts_with_all = ["flush_per_blocks", "flush_per_ms"])]
        flush_per_iops: Option<usize>,
```

downstairs/src/main.rs
```diff
@@ -182,6 +182,45 @@ enum Args {
...
+        extent_size: u64,
+
+        #[clap(long, default_value = "15", action)]
+        extent_count: u64,
+
+        #[clap(long, action)]
+        encrypted: bool,
+
+        // Number of writes to submit at one time to region_write
+        #[clap(short, long, action, default_value = "1")]
```
mkeeter, line 205: Nit: you can use `default_value_t = 1` instead, which makes the default value typed at compile-time.

mkeeter:
You need to add `"small_rng"` to the `rand` line in `Cargo.toml`:
```diff
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,7 @@ opentelemetry = "0.20.0"
 opentelemetry-jaeger = { version = "0.17.0" }
 percent-encoding = "2.3"
 proptest = "1.2.0"
-rand = { version = "0.8.5", features = ["min_const_gen"] }
+rand = { version = "0.8.5", features = ["min_const_gen", "small_rng"] }
 rand_chacha = "0.3.1"
 reedline = "0.23.0"
 reqwest = { version = "0.11", features = ["default", "blocking", "json", "stream"] }
```
Otherwise, it won't build if you try building `crucible-downstairs` alone:
```
➜  crucible git:(dynamometer) cargo run -pcrucible-downstairs -- -h
   Compiling crucible-downstairs v0.0.1 (/Users/mjk/oxide/crucible/downstairs)
error[E0433]: failed to resolve: could not find `SmallRng` in the crate root
  --> downstairs/src/dynamometer.rs:35:26
   |
35 |     let mut rng = crate::SmallRng::from_entropy();
   |                          ^^^^^^^^ could not find `SmallRng` in the crate root

For more information about this error, try `rustc --explain E0433`.
error: could not compile `crucible-downstairs` (lib) due to previous error
```
(it works with other crates, due to feature unification)

downstairs/src/dynamometer.rs
```diff
@@ -0,0 +1,204 @@
+// Copyright 2023 Oxide Computer Company
+use super::*;
+
+pub enum DynoFlushConfig {
+    FlushPerIops(usize),
+    FlushPerBlocks(usize),
+    FlushPerMs(usize),
```
mkeeter, line 7: To make this more strongly typed, we could do `FlushesPerDuration(std::time::Duration)`

downstairs/src/main.rs
```diff
@@ -182,6 +182,45 @@ enum Args {
...
+        #[clap(long, default_value = "100", action)]
+        extent_size: u64,
+
+        #[clap(long, default_value = "15", action)]
+        extent_count: u64,
+
+        #[clap(long, action)]
+        encrypted: bool,
+
+        // Number of writes to submit at one time to region_write
```
mkeeter, line 204: If you make these into doc comments (`///`) instead of regular comments, they will show up in the CLI with the `--help` option:
```
Options:
      --block-size <BLOCK_SIZE>
          [default: 512]
  -d, --data <DIRECTORY>

      --extent-size <EXTENT_SIZE>
          [default: 100]
      --extent-count <EXTENT_COUNT>
          [default: 15]
      --encrypted

  -n, --num-writes <NUM_WRITES>
          Number of writes to submit at one time to region_write [default: 1]
```

downstairs/src/main.rs
```diff
@@ -386,5 +425,64 @@ async fn main() -> Result<()> {
...
+            }
+
+            let flush_config = if flush_per_iops.is_some() {
+                DynoFlushConfig::FlushPerIops(flush_per_iops.unwrap())
+            } else if flush_per_blocks.is_some() {
+                DynoFlushConfig::FlushPerBlocks(flush_per_iops.unwrap())
+            } else if flush_per_ms.is_some() {
+                DynoFlushConfig::FlushPerMs(flush_per_ms.unwrap())
+            } else {
+                panic!("wat");
```
mkeeter, line 480: This is easily reachable by users:
```
cargo run -pcrucible-downstairs -- dynamometer --data tmp
    Finished dev [unoptimized + debuginfo] target(s) in 0.28s
     Running `target/debug/crucible-downstairs dynamometer --data tmp`
Sep 20 14:20:56.483 INFO raising number of open files limit to from 256 to 9223372036854775807
Sep 20 14:20:56.484 INFO Created new region file "tmp/region.json"
None None None
thread 'main' panicked at 'wat', downstairs/src/main.rs:480:17
fish: Job 1, 'cargo run -pcrucible-downstairs…' terminated by signal SIGABRT (Abort)
```

How about something like `bail!("must specify a flush interval");`?

downstairs/src/main.rs
```diff
@@ -182,6 +182,45 @@ enum Args {
         bind_addr: SocketAddr,
     },
     Version,
+    /*
```
mkeeter, line 185: If this is a doc comment (`/// Measure an isolated downstairs`), it shows up in clap's help:
```
Usage: crucible-downstairs <COMMAND>

Commands:
  create
  dump
  export
  run
  repair-api
  serve
  version
  dynamometer  Measure an isolated downstairs
  help         Print this message or the help of the given subcommand(s)
```

downstairs/src/dynamometer.rs
jmpesp: :+1: moved out in 46d7d49

downstairs/src/dynamometer.rs
jmpesp: this is really neat, done in 57854da

downstairs/src/main.rs
jmpesp: see e9f076d

downstairs/src/main.rs
jmpesp: added in 5456783

downstairs/src/main.rs
```diff
@@ -182,6 +182,45 @@ enum Args {
...
+        // Number of writes to submit at one time to region_write
+        #[clap(short, long, action, default_value = "1")]
+        num_writes: usize,
+
+        // Number of samples to exit for
+        #[clap(short, long, action, default_value = "10")]
+        samples: usize,
+
+        // Flush per iops
+        #[clap(long, action, default_value = None)]
```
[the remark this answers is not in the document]
jmpesp: removed in e9f076d3de969cefdea895ca04944cea2f62f864

downstairs/src/main.rs
jmpesp: nice, didn't know about that option! f1c3144

downstairs/src/dynamometer.rs
jmpesp: done in e9f076d3de969cefdea895ca04944cea2f62f864

downstairs/src/main.rs
jmpesp: nice, 3cb447e

downstairs/src/main.rs
jmpesp: I added a `None` variant to test without flushing at all.

downstairs/src/main.rs
jmpesp: see 3cb447e

jmpesp:
> This is in draft because it doesn't include a random write test yet

I removed this TODO and marked this as ready for review. Thanks for the :100: comments Matt :)

mkeeter (approved):

faithanalog (approved):

downstairs/src/main.rs
```diff
@@ -182,6 +183,48 @@ enum Args {
         bind_addr: SocketAddr,
     },
     Version,
+    /// Measure an isolated downstairs
+    Dynamometer {
+        #[clap(long, default_value_t = 512, action)]
```
mkeeter, line 188: Nit: I think the `action` here and below is unnecessary

PR #940, opened 2023-09-22
Measure network round trip time

jmpesp:
Create a network round trip dynamometer. Also, fix how the downstairs dynamometer measures IOPs.


protocol/src/main.rs
```diff
@@ -0,0 +1,234 @@
...
+        #[clap(short, long, default_value_t = 1)]
+        requests_per_write: usize,
+
+        /// Number of samples to exit for
+        #[clap(short, long, default_value_t = 10)]
+        samples: usize,
+    },
+}
+
+#[tokio::main]
```
leftwo, line 46: I don't love that the propocol lib is now a binary.  It might be a little confusing
to someone coming in.  Could we maybe rename the resulting binary `protocol_test` or
`ptest` or something to indicate that?

An additional idea, maybe a README somewhere talking about it.  I've been negligent
as well in documenting what all the stuff in the crucible repo is, and we should start
getting better about leaving breadcrumbs behind to assist new readers on what the
heck all this stuff is.

protocol/src/main.rs
```diff
@@ -0,0 +1,234 @@
...
+use tokio::net::TcpSocket;
+use tokio_util::codec::FramedRead;
+use tokio_util::codec::FramedWrite;
+use uuid::Uuid;
+
+#[derive(Debug, Parser)]
+#[clap(about = "Protocol serialization and deserialization speed test")]
+enum Args {
+    Dynamometer {
+        #[clap(short, long, default_value_t = 512)]
```
leftwo, line 29: This ones missing the doc comment.

protocol/src/main.rs
```diff
@@ -0,0 +1,234 @@
...
+use futures::SinkExt;
+use futures::StreamExt;
+use tokio::net::TcpListener;
+use tokio::net::TcpSocket;
+use tokio_util::codec::FramedRead;
+use tokio_util::codec::FramedWrite;
+use uuid::Uuid;
+
+#[derive(Debug, Parser)]
+#[clap(about = "Protocol serialization and deserialization speed test")]
```
leftwo, line 26: This is really just testing writes.
Would read IO be any different, or is it just the same thing but going in the
other direction?

protocol/src/main.rs
mkeeter: Agreed, I'd even suggest a separate binary crate so that `protocol` remains a pure library crate.

downstairs/src/dynamometer.rs
```diff
@@ -8,6 +8,12 @@ pub enum DynoFlushConfig {
     None,
 }
 
+macro_rules! ceiling_div {
```
mkeeter, line 11: Hopefully [`div_ceil`](https://doc.rust-lang.org/std/primitive.i32.html#method.div_ceil) gets stabilized soon!

downstairs/src/dynamometer.rs
```diff
@@ -84,7 +90,10 @@ pub async fn dynamometer(
                 region.region_write(&writes, JobId(1000), false).await?;
 
                 total_io_time += io_operation_time.elapsed();
-                io_operations_sent += num_writes;
+                io_operations_sent += ceiling_div!(
+                    num_writes * ddef.block_size() as usize,
+                    16 * 1024 * 1024
```
mkeeter, line 95: Can you explain where this number comes from?  It's not obvious to me!

protocol/src/main.rs
mkeeter: The doc comment should also specify the unit (it looks like bytes)

protocol/src/main.rs
```diff
@@ -0,0 +1,234 @@
...
+
+                None => panic!("disconnected!"),
+            }
+
+            total_io_time += io_operation_time.elapsed();
+        }
+
+        io_operations_sent += num_writes;
+        io_operations_sent +=
+            ceiling_div!(num_writes * write_size, 16 * 1024 * 1024);
```
mkeeter, line 182: Same here, not sure where 16 MiB comes from

PR #941, opened 2023-09-23
Use a strong type for upstairs negotiation

mkeeter:
This should be equivalent, but is easier to read.

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -1162,7 +1162,9 @@ where
      * At this point the upstairs looks to see what state the downstairs is
      * currently in.  It will be WaitActive, Faulted, or Offline.
      *
-     * Depending on which state, we will either choose 3 or 4 next.
+     * Depending on which state, we will either choose
+     * NegotiationState::GetLastFlush or NegotiationState::GetExtentVersions
+     * next.
      *
      * For the Offline state, the downstairs was connected and verified
      * and some point after that, the connection was lost.  To handle this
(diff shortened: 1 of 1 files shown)


upstairs/src/lib.rs
```diff
@@ -1162,28 +1178,30 @@ where
...
-     *
+     * NegotiationState::GetLastFlush (offline only)
+     * ------------------------------
      *          Upstairs             Downstairs
-     * 3:       LastFlush(lf)) --->
+     *          LastFlush(lf)) --->
      *                         <---  LastFlushAck(lf)
      *
      * After receiving our last flush, we now move this downstairs state to
      * Replay and skip ahead to step 5.
```
leftwo, line 1188: Step 5 is now `NegotiationState::Done`

upstairs/src/lib.rs
```diff
@@ -1162,28 +1178,30 @@ where
      * also request extent versions and will have to repair this
      * downstairs, skipping over step 3 as well.
      *
-     * (Offline only):
-     *
+     * NegotiationState::GetLastFlush (offline only)
```
leftwo, line 1181: The comment above here should be updated to replace `3` and `4` with 
`NegotiationState::GetLastFlush` and `NegotiationState::GetExtentVersions`

leftwo (approved):
Just some comment cleanup

upstairs/src/lib.rs
mkeeter: Done in 97dfe4778518d3536ecc2adb75bf2b9672296416

upstairs/src/lib.rs
mkeeter: Done in 97dfe4778518d3536ecc2adb75bf2b9672296416

PR #943, opened 2023-09-23
Remove unused async

mkeeter:
This PR enables the `clippy::unused_async` lint, then fixes everything it complains about.

There should be no performance or correctness change, because the functions being edited here – by definition – do not contain a yield point.

There's also a drive-by change of adding `#[derive(Clone)]` to `ExtentInfo`, since it's POD and this removes some boilerplate.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -1,6 +1,7 @@
 // Copyright 2023 Oxide Computer Company
 #![cfg_attr(usdt_need_asm, feature(asm))]
 #![cfg_attr(all(target_os = "macos", usdt_need_asm_sym), feature(asm_sym))]
+#![warn(clippy::unused_async)]
 
 use futures::executor;
 use futures::lock::{Mutex, MutexGuard};
@@ -1981,7 +1982,7 @@ impl Downstairs {
     ) -> Result<Option<Message>> {
         let job = {
             let mut work = self.work_lock(upstairs_connection).await?;
-            let job = work.get_ready_job(job_id).await;
+            let job = work.get_ready_job(job_id);
 
             // `promote_to_active` can clear out the Work struct for this
             // UpstairsConnection, but the tasks can still be working on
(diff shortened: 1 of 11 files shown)


downstairs/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 // Copyright 2023 Oxide Computer Company
 #![cfg_attr(usdt_need_asm, feature(asm))]
 #![cfg_attr(all(target_os = "macos", usdt_need_asm_sym), feature(asm_sym))]
+#![warn(clippy::unused_async)]
```
leftwo, line 4: Instead of updating all the files directly, should we just put this into the CI script:

```
local:pflush$ git diff .github/workflows/rust.yml 
diff --git a/.github/workflows/rust.yml b/.github/workflows/rust.yml
index 65127d7..e278085 100644
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -45,4 +45,4 @@ jobs:
     - uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4
     - uses: dtolnay/rust-toolchain@1.66
     - name: Test Libraries
-      run: cargo clippy --all-targets
+      run: cargo clippy --all-targets -- -W clippy::unused_async
```

Then, we won't forget to add it to other files and async won't sneak back in?

leftwo (approved):

downstairs/src/lib.rs
mkeeter: If I used this strategy, it wouldn't fail locally (either running `cargo clippy` or through IDE hints).

I've instead moved this to `.cargo/config`, which enables it for the entire workspace (and should work for CI too).  Once https://github.com/rust-lang/cargo/issues/12115 is released and we bump the toolchain, we can use that instead.

leftwo: Having a local config file was what I actually wanted, I just did not find that solution in my
limited search.

And, if this runs in CI too, then we don't need to update the `.github/workflows/rust.yml ` file
to include it.

PR #944, opened 2023-09-23
Use atomics for flush and generation state

mkeeter:
This saves a few sync points and should have no correctness impact.


leftwo (approved):
Yeah, this makes sense.  An atomic here is resulting in no difference in behavior.

PR #947, opened 2023-09-23
Cache block size in `Guest`

mkeeter:
Storing the value locally saves us a round trip through the `reqs` queue when calling `BlockIO::read` and `BlockIO::write` from a `Guest`.  This is a hot function, although it may not be the bottleneck right now...

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -9028,6 +9028,13 @@ pub struct Guest {
      */
     bw_tokens: std::sync::Mutex<usize>, // bytes
     bw_limit: Option<usize>,            // bytes per second
+
+    /// Local cache for block size
+    ///
+    /// This is 0 when unpopulated, and non-zero otherwise; storing it locally
+    /// saves a round-trip through the `reqs` queue, and using an atomic means
+    /// it can be read from a `&self` reference.
+    block_size: AtomicU64,
 }
 
 /*
@@ -9060,6 +9067,8 @@ impl Guest {
 
             bw_tokens: std::sync::Mutex::new(0),
             bw_limit: None,
+
+            block_size: AtomicU64::new(0),
         }
     }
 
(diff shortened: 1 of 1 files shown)


upstairs/src/lib.rs
```diff
@@ -9029,6 +9029,13 @@ pub struct Guest {
      */
     bw_tokens: Mutex<usize>, // bytes
     bw_limit: Option<usize>, // bytes per second
+
```
mkeeter, line 9032: I'm sure that Alan's inconsistent comment style alarm is going off here 🚨

I'd like to use proper doc comments, because they are highlighted differently in my editor:

Screenshot 2023-09-23 at 11 20 40 AM

Should I go through and edit the rest of this struct, so it's locally consistent?  Alternatively, I could use `/**` comments here, which are also doc comments (but not recommended by the Rust style guide).

leftwo: 😱

leftwo: Okay..
Here are my thoughts.

Change is coming.  The tide is rising.
I think updating just the comment for the field you change in a struct with other comment styles is
the thing to do here.

If/when we make a mass conversion, I think doing it all at once, with one single change to all files with
only changes to comments could be the way to do it.  The idea being that we don't mask comment 
changes with other changes when looking back through git history.  This is just my idea and we should
discuss it further.

`The measure of intelligence is the ability to change` -Albert Einstein
`The secret of change is to focus all of your energy not on fighting the old, but on building the new` -Socrates
`Our dilemma is that we hate change and love it at the same time; what we really want is for things to remain the same but get better.` -Sydney J. Harris
`Change is the law of life, and those who look only to the past and present are certain to miss the future` -John F. Kennedy

upstairs/src/lib.rs
```diff
@@ -9029,6 +9029,13 @@ pub struct Guest {
...
+    /// Local cache for block size
+    ///
+    /// This is 0 when unpopulated, and non-zero otherwise; storing it locally
+    /// saves a round-trip through the `reqs` queue, and using an atomic means
```
leftwo, line 9036: Is it eaiser/better to have 0 be a sentinel value vs. `Option`?
I'm not sure if the `Option` has the same ergonomics benefit.

mkeeter: You're correct that an `Option` would be more ergonomic / type-safe here.  However, we're relying on atomic operations to make this callable from `&self` functions, and `Option` functions aren't atomic.

PR #957, opened 2023-09-25
Use a fixed size for tag and nonce

mkeeter:
This is a building block for moving context out of SQLite and into flat files, but it's a reasonable enough standalone change that we can PR it now.

### DIFF
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -194,8 +194,8 @@ impl Inner {
             stmt.query_map(params![block, block + count - 1], |row| {
                 let block_index: u64 = row.get(0)?;
                 let hash: i64 = row.get(1)?;
-                let nonce: Option<Vec<u8>> = row.get(2)?;
-                let tag: Option<Vec<u8>> = row.get(3)?;
+                let nonce: Option<[u8; 12]> = row.get(2)?;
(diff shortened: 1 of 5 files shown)


downstairs/src/region.rs
```diff
@@ -189,8 +189,8 @@ impl Inner {
             stmt.query_map(params![block, block + count - 1], |row| {
                 let block_index: u64 = row.get(0)?;
                 let hash: i64 = row.get(1)?;
-                let nonce: Option<Vec<u8>> = row.get(2)?;
-                let tag: Option<Vec<u8>> = row.get(3)?;
+                let nonce: Option<[u8; 12]> = row.get(2)?;
```
leftwo, line 192: Is this a database format change?  Like, can we read a region created with
the old code?

mkeeter: It is backwards compatible, unless someone has deliberately messed up the metadb (which would also break `main`).

To test this, I created regions on `main` with `dsc create  --cleanup --ds-bin=target/release/crucible-downstairs --region-dir=tmp`, then ran `crutest` on `main` to populate some data.  After switching back to this branch, `dsc start` and `crutest` both worked fine.

jmpesp (approved):

PR #958, opened 2023-09-25
Make `reqs` queue synchronous

mkeeter:
Similar to previous PRs, this lock is held for a short time and isn't held across await points, so it can use a `std::sync::Mutex` for efficiency.  (The [example in `tokio::sync::Notify`](https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html) implements a very similar queue using the standard library mutex)

I don't see performance impacts in either direction, running some tests using `crutest` and `pgbench` on a VM.

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -8996,7 +8996,7 @@ pub struct Guest {
      * the submission task tells the listening task that new work has been
      * added.
      */
-    reqs: Mutex<VecDeque<BlockReq>>,
+    reqs: std::sync::Mutex<VecDeque<BlockReq>>,
     notify: Notify,
 
     /*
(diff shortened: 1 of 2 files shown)


mkeeter:
The [failed `test-replay`](https://buildomat.eng.oxide.computer/wg/0/details/01HB6DXV8CYV6NW0KVN8M20CD4/Wjm0Dlnw1z768K1326NxPXNk1wkC5tnG6qBiJ1OyLzNwWzBg/01HB6DYQAF9T849B8V0K6R8QRA) doesn't include many details, unfortunately.  I wonder if there's the possibility for a deadlock when using a synchronous mutex?

leftwo:
Does this need a merge with latest main?  I think James changed some stuff in this same area?

mkeeter:
Closing this in favor of https://github.com/oxidecomputer/crucible/pull/1086, which is a nicer architecture.

Issue #973, opened 2023-09-30
log that an upstairs is read only


leftwo:
If an upstairs is connecting to a read-only downstairs, we should log that somewhere obvious
early in upstairs startup.

Issue #971, opened 2023-09-29
Crucible should print out starting opts


leftwo:
We don't print the VCR any longer in propolis, so crucible upstairs should
report, in a single easy to find log message, what it's opts are when starting.
Don't print the key value though, that should remain a secret.

leftwo:
Ideally, we should also print session in each log line?
Or, improve on propolis side as to make it easier to tell the
difference between a ROP volume and the main volume.

gjcolombo:
Related: oxidecomputer/propolis#536

leftwo:
#973

Issue #486, opened 2022-10-17
Don't print the key in the logs


leftwo:
We should not print the encryption key in the logs:

```
Oct 14 01:44:40.676 INFO Creating Crucible disk from request Volume { id: 1c84b66a-2282-4a90-993b-a2dbd1e9a000, block_size: 512, sub_volumes: [Region { block_size: 512, opts: CrucibleOpts { id: 1c84b66a-2282-4a90-993b-a2dbd1e9a000, target: [[fd00:1122:3344:101::b]:19000, [fd00:1122:3344:101::a]:19000, [fd00:1122:3344:101::7]:19000], lossy: false, flush_timeout: None, key: Some("Yj9fL9Za4xmiZkTydYdm5xqOPn4aPSPk+AfSGAG/qvM="), cert_pem: None, key_pem: None, root_cert_pem: None, control: None, read_only: false }, gen: 0 }], read_only_parent: Some(Url { id: 4bd0d3fa-0516-4967-af88-b644eb9ca950, block_size: 512, url: "http://[fd00:1122:3344:101::1]:54321/media/cloud/focal-server-cloudimg-amd64.raw" }) }
```

jclulow:
Perhaps we should have a newtype for holding it that has Display and Debug implementations that just say something opaque like `[encryption key]`, to avoid any such accidents in future?

andrewjstone:
Keys should also be wrapped with [zeroize](https://docs.rs/zeroize/latest/zeroize/) if they aren't already. You may want to use [secrecy](https://docs.rs/secrecy/latest/secrecy/) instead, which al

jclulow:
> Keys should also be wrapped with [zeroize](https://docs.rs/zeroize/latest/zeroize/) if they aren't already. You may want to use [secrecy](https://docs.rs/secrecy/latest/secrecy/) instead, which also prevents debug logging. I haven't used it myself but came across it reading some code the other day.

We should figure out how to circulate that guidance more broadly I suspect?

leftwo:
> > Keys should also be wrapped with [zeroize](https://docs.rs/zeroize/latest/zeroize/) if they aren't already. You may want to use [secrecy](https://docs.rs/secrecy/latest/secrecy/) instead, which also prevents debug logging. I haven't used it myself but came across it reading some code the other day.
> 
> We should figure out how to circulate that guidance more broadly I suspect?

CP huddle issue created!

jmpesp:
[...]

PR #974, opened 2023-09-30
Log crucible opts on start, shuffle crutest structs

leftwo:
Make crucible upstairs log what options it receives on startup. Don't actually print keys though, just print if they are present.

Updated crutest to have a read-only option to pass to crucible and put all the crutest options in abc order.

### DIFF
--- a/crucible-client-types/src/lib.rs
+++ b/crucible-client-types/src/lib.rs
@@ -81,15 +81,16 @@ impl std::fmt::Display for CrucibleOpts {
         write!(f, " Targets: {:?},", self.target)?;
         write!(f, " lossy: {:?},", self.lossy)?;
         write!(f, " flush_timeout: {:?},", self.flush_timeout)?;
-        write!(f, " read_only: {:?},", self.read_only)?;
         write!(f, " key populated: {}, ", self.key.is_some())?;
         write!(f, " cert_pem populated: {}, ", self.cert_pem.is_some())?;
         write!(f, " key_pem populated: {}, ", self.key_pem.is_some())?;
         write!(
             f,
-            " root_cert_pem populated: {}",
+            " root_cert_pem populated: {}, ",
             self.root_cert_pem.is_some()
         )?;
+        write!(f, " Control: {:?}, ", self.control)?;
+        write!(f, " read_only: {:?}", self.read_only)?;
         Ok(())
     }
 }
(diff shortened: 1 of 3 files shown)


crutest/src/main.rs
```diff
@@ -124,26 +124,37 @@ enum Workload {
 #[clap(name = "client", term_width = 80)]
 #[clap(about = "A Crucible upstairs test client", long_about = None)]
 pub struct Opt {
+    // TLS options
```
leftwo, line 127: These options were put in alphabetical order.

The only new option here is the read_only.  The rest are just moved around.

crutest/src/main.rs
```diff
@@ -162,27 +186,45 @@ pub struct Opt {
...
-
     /// For the verify test, if this option is included we will allow
     /// the write log range of data to pass the verify_volume check.
     #[clap(long, global = true, action)]
     range: bool,
 
+    /// Set the read_only option when starting the upstairs.
+    /// Note that setting this won't prevent you from sending writes to the
+    /// downstairs.  You are responsible for dealing with the fallout.
+    #[clap(long, global = true, action)]
```
leftwo, line 197: This is the new option, all the others were already present.

leftwo:
https://github.com/oxidecomputer/propolis/issues/536
and
https://github.com/oxidecomputer/crucible/issues/973
and
https://github.com/oxidecomputer/crucible/issues/971
and
https://github.com/oxidecomputer/crucible/issues/486

crucible-client-types/src/lib.rs
```diff
@@ -71,3 +72,24 @@ impl CrucibleOpts {
...
+
+/// Display the contents of CrucibleOpts, Only print if keys are populated,
+/// not what the actual contents are.
+impl std::fmt::Display for CrucibleOpts {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Upstairs UUID: {},", self.id)?;
+        write!(f, " Targets: {:?},", self.target)?;
+        write!(f, " lossy: {:?},", self.lossy)?;
+        write!(f, " flush_timeout: {:?},", self.flush_timeout)?;
+        write!(f, " read_only: {:?},", self.read_only)?;
```
gjcolombo, line 84: Is `control` of interest here?

gjcolombo (approved):

crucible-client-types/src/lib.rs
leftwo: It's not.. I think I'm about to make it always on.
But, why not print it till then, so yeah, I'll add it.

Issue #923, opened 2023-09-14
Block size is stored in two places in `RegionDefinition`


mkeeter:
In a `RegionDefinition`, the block size is stored both in `block_size` (as bytes)` and in `extent_size.shift` (as a shift).  We don't ensure that they match.  `block_size` is used for `total_size()`; `extent_size.shift` is used in `extent_from_offset(..)`.  We should remove this duplication, either by making `extent_size` a single `u64` (instead of a `Block`) or removing `block_size`.

mkeeter:
Be warned: this is also written to `region.json`, and will be part of the network-facing API once #1129 goes in, so it's not trivial to change!

PR #982, opened 2023-10-03
Add trait for `ExtentInner`

mkeeter:
This PR is preparation for an alternate SQLite-free extent file format.

It does two things:
- Splits extent info from `region.rs` into a new `extent.rs`
- Makes the extent `Inner` type generic across a new `ExtentInner` trait, then implements that trait for the existing SQLite backend

There should be no change to functionality here!


downstairs/src/extent.rs
```diff
@@ -0,0 +1,829 @@
...
+            let path = extent_path(dir, number);
+
+            /*
+             * Verify there are not existing extent files.
+             */
+            if Path::new(&path).exists() {
+                bail!("Extent file already exists {:?}", path);
+            }
+        }
+        remove_copy_cleanup_dir(dir, number)?;
```
jmpesp, line 367: Maybe this is a comment for the existing code and not this PR, but is this needed here?

downstairs/src/extent.rs
```diff
@@ -0,0 +1,829 @@
...
+        }
+
+        cdt::extent__write__start!(|| {
+            (job_id.0, self.number, writes.len() as u64)
+        });
+
+        let mut inner = self.inner.lock().await;
+        inner.write(job_id, writes, only_write_unwritten, self.iov_max)?;
+
+        cdt::extent__write__file__done!(|| {
```
jmpesp, line 429: This should move inside the sqlite specific impl

jmpesp (changes requested):

downstairs/src/region.rs
```diff
@@ -1,1565 +1,81 @@
...
+    let mut all = some.clone();
+    all.extend(vec![
+        extent_file_name(eid, ExtentType::DbShm),
+        extent_file_name(eid, ExtentType::DbWal),
+    ]);
 
-        // This gets filled one block at a time for hashing
-        let mut block = vec![0; self.block_size as usize];
+    // Either we have some or all.
+    files == some || files == all
```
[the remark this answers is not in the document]
mkeeter: This is identical to the [current implementation](https://github.com/oxidecomputer/crucible/blob/20273bcca1fd5834ebc3e67dfa7020f0e99ad681/downstairs/src/region.rs#L533-L554), but I can change it to a hashset comparison if you want!

(I'm inclined to leave it, because once we switch to the new SQLite-free backend, there's only one file and this becomes trivial)

downstairs/src/extent.rs
mkeeter: This is a [preexisting condition](https://github.com/oxidecomputer/crucible/blame/main/downstairs/src/region.rs#L771).  I agree that it would be a little funky if someone created a extent into a directory that had a left-over copy file, but that _could_ be possible?  Anyways, I think it's harmless.

downstairs/src/extent.rs
mkeeter: Good catch, done in a9eee1a61fc4bb92588a7e1f2a69e10504f4db17

downstairs/src/extent.rs
```diff
@@ -0,0 +1,829 @@
...
+        let replace_dir = replace_dir(dir, number);
+        if !read_only && Path::new(&replace_dir).exists() {
+            info!(
+                log,
+                "Extent {} found replacement dir, finishing replacement",
+                number
+            );
+            move_replacement_extent(dir, number as usize, log)?;
+        }
+
```
leftwo, line 297: The previous check where we validate file size is.. Ah.
It's moved into the SqliteInner open.
And, that makes sense, as the size check and some of the other checks will
depend on what format the file has, and. yeah, all good.

I should just delete this comment, but I wanted to show you all the hard work
I put into this :)

leftwo (approved):
The only actionable thing I has was the panic vs. bail comment.

The rest is my voyage of discovery where I figured out that, yes, you did
not change anything, but moved stuff to the layer that needed it.

downstairs/src/extent.rs
```diff
@@ -0,0 +1,829 @@
...
+use crucible_common::*;
+use repair_client::types::FileType;
+
+use super::*;
+
+#[derive(Debug)]
+pub struct Extent {
+    pub number: u32,
+    read_only: bool,
+    iov_max: usize,
```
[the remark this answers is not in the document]
mkeeter: _Mea culpa_, I forgot about that one.  It's a partial fix to #923

downstairs/src/extent.rs
```diff
@@ -0,0 +1,829 @@
...
+        let inner = {
+            let mut sqlite_path = path.clone();
+            sqlite_path.set_extension("db");
+            if sqlite_path.exists() {
+                let inner = extent_inner_sqlite::SqliteInner::open(
+                    &path, def, number, read_only, log,
+                )?;
+                Box::new(inner)
+            } else {
+                panic!("no SQLite file present at {sqlite_path:?}");
```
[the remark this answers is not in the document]
mkeeter: Good catch, restored the previous error in a6316f90cfc3ff85ae2250a6f2fb808d7d279a59

(Once we add the non-SQLite backend, this behavior will change again, because a missing `.db` file will indicate that we should load the file as a raw extent)

leftwo (approved):
My previous approval still stands!

jmpesp (approved):

PR #984, opened 2023-10-04
Read only downstairs can skip Live Repair

leftwo:
Don't bother trying to live repair a read only downstairs.  If we find a read only downstairs was kicked out for being gone to long, we just let it back in, as a read only region can have no changes, so there should be nothing to repair.

This is literally the code you put into chat last week.
Or, was it Monday? 

I'm trying to come up with a simple test for this without having to do a bunch of work
spinning up real downstairs, but I'm not coming up with any good ideas.

### DIFF
--- a/integration_tests/src/lib.rs
+++ b/integration_tests/src/lib.rs
@@ -2628,7 +2628,10 @@ mod test {
         }
 
         // Old volume is no longer active
-        assert!(!volume.query_is_active().await?);
+        while volume.query_is_active().await? {
+            println!("Waiting for old volume to deactivate");
+            tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+        }
 
         // Read back what we wrote.
         let buffer = Buffer::new(new_volume.total_size().await? as usize);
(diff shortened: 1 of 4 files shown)


upstairs/src/lib.rs
```diff
@@ -2935,6 +2935,12 @@ struct Downstairs {
      */
     live_repair_aborted: ClientData<usize>,
 
+    /**
+     * Times we skipped repairing a downstairs because we are running
+     * as read_only.
+     */
+    rop_lr_skipped: ClientData<usize>,
```
jmpesp, line 2942: It's not required that this is a read-only parent, just that it's read-only. Maybe `ro_lr_skipped` instead?

jmpesp:
Nothing to see here, please move along!

upstairs/src/lib.rs
leftwo: Yeah, that's a better name.

jmpesp (approved):

integration_tests/src/lib.rs
```diff
@@ -2628,7 +2628,10 @@ mod test {
         }
 
         // Old volume is no longer active
-        assert!(!volume.query_is_active().await?);
+        while volume.query_is_active().await? {
```
leftwo, line 2631: I hit this panic while testing, and I think it's just been lucky that we have not seen
it happen before.  Maybe things are faster now, where as before we would
always deactivate before the check happened.

PR #985, opened 2023-10-04
Support multiple downstairs operations in GtoS

jmpesp:
GtoS' purpose is to map the Guest IO request to the downstairs operations required to fulfill it. Part of the work to implement and test guest initiated bulk operations involved using GtoS with many downstairs operations instead of just one, and this required changing GtoS' struct plus fixing `transfer` and `gw_ds_complete` to actually support multiple reads.  Those changes, along with a few quality of life improvements, were pulled out into this commit.

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -8716,7 +8716,7 @@ impl GtoS {
     /*
      * Notify corresponding BlockReqWaiter
      */
-    pub fn notify(self, result: Result<(), CrucibleError>) {
+    fn notify(self, result: Result<(), CrucibleError>) {
         /*
          * If present, send the result to the guest.  If this is a flush
          * issued on behalf of crucible, then there is no place to send
@@ -8730,6 +8730,14 @@ impl GtoS {
         if let Some(req) = self.req {
             req.send_result(result);
         }
+    }
+
+    pub async fn finalize(mut self, result: Result<(), CrucibleError>) {
+        if result.is_ok() {
+            self.transfer().await;
+        }
+
+        self.notify(result);
     }
 }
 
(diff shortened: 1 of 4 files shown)


upstairs/src/lib.rs
```diff
@@ -8676,16 +8677,22 @@ impl GtoS {
...
-            for ds_id in self.completed.iter() {
-                let responses = self.downstairs_buffer.remove(ds_id).unwrap();
+                /*
+                 * Should this panic?  If the caller is requesting a transfer,
+                 * the guest_buffer should exist. If it does not exist, then
+                 * either there is a real problem, or the operation was a write
+                 * or flush and why are we requesting a transfer for those.
+                 */
+                let responses =
+                    self.downstairs_responses.remove(ds_id).unwrap();
```
mkeeter, line 8695: It makes me nervous that this function leaves the object in an invalid state; if someone managed to call `transfer` twice, it would fail.

Looking below, we call `transfer` in a specific context:
```rust
                if result.is_ok() && !gtos_job.guest_buffers.is_empty() {
                    gtos_job.transfer().await;
                }

                gtos_job.notify(result);
```
Could we instead have something like `gtos.finalize(result)`, which does the same combination of `transfer` and `notify` but consumes `mut self` as it does?

upstairs/src/lib.rs
```diff
@@ -8789,22 +8788,26 @@ impl GuestWork {
          * A gw_id that already finished and results were sent back to
          * the guest could still have an outstanding ds_id.
          */
-        if let Some(mut gtos_job) = self.active.remove(&gw_id) {
+        let mut gtos_job_done = false;
+        if let Some(gtos_job) = self.active.get_mut(&gw_id) {
```
mkeeter, line 8792: You could avoid mutation with something like
```rust
        let gtos_job_done = if let Some(gtos_job) = self.active.get_mut(&gw_id) {
            // ...
            gtos_job.submitted.is_empty()
        } else {
            // ...
            false
        };
```

mkeeter (approved):

upstairs/src/lib.rs
jmpesp: agreed, added in 8fa5a530c3fc2389ac51ff7e985c44afb5995af7

upstairs/src/lib.rs
jmpesp: nice! 8fa5a530c3fc2389ac51ff7e985c44afb5995af7

PR #986, opened 2023-10-04
[workspace-hack] use workspace-dotted format and a patch directive

sunshowers:
Similar to https://github.com/oxidecomputer/omicron/pull/4197, this PR
has two changes:

1. Switch to the workspace-dotted format (`.workspace = true`) for
   uniformity with the rest of crucible. This is new in cargo-hakari
   0.9.28.

2. Use a patch directive, which means that the workspace-hack only
   applies while building within this workspace. If another workspace
   imports a crate from here via a git dependency, it will not have the
   workspace-hack applied to it (instead, it will use [this empty crate](https://crates.io/crates/crucible-workspace-hack)
   on crates.io). Thanks so much to @pfmooney for this
   suggestion!

### DIFF
--- a/agent-client/Cargo.toml
+++ b/agent-client/Cargo.toml
@@ -13,5 +13,4 @@ reqwest.workspace = true
 schemars.workspace = true
 serde.workspace = true
 serde_json.workspace = true
-workspace-hack = { version = "0.1", path = "../workspace-hack" }
-
+crucible-workspace-hack = { version = "0.1", path = "../workspace-hack" }
--- a/agent-client/Cargo.toml
+++ b/agent-client/Cargo.toml
@@ -13,4 +13,4 @@ reqwest.workspace = true
 schemars.workspace = true
 serde.workspace = true
 serde_json.workspace = true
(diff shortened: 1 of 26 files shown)


mkeeter (approved):

agent-client/Cargo.toml
```diff
@@ -13,5 +13,4 @@ reqwest.workspace = true
 schemars.workspace = true
 serde.workspace = true
 serde_json.workspace = true
-workspace-hack = { version = "0.1", path = "../workspace-hack" }
-
+crucible-workspace-hack = { version = "0.1", path = "../workspace-hack" }
```
leftwo, line 16: In the omicron PR I saw these lines changed to:
`omicron-workspace-hack.workspace = true`

Can we do the workspace thing here as well, or will that not work for
other reasons?

sunshowers: Oh I missed that, sorry!

sunshowers: Fixed.

leftwo (approved):
Thanks for doing this!

Issue #902, opened 2023-09-01
disk times out and linux gives up on it if you send bunch of big writes and then ask to do a read


faithanalog:
- Tested on both 7b28bfe5c012ccae7e0902f64553c7ac8498a847 and 9aa49abbbe4b32470a182533df91c99ebb6f1aac so far
- Running on a revb, each downstairs with its own u.2 disk.
- 128GiB disk
- 64MiB extents
- 8GiB ram, 16 CPU guest. i dont think this matters that much.

im using this fio job file:

```
[global]
filename=/dev/nvme0n1
iodepth=25
ioengine=aio
time_based
runtime=30
numjobs=1
direct=1
stonewall=1

[write-4M]
bs=4M
rw=write

[read-16K]
bs=16K
rw=read
```

put that in `whatever.fio` on a guest and then run `fio whatever.fio`. Run it a few times. you should either see

- the write job always succeeds

and then, either

- the read job succeeds but it only manages to squeeze out a single 16k IOP.
- the read job fails. you get kernel errors like this. the OS drops the disk and it doesnt even show up in lsblk anymore.

```
[  515.228786] blk_update_request: I/O error, dev nvme0n1, sector 0 op 0x0:(READ) flags 0x0 phys_seg 1 prio class 0
[  515.230145] Buffer I/O error on dev nvme0n1, logical block 0, async page read
[  515.238341] blk_update_request: I/O error, dev nvme0n1, sector 0 op 0x0:(READ) flags 0x0 phys_seg 1 prio class 0
[  515.239697] Buffer I/O error on dev nvme0n1, logical block 0, async page read

[root@nixos:~]# lsblk
NAME   MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
loop0    7:0    0  617M  0 loop /nix/.ro-store
vda    253:0    0  648M  1 disk 
├─vda1 253:1    0  648M  1 part 
└─vda2 253:2    0    3M  1 part 

^ notice no NVME drives!! it's gone!!
```

This is on `main`. I haven't bisected yet but my immediate guess is that this is due to the early ack change. I think that with smaller writes (or fewer writes before a read) it's ok. But if you do a lot of BIG writes, the OS can now just slam the IO queue and then the disk sometimes cannot respond to the read in time because the read depends on all those writes finishing.

[...]

PR #990, opened 2023-10-06
Better backpressure

mkeeter:
This PR makes a few changes to backpressure:
- Backpressure delays are only applied to writes.  Every other guest operation actually waits for the operation to finish, which serves as built-in backpressure (because longer operations will stall the guest for longer)
- The backpressure delay now requires holding a `tokio::sync::Mutex`.  This fixes an issue where _N_ tasks could submit writes simultaneously, which defeated the previous backpressure implementation
- The delay is now computed as the max of two values:
  - The current implementation, which is based on DSW queue length
  - A new delay based on write bytes in flight
- The latter prevents a host from queueing up too many _large writes_, which don't take much space in the queue but take a long time to execute – in some cases, long enough that the host will kick out the disk (#902)

Running `fio` tests in a VM with writes of various sizes, we see that
- small writes are limited by queue length (the black line shows where queue length backpressure kicks in)
- larger writes stabilize with fewer writes in the queue, due to the bytes-in-flight limitation

Screenshot 2023-10-06 at 1 29 56 PM

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -8964,6 +8964,14 @@ pub struct Guest {
 
     /// Backpressure configuration, as a starting point and max delay
     backpressure_config: BackpressureConfig,
+
+    /// Lock held during backpressure delay
+    ///
+    /// Without this lock, multiple tasks could submit jobs to the upstairs and
+    /// wait in parallel, which defeats the purpose of backpressure (since you
+    /// could send arbitrarily many jobs at high speed by sending them from
+    /// different tasks).
+    backpressure_lock: Mutex<()>,
 }
 
 /// Configuration for host-side backpressure
@@ -9008,12 +9016,13 @@ impl Guest {
             bw_limit: None,
 
             block_size: AtomicU64::new(0),
-            backpressure_us: AtomicU64::new(0),
 
+            backpressure_us: AtomicU64::new(0),
             backpressure_config: BackpressureConfig {
                 start: 0.5,
                 max_delay: Duration::from_millis(5),
             },
+            backpressure_lock: Mutex::new(()),
         }
     }
 
@@ -9046,7 +9055,9 @@ impl Guest {
         let bp =
             Duration::from_micros(self.backpressure_us.load(Ordering::SeqCst));
         if bp > Duration::ZERO {
+            let _guard = self.backpressure_lock.lock().await;
             tokio::time::sleep(bp).await;
+            drop(_guard);
         }
 
         self.reqs.lock().await.push_back(BlockReq::new(op, send));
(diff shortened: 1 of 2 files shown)


upstairs/src/lib.rs
```diff
@@ -4769,7 +4792,23 @@ impl Downstairs {
...
+                    | IOop::WriteUnwritten { writes, .. } => {
+                        self.write_bytes_outstanding = self
+                            .write_bytes_outstanding
+                            .checked_sub(
+                                writes
+                                    .iter()
+                                    .map(|w| w.data.len() as u64)
+                                    .sum::<u64>(),
+                            )
+                            .unwrap();
```
leftwo, line 4808: When would we hit this unwrap?  If our accounting was wrong?

upstairs/src/lib.rs
```diff
@@ -8947,21 +9004,43 @@ pub struct Guest {
...
-    start: f32,
-    max_delay: Duration,
+    /// When should backpressure start (in bytes)?
+    bytes_start: u64,
+    /// Scale for byte-based quadratic backpressure
+    bytes_scale: f64,
+
+    /// When should queue-based backpressure start?
+    queue_start: f64,
+    /// Maximum queue-based delay
```
leftwo, line 9042: Will this queue based start consider IO types other than `Write`/`WriteUnwritten` when
deciding to apply backpressure?

I don't see backpressure checks in other IO types, but I'm wondering if there presence
in the work queue would trigger writes to slow down?

leftwo (approved):
Perfect and getting better every day.

upstairs/src/lib.rs
mkeeter: Yup, exactly.  This already caught one bug in development, where I forgot to add bytes from skipped jobs (but still subtracted them here, causing an overflow).

upstairs/src/lib.rs
mkeeter: Good catch – you're correct that if the queue has a bunch of non-write jobs, the queue-based backpressure will take them into account when delaying writes (and backpressure doesn't delay anything else).

This is a little weird, but I'm inclined to leave it as-is; the fundamental goal here is to avoid killing the upstairs due to hitting `MAX_ACTIVE_COUNT`, which counts every job in the queue.

(I'm also not sure how non-write jobs would manage to clog up the queue!)

leftwo: Yeah, I also think it's fine to leave here, just wanted to be sure I understood that
any IO on the queue can impact backpressure.

faithanalog (approved):
I tried very hard to get linux to give up on a disk and call it dead with this PR but wasn't able to. I think we should merge it. We should probably tune the backpressure more in a future PR. right now on an unencrypted dataset I can fill about 2 gigs of buffer, which is a fair chunk of data and also takes about 15 seconds to drain. You won't see buffers get that big with an encrypted dataset right now due to other encryption overhead. At any rate, exact final values we land on are up for debate and may change as perf changes, but the main thing is this PR fixes the hard-fail state of https://github.com/oxidecomputer/crucible/issues/902 so we should get it in

PR #991, opened 2023-10-06
Add raw file backend

mkeeter:
This PR adds a new backend for Crucible Downstairs extent files: a single on-disk file.  The on-disk file includes the raw block contents, a small metadata block, and two slots per block for storing encryption context.

By alternating between slots, we get the same crash consistency behavior as the SQLite database, but with many fewer syncs – we only need to add a `fsync` call if we're about to overwrite a context slot that hasn't yet been synched to disk, i.e. someone has written the same block 3x in a row without a flush.

The new backend implements the same `trait ExtentInner` as the SQLite backend, introduced in #982 .  This means that it's _mostly_ seamless, although I had to tweak a few small things around the edges.

Though this is a change to the on-disk format, we maintain support for existing (SQLite-based) volumes.  Existing volumes fall into two categories, which we handle differently:

- Snapshots are stored on read-only filesystems, so we can't mess with them.  We keep the old `SqliteInner` implementation around and use it in this case.
- Volumes that _aren't_ read-only are migrated from SQLite to raw extent files when they are opened.  ~The migration uses the same strategy as live repair:~
    - ~Creating a separate staging directory and staging new files in there~
    - ~Renaming that directory (to `.migrate` in this case) once everything is staged~
    - ~Copying from the `.migrate` directory if it's present on startup~
    - EDIT: the migration uses a new strategy; see [this comment](https://github.com/oxidecomputer/crucible/pull/991#issuecomment-1753266674)

To simplify the implementation, this change requires a whole-rack update.  Specifically, live repair between SQLite extents and raw file extents is not possible!  We have bumped the `CRUCIBLE_MESSAGE_VERSION` to prevent the system from trying such a thing (though the message format remains identical).

One subtle point: we still have to support `.replace` directories with SQLite files, because a live repair of an older extent could have been interrupted before the system update!

### DIFF
--- a/downstairs/src/extent.rs
+++ b/downstairs/src/extent.rs
@@ -367,7 +367,7 @@ impl Extent {
 
             // Compute supplemental data from the SQLite extent
             let mut inner = extent_inner_sqlite::SqliteInner::open(
-                &path, def, number, read_only, log,
+                dir, def, number, read_only, log,
             )?;
             let ctxs = inner.export_contexts()?;
             let dirty = inner.dirty()?;
@@ -423,12 +423,12 @@ impl Extent {
             if has_sqlite {
                 assert!(read_only || force_sqlite_backend);
                 let inner = extent_inner_sqlite::SqliteInner::open(
-                    &path, def, number, read_only, log,
+                    dir, def, number, read_only, log,
                 )?;
                 Box::new(inner)
             } else {
                 let inner = extent_inner_raw::RawInner::open(
-                    &path, def, number, read_only, log,
+                    dir, def, number, read_only, log,
                 )?;
                 Box::new(inner)
             }
--- a/downstairs/src/extent.rs
+++ b/downstairs/src/extent.rs
@@ -2,10 +2,10 @@
 use std::convert::TryInto;
 use std::fmt;
 use std::fmt::Debug;
-use std::fs::File;
+use std::fs::{File, OpenOptions};
 use tokio::sync::Mutex;
 
-use anyhow::{anyhow, bail, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use nix::unistd::{sysconf, SysconfVar};
 
 use serde::{Deserialize, Serialize};
@@ -252,17 +252,9 @@ pub fn completed_dir<P: AsRef<Path>>(dir: P, number: u32) -> PathBuf {
         .with_extension("completed")
 }
 
-/// Produce a `PathBuf` for the SQLite-to-raw migration directory
-pub fn migrate_dir<P: AsRef<Path>>(dir: P, number: u32) -> PathBuf {
-    extent_dir(dir, number)
-        .join(extent_file_name(number, ExtentType::Data))
-        .with_extension("migrate")
-}
-
 /**
- * Remove temporary directories associated with repair and migration, except for
- * the final .replace or .migrate directory. Replace and migration are handled
- * specifically during extent open.
+ * Remove directories associated with repair except for the replace
+ * directory. Replace is handled specifically during extent open.
  */
 pub fn remove_copy_cleanup_dir<P: AsRef<Path>>(dir: P, eid: u32) -> Result<()> {
     for f in [copy_dir, completed_dir] {
@@ -301,73 +293,112 @@ impl Extent {
         // If there was an unfinished copy, then clean it up here (??)
         remove_copy_cleanup_dir(dir, number)?;
 
-        // There are two possible migrations that could be in progress:
+        // If the replace directory exists for this extent, then it means
+        // a repair was interrupted before it could finish.  We will continue
+        // the repair before we open the extent.
         //
-        // - Extent replacement, which is done from a `.replace` directory and
-        //   calls `move_replacement_extent`; this is part of live repair.
-        // - Extent migration (from SQLite to raw files), which is done from a
-        //   `.migrate` directory and `move_migrate_extent`
-        //
-        // Both of these are supposed to complete fully, but it's possible that
-        // a repair / migration was interrupted before it could finish.  If the
-        // above directories are present, then we are **ready** for repair /
-        // migration, so we will continue it before opening the extent.
-        //
-        // It is not possible for a repair and migration to both be staged;
-        // migration happens later in this function, so the repair will be
-        // completed (right here).
-        if !read_only {
-            let has_replace = replace_dir(dir, number).exists();
-            let has_migrate = migrate_dir(dir, number).exists();
-            if has_replace && has_migrate {
-                bail!(
-                    "cannot have both .replace and .migrate dirs simultaneously"
-                );
-            } else if has_replace {
-                info!(
-                    log,
-                    "Extent {} found replacement dir, finishing replacement",
-                    number
-                );
-                move_replacement_extent(dir, number as usize, log)?;
-            } else if has_migrate {
-                info!(
-                    log,
-                    "Extent {} found migrate dir, finishing migration", number
-                );
-                move_migrate_extent(dir, number as usize, log)?;
-            }
+        // Note that `move_replacement_extent` must support migrating old
+        // (SQLite-based) extents and new (raw file) extents, because it's
+        // possible that an old extent replacement crashed right before we
+        // updated Crucible.
+        let replace_dir = replace_dir(dir, number);
+        if !read_only && Path::new(&replace_dir).exists() {
+            info!(
+                log,
+                "Extent {} found replacement dir, finishing replacement",
+                number
+            );
+            move_replacement_extent(dir, number as usize, log)?;
         }
 
-        // Speaking of migration, it's happening now!  We will migrate
-        // every read-write extent with a SQLite file present.
+        // We will migrate every read-write extent with a SQLite file present.
+        //
+        // We use the presence of the .db file as a marker to whether the extent
+        // data file has been migrated.
+        //
+        // Remember, the goal is to turn 2-4 files (extent data, .db, .db-wal,
+        // db-shm) into a single file containing
+        //
+        //  - The extent data (same as before, at the beginning of the file)
+        //  - Metadata and encryption context slots (after extent data in the
+        //  - same file)
+        //
+        // Here's the procedure:
+        //
+        //  - If the .db file is present
+        //      - Truncate it to the length of the extent data. This is a no-op
+        //        normally, but lets us recover from a partial migration (e.g.
+        //        if we wrote the raw context but crashed before deleting the
+        //        .db file)
+        //      - Open the extent using our existing SQLite extent code
+        //      - Using standard extent APIs, find the metadata and encryption
+        //        context for each block. Encode this in the new raw file format
+        //        (as a Vec<u8>)
+        //      - Close the (SQLite) extent
+        //      - Open the extent data file in append mode, and append the new
+        //        raw metadata + block contexts to the end of the file.
+        //      - Close the extent data file and fsync it to disk
+        //      - Delete the .db file
+        //      - fsync the containing directory (to make the deletion durable;
+        //        not sure if this is necessary)
+        //  - At this point, the .db file is not present and the extent data
+        //    file represents a raw extent that has been persisted to disk
+        //      - If the .db-wal file is present, remove it
+        //      - If the .db-shm file is present, remove it
+        //
+        //  It's safe to fail at any point in this procedure; the next time
+        //  `Extent::open` is called, we will restart the migration (if we
+        //  haven't gotten to deleting the `.db` file).
         let mut has_sqlite = path.with_extension("db").exists();
         if has_sqlite && !read_only {
-            let mut inner = extent_inner_sqlite::SqliteInner::open(
-                &path, def, number, read_only, log,
-            )?;
-            let data = inner.export()?;
+            info!(log, "Migrating extent {number}");
+            // Truncate the file to the length of the extent data
+            {
+                let f =
+                    OpenOptions::new().read(true).write(true).open(&path)?;
+                f.set_len(def.extent_size().value * def.block_size())?;
+            }
+            // Compute metadata and context slots
+            let meta_and_context = {
+                let mut inner = extent_inner_sqlite::SqliteInner::open(
+                    &path, def, number, read_only, log,
+                )?;
+                inner.export_meta_and_context()?
+            };
+            // Append the new raw data, then sync the file to disk
+            {
+                let mut f = OpenOptions::new()
+                    .read(true)
+                    .write(true)
+                    .append(true)
+                    .open(&path)?;
+                f.write_all(&meta_and_context)?;
+                f.sync_all()
+                    .with_context(|| format!("{path:?}: fsync failure"))?;
+            }
 
-            // We'll put the new file into copy_dir first, then rename it to
-            // migrate_dir once it's ready.  The logic here matches the
-            // docstring in `Region::repair_extent`
-            let copy_dir = copy_dir(dir, number);
-            let extent_file_name = extent_file_name(number, ExtentType::Data);
-            let new_file = copy_dir.join(extent_file_name);
-            mkdir_for_file(&new_file)?;
+            // Remove the .db file, because our extent is now a raw extent and
+            // has been persisted to disk.
+            std::fs::remove_file(path.with_extension("db"))?;
 
-            info!(log, "Extent {number} writing migration to {new_file:?}");
-            std::fs::write(&new_file, data)?;
-            sync_path(&new_file, log)?;
-
-            // Rename from .copy to .migrate, then sync the parent directory
-            info!(log, "Extent {number} renaming migration dir");
-            std::fs::rename(copy_dir, migrate_dir(dir, number))?;
+            // Make that removal persistent by synching the parent directory
             sync_path(extent_dir(dir, number), log)?;
 
-            info!(log, "Extent {number} copying the migration over");
-            move_migrate_extent(dir, number as usize, log)?;
+            // We have now migrated from SQLite to raw file extents!
             has_sqlite = false;
+            info!(log, "Done migrating extent {number}");
+        }
+
+        // Clean up spare SQLite files if this is a raw file extent.  These
+        // deletions don't need to be durable, because we're not using their
+        // presence / absence for anything meaningful.
+        if !has_sqlite && !read_only {
+            for ext in ["db-shm", "db-wal"] {
+                let f = path.with_extension(ext);
+                if f.exists() {
+                    std::fs::remove_file(f)?;
+                }
+            }
         }
 
         // Pick the format for the downstairs files.  In most cases, we will be
@@ -579,72 +610,6 @@ impl Extent {
     ) -> tokio::sync::MutexGuard<Box<dyn ExtentInner>> {
         self.inner.lock().await
     }
-}
-
-/// Apply a staged migration directory to the extent file
-///
-/// Migration is only supported from the SQLite extent backend to the raw extent
-/// backend.  It consists of copying the new (raw) extent file, then deleting
-/// the SQLite files.
-pub(crate) fn move_migrate_extent<P: AsRef<Path>>(
-    region_dir: P,
-    eid: usize,
-    log: &Logger,
-) -> Result<(), CrucibleError> {
-    let destination_dir = extent_dir(&region_dir, eid as u32);
-    let extent_file_name = extent_file_name(eid as u32, ExtentType::Data);
-    let migrate_dir = migrate_dir(&region_dir, eid as u32);
-    let completed_dir = completed_dir(&region_dir, eid as u32);
-
-    assert!(Path::new(&migrate_dir).exists());
-    assert!(!Path::new(&completed_dir).exists());
-
-    info!(
-        log,
-        "Copy files from {migrate_dir:?} in {destination_dir:?}"
-    );
-
-    // Setup the original and replacement file names.
-    let new_file = migrate_dir.join(&extent_file_name);
-    let original_file = destination_dir.join(&extent_file_name);
-
-    // Copy the new file on top of the original file.
-    if let Err(e) = std::fs::copy(new_file.clone(), original_file.clone()) {
-        crucible_bail!(
-            IoError,
-            "copy of {new_file:?} to {original_file:?} got: {e:?}",
-        );
-    }
-    sync_path(&original_file, log)?;
-
-    // Remove all of the old files.  It is always valid for an old file to _not_
-    // be present, for two reasons:
-    // - the .db-shm and .db-wal files are always optional
-    // - the .db file may have already been removed by a previous call to
-    //   `move_migrate_extent`, if a later step failed
-    for ext in ["db", "db-shm", "db-wal"] {
-        let original_file = original_file.with_extension(ext);
-        if original_file.exists() {
-            info!(
-                log,
-                "Remove old file {original_file:?} as there is no replacement",
-            );
-            std::fs::remove_file(&original_file)?;
-        }
-    }
-    sync_path(&destination_dir, log)?;
-
-    // After we have all files: rename the migrate dir to indicate that
-    // migration is complete.
-    info!(log, "Move directory  {migrate_dir:?} to {completed_dir:?}");
-    std::fs::rename(migrate_dir, &completed_dir)?;
-
-    sync_path(&destination_dir, log)?;
-
-    std::fs::remove_dir_all(&completed_dir)?;
-
-    sync_path(&destination_dir, log)?;
-    Ok(())
 }
 
 /**
--- a/downstairs/src/extent.rs
+++ b/downstairs/src/extent.rs
@@ -655,7 +655,7 @@ pub(crate) fn move_replacement_extent<P: AsRef<Path>>(
     sync_path(&original_file, log)?;
 
     // We distinguish between SQLite-backend and raw-file extents based on the
-    // presence of the `000.db` file.  We should never do live migration across
+    // presence of the `.db` file.  We should never do live migration across
     // different extent formats; in fact, we should never live-migrate
     // SQLite-backed extents at all, but must still handle the case of
     // unfinished migrations.
--- a/downstairs/src/extent.rs
+++ b/downstairs/src/extent.rs
@@ -43,7 +43,7 @@ pub(crate) trait ExtentInner: Send + Debug {
     ) -> Result<(), CrucibleError>;
 
     fn read(
-        &self,
+        &mut self,
         job_id: JobId,
         requests: &[&crucible_protocol::ReadRequest],
         responses: &mut Vec<crucible_protocol::ReadResponse>,
@@ -60,14 +60,15 @@ pub(crate) trait ExtentInner: Send + Debug {
 
     #[cfg(test)]
     fn get_block_contexts(
-        &self,
+        &mut self,
         block: u64,
         count: u64,
     ) -> Result<Vec<Vec<DownstairsBlockContext>>, CrucibleError>;
 
     /// Sets the dirty flag and updates a block context
     ///
-    /// This should only be called from test functions
+    /// This should only be called from test functions, where we want to
+    /// manually modify block contexts and test associated behavior
     #[cfg(test)]
     fn set_dirty_and_block_context(
         &mut self,
@@ -120,7 +121,17 @@ pub struct ExtentMeta {
 /// Extent version for SQLite-backed metadata
 ///
 /// See [`extent_inner_sqlite::SqliteInner`] for the implementation.
+///
+/// This is no longer used when creating new extents, but we support opening
+/// existing SQLite-based extents because snapshot images are on read-only
+/// volumes, so we can't migrate them.
+#[allow(dead_code)] // used in unit test
 pub const EXTENT_META_SQLITE: u32 = 1;
+
+/// Extent version for raw-file-backed metadata
+///
+/// See [`extent_inner_raw::RawInner`] for the implementation.
+pub const EXTENT_META_RAW: u32 = 2;
 
 impl ExtentMeta {
     pub fn new(ext_version: u32) -> ExtentMeta {
@@ -241,9 +252,16 @@ pub fn completed_dir<P: AsRef<Path>>(dir: P, number: u32) -> PathBuf {
         .with_extension("completed")
 }
 
+pub fn migrate_dir<P: AsRef<Path>>(dir: P, number: u32) -> PathBuf {
+    extent_dir(dir, number)
+        .join(extent_file_name(number, ExtentType::Data))
+        .with_extension("migrate")
+}
+
 /**
- * Remove directories associated with repair except for the replace
- * directory. Replace is handled specifically during extent open.
+ * Remove temporary directories associated with repair and migration, except for
+ * the final .replace or .migrate directory. Replace and migration are handled
+ * specifically during extent open.
  */
 pub fn remove_copy_cleanup_dir<P: AsRef<Path>>(dir: P, eid: u32) -> Result<()> {
     for f in [copy_dir, completed_dir] {
@@ -282,38 +300,91 @@ impl Extent {
         // If there was an unfinished copy, then clean it up here (??)
         remove_copy_cleanup_dir(dir, number)?;
 
-        // If the replace directory exists for this extent, then it means
-        // a repair was interrupted before it could finish.  We will continue
-        // the repair before we open the extent.
-        let replace_dir = replace_dir(dir, number);
-        if !read_only && Path::new(&replace_dir).exists() {
-            info!(
-                log,
-                "Extent {} found replacement dir, finishing replacement",
-                number
-            );
-            move_replacement_extent(dir, number as usize, log)?;
+        // There are two possible migrations that could be in progress:
+        //
+        // - Extent replacement, which is done from a `.replace` directory and
+        //   calls `move_replacement_extent`; this is part of live repair.
+        // - Extent migration (from SQLite to raw files), which is done from a
+        //   `.migrate` directory and `move_migrate_extent`
+        //
+        // Both of these are supposed to complete fully, but it's possible that
+        // a repair / migration was interrupted before it could finish.  If the
+        // above directories are present, then we are **ready** for repair /
+        // migration, so we will continue it before opening the extent.
+        //
+        // It is not possible for a repair and migration to both be staged;
+        // migration happens later in this function, so the repair will be
+        // completed (right here).
+        if !read_only {
+            let has_replace = replace_dir(dir, number).exists();
+            let has_migrate = migrate_dir(dir, number).exists();
+            if has_replace && has_migrate {
+                bail!(
+                    "cannot have both .replace and .migrate dirs simultaneously"
+                );
+            } else if has_replace {
+                info!(
+                    log,
+                    "Extent {} found replacement dir, finishing replacement",
+                    number
+                );
+                move_replacement_extent(dir, number as usize, log)?;
+            } else if has_migrate {
+                info!(
+                    log,
+                    "Extent {} found migrate dir, finishing migration", number
+                );
+                move_migrate_extent(dir, number as usize, log)?;
+            }
+        }
+
+        // Speaking of migration, it's happening now!  We will migrate
+        // every read-write extent with a SQLite file present.
+        let mut has_sqlite = path.with_extension("db").exists();
+        if has_sqlite && !read_only {
+            let mut inner = extent_inner_sqlite::SqliteInner::open(
+                &path, def, number, read_only, log,
+            )?;
+            let data = inner.export()?;
+
+            // We'll put the new file into copy_dir first, then rename it to
+            // migrate_dir once it's ready.  The logic here matches the
+            // docstring in `Region::repair_extent`
+            let copy_dir = copy_dir(dir, number);
+            let extent_file_name = extent_file_name(number, ExtentType::Data);
+            let new_file = copy_dir.join(extent_file_name);
+            mkdir_for_file(&new_file)?;
+
+            info!(log, "Extent {number} writing migration to {new_file:?}");
+            std::fs::write(&new_file, data)?;
+            sync_path(&new_file, log)?;
+
+            // Rename from .copy to .migrate, then sync the parent directory
+            info!(log, "Extent {number} renaming migration dir");
+            std::fs::rename(copy_dir, migrate_dir(dir, number))?;
+            sync_path(extent_dir(dir, number), log)?;
+
+            info!(log, "Extent {number} copying the migration over");
+            move_migrate_extent(dir, number as usize, log)?;
+            has_sqlite = false;
         }
 
         // Pick the format for the downstairs files
         //
         // Right now, this only supports SQLite-flavored Downstairs
-        let inner = {
-            let mut sqlite_path = path.clone();
-            sqlite_path.set_extension("db");
-            if sqlite_path.exists() {
+        let inner: Box<dyn ExtentInner> = {
+            if has_sqlite {
                 let inner = extent_inner_sqlite::SqliteInner::open(
                     &path, def, number, read_only, log,
                 )?;
                 Box::new(inner)
             } else {
-                bail!(
-                    "db file {sqlite_path:?} for extent#{number} is not present"
-                );
+                let inner = extent_inner_raw::RawInner::open(
+                    &path, def, number, read_only, log,
+                )?;
+                Box::new(inner)
             }
         };
-
-        // XXX: schema updates?
 
         let extent = Extent {
             number,
@@ -368,11 +439,37 @@ impl Extent {
         }
         remove_copy_cleanup_dir(dir, number)?;
 
-        let inner = extent_inner_sqlite::SqliteInner::create(dir, def, number)?;
+        // All new extents are created using the raw backend
+        let inner = extent_inner_raw::RawInner::create(dir, def, number)?;
 
         /*
          * Complete the construction of our new extent
          */
+        Ok(Extent {
+            number,
+            read_only: false,
+            iov_max: Extent::get_iov_max()?,
+            inner: Mutex::new(Box::new(inner)),
+        })
+    }
+
+    /// Identical to `create`, but using the SQLite backend
+    ///
+    /// This is only allowed in unit tests
+    #[cfg(test)]
+    pub fn create_sqlite(
+        dir: &Path,
+        def: &RegionDefinition,
+        number: u32,
+    ) -> Result<Extent> {
+        {
+            let path = extent_path(dir, number);
+            if Path::new(&path).exists() {
+                bail!("Extent file already exists {:?}", path);
+            }
+        }
+        remove_copy_cleanup_dir(dir, number)?;
+        let inner = extent_inner_sqlite::SqliteInner::create(dir, def, number)?;
         Ok(Extent {
             number,
             read_only: false,
@@ -399,7 +496,7 @@ impl Extent {
             (job_id.0, self.number, requests.len() as u64)
         });
 
-        let inner = self.inner.lock().await;
+        let mut inner = self.inner.lock().await;
 
         inner.read(job_id, requests, responses, self.iov_max)?;
 
@@ -483,6 +580,72 @@ impl Extent {
     }
 }
 
+/// Apply a staged migration directory to the extent file
+///
+/// Migration is only supported from the SQLite extent backend to the raw extent
+/// backend.  It consists of copying the new (raw) extent file, then deleting
+/// the SQLite files.
+pub(crate) fn move_migrate_extent<P: AsRef<Path>>(
+    region_dir: P,
+    eid: usize,
+    log: &Logger,
+) -> Result<(), CrucibleError> {
+    let destination_dir = extent_dir(&region_dir, eid as u32);
+    let extent_file_name = extent_file_name(eid as u32, ExtentType::Data);
+    let migrate_dir = migrate_dir(&region_dir, eid as u32);
+    let completed_dir = completed_dir(&region_dir, eid as u32);
+
+    assert!(Path::new(&migrate_dir).exists());
+    assert!(!Path::new(&completed_dir).exists());
+
+    info!(
+        log,
+        "Copy files from {migrate_dir:?} in {destination_dir:?}"
+    );
+
+    // Setup the original and replacement file names.
+    let new_file = migrate_dir.join(&extent_file_name);
+    let original_file = destination_dir.join(&extent_file_name);
+
+    // Copy the new file on top of the original file.
+    if let Err(e) = std::fs::copy(new_file.clone(), original_file.clone()) {
+        crucible_bail!(
+            IoError,
+            "copy of {new_file:?} to {original_file:?} got: {e:?}",
+        );
+    }
+    sync_path(&original_file, log)?;
+
+    // Remove all of the old files.  It is always valid for an old file to _not_
+    // be present, for two reasons:
+    // - the .db-shm and .db-wal files are always optional
+    // - the .db file may have already been removed by a previous call to
+    //   `move_migrate_extent`, if a later step failed
+    for ext in ["db", "db-shm", "db-wal"] {
+        let original_file = original_file.with_extension(ext);
+        if original_file.exists() {
+            info!(
+                log,
+                "Remove old file {original_file:?} as there is no replacement",
+            );
+            std::fs::remove_file(&original_file)?;
+        }
+    }
+    sync_path(&destination_dir, log)?;
+
+    // After we have all files: rename the migrate dir to indicate that
+    // migration is complete.
+    info!(log, "Move directory  {migrate_dir:?} to {completed_dir:?}");
+    std::fs::rename(migrate_dir, &completed_dir)?;
+
+    sync_path(&destination_dir, log)?;
+
+    std::fs::remove_dir_all(&completed_dir)?;
+
+    sync_path(&destination_dir, log)?;
+    Ok(())
+}
+
 /**
  * Copy the contents of the replacement directory on to the extent
  * files in the extent directory.
@@ -525,25 +688,14 @@ pub(crate) fn move_replacement_extent<P: AsRef<Path>>(
     }
     sync_path(&original_file, log)?;
 
+    // We distinguish between SQLite-backend and raw-file extents based on the
+    // presence of the `000.db` file.  We should never do live migration across
+    // different extent formats; in fact, we should never live-migrate
+    // SQLite-backed extents at all, but must still handle the case of
+    // unfinished migrations.
     new_file.set_extension("db");
     original_file.set_extension("db");
-    if let Err(e) = std::fs::copy(new_file.clone(), original_file.clone()) {
-        crucible_bail!(
-            IoError,
-            "copy {:?} to {:?} got: {:?}",
-            new_file,
-            original_file,
-            e
-        );
-    }
-    sync_path(&original_file, log)?;
-
-    // The .db-shm and .db-wal files may or may not exist.  If they don't
-    // exist on the source side, then be sure to remove them locally to
-    // avoid database corruption from a mismatch between old and new.
-    new_file.set_extension("db-shm");
-    original_file.set_extension("db-shm");
-    if new_file.exists() {
+    if original_file.exists() {
         if let Err(e) = std::fs::copy(new_file.clone(), original_file.clone()) {
             crucible_bail!(
                 IoError,
@@ -554,35 +706,57 @@ pub(crate) fn move_replacement_extent<P: AsRef<Path>>(
             );
         }
         sync_path(&original_file, log)?;
-    } else if original_file.exists() {
-        info!(
-            log,
-            "Remove old file {:?} as there is no replacement",
-            original_file.clone()
-        );
-        std::fs::remove_file(&original_file)?;
-    }
 
-    new_file.set_extension("db-wal");
-    original_file.set_extension("db-wal");
-    if new_file.exists() {
-        if let Err(e) = std::fs::copy(new_file.clone(), original_file.clone()) {
-            crucible_bail!(
-                IoError,
-                "copy {:?} to {:?} got: {:?}",
-                new_file,
-                original_file,
-                e
+        // The .db-shm and .db-wal files may or may not exist.  If they don't
+        // exist on the source side, then be sure to remove them locally to
+        // avoid database corruption from a mismatch between old and new.
+        new_file.set_extension("db-shm");
+        original_file.set_extension("db-shm");
+        if new_file.exists() {
+            if let Err(e) =
+                std::fs::copy(new_file.clone(), original_file.clone())
+            {
+                crucible_bail!(
+                    IoError,
+                    "copy {:?} to {:?} got: {:?}",
+                    new_file,
+                    original_file,
+                    e
+                );
+            }
+            sync_path(&original_file, log)?;
+        } else if original_file.exists() {
+            info!(
+                log,
+                "Remove old file {:?} as there is no replacement",
+                original_file.clone()
             );
+            std::fs::remove_file(&original_file)?;
         }
-        sync_path(&original_file, log)?;
-    } else if original_file.exists() {
-        info!(
-            log,
-            "Remove old file {:?} as there is no replacement",
-            original_file.clone()
-        );
-        std::fs::remove_file(&original_file)?;
+
+        new_file.set_extension("db-wal");
+        original_file.set_extension("db-wal");
+        if new_file.exists() {
+            if let Err(e) =
+                std::fs::copy(new_file.clone(), original_file.clone())
+            {
+                crucible_bail!(
+                    IoError,
+                    "copy {:?} to {:?} got: {:?}",
+                    new_file,
+                    original_file,
+                    e
+                );
+            }
+            sync_path(&original_file, log)?;
+        } else if original_file.exists() {
+            info!(
+                log,
+                "Remove old file {:?} as there is no replacement",
+                original_file.clone()
+            );
+            std::fs::remove_file(&original_file)?;
+        }
     }
     sync_path(&destination_dir, log)?;
 
--- a/downstairs/src/extent.rs
+++ b/downstairs/src/extent.rs
@@ -281,6 +281,7 @@ impl Extent {
         def: &RegionDefinition,
         number: u32,
         read_only: bool,
+        backend: Backend,
         log: &Logger,
     ) -> Result<Extent> {
         /*
@@ -349,7 +350,13 @@ impl Extent {
         //  `Extent::open` is called, we will restart the migration (if we
         //  haven't gotten to deleting the `.db` file).
         let mut has_sqlite = path.with_extension("db").exists();
-        if has_sqlite && !read_only {
+        let force_sqlite_backend = match backend {
+            Backend::RawFile => false,
+            #[cfg(any(test, feature = "integration-tests"))]
+            Backend::SQLite => true,
+        };
+        let should_migrate = has_sqlite && !read_only && !force_sqlite_backend;
+        if should_migrate {
             info!(log, "Migrating extent {number}");
             // Truncate the file to the length of the extent data
             {
@@ -414,7 +421,7 @@ impl Extent {
         // as-is.
         let inner: Box<dyn ExtentInner> = {
             if has_sqlite {
-                assert!(read_only);
+                assert!(read_only || force_sqlite_backend);
                 let inner = extent_inner_sqlite::SqliteInner::open(
                     &path, def, number, read_only, log,
                 )?;
@@ -463,6 +470,7 @@ impl Extent {
         dir: &Path,
         def: &RegionDefinition,
         number: u32,
+        backend: Backend,
     ) -> Result<Extent> {
         /*
          * Store extent data in files within a directory hierarchy so that
@@ -480,8 +488,15 @@ impl Extent {
         }
         remove_copy_cleanup_dir(dir, number)?;
 
-        // All new extents are created using the raw backend
-        let inner = extent_inner_raw::RawInner::create(dir, def, number)?;
+        let inner: Box<dyn ExtentInner> = match backend {
+            Backend::RawFile => {
+                Box::new(extent_inner_raw::RawInner::create(dir, def, number)?)
+            }
+            #[cfg(any(test, feature = "integration-tests"))]
+            Backend::SQLite => Box::new(
+                extent_inner_sqlite::SqliteInner::create(dir, def, number)?,
+            ),
+        };
 
         /*
          * Complete the construction of our new extent
@@ -490,32 +505,7 @@ impl Extent {
             number,
             read_only: false,
             iov_max: Extent::get_iov_max()?,
-            inner: Mutex::new(Box::new(inner)),
-        })
-    }
-
-    /// Identical to `create`, but using the SQLite backend
-    ///
-    /// This is only allowed in unit tests
-    #[cfg(any(test, feature = "integration-tests"))]
-    pub fn create_sqlite(
-        dir: &Path,
-        def: &RegionDefinition,
-        number: u32,
-    ) -> Result<Extent> {
-        {
-            let path = extent_path(dir, number);
-            if Path::new(&path).exists() {
-                bail!("Extent file already exists {:?}", path);
-            }
-        }
-        remove_copy_cleanup_dir(dir, number)?;
-        let inner = extent_inner_sqlite::SqliteInner::create(dir, def, number)?;
-        Ok(Extent {
-            number,
-            read_only: false,
-            iov_max: Extent::get_iov_max()?,
-            inner: Mutex::new(Box::new(inner)),
+            inner: Mutex::new(inner),
         })
     }
 
--- a/downstairs/src/extent.rs
+++ b/downstairs/src/extent.rs
@@ -77,7 +77,7 @@ pub(crate) trait ExtentInner: Send + Debug {
 }
 
 /// BlockContext, with the addition of block index and on_disk_hash
-#[derive(Clone)]
+#[derive(Copy, Clone)]
 pub struct DownstairsBlockContext {
     pub block_context: BlockContext,
 
@@ -357,30 +357,29 @@ impl Extent {
                     OpenOptions::new().read(true).write(true).open(&path)?;
                 f.set_len(def.extent_size().value * def.block_size())?;
             }
-            // Compute metadata and context slots
-            let meta_and_context = {
-                let mut inner = extent_inner_sqlite::SqliteInner::open(
-                    &path, def, number, read_only, log,
-                )?;
-                let ctxs = inner.export_contexts()?;
-                let dirty = inner.dirty()?;
-                let flush_number = inner.flush_number()?;
-                let gen_number = inner.gen_number()?;
+
+            // Compute supplemental data from the SQLite extent
+            let mut inner = extent_inner_sqlite::SqliteInner::open(
+                &path, def, number, read_only, log,
+            )?;
+            let ctxs = inner.export_contexts()?;
+            let dirty = inner.dirty()?;
+            let flush_number = inner.flush_number()?;
+            let gen_number = inner.gen_number()?;
+            drop(inner);
+
+            // Reopen the file and import those changes
+            {
+                let mut f =
+                    OpenOptions::new().read(true).write(true).open(&path)?;
                 extent_inner_raw::RawInner::import(
+                    &mut f,
+                    def,
                     ctxs,
                     dirty,
                     flush_number,
                     gen_number,
-                )?
-            };
-            // Append the new raw data, then sync the file to disk
-            {
-                let mut f = OpenOptions::new()
-                    .read(true)
-                    .write(true)
-                    .append(true)
-                    .open(&path)?;
-                f.write_all(&meta_and_context)?;
+                )?;
                 f.sync_all()
                     .with_context(|| format!("{path:?}: fsync failure"))?;
             }
--- a/downstairs/src/extent.rs
+++ b/downstairs/src/extent.rs
@@ -362,7 +362,16 @@ impl Extent {
                 let mut inner = extent_inner_sqlite::SqliteInner::open(
                     &path, def, number, read_only, log,
                 )?;
-                inner.export_meta_and_context()?
+                let ctxs = inner.export_contexts()?;
+                let dirty = inner.dirty()?;
+                let flush_number = inner.flush_number()?;
+                let gen_number = inner.gen_number()?;
+                extent_inner_raw::RawInner::import(
+                    ctxs,
+                    dirty,
+                    flush_number,
+                    gen_number,
+                )?
             };
             // Append the new raw data, then sync the file to disk
             {
--- a/downstairs/src/extent.rs
+++ b/downstairs/src/extent.rs
@@ -332,8 +332,7 @@ impl Extent {
         //        .db file)
         //      - Open the extent using our existing SQLite extent code
         //      - Using standard extent APIs, find the metadata and encryption
-        //        context for each block. Encode this in the new raw file format
-        //        (as a Vec<u8>)
+        //        context for each block. Append this to the existing data file.
         //      - Close the (SQLite) extent
         //      - Open the extent data file in append mode, and append the new
         //        raw metadata + block contexts to the end of the file.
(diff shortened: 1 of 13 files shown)


mkeeter:
Performance notes: this is a ~30% speedup running `pgbench` on Alpine Linux, with an encrypted 128 GiB volume (4K blocks, 16384 blocks/extent).

Screenshot 2023-10-06 at 3 46 20 PM

On the same system, it's roughly a 67% speedup for 4K random writes using `fio`.  However, it's slower for 1M and 4M writes, for reasons I haven't yet ascertained.
```
main:
  WRITE: bw=15.2MiB/s (15.9MB/s), 15.2MiB/s-15.2MiB/s (15.9MB/s-15.9MB/s), io=910MiB (954MB), run=60009-60009msec
  WRITE: bw=371MiB/s (389MB/s), 371MiB/s-371MiB/s (389MB/s-389MB/s), io=21.8GiB (23.4GB), run=60064-60064msec
  WRITE: bw=348MiB/s (365MB/s), 348MiB/s-348MiB/s (365MB/s-365MB/s), io=20.5GiB (22.0GB), run=60263-60263msec

raw-file-extents:
  WRITE: bw=25.5MiB/s (26.7MB/s), 25.5MiB/s-25.5MiB/s (26.7MB/s-26.7MB/s), io=1529MiB (1603MB), run=60005-60005msec
  WRITE: bw=311MiB/s (326MB/s), 311MiB/s-311MiB/s (326MB/s-326MB/s), io=18.2GiB (19.6GB), run=60076-60076msec
  WRITE: bw=342MiB/s (358MB/s), 342MiB/s-342MiB/s (358MB/s-358MB/s), io=20.1GiB (21.6GB), run=60301-60301msec
```
Here's the `fio` file I used for these tests:
```toml
[global]
filename=/dev/nvme0n1
iodepth=25
ioengine=aio
time_based
runtime=60
numjobs=1
direct=1
stonewall=1

[randwrite-4K]
bs=4K
rw=randwrite

[randwrite-1M]
bs=1M
rw=randwrite

[randwrite-4M]
bs=4M
rw=randwrite
```

mkeeter:
There are a few unit tests for automatic migration, but I also did an end-to-end test:

- Created a 128 GiB SQLite-flavored volume using `crucible-downstairs` / `dsc` on `main` (same specs as above)
- In a `propolis-standalone` VM, initialized that disk as ext4 and wrote 4GiB of random data to `/mnt/lol.dat`
- Recorded the checksum of that random file
- Exited the VM and `propolis-standalone`
- Restarted 3x `crucible-downstairs` on this branch (`raw-file-extents`).  This triggers automatic migration!
- Restarted the VM
- Mounted `/dev/nvme0n1` to `/mnt`
- Calculated the checksum of `/mnt/lol.dat`

Sure enough, the checksum matched after the migration.

However, the migration was a bit slow: it took 8 minutes for the `crucible-downstairs` to come up.  **I'm not sure if this is a problem!**

Looking at the flamegraph, it appears to be mostly Good Honest Disk I/O in `std::fs::copy`:

Screenshot 2023-10-06 at 5 24 05 PM

Right now, the procedure is to 
1) Compute the new raw extent file in a `.copy` directory, then rename it to `.migrate`
2) Copy that file into place from a `.migrate` directory
3) Rename the `.migrate` directory to `.complete`, then delete it

(with syncs in various places)

This makes sense for the SQLite repair process, which has multiple files to keep track of.  However, we only care about a single file here.  I'm wondering whether we could simply rename the raw file (instead of copying it), which I _think_ should be atomic.

mkeeter:
I've pushed a much faster migration procedure in dafe220c575b4226d40a3e264909ec7e71eeecf0, which is detailed in a [long block comment](https://github.com/oxidecomputer/crucible/blob/dafe220c575b4226d40a3e264909ec7e71eeecf0/downstairs/src/extent.rs#L314-L351).  This brings migration time down to 57 seconds for a 128 GiB disk, with nearly all that time being spent in `fully_rehash_and_clean_all_stale_contexts`:

Screenshot 2023-10-09 at 8 12 33 AM

mkeeter:
...and one more optimization in 553704aadcdd2c52b8367ddb2f1ef8af7a4c33dd, which brings migration for the 128 GiB disk down to 9 seconds.  That seems fast enough to me.

downstairs/src/extent_inner_sqlite.rs
```diff
@@ -508,6 +505,77 @@ impl ExtentInner for SqliteInner {
...
+    /// Returns the metadata and context slots, which should be positioned
+    /// directly after the raw block data in memory.
+    pub fn export_meta_and_context(
+        &mut self,
+    ) -> Result<Vec<u8>, CrucibleError> {
+        // Check whether we need to rehash.  This is theoretically represented
+        // by the `dirty` bit, but we're being _somewhat_ paranoid and manually
+        // forcing a rehash if any blocks have multiple contexts stored.
+        let ctxs = self.get_block_contexts(0, self.extent_size.value)?;
+        let needs_rehash = ctxs.iter().any(|c| c.len() > 1);
```
mkeeter, line 519: Feedback requested: is this paranoia justified, or should I just call `self.fully_rehash_and_clean_all_stale_contexts(false)?` and trust the dirty flag?

downstairs/src/extent_inner_sqlite.rs
```diff
@@ -508,6 +505,77 @@ impl ExtentInner for SqliteInner {
...
+
+        use crate::{
+            extent::EXTENT_META_RAW,
+            extent_inner_raw::{
+                OnDiskDownstairsBlockContext, OnDiskMeta,
+                BLOCK_CONTEXT_SLOT_SIZE_BYTES, BLOCK_META_SIZE_BYTES,
+            },
+        };
+
+        // Record the metadata region, which will be right after raw block data
```
leftwo, line 537: Nit, I try not to use `extent` and `region` unless I'm specifically talking about Crucible's specific `extent` or `region`.  I might suggest "Record the metadata section, .." or another word of your choice here.

downstairs/src/region.rs
```diff
@@ -1592,6 +1615,63 @@ pub(crate) mod test {
         // Make copy directory for this extent
         let cp = Region::create_copy_dir(&dir, 1)?;
 
+        // We are simulating the copy of files from the "source" repair
```
leftwo, line 1618: Is the comment up on line 1599 still correct for this test?

downstairs/src/region.rs
```diff
@@ -1735,67 +1932,14 @@ pub(crate) mod test {
...
-            "001.db-wal".to_string(),
-        ];
+        let good_files: Vec<String> = vec!["001".to_string()];
 
         assert!(!validate_repair_files(2, &good_files));
     }
 
-    #[test]
-    fn validate_repair_files_too_good() {
-        // Duplicate file in list
```
leftwo, line 1777: This test did catch a case where I sent the same name twice.  You don't think we should have something
similar here?

leftwo (commented):
Just really some minor things.

Have you ran any of the tests in `./tools` directory on this?  I'm going to take the
nightly-build from the CI and put it on a bench gimlet and let things run, but I'm
not expecting to find any issues, this all seems really solid.

downstairs/src/extent_inner_raw.rs
```diff
@@ -0,0 +1,1329 @@
...
+    pub fn open(
+        path: &Path,
+        def: &RegionDefinition,
+        extent_number: u32,
+        read_only: bool,
+        log: &Logger,
+    ) -> Result<Self> {
+        let bcount = def.extent_size().value;
+        let size = def.block_size().checked_mul(bcount).unwrap()
+            + BLOCK_CONTEXT_SLOT_SIZE_BYTES * bcount * 2
```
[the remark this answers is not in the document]
mkeeter: Yes, done in bd01c1833c8ac89

leftwo:
Another win for these changes can be seen in the region creation test.
We can create larger regions much faster now:

Stock bits:
```
  SECONDS REGION_SIZE  EXTENT_SIZE     ES     EC   BS
  171.751  750.00 GiB    16.00 MiB   4096  48000 4096
   54.902  750.00 GiB    32.00 MiB   8192  24000 4096
   19.971  750.00 GiB    64.00 MiB  16384  12000 4096
    7.886  750.00 GiB   128.00 MiB  32768   6000 4096
  301.727 1024.00 GiB    16.00 MiB   4096  65536 4096
   90.760 1024.00 GiB    32.00 MiB   8192  32768 4096
   30.534 1024.00 GiB    64.00 MiB  16384  16384 4096
   11.440 1024.00 GiB   128.00 MiB  32768   8192 4096
```
These changes:
```
  SECONDS REGION_SIZE  EXTENT_SIZE     ES     EC   BS
   53.728  750.00 GiB    16.00 MiB   4096  48000 4096
   28.935  750.00 GiB    32.00 MiB   8192  24000 4096
   17.791  750.00 GiB    64.00 MiB  16384  12000 4096
   10.442  750.00 GiB   128.00 MiB  32768   6000 4096
   74.311 1024.00 GiB    16.00 MiB   4096  65536 4096
   38.683 1024.00 GiB    32.00 MiB   8192  32768 4096
   23.020 1024.00 GiB    64.00 MiB  16384  16384 4096
   14.461 1024.00 GiB   128.00 MiB  32768   8192 4096
```

downstairs/src/extent_inner_raw.rs
```diff
@@ -0,0 +1,1329 @@
...
+        // Second context was appended
+        assert_eq!(ctx.block_context.encryption_context.unwrap().nonce, blob1);
+        assert_eq!(ctx.block_context.encryption_context.unwrap().tag, blob2);
+        assert_eq!(ctx.block_context.hash, 1024);
+        assert_eq!(ctx.on_disk_hash, 65536);
+
+        Ok(())
+    }
+
+    #[tokio::test]
```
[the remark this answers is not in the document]
mkeeter: Yeah, I tried to copy over tests that were still relevant!

downstairs/src/extent_inner_sqlite.rs
mkeeter: Good idea, done in db9c6548cee97ae25d0bb5e6838bb0e74700ce44

downstairs/src/extent.rs
```diff
@@ -295,26 +311,113 @@ impl Extent {
...
-                    "db file {sqlite_path:?} for extent#{number} is not present"
-                );
+                let inner = extent_inner_raw::RawInner::open(
+                    &path, def, number, read_only, log,
+                )?;
+                Box::new(inner)
             }
         };
 
-        // XXX: schema updates?
```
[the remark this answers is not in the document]
mkeeter: 🚀

downstairs/src/extent.rs
```diff
@@ -525,25 +654,14 @@ pub(crate) fn move_replacement_extent<P: AsRef<Path>>(
     }
     sync_path(&original_file, log)?;
 
+    // We distinguish between SQLite-backend and raw-file extents based on the
+    // presence of the `000.db` file.  We should never do live migration across
```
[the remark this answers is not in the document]
mkeeter: Yeah, rephrased in c5ecb20085b17ba3b92eee88b995c26dd1be3ead

downstairs/src/extent_inner_raw.rs
```diff
@@ -0,0 +1,1329 @@
...
+    fn set_block_context(
+        &mut self,
+        block_context: &DownstairsBlockContext,
+    ) -> Result<bool> {
+        let block = block_context.block as usize;
+        // Select the inactive slot
+        let slot = !self.get_block_context_slot(block)?.unwrap_or(true);
+
+        // If the context slot that we're about to write into hasn't been
+        // synched to disk yet, we must sync it first.  This prevents subtle
```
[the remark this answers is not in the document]
mkeeter: Yes, this is checked in `test_auto_sync` below (reading `self.sync_index` and making sure it changes when we expect).

downstairs/src/extent.rs
```diff
@@ -285,6 +296,11 @@ impl Extent {
         // If the replace directory exists for this extent, then it means
         // a repair was interrupted before it could finish.  We will continue
         // the repair before we open the extent.
+        //
```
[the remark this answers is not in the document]
mkeeter: Done in 06151194d682629eebdaf88af8040032b6818e60

downstairs/src/extent.rs
```diff
@@ -295,26 +311,113 @@ impl Extent {
...
+
+        // Pick the format for the downstairs files.  In most cases, we will be
+        // using the raw extent format, but for read-only snapshots, we're stuck
+        // with the SQLite backend.
+        let inner: Box<dyn ExtentInner> = {
+            if has_sqlite {
                 let inner = extent_inner_sqlite::SqliteInner::open(
                     &path, def, number, read_only, log,
                 )?;
                 Box::new(inner)
```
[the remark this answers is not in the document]
mkeeter: Sure, done in 02dfdb3e71b07b487d8c1345d7641c8b15ed7d0a

downstairs/src/extent.rs
```diff
@@ -295,26 +311,113 @@ impl Extent {
...
+            for ext in ["db-shm", "db-wal"] {
+                let f = path.with_extension(ext);
+                if f.exists() {
+                    std::fs::remove_file(f)?;
+                }
+            }
+        }
+
+        // Pick the format for the downstairs files.  In most cases, we will be
+        // using the raw extent format, but for read-only snapshots, we're stuck
```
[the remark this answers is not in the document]
mkeeter: Yeah, clarified in 02dfdb3e71b07b487d8c1345d7641c8b15ed7d0a

downstairs/src/region.rs
mkeeter: Now that we're down to a single file, that condition is checked in `validate_repair_files_duplicate`.  I also just added `validate_repair_files_old`, which ensures that we fail if the repair file list includes a `.db` file.

downstairs/src/extent_inner_raw.rs
```diff
@@ -0,0 +1,1329 @@
...
+
+        cdt::extent__write__file__done!(|| {
+            (job_id.0, self.extent_number, writes.len() as u64)
+        });
+
+        Ok(())
+    }
+
+    fn read(
+        &mut self,
```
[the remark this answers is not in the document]
mkeeter: Interesting, you're correct that these are (currently) identical!  Let me think about whether we should keep them separate, or try combining them...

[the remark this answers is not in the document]
leftwo: I don't think it is a bad thing though.

My hope is that we can, someday, remove support for the SQL backend, and keeping them in two different
places might make that work easier.

Though, if we add a third backend, then maybe it makes sense to abstract them.

I'm certainly not making that a requirement for this PR.

jmpesp (changes requested):
Here's a first pass for the review. I have unanswered questions about how ZFS behaves that I'll follow up with out of band.

downstairs/src/extent_inner_raw.rs
```diff
@@ -0,0 +1,1325 @@
...
+            // TODO it would be nice if we could profile what % of time we're
+            // spending on hashes locally vs writing to disk
+            let on_disk_hash = integrity_hash(&[&write.data[..]]);
+
+            let next_slot =
+                self.set_block_context(&DownstairsBlockContext {
+                    block_context: write.block_context,
+                    block: write.offset.value,
+                    on_disk_hash,
+                })?;
```
jmpesp, line 252: If any of the extent data writes below fail, then the downstairs will re-execute this whole function for the same `writes` argument. It looks like `set_block_context` is destructive for the current contexts, which means if this function is retried two times (failing each time during the extent data write) then it will overwrite the correct context for all N blocks in `writes`.

jmpesp (changes requested):

downstairs/src/extent_inner_raw.rs
```diff
@@ -0,0 +1,1325 @@
...
+///
+/// This must be large enough to contain an `Option<DownstairsBlockContext>`
+/// serialized using `bincode`.
+pub const BLOCK_CONTEXT_SLOT_SIZE_BYTES: u64 = 64;
+
+/// Size of metadata region
+///
+/// This must be large enough to contain an `OnDiskMeta` serialized using
+/// `bincode`.
+pub const BLOCK_META_SIZE_BYTES: u64 = 64;
```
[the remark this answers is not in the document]
mkeeter: Good idea, done in 896dac97903a15cf1b1cefe1f7133973aae5fb85

downstairs/src/extent.rs
```diff
@@ -295,26 +310,115 @@ impl Extent {
...
+        //  `Extent::open` is called, we will restart the migration (if we
+        //  haven't gotten to deleting the `.db` file).
+        let mut has_sqlite = path.with_extension("db").exists();
+        if has_sqlite && !read_only {
+            info!(log, "Migrating extent {number}");
+            // Truncate the file to the length of the extent data
+            {
+                let f =
+                    OpenOptions::new().read(true).write(true).open(&path)?;
+                f.set_len(def.extent_size().value * def.block_size())?;
```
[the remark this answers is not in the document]
mkeeter: I don't think so!  This call to `set_len` should only ever truncate the file down to the raw data size.

The steps are
1) Truncate file
2) Write new metadata and context region
3) `fsync` the file
4) Remove the `.db` file, indicating that the raw file now contains has metadata and context
5) `fsync` the parent directory to make that removal persistent

If we don't make it to step 4 + 5, we'll re-run steps 1-3 (because the `.db` file will still be present).  I believe that steps 1-3 are safe to run repeatedly with crashes at any point: truncating the file is idempotent, and resets the effects of step 2.

downstairs/src/extent_inner_raw.rs
```diff
@@ -0,0 +1,1325 @@
...
+         *
+         * If "only_write_unwritten" is true, then we only issue a write for
+         * a block if that block has not been written to yet.  Note
+         * that we can have a write that is "sparse" if the range of
+         * blocks it contains has a mix of written an unwritten
+         * blocks.
+         *
+         * We define a block being written to or not has if that block has
+         * `Some(...)` with a matching checksum serialized into a context slot
+         * or not. So it is required that a written block has a checksum.
```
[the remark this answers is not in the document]
mkeeter: "So it is required that a written block as a checksum."?

(this doesn't make sense to me)

downstairs/src/extent_inner_raw.rs
mkeeter: This is a little subtle, but I think it's correct.

- Let's say that block 0 is using context slot A.
- `set_block_context` will write to context slot B, but _will not_ update `self.active_context`.
- If we then fail the extent data write, we will then actively read the data back from the file and check which context slot is valid
  - If the new data was written, according to reading back from the file (but then the write returned an error code?), we will update the context slot to slot B
  - If the new data wasn't written, we leave the context slot at slot A.  This means that we're in the same place that we started, where the data in the file corresponds to context slot A

downstairs/src/extent_inner_raw.rs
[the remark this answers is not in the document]
jmpesp: Maybe this line was updated after this comment was made haha, I was highlighting the first "has" here:

> We define a block being written to or not has if that block has `Some(...)` with a matching checksum ...

mkeeter:
I did a migration test for a fully populated 128 GiB volume (4KiB block size, 16384 blocks per extent).

To populate the volume, I ran `dd if=/dev/urandom of=/dev/nvme0n1 bs=4194304 count=32768`

Using `kill -USR1 $PID` to track the progress of `dd`, I saw about 259 MiB/sec.

When starting the `raw-file-backend` Crucible downstairs, it took 57 seconds to migrate the data over – basically all in SQLite!

mkeeter:
Migrating a dense 128 GiB region is now down to 42 seconds after 14c8eb788cf74335951141c6f484ff6e5016c039

mkeeter:
I forgot to post this at the time, but reorganizing context slots into AAAAAABBBBBBB (instead of ABABABABAB) + writing them in bulk fixed our `fio` performance numbers for larger writes:

```
main:
  WRITE: bw=15.2MiB/s (15.9MB/s), 15.2MiB/s-15.2MiB/s (15.9MB/s-15.9MB/s), io=910MiB (954MB),
  WRITE: bw=371MiB/s (389MB/s), 371MiB/s-371MiB/s (389MB/s-389MB/s), io=21.8GiB (23.4GB),
  WRITE: bw=348MiB/s (365MB/s), 348MiB/s-348MiB/s (365MB/s-365MB/s), io=20.5GiB (22.0GB),

raw-file-backend
  WRITE: bw=28.3MiB/s (29.7MB/s), 28.3MiB/s-28.3MiB/s (29.7MB/s-29.7MB/s), io=1698MiB (1781MB), 
  WRITE: bw=370MiB/s (388MB/s), 370MiB/s-370MiB/s (388MB/s-388MB/s), io=21.7GiB (23.3GB),
  WRITE: bw=372MiB/s (390MB/s), 372MiB/s-372MiB/s (390MB/s-390MB/s), io=21.9GiB (23.5GB),
```

downstairs/src/extent_inner_sqlite.rs
jmpesp: I'd say this is justified - `sqlite` could have a bug that creates this edge case.

downstairs/src/extent_inner_raw.rs
```diff
@@ -0,0 +1,1757 @@
...
+                for i in 0..8 {
+                    active_context.push(if b & (1 << i) == 0 {
+                        ContextSlot::A
+                    } else {
+                        ContextSlot::B
+                    });
+                }
+            }
+            // It's possible that block count isn't a multiple of 8; in that
+            // case, shrink down the active context array.
```
jmpesp, line 675: We could assert this, what do you think?

jmpesp (commented):
I'm concerned that the on-disk format of the raw file metadata slots is implemented manually in several places and not through for example some shared struct. I'd prefer replacing all the calls to pread/pwrite + calls bincode serialize_into/deserialize with some functions instead.

Something like this would be similar to what is written to disk:

```rust
struct SerializableOnDiskMetadata {
  a_slots: Vec<Option<OnDiskDownstairsBlockContext>>,
  b_slots: Vec<Option<OnDiskDownstairsBlockContext>>,
  active_contexts: Vec<ContextSlot>,
  metadata: OnDiskMeta,
};
```

Then you could imagine:

```rust
pub fn export_meta_and_context(
  &mut self
) -> result<Vec<u8>, CrucibleError> {
  let ctxs = self.get_block_contexts(0, self.extent_size.value)?;
  let needs_rehash = ctxs.iter().any(|c| c.len() > 1);

  let ctxs = if needs_rehash {
    self.fully_rehash_and_clean_all_stale_contexts(true)?;
    self.get_block_contexts(0, self.extent_size.value)?
  } else {
    ctxs
  };

  let ctxs: Vec<Option<OnDiskDownstairsBlockContext>> = 
    ctxs.into_iter().map(|c| {
      match c.len() {
        0 => None,
        1 => Some(OnDiskDownstairsBlockContext {
          block_context: c[0].block_context,
          on_disk_hash: c[0].on_disk_hash,
        }),
        i => panic!("invalid context count: {i}"),
      }
    })
    .collect();

  SerializableOnDiskMetadata::buf_from_contexts(
    ctxs,
    self.dirty()?,
    self.flush_number()?,
    self.gen_number()?,
  )
}
```

where

```rust
impl SerializableOnDiskMetadata {
  pub fn buf_from_contexts(ctxs: Vec<OnDiskDownstairsBlockContext>, dirty: bool, flush_number: u64, gen_number: u64) -> Result<Vec<u8>, CrucibleError> {
    let block_count = ctxs.len();

    let struct = SerializableOnDiskMetadata {
      a_slots: ctxs,
      b_slots: vec![None; block_count],
      active_contexts: vec![ContextSlot::A; block_count],
      metadata: OnDiskMeta {
        dirty,
        flush_number,
        gen_number,
        ext_version: EXTENT_META_RAW, // new extent version for raw files
      }
    };

    struct.write_to_buf()
  }
}
```

Here, `write_to_buf` wraps creating the Vec at the right size and writing in the correct order using the correct methods.

We'd also want to support grabbing only parts of the metadata file, and updating _only_ a subset of the slots (aka a run of blocks that are being read or written to). This would require using the struct's functions but not the struct itself. Taking a block of code from `open`:

```rust
        // Read the active context bitpacked array and metadata.  The former is
        // only valid if `dirty` is false in the metadata, but that's the most
        // common case, so we'll optimize for it.
        let (active_contexts, metadata) = SerializableOnDiskMetadata::meta_from_file(file.as_raw_fd())?;

        let active_contexts = if !metadata.dirty {
          active_contexts
        } else {
          let (a_slots, b_slots) = SerializableOnDiskMetadata::slots_from_file(file.as_raw_fd())?;

          // the code that compares the on-disk extent data with slots
        };
```

and another example:

```rust
    fn set_block_contexts_contiguous(
        &mut self,
        block_contexts: &[DownstairsBlockContext],
    ) -> Result<u64> {
        for (a, b) in block_contexts.iter().zip(block_contexts.iter().skip(1)) {
            assert_eq!(a.block + 1, b.block, "blocks must be contiguous");
        }

        let mut writes = 0u64;

        for (slot, group) in block_contexts
            .iter()
            .group_by(|block_context| {
                // We'll be writing to the inactive slot
                !self.active_context[block_context.block as usize]
            })
            .into_iter()
        {
            let mut group = group.peekable();
            let block_start = group.peek().unwrap().block;

            let ctxs = group.iter().map(|block_context| OnDiskDownstairsBlockContext {
              block_context: block_context.block_context,
              on_disk_hash: block_context.on_disk_hash,
            }).collect();

            writes += SerializableOnDiskMetadata::partial_write_slot_contexts_contiguous(
              self.file.as_raw_fd(),
              slot,
              block_start,
              ctxs,
            )?;
        }

        if let Some(writes) = writes.checked_sub(1) {
            self.extra_syscall_count += writes;
            self.extra_syscall_denominator += 1;
        }

        Ok(writes)
    }
```

Another question - would we want to version `SerializableOnDiskMetadata` as well? I know it already stores extent version in `OnDiskMeta` but this is an opportunity to add a version now where we may want one later.

What are your thoughts?

downstairs/src/extent_inner_raw.rs
```diff
@@ -0,0 +1,1757 @@
...
+        dir: &Path,
+        def: &RegionDefinition,
+        extent_number: u32,
+    ) -> Result<Self> {
+        let path = extent_path(dir, extent_number);
+        let bcount = def.extent_size().value;
+        let size = def.block_size().checked_mul(bcount).unwrap()
+            + BLOCK_META_SIZE_BYTES
+            + (bcount + 7) / 8
+            + BLOCK_CONTEXT_SLOT_SIZE_BYTES * bcount * 2;
```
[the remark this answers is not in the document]
mkeeter: Moved to `RawInner::file_size` in b241df29112c17af4637156de423751213edd51f

downstairs/src/extent_inner_raw.rs
mkeeter: No, having a block count that isn't a multiple of 8 is totally fine – for example, all of our tests call `new_region_options`, which builds an extent with 10 blocks.

downstairs/src/extent_inner_raw.rs
```diff
@@ -0,0 +1,1757 @@
...
+
+        let m = OnDiskMeta {
+            dirty: true,
+            gen_number: u64::MAX,
+            flush_number: u64::MAX,
+            ext_version: u32::MAX,
+        };
+        let mut meta_buf = [0u8; BLOCK_META_SIZE_BYTES as usize];
+        bincode::serialize_into(meta_buf.as_mut_slice(), &Some(m)).unwrap();
+    }
```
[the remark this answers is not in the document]
mkeeter: Good catch – I swore I wrote them, but can't find them in my various experimental branches.

Done in b0482863e6384e1f611eb38be0a0dca244612240

downstairs/src/extent_inner_sqlite.rs
```diff
@@ -508,6 +505,91 @@ impl ExtentInner for SqliteInner {
...
+        let meta = OnDiskMeta {
+            dirty,
+            flush_number,
+            gen_number,
+            ext_version: EXTENT_META_RAW, // new extent version for raw files
+        };
+        let mut meta_buf = [0u8; BLOCK_META_SIZE_BYTES as usize];
+        bincode::serialize_into(meta_buf.as_mut(), &meta)
+            .map_err(|e| CrucibleError::IoError(e.to_string()))?;
+        buf.extend(meta_buf);
```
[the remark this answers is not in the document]
mkeeter: Fixed in 634dae27d650bb420d792495031ea22b90195318

mkeeter:
> I'm concerned that the on-disk format of the raw file metadata slots is implemented manually in several places and not through for example some shared struct. I'd prefer replacing all the calls to pread/pwrite + calls bincode serialize_into/deserialize with some functions instead.

I agree with the goal of having a single struct that isolates details of the raw file format.  My goal was for that to be `struct RawInner`, rather than a new `SerializableOnDiskMetadata`.  This is why `RawInner` has helper functions like `meta_offset` and `context_slot_offset` – it's already the thing that knows how the file is laid out on disk, so I wanted to keep all that info within the type.

Previously, the SQLite export _also_ had to know about the file format.  I changed that in 634dae27d650bb420d792495031ea22b90195318, making the export use normal types, and adding a new `RawInner::import` to generate the meta + context data.

At the end of the day, _someone_ has to be calling `pread` / `pwrite` 😄.  Let me do an audit of `RawInner` and see if there are other places where I could isolate and deduplicate those bits, and then we can talk through remaining concerns.

mkeeter:
@jmpesp After some poking, I _did_ end up making a new class (`RawLayout`) and isolated all of the syscalls into there.  Take a look at 2c41131f109961ef3e2a23d6f69495915548d0ac and see if that lines up with what you had in mind!

downstairs/src/extent_inner_raw.rs
```diff
@@ -0,0 +1,2028 @@
...
+                    block_contexts: vec![block_context]
+                }]
+            );
+        }
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_defragment() -> Result<()> {
```
jmpesp, line 1903: I still think there should be more defragment tests, exercising all the possible cases that can be seen.

integration_tests/src/lib.rs
```diff
@@ -2237,6 +2309,283 @@ mod test {
...
+    }
+
+    #[tokio::test]
+    async fn integration_test_sqlite_migration() -> Result<()> {
+        // Test using an old SQLite backend as a read-only parent.
+
+        const BLOCK_SIZE: usize = 512;
+
+        // boot three downstairs, write some data to them, then change to
+        // read-only.
```
jmpesp, line 2509: `s/read-only/read-write/`

jmpesp (changes requested):

downstairs/src/extent.rs
```diff
@@ -295,26 +311,129 @@ impl Extent {
...
+        //
+        //  - If the .db file is present
+        //      - Truncate it to the length of the extent data. This is a no-op
+        //        normally, but lets us recover from a partial migration (e.g.
+        //        if we wrote the raw context but crashed before deleting the
+        //        .db file)
+        //      - Open the extent using our existing SQLite extent code
+        //      - Using standard extent APIs, find the metadata and encryption
+        //        context for each block. Encode this in the new raw file format
+        //        (as a Vec<u8>)
```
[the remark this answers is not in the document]
mkeeter: Good catch, fixed in 579984d14f93edba512b8e0df55e155256234a08

downstairs/src/extent_inner_raw.rs
```diff
@@ -0,0 +1,2028 @@
...
+            .ok_or(CrucibleError::MissingContextSlot(block))?;
+        self.active_context[block as usize] = value;
+        Ok(())
+    }
+
+    fn set_block_contexts(
+        &mut self,
+        block_contexts: &[DownstairsBlockContext],
+    ) -> Result<(), CrucibleError> {
+        // If any of these block contexts will be overwriting an unsyched
```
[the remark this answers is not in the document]
mkeeter: oops, fixed in a607fab2cdd99049729ba284c2b676bf79bc674e

downstairs/src/extent_inner_raw.rs
```diff
@@ -0,0 +1,2028 @@
...
+        &self,
+        file: &File,
+        block_start: u64,
+        iter: I,
+        slot: ContextSlot,
+    ) -> Result<(), CrucibleError>
+    where
+        I: Iterator<Item = Option<&'a DownstairsBlockContext>>,
+    {
+        let mut buf = self.buf.take();
```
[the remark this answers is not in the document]
mkeeter: Yup, exactly.  This is just to prevent allocation churn, so I didn't want to get too fancy.

integration_tests/src/lib.rs
mkeeter: Fixed in 5f88dba07ad019e12d3b486186936f3b7e76334f

downstairs/src/extent_inner_raw.rs
```diff
@@ -0,0 +1,2028 @@
...
+    pub fn import(
+        file: &mut File,
+        def: &RegionDefinition,
+        ctxs: Vec<Option<DownstairsBlockContext>>,
+        dirty: bool,
+        flush_number: u64,
+        gen_number: u64,
+    ) -> Result<(), CrucibleError> {
+        let layout = RawLayout::new(def.extent_size());
+        let block_count = layout.block_count() as usize;
```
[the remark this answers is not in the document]
mkeeter: Okay, done in 9c7c7c2e7afd8fd462a79416cb7647b0806af199

downstairs/src/extent_inner_raw.rs
```diff
@@ -0,0 +1,2028 @@
...
+        Ok(out)
+    }
+
+    /// Write out the active context array and metadata section of the file
+    ///
+    /// This is done in a single write, so it should be atomic.
+    fn write_active_context_and_metadata(
+        &self,
+        file: &File,
+        active_context: &[ContextSlot],
```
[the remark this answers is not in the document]
mkeeter: Done in fa37c53d51deeec889b1fe6ff57cedea1d089fee (with an assertion, since screwing this up is bad)

downstairs/src/extent_inner_raw.rs
```diff
@@ -0,0 +1,2028 @@
...
+
+    fn write_context_slots_contiguous<'a, I>(
+        &self,
+        file: &File,
+        block_start: u64,
+        iter: I,
+        slot: ContextSlot,
+    ) -> Result<(), CrucibleError>
+    where
+        I: Iterator<Item = Option<&'a DownstairsBlockContext>>,
```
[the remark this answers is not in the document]
mkeeter: I experimented with changing this signature, with mixed results.  It turns out that we use the possibility of writing `None` in a few places:

- When migrating instances, to write the context slots in bulk
- When defragmenting context slots, for the same reason

It's possible to work around this, but it makes this code much trickier (see the diff below).

Because this is a private function – no one outside this module can call it – I'd vote to leave it as is, but would reconsider if you feel strongly about it.

For reference, here's the diff:

```diff
diff --git a/downstairs/src/extent_inner_raw.rs b/downstairs/src/extent_inner_raw.rs
index b8459d8..6d8a624 100644
--- a/downstairs/src/extent_inner_raw.rs
+++ b/downstairs/src/extent_inner_raw.rs
@@ -545,18 +545,22 @@ impl RawInner {
         assert_eq!(block_count, ctxs.len());
 
         file.set_len(layout.file_size())?;
-        layout.write_context_slots_contiguous(
-            file,
-            0,
-            ctxs.iter().map(Option::as_ref),
-            ContextSlot::A,
-        )?;
-        layout.write_context_slots_contiguous(
-            file,
-            0,
-            std::iter::repeat(None).take(block_count),
-            ContextSlot::B,
-        )?;
+        for chunk in ctxs
+            .iter()
+            .group_by(|b| b.is_some())
+            .into_iter()
+            .filter(|(some, _group)| *some)
+            .map(|(_some, group)| group)
+        {
+            layout.write_context_slots_contiguous(
+                file,
+                0,
+                chunk.map(|b| b.as_ref().unwrap()),
+                ContextSlot::A,
+            )?;
+        }
+        // `ContextSlot::B` is already initialized to all zeros, which
+        // deserializes to `None`.
 
         layout.write_active_context_and_metadata(
             file,
@@ -914,10 +918,7 @@ impl RawInner {
             let mut group = group.peekable();
             let start = group.peek().unwrap().block;
             self.layout.write_context_slots_contiguous(
-                &self.file,
-                start,
-                group.map(Option::Some),
-                slot,
+                &self.file, start, group, slot,
             )?;
             writes += 1;
         }
@@ -1113,12 +1114,25 @@ impl RawInner {
                     self.sync_index + 1;
             }
         }
-        let r = self.layout.write_context_slots_contiguous(
-            &self.file,
-            counter.min_block,
-            dest_slots.iter().map(|v| v.as_ref()),
-            !copy_from,
-        );
+
+        let mut r = Ok(());
+        for chunk in dest_slots
+            .iter()
+            .group_by(|b| b.is_some())
+            .into_iter()
+            .filter(|(some, _group)| *some)
+            .map(|(_some, group)| group)
+        {
+            if let Err(e) = self.layout.write_context_slots_contiguous(
+                &self.file,
+                0,
+                chunk.map(|b| b.as_ref().unwrap()),
+                !copy_from,
+            ) {
+                r = Err(e);
+                break;
+            }
+        }
 
         // If this write failed, then try recomputing every context slot
         // within the unknown range
@@ -1251,7 +1265,7 @@ impl RawLayout {
         slot: ContextSlot,
     ) -> Result<(), CrucibleError>
     where
-        I: Iterator<Item = Option<&'a DownstairsBlockContext>>,
+        I: Iterator<Item = &'a DownstairsBlockContext>,
     {
         let mut buf = self.buf.take();
         buf.clear();
@@ -1259,11 +1273,11 @@ impl RawLayout {
         for block_context in iter {
             let n = buf.len();
             buf.extend([0u8; BLOCK_CONTEXT_SLOT_SIZE_BYTES as usize]);
-            let d = block_context.map(|b| OnDiskDownstairsBlockContext {
-                block_context: b.block_context,
-                on_disk_hash: b.on_disk_hash,
-            });
-            bincode::serialize_into(&mut buf[n..], &d).unwrap();
+            let d = OnDiskDownstairsBlockContext {
+                block_context: block_context.block_context,
+                on_disk_hash: block_context.on_disk_hash,
+            };
+            bincode::serialize_into(&mut buf[n..], &Some(d)).unwrap();
         }
         let offset = self.context_slot_offset(block_start, slot);
         nix::sys::uio::pwrite(file.as_raw_fd(), &buf, offset as i64).map_err(
```

downstairs/src/extent_inner_raw.rs
```diff
@@ -0,0 +1,2373 @@
...
+        // within the unknown range
+        if r.is_err() {
+            for block in counter.min_block..=counter.max_block {
+                self.recompute_slot_from_file(block).unwrap();
+            }
+        } else {
+            for block in counter.min_block..=counter.max_block {
+                self.active_context[block as usize] = !copy_from;
+            }
+        }
```
jmpesp, line 1133: At this point in the function (if there was no error in writing the context slots), the in memory active context array is different than what active contexts were previously serialized to the file. The second of the following asserts will fire during the defragment tests:

```patch
diff --git a/downstairs/src/extent.rs b/downstairs/src/extent.rs
index 3457512..bbe4cd9 100644
--- a/downstairs/src/extent.rs
+++ b/downstairs/src/extent.rs
@@ -77,7 +77,7 @@ pub(crate) trait ExtentInner: Send + Debug {
 }
 
 /// BlockContext, with the addition of block index and on_disk_hash
-#[derive(Copy, Clone)]
+#[derive(Debug, Copy, Clone, PartialEq)]
 pub struct DownstairsBlockContext {
     pub block_context: BlockContext,

diff --git a/downstairs/src/extent_inner_raw.rs b/downstairs/src/extent_inner_raw.rs
index 0e4f8ea..a35b1ca 100644
--- a/downstairs/src/extent_inner_raw.rs
+++ b/downstairs/src/extent_inner_raw.rs
@@ -1131,6 +1140,10 @@ impl RawInner {
                 self.active_context[block as usize] = !copy_from;
             }
         }
+
+        assert!(!self.dirty);
+        assert_eq!(self.active_context, self.layout.get_active_contexts(&self.file)?);
+
         r.map(|_| ())
     }
 ```

Also the dirty bit is not set (as we just did a flush): if there is a crash now, `Extent::open` will see that the dirty bit is not set and deserialize the active contexts that were previously written to the file, and _not_ validate those against the block's data on disk.

However, I _think_ this is ok: we only defragment after a flush, so we know what that that the block data contents are on durable storage and that the active context slot is correct and matches that. Defragmentation only copies the contents of the active context slot to the inactive context slot, so after defragmentation both slot A and B contain the same (valid) context. The context will be correct for the block data no matter which one is picked.

The following proves that:

```patch
diff --git a/downstairs/src/extent.rs b/downstairs/src/extent.rs
index 3457512..bbe4cd9 100644
--- a/downstairs/src/extent.rs
+++ b/downstairs/src/extent.rs
@@ -77,7 +77,7 @@ pub(crate) trait ExtentInner: Send + Debug {
 }
 
 /// BlockContext, with the addition of block index and on_disk_hash
-#[derive(Copy, Clone)]
+#[derive(Debug, Copy, Clone, PartialEq)]
 pub struct DownstairsBlockContext {
     pub block_context: BlockContext,

diff --git a/downstairs/src/extent_inner_raw.rs b/downstairs/src/extent_inner_raw.rs
index 0e4f8ea..585e18d 100644
--- a/downstairs/src/extent_inner_raw.rs
+++ b/downstairs/src/extent_inner_raw.rs
@@ -1065,6 +1065,13 @@ impl RawInner {
             max_block: 0,
         };
         let mut b_count = a_count;
+
+        // PARANOID: make sure these match
+        assert_eq!(
+            self.active_context,
+            self.layout.get_active_contexts(&self.file)?,
+        );
+
         for (i, s) in self.active_context.iter().enumerate() {
             let count = match s {
                 ContextSlot::A => &mut a_count,
@@ -1086,6 +1093,7 @@ impl RawInner {
         assert!(counter.count > 0);
         assert!(counter.max_block >= counter.min_block);
         let num_slots = counter.max_block + 1 - counter.min_block;
+        assert!(num_slots > 0);
 
         // Read the source context slots from the file
         let source_slots = self.layout.read_context_slots_contiguous(
@@ -1113,6 +1121,7 @@ impl RawInner {
                     self.sync_index + 1;
             }
         }
+
         let r = self.layout.write_context_slots_contiguous(
             &self.file,
             counter.min_block,
@@ -1131,6 +1140,94 @@ impl RawInner {
                 self.active_context[block as usize] = !copy_from;
             }
         }
+
+        // Here, self.active_context does not match what is on disk, and dirty
+        // is not set. If there is a crash now, Extent::open will be called, and
+        // active context will be read from the disk and _not_ validated against
+        // the file data. This is ok if we defragment only after a flush has
+        // successfully occurred: due to how defragmentation works, when there
+        // is a mismatch between what is in memory now and what is on disk, the
+        // two context slots on disk are duplicates of each other, and match the
+        // file data on disk (because we just did a successful flush).
+        //
+        // Go through _every_ block, and debug assert that when there is a
+        // mismatch the two context slots on disk are duplicates of each other,
+        // and match the file data that is on disk.
+        assert!(!self.dirty);
+        assert_ne!(self.active_context, self.layout.get_active_contexts(&self.file)?);
+
+        let block_size = self.extent_size.block_size_in_bytes();
+        for (i, (in_memory_slot, on_disk_slot)) in std::iter::zip(self.active_context.iter(), self.layout.get_active_contexts(&self.file)?.iter()).enumerate() {
+            let block = i as u64;
+
+            let (buf, hash) = {
+                let mut buf = vec![0; block_size as usize];
+
+                nix::sys::uio::pread(
+                    self.file.as_raw_fd(),
+                    &mut buf,
+                    (block_size as u64 * block) as i64,
+                )
+                .map_err(|e| {
+                    CrucibleError::IoError(format!(
+                        "extent {}: reading block {} data failed: {e}",
+                        self.extent_number,
+                        block,
+                    ))
+                })?;
+
+                let hash = integrity_hash(&[&buf]);
+
+                (buf, hash)
+            };
+
+            if in_memory_slot != on_disk_slot {
+                // The in memory active slot does not match the on-disk active
+                // slot. Make sure that the on disk contexts are duplicates, and
+                // that they both match the on disk file data.
+                let a_context = self.layout.read_context_slots_contiguous(
+                    &self.file,
+                    block,
+                    1,
+                    ContextSlot::A,
+                )?;
+                let b_context = self.layout.read_context_slots_contiguous(
+                    &self.file,
+                    block,
+                    1,
+                    ContextSlot::B,
+                )?;
+                debug_assert_eq!(a_context, b_context);
+
+                if let Some(a_context) = a_context[0] {
+                    // If the slot is Some, then there is encrypted data there.
+                    debug_assert_eq!(a_context.block, block);
+                    debug_assert_eq!(a_context.on_disk_hash, hash);
+                } else {
+                    // If the slot is None, then the data should be all zeros
+                    debug_assert_eq!(buf, vec![0u8; block_size as usize]);
+                }
+            } else {
+                // The in memory active slot matches the on disk active slot.
+                // Validate the on disk active slot while we're here.
+                let on_disk_context = self.layout.read_context_slots_contiguous(
+                    &self.file,
+                    block,
+                    1,
+                    *on_disk_slot,
+                )?;
+
+                if let Some(on_disk_context) = on_disk_context[0] {
+                    // If the slot is Some, then there is encrypted data there.
+                    debug_assert_eq!(on_disk_context.block, block);
+                    debug_assert_eq!(on_disk_context.on_disk_hash, hash);
+                } else {
+                    // If the slot is None, then the data should be all zeros
+                    debug_assert_eq!(buf, vec![0u8; block_size as usize]);
+                }
+            }
+        }
+
         r.map(|_| ())
     }
 }
```

What do you think?

mkeeter: Excellent analysis, I agree with both parts (the cached context slots on disk disagree with active context slots in-memory, despite dirty being false; and this is still safe, because the cached context slots contain valid contexts)

mkeeter: I also added another assertion in 5ed99f7c00aee9b0cfb3f9c4bf7d8d4b42e99959 to make sure no one tries to call `defragment` on a dirty extent.

jmpesp:
Thinking about all possible messages that the Upstairs could send, there's a strange bug with the raw backend that occurs if the Upstairs sends multiple writes to the same offset in the same write message.

The sqlite backend records every one and return them in the read response (note this requires commenting out the assertion that there is no existing value in `dirty_blocks`):

```diff
diff --git a/downstairs/src/extent_inner_sqlite.rs b/downstairs/src/extent_inner_sqlite.rs
index 34056c7..bc4b806 100644
--- a/downstairs/src/extent_inner_sqlite.rs
+++ b/downstairs/src/extent_inner_sqlite.rs
@@ -474,7 +474,7 @@ impl ExtentInner for SqliteInner {
                 let prev = self
                     .dirty_blocks
                     .insert(write.offset.value as usize, Some(*h));
-                assert_eq!(prev, Some(None));
+                //assert_eq!(prev, Some(None));
             }
         }
         Ok(())
@@ -1665,4 +1665,67 @@ mod test {
 
         Ok(())
     }
+
+    /// Test that multiple writes to the same location work
+    #[test]
+    fn test_multiple_writes_to_same_location_sqlite() -> Result<()> {
+        let dir = tempdir()?;
+        let mut inner =
+            SqliteInner::create(dir.as_ref(), &new_region_definition(), 0)
+                .unwrap();
+
+        // Write the same block four times in the same write command.
+
+        let writes: Vec<_> = (0..4)
+            .map(|i| {
+                let data = Bytes::from(vec![i as u8; 512]);
+                let hash = integrity_hash(&[&data[..]]);
+
+                crucible_protocol::Write {
+                    eid: 0,
+                    offset: Block::new_512(0),
+                    data,
+                    block_context: BlockContext {
+                        encryption_context: None,
+                        hash,
+                    },
+                }
+            })
+            .collect();
+
+        let write_refs: Vec<_> = writes.iter().collect();
+
+        inner.write(JobId(30), &write_refs, false, IOV_MAX_TEST)?;
+
+        // Block 0 should be 0x03 repeated.
+        let mut resp = Vec::new();
+        let read = ReadRequest {
+            eid: 0,
+            offset: Block::new_512(0),
+        };
+
+        inner.read(JobId(31), &[&read], &mut resp, IOV_MAX_TEST)?;
+
+        let data = Bytes::from(vec![0x03; 512]);
+
+        assert_eq!(resp.len(), 1);
+        assert_eq!(resp[0].data, BytesMut::from(data.as_ref()));
+
+        // All block contexts should be returned
+        assert_eq!(
+            resp[0]
+                .block_contexts
+                .iter()
+                .cloned()
+                .collect::<HashSet<_>>(),
+            HashSet::from([
+                writes[0].block_context,
+                writes[1].block_context,
+                writes[2].block_context,
+                writes[3].block_context,
+            ]),
+        );
+
+        Ok(())
+    }
 }
diff --git a/protocol/src/lib.rs b/protocol/src/lib.rs
index 9c559b8..3d8c55c 100644
--- a/protocol/src/lib.rs
+++ b/protocol/src/lib.rs
@@ -156,8 +156,7 @@ impl ReadResponse {
     }
 }
 
-#[allow(clippy::derive_partial_eq_without_eq)]
-#[derive(Copy, Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Hash)]
 pub struct BlockContext {
     /// If this is a non-encrypted write, then the integrity hasher has the
     /// data as an input:
@@ -182,8 +181,7 @@ pub struct BlockContext {
     pub encryption_context: Option<EncryptionContext>,
 }
 
-#[allow(clippy::derive_partial_eq_without_eq)]
-#[derive(Copy, Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Hash)]
 pub struct EncryptionContext {
     pub nonce: [u8; 12],
     pub tag: [u8; 16],
```

As I understand it, given this input the raw backend should ping pong between context slots (flushing a few times) but ultimately return only the last one. However, this test fails:

```diff
diff --git a/downstairs/src/extent_inner_raw.rs b/downstairs/src/extent_inner_raw.rs
index 6ed9e15..fbdf785 100644
--- a/downstairs/src/extent_inner_raw.rs
+++ b/downstairs/src/extent_inner_raw.rs
@@ -2372,4 +2372,65 @@ mod test {
         let mut meta_buf = [0u8; BLOCK_META_SIZE_BYTES as usize];
         bincode::serialize_into(meta_buf.as_mut_slice(), &Some(m)).unwrap();
     }
+
+    /// Test that multiple writes to the same location work
+    #[test]
+    fn test_multiple_writes_to_same_location_raw() -> Result<()> {
+        let dir = tempdir()?;
+        let mut inner =
+            RawInner::create(dir.as_ref(), &new_region_definition(), 0)
+                .unwrap();
+
+        // Write the same block four times in the same write command.
+
+        let writes: Vec<_> = (0..4)
+            .map(|i| {
+                let data = Bytes::from(vec![i as u8; 512]);
+                let hash = integrity_hash(&[&data[..]]);
+
+                crucible_protocol::Write {
+                    eid: 0,
+                    offset: Block::new_512(0),
+                    data,
+                    block_context: BlockContext {
+                        encryption_context: None,
+                        hash,
+                    },
+                }
+            })
+            .collect();
+
+        let write_refs: Vec<_> = writes.iter().collect();
+
+        inner.write(JobId(30), &write_refs, false, IOV_MAX_TEST)?;
+
+        // Block 0 should be 0x03 repeated.
+        let mut resp = Vec::new();
+        let read = ReadRequest {
+            eid: 0,
+            offset: Block::new_512(0),
+        };
+
+        inner.read(JobId(31), &[&read], &mut resp, IOV_MAX_TEST)?;
+
+        let data = Bytes::from(vec![0x03; 512]);
+        let hash = integrity_hash(&[&data[..]]);
+        let block_context = BlockContext {
+            encryption_context: None,
+            hash,
+        };
+
+        assert_eq!(
+            resp,
+            vec![ReadResponse {
+                eid: 0,
+                offset: Block::new_512(0),
+                data: BytesMut::from(data.as_ref()),
+                // Only the most recent block context should be returned
+                block_contexts: vec![block_context],
+            }]
+        );
+
+        Ok(())
+    }
 }
```

because _no_ contexts are returned:

```
thread 'extent_inner_raw::test::test_multiple_writes_to_same_location_raw' panicked at 'assertion failed: `(left == right)`
  left: `[ReadResponse { eid: 0, offset: Block { value: 0, shift: 9 }, data: b"\x03\x03\x03...<snip>", block_contexts: [] }]`,
 right: `[ReadResponse { eid: 0, offset: Block { value: 0, shift: 9 }, data: b"\x03\x03\x03...<snip>", block_contexts: [BlockContext { hash: 12464899016011810304, encryption_context: None }] }]`', downstairs/src/extent_inner_raw.rs:2423:9
```

downstairs/src/extent_inner_raw.rs
```diff
@@ -0,0 +1,2375 @@
...
+        block_start: u64,
+        block_count: u64,
+        slot: ContextSlot,
+    ) -> Result<Vec<Option<DownstairsBlockContext>>, CrucibleError> {
+        let mut buf = self.buf.take();
+        buf.resize((BLOCK_CONTEXT_SLOT_SIZE_BYTES * block_count) as usize, 0u8);
+
+        let offset = self.context_slot_offset(block_start, slot);
+        nix::sys::uio::pread(file.as_raw_fd(), &mut buf, offset as i64)
+            .map_err(|e| {
```
leftwo, line 1294: We don't call `buf.clear()` here, 
Is their any chance the `pread` will return with a partially filled buffer?
It's only us who use the buffer in the downstairs, so there should not be
a chance for someone to get old data from a previous read, but just
wondering here if that is a possibility.

downstairs/src/extent_inner_raw.rs
```diff
@@ -0,0 +1,2375 @@
...
+        {
+            // Unpack bits from each byte
+            active_context.push(if bit {
+                ContextSlot::A
+            } else {
+                ContextSlot::B
+            });
+        }
+        assert_eq!(active_context.len(), self.block_count() as usize);
+        Ok(active_context)
```
leftwo, line 1398: You did not `self.buf.set(buf);` here. Is that on purpose?

downstairs/src/region.rs
```diff
@@ -28,25 +28,12 @@ use crate::extent::{
 
 /**
  * Validate a list of sorted repair files.
- * There are either two or four files we expect to find, any more or less
- * and we have a bad list.  No duplicates.
+ *
```
leftwo, line 31: We could probably update this comment to be "Validate the repair file"

leftwo (approved):
Just a few small questions/comments.
Ready to see this one go back!

mkeeter:
@jmpesp That's a good find.  The strategy of "write context slots, then write data" breaks down if the same block is written more than once, because (like you said) we can only store 2 context slots simultaneously, not _N_.

I've added that test and implemented a fix in 06fa0d9fa35f940e8a9672ccea96b278f08d3d30.  The main idea is
- Check whether a write includes duplicate blocks
- If so, split it into multiple rounds of "write context slots, then write data", where each round contains non-duplicate sets of blocks.  This is slower, but is guaranteed to be crash consistent.

The implementation is a _little_ clever, because I wanted to avoid allocation in this path, so I'm using a scratch buffer and rolling index to do a fast uniqueness check (which may rarely return false negatives).  False negatives are fine, because falling back to the slower path is still correct.

downstairs/src/extent_inner_raw.rs
mkeeter: Good question!  Quote [the standard](https://pubs.opengroup.org/onlinepubs/9699919799/functions/pread.html),

> The value returned may be less than nbyte if the number of bytes left in the file is less than nbyte, if the read() request was interrupted by a signal, or if the file is a pipe or FIFO or special file and has fewer than nbyte bytes immediately available for reading. For example, a read() from a file associated with a terminal may return one typed line of data.

The only concerning part here is "interrupted by a signal"; I think everything else isn't relevant.  Do you think we need to wrap every `pread(v)/pwrite(v)` in a little loop?  Right now, the SQLite backend is using `pread` without any special handling.

downstairs/src/extent_inner_raw.rs
mkeeter: Nope, fixed in a8b3705f33851d1c4a6d66916b066f0d5e299b02

downstairs/src/extent_inner_raw.rs
mkeeter: I've added some handling of this in 2a5ed27c32caa2922e620a5e5da7cb467ceeff4f

Specifically, `pread` and `pwrite` will now loop until the full buffer is read / written, and `preadv` / `pwritev` will return an `IoError` (because I didn't want to write that handling logic yet).

downstairs/src/extent_inner_raw.rs
mkeeter: Yup, added a comment in 971ed1e97e24e38bbd24e6991abde4467126b5db

downstairs/src/extent_inner_raw.rs
leftwo: Looks good, and I think not having the error retry loop is fine as I don't believe we have ever seen this error, and if we do, it's probably something catastrophic going on anyway.

downstairs/src/extent_inner_raw.rs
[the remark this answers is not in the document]
jmpesp: Circling back here, are those two cases still true?

- When migrating, the context slots start as all zeroes, which deserialize as None.
- After defragmenting, both context slots contain the same data - the current code doesn't assign None to overwrite one of them.

downstairs/src/extent_inner_raw.rs
jmpesp: :+1:

mkeeter:
After the last two commits (868bca4c5d87c413bdb005da6c3e962c1a3361a5 and 0d4f5731cfd27fbaa628e9c593d25457b033752a), we're using significantly less RAM than `main`.

| | `main` | `raw-file-backend`|
|-|-|-|
| startup | 394M | 136M |
| after running tests | 2708M | 448M |

(measured using the `RES` column in `htop`)

The tests are my usual suite of `fio` + `pgbench`.

jmpesp:
Let's say for whatever reason the Upstairs is talking to one downstairs that's a sqlite backend, and two others that are both raw backends. Both are message version 5. After an unclean shutdown, or if a write lands on one but not the other before shutdown, the Upstairs may decide to reconcile when reconnecting to the set based on a mismatch here:

https://github.com/oxidecomputer/crucible/blob/6df8a67a74ccf840029f9bfa0b63e4696f3bd89b/upstairs/src/lib.rs#L6678-L6707

If that's the case, it'll start the repair process, which cannot succeed due to backend differences. I'm thinking that the downstairs needs to report which backend it's using to the Upstairs, but if this is the case then the Upstairs has to panic - repair is required but can't happen.

What do you think? Is this too contrived a scenario?

mkeeter:
@jmpesp I think the initial situation ("the Upstairs is talking to one downstairs that's a sqlite backend, and two others that are both raw backends") is too contrived.  The downstairs should always be running the raw file backend, _unless_ it's a previously-existing SQLite-backed volume **and** `read_only` is true when we open it (since we automatically migrate existing SQLite-backed extents if they aren't read-only).

In other words, to arrive in that initial situation would require the three Downstairs to disagree about whether a volume is `read_only`; in that situation, I suspect many other things will also go wrong!

faithanalog:
My testing setup
- 2048 extents
- 131072 blocks per extent
- 512-byte block
- 128GiB disk
- no disk encryption (so no encryption overhead)
- all fio tests
	- 120 seconds
	- direct IO
	- io_uring
	- io depth 48
	- `sync` between fio invocations

### observations of interest

- CPU usage is high on small writes
	- randwrite-16k: each downstairs is using 10-14 cores worth of CPU, 30% of total gimlet when you combine all 3
	- We are getting significant downstairs parallelization!!!
	- I think we can limit this by configuring the number of tokio workers if we want, but do we want to?
		- limiting this way might reduce burst random IOPS though
		- consider that most people are not doing random writes across all extents on the disk as fast as possible, especially not for long periods.  

- ram
	- measured by RES column in `htop` (NOT VIRT)
	- startup: 450MiB / downstairs
	- during initial long sustained sequential write, grows to 2.7GiB and holds steady
	- By the end of running all FIO tests, ram usage was different on 3 downstairs
		- 6.6 GiB / 4.4 GiB / 4.8 GiB
		- Why is it different? I don't know yet.
		- Not sure where the memory usage is coming from. Cached metadata in ram should be a small fraction of this size.
	- unclear whether RAM usage grows unbounded or eventually caps out.
	- (TODO: is it only on write ops?)

- high backpressure under 4M randrw
	- backpressure capped at 76,000
	- one of the downstairs was moved to `offline` for 2 seconds
		- reconnected, did not go into live repair when it came back
	- took 35 seconds to drain
	- pure-write patterns do not end up in this backpressure situation
	- smaller randrws (4k, 16k) do not cause this
	- Unless you can invent a scenario where a customer will hit this, maybe not an immediate concern.

- I managed to cause a OS-level disk error
	- error was NOT on latest commit. on 622ef9b52e9fe4288d0bc86f3566b5219be5ca5e
		- all other test results in this comment are on latest though.
	- I have had it happen twice but it doesn't happen every time.
	- TODO: can i reproduce on latest?
		- going to re-run the benches a bunch to see if it triggers
	- Previously I was only running pure-read and pure-write tests. With the move to iodriver I started including rw which does interleaved reads and writes
	- UPW stuck at 93, not changing
	- DSW stuck at 6281, not changing
	- BAKPR stuck at 40836, not changing
	- Propolis had to be force-killed
	- Sequence:
		- longwrite - special 300 second 4M write followed by brief 4k read
		- randread-16k
		- randread-4M
		- randread-4k
		- randrw-16k
		- randrw-4M
		- randrw-4k <-- error

### numbers

- Long sequential 4M write:
	- 509MiB/s
	- only 5ms latency cost incurred on read after the write finished!
- rand read:
	- 4K: 8292 IOPS
	- 16K: 4546 IOPS
	- 4M: 25 IOPS (106MiB/s)
- rand write:
	- 4K: 5194 IOPS
	- 16K: 3700 IOPS
	- 4M: 56 IOPS (224MiB/s)
- rand rw:
	- 4K: 2751/2751 rw IOPS
		- notice the falloff vs pure read or pure write 4K
	- 16K: 2469/2469 rw IOPS
	- 4M: 10/10 rw IOPS (40MiB/s duplex)
- seq read:
	- 4K: 8035 IOPS
	- 16K: 4199 IOPS
	- 4M: 24 IOPS (99.4MiB/s)
- seq write:
	- 4K: 10,600 IOPS
	- 16K: 9006 IOPS
	- 4M: 104 IOPS (417MiB/s)
		- lower than initial longwrite - TODO: overhead from region with data already written to it, or overhead of long-running downstairs process?
- seq rw:  
	- 4K: 2795/2795 rw IOPS
	- 16K: 4321/4321 rw IOPS 
	- 4M: 22/22 rw IOPS (much better than rand rw)


- pgbench
	- unsure how to interpret these numbers but 520 tps on stock pgbench settings/test

jmpesp:
@mkeeter another contrived scenario: one sqlite backend, two raw backends, both read-write this time. I was concerned that the Upstairs would panic if it received N block contexts from the sqlite backend but only one from the raw backends, if there was a mismatch in length of `read_response_hashes` here:

https://github.com/oxidecomputer/crucible/blob/6df8a67a74ccf840029f9bfa0b63e4696f3bd89b/upstairs/src/lib.rs#L4439-L4467

But this isn't the case. I reproduced this with a (admittedly contrived!) integration test and the Upstairs filters out all but the correct hash [here](https://github.com/oxidecomputer/crucible/blob/6df8a67a74ccf840029f9bfa0b63e4696f3bd89b/upstairs/src/lib.rs#L4222-L4226) and [here](https://github.com/oxidecomputer/crucible/blob/6df8a67a74ccf840029f9bfa0b63e4696f3bd89b/upstairs/src/lib.rs#L4254-L4259).

__However__, I agree with what you said about the situation being too contrived though. If we were _not_ automatically upgrading every read-write downstairs this way, then it could happen, but every sqlite backend should be read-only after this commit lands. We'll have:

- old read-only downstairs that use the sqlite backend
- new read-write downstairs that will use the raw backend
- new read-only downstairs that are created from those, which use the raw backend

The only way we would end up with a read-write downstairs using the sqlite backend after this commit lands is if a failure to migrate a read-write downstairs to the raw backend is ignored by the Downstairs, and it starts anyway. This isn't what we do though!

jmpesp (approved):

mkeeter:
I did some more poking at RAM, and generally feel good about merging this.

Running the same volume shape as @faithanalog (128 GiB, 64 MiB extents, 512-byte blocks, unencrypted), I did get `RES` usage up to 2.7 GiB running some `fio` tests.  Grabbing a core with `gcore` and looking at it with `::umastat` in `mdb`, here's the relevant stuff:

```
cache                        buf     buf     buf     buf  memory     alloc alloc
name                        size  in use  in ptc   total  in use   succeed  fail
------------------------- ------ ------- ------- ------- ------- --------- -----
umem_magazine_1               16    3848       -    4032     64K      4014     0
umem_magazine_3               32    1039       -    2142     68K      3474     0
umem_magazine_7               64     219       -     378     24K      2533     0
umem_magazine_15             128     184       -    1054    136K      1325     0
umem_magazine_31             256     135       -     690    184K       961     0
umem_magazine_47             384     128       -     350    140K       730     0
umem_magazine_63             512     128       -     455    260K       459     0
umem_magazine_95             768   23738       -   23745   18.6M     23743     0
umem_magazine_143           1152      10       -      12     16K        10     0
umem_slab_cache               56  374363       -  374409   23.2M    374885     0
umem_bufctl_cache             24 2235372       - 2235744   69.3M   2238899     0
umem_bufctl_audit_cache      192       0       -       0       0         0     0
umem_alloc_8                   8       0       0       0       0         0     0
umem_alloc_16                 16   78754   78726   78876   1.22M     78754     0
umem_alloc_32                 32     599     452    3780    120K       599     0
umem_alloc_48                 48  368795  368505  368844   17.2M    368795     0
umem_alloc_64                 64     748     222     756     48K       748     0
umem_alloc_80                 80    2746     428    2750    220K      2746     0
umem_alloc_96                 96     248     115     252     24K       248     0
umem_alloc_112               112    2529     196    2556    284K      2529     0
umem_alloc_128               128   44016   41838   44020   5.55M     44016     0
umem_alloc_160               160     262     257     275     44K       262     0
umem_alloc_192               192     298     297     315     60K       298     0
umem_alloc_224               224    7025    7021    7038   1.53M      7025     0
umem_alloc_256               256     258     258     270     72K       258     0
umem_alloc_320               320     317     315     324    108K       317     0
umem_alloc_384               384     210     171     210     84K       210     0
umem_alloc_448               448     279     279     279    124K       279     0
umem_alloc_512               512       1       -     272    136K    446022     0
umem_alloc_640               640       2       - 2218494   1.41G 110102664     0
umem_alloc_768               768       2       -     405    324K      3761     0
umem_alloc_896               896       0       -   10710   9.30M    862946     0
umem_alloc_1152             1152       4       -     357    408K    336394     0
umem_alloc_1344             1344       1       -     198    264K       440     0
umem_alloc_1600             1600       0       -     270    432K     70220     0
umem_alloc_2048             2048       0       -     274    548K     59246     0
umem_alloc_2304             2304     129       -     434    992K    276587     0
umem_alloc_2688             2688       0       -      78    208K       175     0
umem_alloc_4096             4096       0       -     156    624K    128351     0
umem_alloc_4544             4544       6       -     944   4.15M    279220     0
umem_alloc_8192             8192       1       -     157   1.23M    128585     0
umem_alloc_9216             9216       1       -     188   1.65M    275021     0
umem_alloc_12288           12288       0       -      84   1008K       152     0
umem_alloc_16384           16384       3       -     189   2.95M    128901     0
umem_alloc_24576           24576       1       -     249   5.84M    271548     0
umem_alloc_32768           32768       0       -     254   7.94M    126536     0
umem_alloc_40960           40960       1       -     244   9.53M    300132     0
umem_alloc_49152           49152       0       -      78   3.66M     68793     0
umem_alloc_57344           57344       0       -     325   17.8M     74740     0
umem_alloc_65536           65536       0       -     130   8.12M     50137     0
umem_alloc_73728           73728       0       -      93   6.54M      2246     0
umem_alloc_81920           81920       0       -     178   13.9M      2404     0
umem_alloc_90112           90112       0       -      53   4.55M       120     0
umem_alloc_98304           98304       0       -     173   16.2M      2956     0
umem_alloc_106496         106496       0       -     190   19.3M     47781     0
umem_alloc_114688         114688       0       -       0       0         0     0
umem_alloc_122880         122880       0       -     195   22.9M     50837     0
umem_alloc_131072         131072       0       -       0       0         0     0
------------------------- ------ ------- ------- ------- ------- --------- -----
Total [umem_internal]                                       112M   2651033     0
Total [umem_default]                                       1.59G 114603999     0
------------------------- ------ ------- ------- ------- ------- --------- -----

vmem                         memory     memory    memory     alloc alloc
name                         in use      total    import   succeed  fail
------------------------- --------- ---------- --------- --------- -----
sbrk_top                      2.37G      3.12G         0    647807  6463
    sbrk_heap                 2.37G      2.37G     2.37G    647807     0
        vmem_internal          130M       130M      130M     31512     0
            vmem_seg           123M       123M      123M     31482     0
            vmem_hash         6.66M      6.66M     6.66M        27     0
            vmem_vmem         45.1K      54.0K       36K        15     0
        umem_internal          129M       129M      129M     28790     0
            umem_cache         958K      1.13M     1.13M        58     0
            umem_hash         16.1M      16.1M     16.1M        55     0
        umem_log                  0          0         0         0     0
        umem_firewall_va          0          0         0         0     0
            umem_firewall         0          0         0         0     0
        umem_oversize          518M       534M      534M    169418     0
        umem_memalign         96.3K       116K      116K    219360     0
        umem_default          1.59G      1.59G     1.59G    381666     0
------------------------- --------- ---------- --------- --------- -----
```

The vast majority of RAM is in this line:
```
cache                        buf     buf     buf     buf  memory     alloc alloc
name                        size  in use  in ptc   total  in use   succeed  fail
------------------------- ------ ------- ------- ------- ------- --------- -----
umem_alloc_640               640       2       - 2218494   1.41G 110102664     0
```

This makes sense, because we've got a 512-byte block size.  Note that the `buf in use` column is very small; this is entirely the allocator deciding not to give back 640-byte chunks back to the OS.

The second largest RAM hog is
```
vmem                         memory     memory    memory     alloc alloc
name                         in use      total    import   succeed  fail
------------------------- --------- ---------- --------- --------- -----
        umem_oversize          518M       534M      534M    169418     0
```

Each extent contains 2x one-byte-per-block `Vec` of block metadata.  There are 131072 blocks per extent, so it makes sense that it would fall into the `umem_oversize` bucket; it can't fit into `umem_alloc_131072`, since the allocator takes a little extra room for metadata.  Anyways, multiplying 131072 bytes/vec * 2 vec/extent * 2048 extents = 512 MiB, which neatly explains `umem_oversize` at `518M`.

------------------------------------------------------------

Running the same tests on `main` got up to 2.7 GiB of RAM used, then `fio` failed due to hitting the Downstairs IO limit (`IO_OUTSTANDING_MAX`).  We probably need to tune backpressure on unencrypted extents, but that's orthogonal to this PR; I was mostly checking that it doesn't make things _worse_.

faithanalog (approved):
I like the state of this change right now. I've been following the meta-discussion around this; I've also read through the new `extent_inner_raw` implementation and it makes sense to me (love the strategy for defragmentation!). I don't think I can fully comb through every part of this change (particularly version migration) right now, and I don't want to hold up the PR over that- if james and alan like it, that's good by me.

I *have* gotten into a couple Weird states while hammering the storage stack with fio jobs, but most of the time I don't hit those states, which is a good sign.

Because I can't deterministically reproduce them, I currently have no way of knowing if those weird states are a consequence of changes in this PR, or if they're caused by bugs that are already in main. Or both: this PR making bugs in other codepaths easier to hit simply by speeding things up, also an option. But from my understanding of this code, this PR does not really strike me as a likely source of deadlocks.

I'm going to keep chasing down trying to reproduce those ghosts because I do not like them. I don't feel compelled to hold this PR back because of them, unless someone else has a particular theory about them.

PR #1000, opened 2023-10-13
Use regions_dataset path for apply_smf

### LINKED ISSUE #999: Downstairs `config/directory` property has region id twice
```
root@oxz_crucible_59bfadb8-2006-4d3b-8506-c1a7aa9ac647:~# svccfg -s svc:/oxide/crucible/downstairs:downstairs-4152fbaf-234e-4ec3-be96-afe2abdaf214 listprop config/directory
config/directory  astring  /data/regions/4152fbaf-234e-4ec3-be96-afe2abdaf214/4152fbaf-234e-4ec3-be96-afe2abdaf214
```

jmpesp:
`apply_smf`'s `dataset` argument is meant to be for the regions dataset,
not the specific region dataset.
    
Additionally, panic if regions_dataset.path() returns Err: the agent
requires a valid regions dataset and path to work.

Fixes #999

### DIFF
--- a/agent/src/main.rs
+++ b/agent/src/main.rs
@@ -1532,8 +1532,8 @@ fn worker(
         Ok(regions_dataset_path) => regions_dataset_path,
         Err(e) => {
             panic!(
-               "Cannot get regions_dataset_path for {:?}: {}",
-               regions_dataset, e,
+                "Cannot get regions_dataset_path for {:?}: {}",
+                regions_dataset, e,
             );
         }
     };
--- a/agent/src/main.rs
+++ b/agent/src/main.rs
@@ -77,6 +77,7 @@ enum Args {
     },
 }
 
+#[derive(Debug)]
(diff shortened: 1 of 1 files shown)


agent/src/main.rs
```diff
@@ -1527,6 +1528,16 @@ fn worker(
     downstairs_prefix: String,
     snapshot_prefix: String,
 ) {
+    let regions_dataset_path = match regions_dataset.path() {
+        Ok(regions_dataset_path) => regions_dataset_path,
+        Err(e) => {
+            panic!(
```
leftwo, line 1534: So, this will panic right at the start if the specific regions dataset path does not
exist before we even begin then, right?

leftwo (approved):

agent/src/main.rs
jmpesp: Right

PR #1004, opened 2023-10-17
Don't check for a snapshot that's gone

jmpesp:
When Nexus is trying to delete its snapshots, it will send two calls to the agent: one to delete the read-only downstairs booted for a snapshot (called a "running snapshot" in the agent), and one to delete the snapshot itself. Multiple requests from a user(s) to delete a snapshot can queue up multiple identical snapshot-delete sagas that run one after each other. With the current agent this results in all snapshot-delete sagas after the first experiencing 404s during the DELETE call for a running snapshot, because the agent checks to see if the underlying snapshot exists before accepting that delete request.

Given that Nexus is calling

    DELETE /crucible/0/regions/{id}/snapshots/{name}/run

it seemed natural when writing the agent to return a 404 if the snapshot didn't exist: this is a nested path, and that's an invalid request if we're a traditional REST server. The agent isn't a traditional REST server though - because it is an external service that is contacted as part of sagas, there are additional requirements on it.

One of those requirements is idempotency: if a repeated request is submitted, then the agent's response should be identical to the first response. This PR changes that to extend further: the agent keeps enough state (serialized in crucible.json) to return an identical response even if the snapshot is gone, and even if the region is gone. It's able to do this because the agent does not currently remove records for deleted objects.

### DIFF
--- a/agent/src/datafile.rs
+++ b/agent/src/datafile.rs
@@ -339,9 +339,11 @@ impl DataFile {
 
         match running_snapshots.get_mut(&request.name) {
             None => {
+                // A read-only downstairs never existed for this snapshot,
+                // return OK.
                 info!(
                     self.log,
-                    "not running for region {} snapshot {}, returning Ok",
+                    "running snapshot never existed for region {} snapshot {}, returning Ok",
                     request.id.0,
                     request.name
                 );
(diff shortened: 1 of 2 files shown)


agent/src/datafile.rs
```diff
@@ -339,9 +339,11 @@ impl DataFile {
 
         match running_snapshots.get_mut(&request.name) {
             None => {
+                // A read-only downstairs never existed for this snapshot,
```
leftwo, line 342: Is the name "running_snapshots" still correct here?
This could be a current snapshot or one that we already deleted, right?

leftwo (commented):
How long will we keep deleted snapshots around?
Do we want to consider deleting ones older than X number of days?
This change handles situations where multiple deletes are in flight, but after X number
of days, do we still need to remember deleted snapshots?

jmpesp:
> How long will we keep deleted snapshots around? Do we want to consider deleting ones older than X number of days? This change handles situations where multiple deletes are in flight, but after X number of days, do we still need to remember deleted snapshots?

As it stands now the agent remembers every region and running snapshot it's ever seen. I imagine that a future Nexus garbage collection could also invoke an agent garbage collection of sorts.

leftwo (approved):
Let's file an issue for garbage collection, as the number of snapshots is only
going to continue to grow and we have to be sure to clean it out before
things get too big.

PR #1011, opened 2023-10-25
Antagonize the Crucible Agent

jmpesp:
The `agent-antagonist` will stress test the Crucible agent by continuously requesting and deleting regions, randomly taking and deleting snapshots of those regions.

### DIFF
--- a/agent-antagonist/src/main.rs
+++ b/agent-antagonist/src/main.rs
@@ -77,6 +77,290 @@ fn get_client(agent: &SocketAddr) -> CrucibleAgentClient {
     CrucibleAgentClient::new_with_client(&format!("http://{}", agent), client)
 }
 
+async fn main_thread(log: Logger, agent: SocketAddr, dataset: String) {
+    'outer: loop {
+        // Create a 1 GB region
+        let region_id = Uuid::new_v4();
+
+        let region_request = CreateRegion {
+            block_size: 512,
+            extent_count: 16,
+            extent_size: 131072,
+            id: RegionId(region_id.to_string()),
+            encrypted: true,
+            cert_pem: None,
+            key_pem: None,
+            root_pem: None,
+        };
+
+        loop {
+            info!(log, "requesting region {:?}", region_id);
+            let client = get_client(&agent);
+            let r = client.region_create(&region_request).await;
+            drop(client);
+
+            let region = match r {
+                Ok(region) => {
+                    info!(log, "requested region {:?} ok", region_id,);
+                    region
+                }
+
+                Err(e) => {
+                    error!(
+                        log,
+                        "requesting region {:?} failed: {:?}", region_id, e,
+                    );
+                    break 'outer;
+                }
+            };
+
+            match region.state {
+                RegionState::Requested => {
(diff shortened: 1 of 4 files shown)


agent-antagonist/src/main.rs
```diff
@@ -0,0 +1,349 @@
...
+    match args {
+        Args::Run {
+            agent,
+            dataset,
+            tasks,
+        } => {
+            let log = build_logger();
+
+            let agent = if let Some(agent) = agent {
+                info!(log, "using agent at argument {:?}", agent);
```
leftwo, line 92: Nit: Agent at address?

agent-antagonist/src/main.rs
```diff
@@ -0,0 +1,349 @@
...
+                )?;
+                let listen_port =
+                    listen_port.split(' ').last().unwrap().to_string();
+                info!(log, "using listen_port {:?}", listen_port);
+
+                format!("[{}]:{}", listen_addr, listen_port).parse()?
+            };
+
+            let dataset = if let Some(dataset) = dataset {
+                info!(log, "using dataset at argument {:?}", dataset);
```
leftwo, line 119: Also Nit, using dataset {:?} maybe?  Maybe for the other one too?

agent-antagonist/src/main.rs
```diff
@@ -0,0 +1,349 @@
...
+                                }
+
+                                Err(e) => {
+                                    info!(log, "get region {:?} snapshots failed with {:?}", region_id, e);
+                                    break 'outer;
+                                }
+                            }
+
+                            let client = get_client(&agent);
+                            let r = client.region_get_snapshot(&RegionId(region_id.to_string()), &snapshot_id.to_string()).await;
```
leftwo, line 218: Did `cargo fmt` leave this line longer than 80 characters?  Is it longer than 80?

leftwo (approved):
This is cool, and much needed to it sounds like.

I would like to see this running in CI, but to do that we need a way to loop for some time
or iteration count and that could be a follow on PR along with the buildomat stuff to
run it.

agent-antagonist/src/main.rs
jmpesp: It's definitely longer, I wonder why fmt didn't pick it up?

agent-antagonist/src/main.rs
jmpesp: removed, see 1137af4

agent-antagonist/src/main.rs
jmpesp: removed, see 1137af4

leftwo (commented):
Did you move the entire loop to a new function just to
reduce the lines to < 80 characters?

leftwo:
> Did you move the entire loop to a new function just to reduce the lines to < 80 characters?

If it's just a move, then go ahead and merge this with main.
But, if there are other changes, let me know and I'll actually look at it. :)

jmpesp:
> Did you move the entire loop to a new function just to reduce the lines to < 80 characters?

I did! Merging now.

PR #1012, opened 2023-10-27
Remove block size (redundant with `extent_size.shift`)

### LINKED ISSUE #923: Block size is stored in two places in `RegionDefinition`

mkeeter:
Fixes #923


mkeeter:
On second though, this requires a message version bump.

PR #1013, opened 2023-10-27
The Pantry should reject non-block sized writes

jmpesp:
Instead of panicking, the Pantry should reject non-block sized writes with a 400 level response. The panic came from Volume, which has been corrected to return a CrucibleError instead.

Unfortunately everything the Pantry returns is currently a 500: this commit also defines different HttpError conversions for CrucibleError, and the Pantry now returns those.

This commit also removes the `impl Into<std::io::Error> for CrucibleError`: the author of that code didn't realize that `impl From` works as long as one of the types is in the current crate!


pantry/src/pantry.rs
```diff
@@ -171,19 +184,26 @@ impl PantryEntry {
...
             })
-            .await?;
+            .await
+            .map_err(|e| CrucibleError::GenericError(e.to_string()))?;
 
             let content_length = response
                 .headers()
                 .get(reqwest::header::CONTENT_LENGTH)
                 .ok_or("no content length!")
                 .map_err(|e| anyhow!(e))?;
```
leftwo, line 194: Will this ? turn into a `CrucibleError`?  Do we care if it does not?

upstairs/src/volume.rs
```diff
@@ -481,6 +481,12 @@ impl Volume {
             return Ok(());
         }
 
+        let bs = self.get_block_size().await?;
```
leftwo, line 484: Do we want the same check for the read path as well?

leftwo (approved):
Just a few small questions

upstairs/src/volume.rs
jmpesp: Yes! see 156eb29

pantry/src/pantry.rs
jmpesp: It will turn into a `CrucibleError::GenericError`, yeah, which will get returned by the Pantry as a 500.

PR #1019, opened 2023-11-08
Don't hold the Downstairs lock while doing encryption

mkeeter:
# Test setup
128 GiB disk, 64 KiB extents, 4K blocks, encrypted, running `fio` on an Alpine VM with the following script:

```toml
[global]
filename=/dev/nvme0n1
iodepth=25
ioengine=aio
time_based
runtime=60
numjobs=1
direct=1
stonewall=1

[randwrite-4K]
bs=4K
rw=randwrite

[randwrite-1M]
bs=1M
rw=randwrite

[randwrite-4M]
bs=4M
rw=randwrite
```

# Before
(`main` at da534e73380f3cc53ca0de073e1ea862ae32109b)
```
Run status group 0 (all jobs):
  WRITE: bw=16.1MiB/s (16.9MB/s), 16.1MiB/s-16.1MiB/s (16.9MB/s-16.9MB/s), io=967MiB (1014MB), run=60008-60008msec

Run status group 1 (all jobs):
  WRITE: bw=367MiB/s (385MB/s), 367MiB/s-367MiB/s (385MB/s-385MB/s), io=21.6GiB (23.1GB), run=60061-60061msec

Run status group 2 (all jobs):
  WRITE: bw=349MiB/s (365MB/s), 349MiB/s-349MiB/s (365MB/s-365MB/s), io=20.5GiB (22.0GB), run=60270-60270msec
```

# After
```
Run status group 0 (all jobs):
  WRITE: bw=15.1MiB/s (15.8MB/s), 15.1MiB/s-15.1MiB/s (15.8MB/s-15.8MB/s), io=904MiB (948MB), run=60036-60036msec

Run status group 1 (all jobs):
  WRITE: bw=457MiB/s (479MB/s), 457MiB/s-457MiB/s (479MB/s-479MB/s), io=26.8GiB (28.7GB), run=60045-60045msec

Run status group 2 (all jobs):
  WRITE: bw=499MiB/s (523MB/s), 499MiB/s-499MiB/s (523MB/s-523MB/s), io=29.3GiB (31.5GB), run=60190-60190msec
```

# Conclusion
free real estate

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -5954,6 +5954,20 @@ impl Upstairs {
         let ddef = self.ddef.lock().await.get_def().unwrap();
 
         /*
+         * Verify IO is in range for our region.  If not give up now and
+         * report error.
+         */
+        match ddef.validate_io(offset, data.len()) {
+            Ok(()) => {}
+            Err(e) => {
+                if let Some(req) = req {
+                    req.send_err(e);
+                }
+                return Err(());
+            }
+        }
+
+        /*
          * Given the offset and buffer size, figure out what extent and
          * byte offset that translates into. Keep in mind that an offset
          * and length may span two extents.
@@ -6031,20 +6045,6 @@ impl Upstairs {
 
         // While we've got the lock, update our current backpressure
         self.set_backpressure_with_downstairs(&downstairs);
-
-        /*
-         * Verify IO is in range for our region.  If not give up now and
-         * report error.
-         */
-        match ddef.validate_io(offset, data.len()) {
-            Ok(()) => {}
-            Err(e) => {
-                if let Some(req) = req {
-                    req.send_err(e);
-                }
-                return Err(());
-            }
-        }
 
         self.set_flush_need();
 
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -5951,33 +5951,7 @@ impl Upstairs {
             return Err(());
         }
 
-        /*
-         * Get the next ID for the guest work struct we will make at the
-         * end. This ID is also put into the IO struct we create that
-         * handles the operation(s) on the storage side.
-         */
-        let mut gw = self.guest.guest_work.lock().await;
-        let mut downstairs = self.downstairs.lock().await;
         let ddef = self.ddef.lock().await.get_def().unwrap();
-
-        // While we've got the lock, update our current backpressure
-        self.set_backpressure_with_downstairs(&downstairs);
-
-        /*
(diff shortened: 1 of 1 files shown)


upstairs/src/lib.rs
```diff
@@ -5951,34 +5951,8 @@ impl Upstairs {
...
         }
 
-        /*
-         * Get the next ID for the guest work struct we will make at the
-         * end. This ID is also put into the IO struct we create that
-         * handles the operation(s) on the storage side.
-         */
-        let mut gw = self.guest.guest_work.lock().await;
-        let mut downstairs = self.downstairs.lock().await;
         let ddef = self.ddef.lock().await.get_def().unwrap();
```
leftwo, line 5954: The lock ordering here might cause us a problem.
All the other places where we get the three locks (gw, downstairs, ddef), they always go in
that order.  If we get ddef here, but someone else has gw and or ds and then tries to get
ddef, we will deadlock.

Some options,
1) If we need ddef, maybe we move it up to the first lock selected?
2) For this use case, we might be able to get the fields we need out of ddef and
then release the lock?  At the moment the `ddef` fileds we use here should not be
changing once they are initially set.  At some point we may want to grow a region, and
we will have to deal with that, but with the current setup, we set it once and panic if it
ever changes.

mkeeter: The `ddef` lock isn't actually held here: `get_def()` gets a copy of the `RegionDefinition`, and then the lock is dropped at the end of the line (so we're already doing your Option 2).

leftwo: Ah, well then, no deadlock.

We might have to give this area some attention if we change how the `ddef` is updated, but we can
burn that bridge when we get to it.

leftwo (approved):

upstairs/src/lib.rs
```diff
@@ -5951,34 +5951,8 @@ impl Upstairs {
...
 
-        // While we've got the lock, update our current backpressure
-        self.set_backpressure_with_downstairs(&downstairs);
-
-        /*
-         * Verify IO is in range for our region.  If not give up now and
-         * report error.
-         */
-        match ddef.validate_io(offset, data.len()) {
```
jmpesp, line 5970: This needs to be before `extent_from_offset`, otherwise that function could panic.

jmpesp (changes requested):

upstairs/src/lib.rs
mkeeter: Good catch, done in 0bfdceec15768579fc0f29b831c6e2ef945f5cf5

jmpesp (approved):

PR #1021, opened 2023-11-09
Decrypt without holding the downstairs lock

jmpesp:
Same sort of free real estate as b960fe88! Do not decrypt while holding the downstairs lock.

Also, filter out invalid block contexts when validating read responses. This is then later relied upon (in the case when comparing different downstairs' read responses) when comparing `job.read_response_hashes` to what was just returned: `job.read_response_hashes` is `Vec<Option<u64>>`
 not `Vec<Vec<u64>>`, so zero or one valid block context is turned into
an Option. Importantly, the result is not `collect`ed (except when panicking!), saving an allocation.

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -3977,10 +3977,8 @@ impl Downstairs {
         response: &mut ReadResponse,
         log: &Logger,
     ) -> Result<Option<u64>, CrucibleError> {
-        // check integrity hashes - make sure at least one is correct.
-        let mut valid_hash = None;
-
         if !response.block_contexts.is_empty() {
+            // check integrity hashes - make sure at least one is correct.
             let mut successful_hash = false;
             let computed_hash = integrity_hash(&[&response.data[..]]);
 
@@ -3998,7 +3996,8 @@ impl Downstairs {
                 response
                     .block_contexts
                     .retain(|context| context.hash == computed_hash);
-                valid_hash = Some(computed_hash);
+
+                Ok(Some(computed_hash))
             } else {
                 // No integrity hash was correct for this response
                 error!(log, "No match computed hash:0x{:x}", computed_hash,);
@@ -4024,9 +4023,9 @@ impl Downstairs {
             //
             // XXX if it's not a blank block, we may be under attack?
             assert!(response.data[..].iter().all(|&x| x == 0));
-        }
 
-        Ok(valid_hash)
+            Ok(None)
+        }
     }
 
     /// Returns:
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -3982,19 +3982,24 @@ impl Downstairs {
 
         if !response.block_contexts.is_empty() {
             let mut successful_hash = false;
-
             let computed_hash = integrity_hash(&[&response.data[..]]);
 
             // The most recent hash is probably going to be the right one.
             for context in response.block_contexts.iter().rev() {
                 if computed_hash == context.hash {
                     successful_hash = true;
-                    valid_hash = Some(context.hash);
                     break;
                 }
             }
 
-            if !successful_hash {
+            if successful_hash {
+                // Filter out contexts that don't match, and return the
+                // successful hash.
+                response
+                    .block_contexts
+                    .retain(|context| context.hash == computed_hash);
+                valid_hash = Some(computed_hash);
+            } else {
                 // No integrity hash was correct for this response
                 error!(log, "No match computed hash:0x{:x}", computed_hash,);
                 for context in response.block_contexts.iter().rev() {
@@ -4061,7 +4066,6 @@ impl Downstairs {
         }
 
         let mut valid_hash = None;
-
         let mut successful_decryption = false;
         let mut successful_hash = false;
 
@@ -4152,6 +4156,7 @@ impl Downstairs {
                     computed_hash
                 );
             }
+
             // no hash was correct
             Err(CrucibleError::HashMismatch)
         } else if !successful_decryption {
@@ -4159,8 +4164,17 @@ impl Downstairs {
             error!(log, "Decryption failed with correct hash");
             Err(CrucibleError::DecryptionError)
         } else {
-            // Ok!
-            Ok(valid_hash)
+            let Some(valid_hash) = valid_hash else {
+                panic!("valid hash is None!");
+            };
+
+            // Filter out contexts that don't match, and return the successful
+            // hash.
+            response
+                .block_contexts
+                .retain(|context| context.hash == valid_hash);
+
+            Ok(Some(valid_hash))
         }
     }
 
@@ -4177,8 +4191,7 @@ impl Downstairs {
         &mut self,
         ds_id: JobId,
         client_id: ClientId,
-        responses: Result<Vec<ReadResponse>, CrucibleError>,
-        encryption_context: &Option<Arc<EncryptionContext>>,
+        read_data: Result<Vec<ReadResponse>, CrucibleError>,
         up_state: UpState,
         extent_info: Option<ExtentInfo>,
     ) -> Result<bool> {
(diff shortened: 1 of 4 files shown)


upstairs/src/lib.rs
```diff
@@ -4152,15 +4156,25 @@ impl Downstairs {
...
             // no hash was correct
             Err(CrucibleError::HashMismatch)
         } else if !successful_decryption {
             // no hash + encryption context combination decrypted this block
             error!(log, "Decryption failed with correct hash");
             Err(CrucibleError::DecryptionError)
         } else {
-            // Ok!
-            Ok(valid_hash)
+            let Some(valid_hash) = valid_hash else {
```
mkeeter, line 4167: Nit: it's a little weird to have both `valid_hash: Option<u64>` and `successful_hash: bool`, since they reflect the same data (`successful_hash == valid_hash.is_some()`).

I'd suggest removing `successful_hash`, then restructure this conditional to something like

```rust
if let Some(valid_hash) = valid_hash {
    if successful_decryption {
        // .. happy case
    } else {
        // .. decryption failed
    }
} else {
    // ...missing hash case
}
```

upstairs/src/lib.rs
```diff
@@ -3982,19 +3982,24 @@ impl Downstairs {
...
             }
 
-            if !successful_hash {
+            if successful_hash {
+                // Filter out contexts that don't match, and return the
+                // successful hash.
+                response
+                    .block_contexts
+                    .retain(|context| context.hash == computed_hash);
+                valid_hash = Some(computed_hash);
```
mkeeter, line 4001: Nit: is `valid_hash` even needed?  It seems like you could write this conditional as
```suggestion
                // Filter out contexts that don't match, and return the
                // successful hash.
                response
                    .block_contexts
                    .retain(|context| context.hash == computed_hash);
                Some(computed_hash)
```
then remove the `Ok(valid_hash)` at the end of the function (deleting `valid_hash` entirely).

upstairs/src/lib.rs
```diff
@@ -4438,14 +4379,36 @@ impl Downstairs {
...
+                     * has the same list of hashes
                      */
                     let read_data: Vec<ReadResponse> = read_data.unwrap();
                     assert!(!read_data.is_empty());
-                    if job.read_response_hashes != read_response_hashes {
+
+                    let mismatch = std::iter::zip(
+                        job.read_response_hashes.iter(),
+                        read_data.iter().map(|response| {
+                            response.block_contexts.get(0).map(|ctx| ctx.hash)
```
mkeeter, line 4390: This is used enough that it could be a function in `ReadResponse`, e.g. `first_hash(&self) -> Option<u64>`

upstairs/src/lib.rs
```diff
@@ -4505,8 +4468,22 @@ impl Downstairs {
...
+                        job.read_response_hashes = read_data
+                            .iter()
+                            .map(|read_response| {
+                                // Each entry in a validated list of read
+                                // responses will have either an empty list of
+                                // hashes (meaning the data block was all
+                                // zeros), or one valid context (because the
+                                // rest were filtered out). Turn this into an
+                                // Option here.
+                                read_response.hashes().first().copied()
```
mkeeter, line 4481: We could use the hypothetical `first_hash()` function here too!

upstairs/src/lib.rs
```diff
@@ -4517,20 +4494,45 @@ impl Downstairs {
...
                          */
                         debug!(
                             self.log,
                             "[{}] Read already AckReady {}",
                             client_id,
                             job.ds_id
                         );
-                        if job.read_response_hashes != read_response_hashes {
+
+                        let mismatch = std::iter::zip(
```
mkeeter, line 4508: This code duplicates stuff in the `ack_status == AckStatus::Acked` condition above.  Perhaps it should be a function on `job`, e.g. `DownstairsIO::check_read_response_hashes(&self, read_data: &[ReadResponse]))`?

upstairs/src/lib.rs
jmpesp: Agreed, done in c1c388b

upstairs/src/lib.rs
jmpesp: I think so but only when validating encrypted block contexts, where each `computed_hash` will be different. In the unencrypted case `computed_hash` is computed once and checked for each context.

upstairs/src/lib.rs
jmpesp: agreed, done in 6eb6d22

upstairs/src/lib.rs
jmpesp: added in 6eb6d22 :)

upstairs/src/lib.rs
jmpesp: I like it, also added in 6eb6d22

mkeeter (approved):

upstairs/src/test.rs
```diff
@@ -5093,6 +4975,7 @@ pub(crate) mod up_test {
     }
 
     #[tokio::test]
+    #[should_panic]
```
leftwo, line 4978: If we should panic here, then we unwrap on `encrypt_in_placed` won't this give us a false negative?

upstairs/src/test.rs
```diff
@@ -5151,21 +5035,13 @@ pub(crate) mod up_test {
...
-                )
-            }));
-        assert!(result.is_err());
+        let _result = upstairs
+            .process_ds_operation(next_id, ClientId::new(0), response, None)
+            .await;
     }
 
     #[tokio::test]
+    #[should_panic]
```
leftwo, line 5044: Same `should_panic` here,  If there is any `unwrap()` before the expected panic, I think this 
tests will pass for the wrong reasons.

leftwo (commented):
A high level question, if we are doing the decryption outside the lock, do we have
to worry about two reads coming back at close to the same time and then racing for
who is the first?  My question is based on my lazyness on not figuring it out for myself.

upstairs/src/lib.rs
mkeeter: I was thinking about also moving the rest of the stuff (including the logging and panic) into that function, but don't feel strongly about it.

upstairs/src/test.rs
jmpesp: You're right, see 707f4d9 where I rolled this back

upstairs/src/test.rs
jmpesp: Same, see 707f4d9

upstairs/src/lib.rs
mkeeter: This function is called `validate_unencrypted_read_response`, so I don't think the encrypted case matters here 😄

jmpesp:
> A high level question, if we are doing the decryption outside the lock, do we have to worry about two reads coming back at close to the same time and then racing for who is the first?

I don't think so. The decryption context doesn't hold any state that feeds forward from one decryption to the next (unlike something stateful like TLS). The locking of the Downstairs struct eventually does happen which serializes the modifications. If we had a race problem before this commit then that'd be an existing bug that I believe we'd have seen already.

upstairs/src/lib.rs
jmpesp: AH I see what you mean now! e4dd3fb

upstairs/src/lib.rs
```diff
@@ -4128,7 +4129,21 @@ impl Downstairs {
             }
         }
 
-        if !successful_hash {
+        if let Some(valid_hash) = valid_hash {
```
leftwo, line 4132: Since you are here already, could we correct the comment up at line 4111, or at least mention that
it is only for the SQL backend? Thanks!

leftwo (approved):

upstairs/src/lib.rs
jmpesp: Rgr, done in 5d46ff4

Issue #968, opened 2023-09-27
Downstairs disconnect during LiveRepair can result in LiveRepair hanging forever


askfongjojo:
I found a VM I had running for weeks throwing a bunch of EXT4-fs errors on the serial console:
<img width="1328" alt="ext4-fs-errors" src="https://github.com/oxidecomputer/crucible/assets/10874909/171b2f17-72eb-4634-b06b-114979928a2f">
The same problem happened to be reported by a customer today (the symptoms match but the root cause may differ).

The propolis log shows that it's waiting for a repair that is still going on:
```
BRM42220051 # zlogin oxz_propolis-server_38d396d5-eead-446b-a21f-b3c67d3c0479
[Connected to zone 'oxz_propolis-server_38d396d5-eead-446b-a21f-b3c67d3c0479' pts/13]
The illumos Project     helios-2.0.22187        September 2023
root@oxz_propolis-server_38d396d5-eead-446b-a21f-b3c67d3c0479:~# tail /var/svc/log/system-illumos-propolis-server\:default.log | looker
19:34:56.388Z WARN propolis-server (crucible-94a61313-ccf2-469a-9614-81c667dae72f): Upstairs repair task running, trying again later
19:34:56.388Z INFO propolis-server (crucible-94a61313-ccf2-469a-9614-81c667dae72f): Live Repair in progress, try again
19:35:56.392Z INFO propolis-server (crucible-94a61313-ccf2-469a-9614-81c667dae72f): Checking if live repair is needed
19:35:56.392Z WARN propolis-server (crucible-94a61313-ccf2-469a-9614-81c667dae72f): Upstairs repair task running, trying again later
19:35:56.392Z INFO propolis-server (crucible-94a61313-ccf2-469a-9614-81c667dae72f): Live Repair in progress, try again
19:36:56.397Z INFO propolis-server (crucible-94a61313-ccf2-469a-9614-81c667dae72f): Checking if live repair is needed
19:36:56.397Z WARN propolis-server (crucible-94a61313-ccf2-469a-9614-81c667dae72f): Upstairs repair task running, trying again later
19:36:56.397Z INFO propolis-server (crucible-94a61313-ccf2-469a-9614-81c667dae72f): Live Repair in progress, try again
```
But the downstairs serving the volume look healthy when I checked. One of them did report a repair that happened two days ago:
```
[...]

PR #1023, opened 2023-11-15
Abort a live repair

leftwo:
A fix for [issue 968](https://github.com/oxidecomputer/crucible/issues/968)
A live repair can get stuck waiting on a channel that will never respond.

There is a window of time when the repair task has sent a repair job to
all three downstairs and is waiting for a response that all three downstairs
have handled that operation.  If the two good downstairs respond, then the third
downstairs disconnects before answering, the live repair task has no way of knowing
and will wait forever for an answer to a job that will never come.
If the downstairs that disconnects comes back, it will attempt to enter LiveRepair, but
will be blocked as a LiveRepair is in progress.

This PR adds a few things,
A new channel that LiveRepair will listen on that can be used to signal an abort.
A new `AckState`: `Aborted`
A new `LiveRepair` `RepairCheck`: `RepairAborted`

The "test_fail_live_repair.sh" was fixed to test this scenario.

The "test_fail_live_repair.sh" was added to the nightly test list.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -953,12 +953,10 @@ where
      */
     while job_channel_rx.recv().await.is_some() {
         // Add a little time to completion for this operation.
-        /*
         if ads.lock().await.lossy && random() && random() {
             info!(ads.lock().await.log, "[lossy] sleeping 1 second");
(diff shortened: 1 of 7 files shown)


leftwo:
With the one-true-task PR landed, the work here is obsoleted and no longer needed.

Issue #948, opened 2023-09-24
On startup, a non-repair can finish before a downstairs task arrives to repair


leftwo:
It's possible (I believe) that the reconciliation task has no work to do, and 
can get to the end of reconciliation and mark a downstairs as `Active` before
that downstairs task has even arrived at `cmd_loop`.

Since `Active` is not a state `cmd_loop` expects to find a downstairs in, we can trigger
this panic:

```
                panic!(
                    "[{}] {} Downstairs in invalid state: {}",
                    up_coms.client_id, up.uuid, bad_state,
                );
```

mkeeter:
This happens frequently enough in CI to be annoying, e.g.

https://buildomat.eng.oxide.computer/wg/0/artefact/01HB9GNPDNEHC67HQYE3ZKDBR6/jdafiwf3gPbHY5M43SQImUC0yjr7uOXA22twFIiVqM7ZAzUC/01HB9GP96XZSFFFH6WNJESYCSW/01HB9JJE219KABFF7JGCTAJSB3/test_up_out.txt

leftwo:
With the re-write in https://github.com/oxidecomputer/crucible/pull/1058 the conditions no
longer exist for this issue to happen.

PR #1039, opened 2023-11-16
Add a `ProducerKind` field to `oximeter` registration requests.

bnaecker:


bnaecker:
This is the first Crucible in a sequence intended to add a new parameter to a Nexus internal API. The first PR is https://github.com/oxidecomputer/omicron/pull/4497, after which I'll merge this one and a corresponding update in Propolis. That PR adds an optional parameter indicating the _kind_ of metric producer one registers as. A follow-up commit in Omicron will make that parameter optional, once the out-of-tree consumers all supply it.

upstairs/src/active_jobs.rs
```diff
@@ -354,7 +354,9 @@ impl BlockMap {
         // We record every job, even if it's empty
         self.job_to_range.insert(job, r);
 
-        let Some(r) = Self::blocks_to_range(r) else { return; };
+        let Some(r) = Self::blocks_to_range(r) else {
```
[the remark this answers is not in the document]
bnaecker: It is. I needed to bump the toolchain to deal with what appears to be a bug in 1.70. If you run `cargo update -p oximeter` on `main` using the existing pinned toolchain, you'll see:

```console
bnaecker@shale : ~/crucible $ cargo update -p oximeter
    Updating git repository `https://github.com/oxidecomputer/omicron`
    Updating crates.io index
    Updating git repository `https://github.com/oxidecomputer/dropshot`
    Updating git repository `https://github.com/oxidecomputer/progenitor`
error: no matching package named `omicron-common` found
location searched: https://github.com/oxidecomputer/omicron?branch=main
required by package `crucible-agent v0.0.1 (/home/bnaecker/crucible/agent)`
```

I am not sure what exactly is going on here, and was going to raise bumping the toolchain as a "rider" on this PR in case there are strong opinions against doing that.

bnaecker:
[This failure](https://github.com/oxidecomputer/crucible/actions/runs/6895522085/job/18759647396?pr=1039) is due to a bump in the `openapiv3` dependency. I'm not sure that's required, but I think the update of `openapi-lint` includes a bump to that newer version, and thus introduces a "are multiple versions of this crate used?" error unless the direct dependency is also updated.

I don't imagine we care so much about the dependency, but I _do_ imagine we don't want to change the naming of these enum variants in the OpenAPI document. I'm not sure who consumes that document though.

Cargo.toml
```diff
@@ -105,10 +105,10 @@ uuid = { version = "1", features = [ "serde", "v4" ] }
 
 # git
 dropshot = { git = "https://github.com/oxidecomputer/dropshot", branch = "main", features = [ "usdt-probes" ] }
-omicron-common = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
+omicron-common = { git = "https://github.com/oxidecomputer/omicron", branch = "add-oximeter-producer-kind" }
```
bnaecker, line 108: Note to self -- point to `main` once the corresponding Omicron PR merges.

leftwo: Ah, this pr: https://github.com/oxidecomputer/omicron/pull/4497

bnaecker: Right. That PR introduces a new optional field, to avoid breaking builds right away. I'm adding the optional field to the out-of-tree repos now. Then I'll make that field required in Omicron, and then change the downstream consumers to follow.

dsc/src/main.rs
```diff
@@ -930,6 +930,7 @@ struct MonitorInfo {
 
 /// State of a downstairs.
 #[derive(Debug, Copy, Clone, Deserialize, Serialize, JsonSchema)]
+#[serde(rename_all = "snake_case")]
```
bnaecker, line 933: Flagging this for attention, any older client will have its requests rejected after this change.

openapi/crucible-pantry.json
```diff
@@ -5,7 +5,7 @@
     "version": "0.0.0"
   },
   "paths": {
-    "/crucible/pantry/0/job/{id}/is_finished": {
+    "/crucible/pantry/0/job/{id}/is-finished": {
```
bnaecker, line 8: Same note here, this requires a new generated client. I think that is ok for Omicron, since it will pull in the new Crucible dep (via Propolis) and use the new generated client all at once. But I would like some attention on it.

pantry/src/server.rs
```diff
@@ -72,7 +72,7 @@ struct JobPollResponse {
 /// Poll to see if a Pantry background job is done
 #[endpoint {
     method = GET,
-    path = "/crucible/pantry/0/job/{id}/is_finished",
+    path = "/crucible/pantry/0/job/{id}/is-finished",
```
bnaecker, line 75: Same note here, I think Omicron uses the generated client and will be updated wholesale at the same time the pantry server is.

openapi/crucible-pantry.json
leftwo: Omicron will need this line changed, I think:
https://github.com/oxidecomputer/omicron/blob/4951c339a8987c757596207c81afdfada8bbf7c7/sled-agent/src/sim/http_entrypoints_pantry.rs#L99

leftwo (approved):
Yeah, let's get this party started.

bnaecker:
Thanks @leftwo, just waiting on approval of the [first PR in Omicron](https://github.com/oxidecomputer/omicron/pull/4497)

bnaecker:
Ok, Omicron side is set to automerge. When that completes, I'll update this to point back at `main` instead of the feature branch.

leftwo (approved):
Yeah.
You might want to fix that `cargo fmt` change though.

bnaecker:
Thanks Alan. I took a quick glance and thought it might be GH running actions against the hypothetical merge commit, but I’m probably wrong. I’ll look in detail tomorrow and resolve it before merging.

leftwo:
> Thanks Alan. I took a quick glance and thought it might be GH running actions against the hypothetical merge commit, but I’m probably wrong. I’ll look in detail tomorrow and resolve it before merging.

Yeah, I tried `cargo fmt` on a local copy of the branch and it found nothing to complain about, so I think you
are right.

bnaecker:
@leftwo Same here, `cargo fmt` on the PR branch itself is fine. I made a test branch from the PR, and merged in `main`, and _that_ fails with the format error. So this is a "real" failure in that merging this to `main` will result in a format error. I'll merge `main` into the PR now, which should resolve that. Assuming it does, I'll merge. Thanks!

bnaecker:
Hmm, this last test seems to have hit a panic [here](https://buildomat.eng.oxide.computer/wg/0/details/01HFSGH7WXKYCCZ7288V4ECRMC/qns3xwnb5DOzq3olWpXFJyCz5whvDNd5jQDE7zeincv2gUvK/01HFSGHRSBG84NYEKVQJM1EYGN#S875). I don't _think_ this is related, but I'll look more now.

leftwo:
> Hmm, this last test seems to have hit a panic [here](https://buildomat.eng.oxide.computer/wg/0/details/01HFSGH7WXKYCCZ7288V4ECRMC/qns3xwnb5DOzq3olWpXFJyCz5whvDNd5jQDE7zeincv2gUvK/01HFSGHRSBG84NYEKVQJM1EYGN#S875). I don't _think_ this is related, but I'll look more now.

I believe it's unrelated, and I think it's this issue: https://github.com/oxidecomputer/crucible/issues/948

bnaecker:
@leftwo has suggested this could be https://github.com/oxidecomputer/crucible/issues/948. I'm going to rerun that failed check, and go from there.

bnaecker:
Ok, the rerun was successful, so this is going in! Thanks @leftwo.

PR #1045, opened 2023-11-29
Oximeter producer-kind parameter is now required

bnaecker:


bnaecker:
This is the second step in Crucible for handling https://github.com/oxidecomputer/omicron/pull/4497. It updates the OpenAPI document for Nexus's internal API, and adapts to the fact that the `ProducerKind` parameter is now required rather than optional.

leftwo (approved):

PR #1048, opened 2023-12-08
Buffer should destructure to Vec when single-referenced

pfmooney:
Until now, Buffer offered no means of extracting the internal `Vec<u8>` when, say, a Volume::read() operation had completed.  This made it impossible for Crucible consumers to reuse the `Vec` buffer, forcing an otherwise unnecessary allocation.

### DIFF
--- a/crudd/src/main.rs
+++ b/crudd/src/main.rs
@@ -180,7 +180,7 @@ async fn cmd_read<T: BlockIO>(
         // So say we have an offset of 5. we're misaligned by 5 bytes, so we
         // read 5 bytes we don't need. we skip those 5 bytes then write
         // the rest to the output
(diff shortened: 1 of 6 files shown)


pfmooney:
If folks think there's no reason to ever handle the `into_vec()` failure case, I'll make it `panic()` with a more explicit error if there are extra references remaining, rather than `unwrap()`/`expect()`s all over the place.

pfmooney:
The `BlockReq` struct makes clones of `Buffer` through its use, making it somewhat challenging to ensure that all potential holders of the buffer are cleaned up at request completion, at least presently.

Since this isn't a pressing concern, I'm going to back-burner it until some of the other large refactoring work is done before diving back in.

PR #1064, opened 2023-12-18
Buffer should destructure to Vec when single-referenced

pfmooney:
Until now, Buffer offered no means of extracting the internal `Vec<u8>` when, say, a Volume::read() operation had completed.  This made it impossible for Crucible consumers to reuse the `Vec` buffer, forcing an otherwise unnecessary allocation.

There are a few other changes I believe need to be made before consumers like Propolis can start reusing their buffer `Vec`, rather than reallocating after every request, but I believe this is a step in the right direction for that.

### DIFF
--- a/upstairs/src/block_req.rs
+++ b/upstairs/src/block_req.rs
@@ -11,41 +11,34 @@ use tokio::sync::oneshot;
 #[derive(Debug)]
 pub(crate) struct BlockReq {
     pub op: BlockOp,
-    sender: oneshot::Sender<Result<(), CrucibleError>>,
+    pub res: BlockRes,
 }
 
-impl BlockReq {
-    pub fn new(
-        op: BlockOp,
-        sender: oneshot::Sender<Result<(), CrucibleError>>,
-    ) -> BlockReq {
-        Self { op, sender }
-    }
-
-    /// Return a copy of the block op
-    pub fn op(&self) -> BlockOp {
-        self.op.clone()
-    }
-
-    /// Consume this BlockReq and send Ok to the receiver
+#[must_use]
+#[derive(Debug)]
+pub(crate) struct BlockRes(Option<oneshot::Sender<Result<(), CrucibleError>>>);
+impl BlockRes {
+    /// Consume this BlockRes and send Ok to the receiver
     pub fn send_ok(self) {
         self.send_result(Ok(()))
     }
 
-    /// Consume this BlockReq and send an Err to the receiver
+    /// Consume this BlockRes and send an Err to the receiver
     pub fn send_err(self, e: CrucibleError) {
         self.send_result(Err(e))
     }
 
-    /// Consume this BlockReq and send a Result to the receiver
-    pub fn send_result(self, r: Result<(), CrucibleError>) {
+    /// Consume this BlockRes and send a Result to the receiver
+    fn send_result(mut self, r: Result<(), CrucibleError>) {
         // XXX this eats the result!
-        let _ = self.sender.send(r);
+        let _ = self.0.take().expect("sender was populated").send(r);
     }
-
-    /// Consume this BlockReq and return the inner oneshot sender
-    pub fn take_sender(self) -> oneshot::Sender<Result<(), CrucibleError>> {
-        self.sender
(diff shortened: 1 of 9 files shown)


upstairs/src/block_req.rs
```diff
@@ -11,41 +11,34 @@ use tokio::sync::oneshot;
...
     }
-
-    /// Consume this BlockReq and return the inner oneshot sender
-    pub fn take_sender(self) -> oneshot::Sender<Result<(), CrucibleError>> {
-        self.sender
+}
+impl Drop for BlockRes {
+    fn drop(&mut self) {
+        // Dropping a BlockRes without issuing a completion would mean the
+        // associated waiter would be stuck waiting forever for a result.
```
jmpesp, line 40: I like this check! But: instead of asserting, would a `send_err` be appropriate?

upstairs/src/upstairs.rs
```diff
@@ -945,37 +933,34 @@ impl Upstairs {
...
                 info!(self.log, "{} active request set", self.cfg.upstairs_id);
             }
-            UpstairsState::GoActive { .. } => {
+            UpstairsState::GoActive(..) => {
                 info!(
                     self.log,
                     "{} request to activate upstairs already going active",
                     self.cfg.upstairs_id
                 );
-                let _ = reply.send(Err(CrucibleError::UpstairsAlreadyActive));
```
jmpesp, line 963: The error changed here

jmpesp (changes requested):

upstairs/src/lib.rs
```diff
@@ -1681,7 +1693,6 @@ pub enum BlockOp {
     },
     Deactivate,
     // Management commands
-    RepairOp,
```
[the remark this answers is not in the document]
leftwo: I'm sure I put that there, but I don't remember what I was planning for when I did it.
I also don't see it's used anywhere, so I think it's okay to remove.

upstairs/src/block_req.rs
pfmooney: It could be.  I'll leave that decision up to y'all.  Not dealing with a `BlockRes` before it's dropped is definitely a programmer error (vs operational error), but since it's not being enforced by the compiler, I can appreciate not wanting an assertion.  That said, having an assert helped me track down where such drops were occurring.

upstairs/src/lib.rs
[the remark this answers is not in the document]
pfmooney: I assumed removal was safe since, as @leftwo pointed out, it wasn't used anywhere.

upstairs/src/upstairs.rs
pfmooney: Nice catch.  I must have missed that during the rebase.

upstairs/src/block_req.rs
pfmooney: Some additional context: propolis has been using a similar pattern for its [block Request](https://github.com/oxidecomputer/propolis/blob/85dc6db36709981c1d35b0651a18e20caee0b73f/lib/propolis/src/block/mod.rs#L176-L182) abstraction.

upstairs/src/lib.rs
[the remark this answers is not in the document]
jmpesp: :+1:

upstairs/src/block_req.rs
leftwo: Sorry for my ignorance here,
Who is dropping the `BlockRes` in this situation?  Is it the crucible upstairs side that would
have done the drop?

pfmooney: Yeah, it would be upstairs which could drop a `BlockRes` rather than properly "disposing" of it by calling `send_ok` or `send_err`.

jmpesp (approved):
I deployed and tested this, it looks good! Thanks @pfmooney

upstairs/src/block_req.rs
leftwo: I like panic for programmer errors (and, yeah, I benefit so often by catching myself).

PR #1065, opened 2023-12-19
Drive hash / decrypt tests from Upstairs::apply

mkeeter:
This PR converts hash / decryption read tests to run through `Upstairs::apply`, instead of using internal `Downstairs` APIs.

It is in preparation for moving decryption off-thread (and I've got a branch which does that in the wings), but this is a standalone building block that shouldn't be too controversial.

I also made some small changes to tests where relevant:
- `work_read_hash_mismatch_ack` and `work_read_hash_mismatch_third_ack` are removed, because jobs are now acked immediately (so the in-between point that those tests are checking doesn't exist any more)
- `work_read_hash_mismatch_inside` wasn't actually testing the case of hashes vecs with different lengths (despite the comment); it was changed to actually check that by returning reads including 1 vs 2 hashes.
- `bad_decryption_means_panic` wasn't actually checking decryption: the `Upstairs` was constructed without an encryption context, so it was failing at the hash match stage.

In all of the tests that are expected to panic, we now catch the panic string and do a quick substring check.  (This is how I found the issue with `bad_decryption_means_panic`)


upstairs/src/upstairs.rs
```diff
@@ -3504,4 +3508,641 @@ pub(crate) mod test {
             assert_eq!(c.state(), DsState::New);
         }
     }
+
+    #[tokio::test]
+    async fn good_decryption() {
```
leftwo, line 3513: This test is called "good_decryption" but the comment on line 3527 says we
are going to fail decryption?  which is it :)

upstairs/src/upstairs.rs
```diff
@@ -3504,4 +3508,641 @@ pub(crate) mod test {
...
+            .unwrap();
+        assert!(
+            r.contains("DecryptionError"),
+            "panic for the wrong reason: {r}"
+        );
+    }
+
+    #[tokio::test]
+    async fn bad_hash_on_encrypted_read_panic() {
+        let mut up = make_encrypted_upstairs();
```
leftwo, line 3647: Could we add a comment here on what makes this test different than the
`bad_read_hash_makes_panic` test that follows?  I think this test is altering
the tag?

leftwo (approved):
Just a few comment nits, but this is good.

upstairs/src/upstairs.rs
mkeeter: Good catch, this test was misleading (it altered the tag, but _also_ had a bad hash).  I removed the tag changes, since those are tested elsewhere, and added a comment.

upstairs/src/upstairs.rs
mkeeter: Oops, fixed

mkeeter:
This failed `test-up-encrypted` in CI [here](https://github.com/oxidecomputer/crucible/pull/1065/checks?check_run_id=20095256238).  I'm suspicious that this branch may not be at fault, because I saw a similar failure in [`offload-write-encryption` here](https://github.com/oxidecomputer/crucible/pull/1066#issuecomment-1874179836).

PR #1066, opened 2023-12-19
Offload write encryption

mkeeter:
This PR moves write encryption out of the main task.

Doing so is harder than it sounds, because we want to **preserve ordering** of incoming `BlockReq`, but only `BlockReq::Write` and `BlockReq::WriteUnwritten` need this encryption offload.

I turned to `FuturesOrdered`, which allows you to store a set of futures which resolve in FIFO order.  For most `BlockReq`, we await a `futures::future::ready(..)`, which simply provides the same `BlockReq` back to us.  For writes, we offload the work using `rayon::spawn`, then return the result through a `tokio::sync::oneshot` channel.  This strategy is recommended by [this blog post by a Tokio maintainer](https://ryhl.io/blog/async-what-is-blocking/).

We can then store `Either<Ready, oneshot::Receiver>` in a `FuturesOrdered`, which works out nicely.

The PR introduces a new generic `DeferredQueue` to organize these futures.  (The `DeferredQueue` will also be used to defer read decryption in a subsequent PR).  When a `BlockReq` comes in, we put it into this queue*; futures being available on the queue are another event source for the `select / apply` loop.

This is a significant speedup for large random writes:

| IO | `main` | `offload-write-encryption` |
|-|-|-|
| 4K random writes | 27.2 MiB/s | 26.7 MiB/s |
| 1M random writes | 609 MiB/s | 710 MiB/s |
| 4M random writes | 629 MiB/s | 890 MiB/s |

I'm not sure if the performance delta for 4K writes is significant; if folks are worried, we could special-case sufficiently small writes to be executed in the main task.

*as an optimization, we execute non-write `BlockOp`s immediately if there is nothing in the queue, to avoid an extra event

### DIFF
--- a/upstairs/src/deferred.rs
+++ b/upstairs/src/deferred.rs
@@ -1,4 +1,4 @@
-// Copyright 2023 Oxide Computer Company
+// Copyright 2024 Oxide Computer Company
 //! Tools to defer encryption work to a separate thread pool
 use std::sync::Arc;
 
--- a/upstairs/src/deferred.rs
+++ b/upstairs/src/deferred.rs
@@ -1,0 +1,186 @@
+// Copyright 2023 Oxide Computer Company
+//! Tools to defer encryption work to a separate thread pool
+use std::sync::Arc;
+
+use crate::{
+    upstairs::UpstairsConfig, BlockContext, BlockReq, BlockRes, ImpactedBlocks,
+};
+use bytes::Bytes;
+use crucible_common::{integrity_hash, CrucibleError, RegionDefinition};
+use futures::{
+    future::{Either, Ready},
+    stream::FuturesOrdered,
+    StreamExt,
+};
+use tokio::sync::oneshot;
+
+/// Future stored in a [`DeferredQueue`]
+///
+/// This is either an immediately-ready `T` or a oneshot channel which returns a
+/// `T` when an off-task job finishes.
+type DeferredQueueFuture<T> =
+    Either<Ready<Result<T, oneshot::error::RecvError>>, oneshot::Receiver<T>>;
+
+/// A `DeferredQueue` stores pending work (optionally executed off-task)
+pub(crate) struct DeferredQueue<T> {
+    /// Ordered stream of deferred futures
+    stream: FuturesOrdered<DeferredQueueFuture<T>>,
+
+    /// Stores whether it is known that there are no futures in `self.stream`
+    ///
+    /// This is tracked separately because `FuturesOrdered::next` will
+    /// immediately return `None` if the queue is empty; we don't want that when
+    /// it's one of many options in a `tokio::select!`.
+    empty: bool,
+}
+
+impl<T> DeferredQueue<T> {
+    /// Build a new empty `FuturesOrdered`
+    pub fn new() -> Self {
+        Self {
+            stream: FuturesOrdered::new(),
+            empty: true,
+        }
+    }
+
+    /// Stores a new future in the queue, marking it as non-empty
+    pub fn push_back(&mut self, f: DeferredQueueFuture<T>) {
+        self.stream.push_back(f);
+        self.empty = false;
+    }
+
+    /// Returns the next future from the queue
+    ///
+    /// If the future is `None`, then the queue is marked as empty
+    ///
+    /// This function is cancel safe: if a result is taken from the internal
+    /// `FuturesOrdered`, then it guaranteed to be returned.
+    pub async fn next(&mut self) -> Option<T> {
+        // Early exit if we know the stream is empty
+        if self.empty {
+            return None;
+        }
+
+        // Cancel-safety: there can't be any yield points after this!
+        let t = self.stream.next().await;
+        self.empty |= t.is_none();
+
+        // The oneshot is managed by a worker thread, which should never be
+        // dropped, so we don't expect the oneshot
+        t.map(|t| t.expect("oneshot failed"))
+    }
+
+    /// Check whether the queue is known to be empty
+    ///
+    /// It is possible for this to return `false` if the queue is actually
+    /// empty; in that case, a subsequent call to `next()` will return `None`
+    /// and *later* calls to `is_empty()` will return `true`.
+    pub fn is_empty(&self) -> bool {
+        self.empty
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+/// All of the information needed to encrypt a write operation
+///
+/// The `DeferredWrite` is standalone so that it can either be executed locally
+/// or in a separate worker thread.
+pub(crate) struct DeferredWrite {
+    pub ddef: RegionDefinition,
+    pub impacted_blocks: ImpactedBlocks,
+    pub data: Bytes,
+    pub res: Option<BlockRes>,
+    pub is_write_unwritten: bool,
+    pub cfg: Arc<UpstairsConfig>,
+}
+
+/// Result of a deferred `BlockReq`
+///
+/// In most cases, this is simply the original `BlockReq` (stored in
+/// `DeferredBlockReq::Other`).  The exception is `BlockReq::Write` and
+/// `BlockReq::WriteUnwritten`, which require encryption; in these cases,
+/// encryption is done off-thread and the result is a `DeferredBlockReq::Write`.
+#[derive(Debug)]
+pub(crate) enum DeferredBlockReq {
+    Write(EncryptedWrite),
+    Other(BlockReq),
+}
+
+#[derive(Debug)]
+pub(crate) struct EncryptedWrite {
+    pub writes: Vec<crucible_protocol::Write>,
+    pub impacted_blocks: ImpactedBlocks,
+    pub res: Option<BlockRes>,
+    pub is_write_unwritten: bool,
+}
+
+impl DeferredWrite {
+    pub fn run(self) -> Option<EncryptedWrite> {
+        // Build up all of the Write operations, encrypting data here
+        let mut writes: Vec<crucible_protocol::Write> =
+            Vec::with_capacity(self.impacted_blocks.len(&self.ddef));
+
+        let mut cur_offset: usize = 0;
+        let byte_len: usize = self.ddef.block_size() as usize;
+        for (eid, offset) in self.impacted_blocks.blocks(&self.ddef) {
+            let (sub_data, encryption_context, hash) = if let Some(context) =
+                &self.cfg.encryption_context
+            {
+                // Encrypt here
+                let mut mut_data = self
+                    .data
+                    .slice(cur_offset..(cur_offset + byte_len))
+                    .to_vec();
+
+                let (nonce, tag, hash) =
+                    match context.encrypt_in_place(&mut mut_data[..]) {
+                        Err(e) => {
+                            if let Some(res) = self.res {
+                                res.send_err(CrucibleError::EncryptionError(
+                                    e.to_string(),
+                                ));
+                            }
+                            return None;
+                        }
+
+                        Ok(v) => v,
+                    };
+
+                (
+                    Bytes::copy_from_slice(&mut_data),
+                    Some(crucible_protocol::EncryptionContext {
+                        nonce: nonce.into(),
+                        tag: tag.into(),
+                    }),
+                    hash,
+                )
+            } else {
+                // Unencrypted
+                let sub_data =
+                    self.data.slice(cur_offset..(cur_offset + byte_len));
+                let hash = integrity_hash(&[&sub_data[..]]);
+
+                (sub_data, None, hash)
+            };
+
+            writes.push(crucible_protocol::Write {
+                eid,
+                offset,
+                data: sub_data,
+                block_context: BlockContext {
+                    hash,
+                    encryption_context,
+                },
+            });
+
+            cur_offset += byte_len;
+        }
+        Some(EncryptedWrite {
+            writes,
+            impacted_blocks: self.impacted_blocks,
+            res: self.res,
+            is_write_unwritten: self.is_write_unwritten,
+        })
+    }
+}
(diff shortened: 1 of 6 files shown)


upstairs/src/upstairs.rs
```diff
@@ -1143,54 +1239,91 @@ impl Upstairs {
...
+            .and_then(DeferredWrite::run)
+        {
+            self.apply_guest_request(DeferredBlockReq::Write(w)).await
+        }
     }
 
     /// Submits a new write job to the upstairs, optionally without a `BlockReq`
     ///
     /// # Panics
     /// If `req` is `None` and this isn't running in the test suite
```
leftwo, line 1282: The panic check is gone here, is that on purpose?

Ah, it happens still, just in the `compute_deferred_write()` call

upstairs/src/upstairs.rs
```diff
@@ -1143,54 +1239,91 @@ impl Upstairs {
...
-    async fn submit_write_inner(
+    fn submit_deferred_write_inner(
         &mut self,
         offset: Block,
         data: Bytes,
         req: Option<BlockReq>,
         is_write_unwritten: bool,
     ) {
+        // It's possible for the write to be invalid out of the date, in which
```
leftwo, line 1290: Is this a typo here?  Maybe `invalid out of the gate`?
Or, something else?

leftwo (commented):
Why do we need to preserve the ordering of incoming `BlockReq` ?

mkeeter:
This failed a CI check.  Since the Github UI makes it easy to lose this things, I'm linking it here for reference:
https://github.com/oxidecomputer/crucible/pull/1065/checks?check_run_id=20095256238

upstairs/src/deferred.rs
```diff
@@ -0,0 +1,184 @@
...
+/// `T` when an off-task job finishes.
+type DeferredQueueFuture<T> =
+    Either<Ready<Result<T, oneshot::error::RecvError>>, oneshot::Receiver<T>>;
+
+/// A `DeferredQueue` stores pending work (optionally executed off-task)
+pub(crate) struct DeferredQueue<T> {
+    /// Ordered stream of deferred futures
+    stream: FuturesOrdered<DeferredQueueFuture<T>>,
+
+    /// Stores whether it is know that there are no futures in `self.stream`
```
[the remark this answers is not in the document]
mkeeter: Fixed in 9c836a3d1dc7d056b619b61210e99866706c3867

upstairs/src/upstairs.rs
mkeeter: Yup, this was meant to be "out of the gate".  Fixed in 9c836a3d1dc7d056b619b61210e99866706c3867

upstairs/src/upstairs.rs
```diff
@@ -377,6 +392,26 @@ impl Upstairs {
...
+
+                    // The innermost Option is None if the deferred task handled
+                    // the request on its own (and replied to the `BlockReq`
+                    // already). This happens if encryption fails, which would
+                    // be odd, but possible?
+                    Some(None) => UpstairsAction::NoOp,
+
+                    // The outer Option is None if the FuturesOrdered is empty
+                    None => {
+                        assert!(self.deferred_reqs.is_empty());
```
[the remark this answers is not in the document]
mkeeter: Yes, setting this flag is a side-effect of calling `deferred_reqs.next()` when the list is empty (documented [here](https://github.com/mkeeter/crucible/blob/9c836a3d1dc7d056b619b61210e99866706c3867/upstairs/src/deferred.rs#L52)), so this is a cheap sanity check.  I added a comment to that effect in ee6e51a91eef5f54ce0f2ac020813f3aa0edba70

mkeeter:
@leftwo 

> Why do we need to preserve the ordering of incoming `BlockReq`?

This is preserving the order of `BlockReq` in the period between (1) being sent by the guest and (2) being added to the active jobs list (which establishes inter-job dependencies).

We need to preserve order for correctness.  It would be wrong if the guest sent
```
1. Do a big, encrypted write
2. Read from that same block
```
and we reordered that to
```
2. Read from that same block
1. Do the big encrypted write (delayed because encryption is slow)
```

upstairs/src/upstairs.rs
```diff
@@ -437,8 +474,13 @@ impl Upstairs {
             UpstairsAction::Control(c) => {
                 self.on_control_req(c).await;
             }
+            UpstairsAction::NoOp => (),
         }
 
+        // Check whether we need to mark a Downstairs as faulted because too
+        // many jobs have piled up.
+        self.gone_too_long();
```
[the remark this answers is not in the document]
mkeeter: Yeah, the `Guest` action will now either
- Immediately handle the `BlockReq` (if it's not a write _and_ there are no deferred actions in the queue), **or**
- Defer that `BlockReq` into the queue (if it's a write, _or_ other `BlockReq`'s are already deferred)

We'd want to check `gone_too_long` in the first case but not the second; in the second case, we'd want to check `gone_too_long` when the deferred `BlockReq` is resolved (from `UpstairsAction::DeferredBlockReq`).

Instead of trying to keep track of this, I opted to move `gone_too_long` into the "invariant checking" section of `Upstairs::apply`.  It should be a cheap O(1) check, so this seemed like the simplest option.

upstairs/src/upstairs.rs
mkeeter: That sounds right to me.  At some point I'd like to remove the option of sending jobs without a corresponding `BlockReq`, to avoid all of this Bonus Code (which is only used in the test suite).

mkeeter:
While rebasing this branch onto `main`, I realized that the changes from https://github.com/oxidecomputer/crucible/pull/1064 make this a little trickier.  The spawned encryption threads themselves own the `BlockReq` / `BlockRes`, but there's no obvious way to ensure that they're cleaned up, so we could potentially panic during shutdown.

I opted to implement `Drop` for `BlockRes` itself, ensuring that the error is always sent.  This was discussed in the [original PR](https://github.com/oxidecomputer/crucible/pull/1064#discussion_r1434206094), and no one was _strongly_ opposed, but @leftwo expressed support for the previous (panicking) implementation.  @jmpesp / @pfmooney I'd be curious to hear your thoughts on this.

pfmooney:
I'm not fundamentally opposed to changing the Drop behavior, and thus the ergonomics, of `BlockRes` if the current behavior is making it inconvenient to implement your changes.  That being said, I think there are a few other considerations which could be made:

1. If there are situations where it's awkward to chase down all the potential drop paths (like when the `BlockRes` is stored in a status enum like you mentioned), we could add a method to it to set a "here's the error to emit if I'm dropped w/o sending a status" field.  Armed with that, logic where minding the `BlockRes` drop is a challenge could use such an out, where "normal" IO logic would still be beholden to the "don't drop this without sending a status" requirements.

2. If you think the existing drop behavior really is just a hassle and a footgun, and switch to automatically sending an error for dropped-prior-to-response `BlockRes` instances, I would at least add a USDT probe there to fire when such events occur.

leftwo:
> Philosophically, if our goal is to enforce "one must always reply to a BlockReq", we can either
>
> Panic if a BlockReq is not replied to when it is dropped (the current implementation), then add code to reply everywhere 
> (and add new fixes if we see a panic during runtime). This forces us to think about it, and is error-prone; for example, we 
> had forgotten to reply to the BlockReq held in the UpstairsState until
> [this change](https://github.com/oxidecomputer/crucible/commit/0e31f5e4b2978a11998ab752c91d296433aaae97#diff-f56c2bf9a45d9c7ff7ead33ca9af7edf890c04f1c7af2cca76543c8407418fa6R194)
>
> Send an error reply automatically on drop (the proposed implementation). This is impossible to screw up. Are there 
> cases where this causes incorrect behavior or makes the programmer's life harder?

To take a step back here, are there valid situations where we would not reply to a `BlockReq`?  That someone has sent a 
guest IO and not bothered to wait for or check the result?

In general I'm in favor of the panic, but I want to be clear on what the panic means.
I'm also concerned that we could be dropping a `BlockReq` and not knowing it.  I do understand that keeping track
of things comes with added work for the programmer, but if that added work prevents future us from doing something
that leaves unresponded IOs out there, it might be hard to find.  Maybe I'm just not understanding the situation we
are in for this panic.

mkeeter:
@leftwo 

> To take a step back here, are there valid situations where we would not reply to a BlockReq? That someone has sent a
guest IO and not bothered to wait for or check the result?

I don't know of any!  It's harmless to send a message to a oneshot whose `Receiver` has been dropped (`send` returns an error, but we ignore it, which I think is the correct behavior).  If we were _deliberately_ not replying to a `BlockReq` anywhere in our existing code, it would panic, so I hope we're not doing that.

> I'm also concerned that we could be dropping a BlockReq and not knowing it. I do understand that keeping track
of things comes with added work for the programmer, but if that added work prevents future us from doing something
that leaves unresponded IOs out there, it might be hard to find. 

Automatically replying to the `BlockReq` on drop (i.e. the proposed new behavior) **universally protects us** from leaving unresponded IOs out there.  The current behavior guarantees that we will panic if our code happens to leave an unresponded IO out there, but that's still a potential availability issue that we'd have to catch and fix.

Basically, I don't see the benefit of manually searching through every place where an unanswered  `BlockReq` is dropped and adding code that sends an error; if our goal is to always send an error when an unanswered `BlockReq` is dropped, then we should just... do that.

If that's not our goal, then I'm pretty sure the existing panic isn't what we want either!

@pfmooney 

> If you think the existing drop behavior really is just a hassle and a footgun, and switch to automatically sending an error for dropped-prior-to-response BlockRes instances, I would at least add a USDT probe there to fire when such events occur.

I like this idea and will add it!  As a aside, callers will see the error (the new `CrucibleError::BlockReqDropped`), so it's not like the information is lost.

leftwo:
Thanks for your patients here with me @mkeeter 

>> To take a step back here, are there valid situations where we would not reply to a BlockReq? That someone has sent a
guest IO and not bothered to wait for or check the result?
> I don't know of any! It's harmless to send a message to a oneshot whose Receiver has been dropped (send returns an error, but we ignore it, which I think is the correct behavior). 

I agree it's harmless to send a message, my concern is why this Receiver has been dropped, and is it an indication
that there is a problem with the other side (which I believe is propolis here?) or did the programmer on the other
side forget to wait on the `BlockReq`?  However, I think getting an error when trying to send the completion to
the `BlockReq` is different than not sending a completion at all.

>If we were deliberately not replying to a BlockReq anywhere in our existing code, it would panic, so I hope we're not doing that.

This is the protection I'm looking for, so maybe this is all off in the weeds.

>> I'm also concerned that we could be dropping a BlockReq and not knowing it. I do understand that keeping track
>> of things comes with added work for the programmer, but if that added work prevents future us from doing something
>> that leaves unresponded IOs out there, it might be hard to find.

> Automatically replying to the BlockReq on drop (i.e. the proposed new behavior) universally protects us from leaving unresponded IOs out there. 

Yes, leaving no IO behind is good, and I see how this proposal protects us from that.  But I think this also means we now expect on the drop semantics to automatically send error responses for us for our `BlockReq` if a response is not set?

> The current behavior guarantees that we will panic if our code happens to leave an unresponded IO out there, but that's still a potential availability issue that we'd have to catch and fix.

>Basically, I don't see the benefit of manually searching through every place where an unanswered BlockReq is dropped and adding code that sends an error; if our goal is to always send an error when an unanswered BlockReq is dropped, then we should just... do that, automatically, in the Drop implementation, so there's no room for human mistakes.

Well, maybe I don't understand things correctly here (totally possible) but are those places where we drop and
don't respond to a `BlockReq` bugs?  Are we (crucible) suppose to be responding and cleaning up all requests?

> If that's not our goal, then I'm pretty sure the existing panic isn't what we want either!

That is also quite possible.

mkeeter:
> I agree it's harmless to send a message, my concern is why this Receiver has been dropped, and is it an indication
that there is a problem with the other side (which I believe is propolis here?) or did the programmer on the other
side forget to wait on the BlockReq? However, I think getting an error when trying to send the completion to
the BlockReq is different than not sending a completion at all.

I agree that those two things are different, and I'm focused on everything that leads up to calling `send(..)` here.  I'm not proposing any changes to how we handle errors returned from that call.

(For more context, the "other side" is actually still within Crucible: it's the APIs within `struct Guest`, which (1) push a `BlockReq` to the queue then (2) wait on the corresponding `BlockReqWaiter`.  It's hard to imagine the oneshot failing other than during program exit, because `Guest::send` and `BlockReqWaiter::wait` are always called one after the other)

> Well, maybe I don't understand things correctly here (totally possible) but are those places where we drop and
don't respond to a BlockReq bugs? Are we (crucible) suppose to be responding and cleaning up all requests?

Yes to both.  In our current code, if we **ever** drop a `BlockReq` without sending a reply, then `crucible-upstairs` will panic (and take out whatever's using it as a library).

mkeeter:
After a bunch of discussion, I moved the `BlockReq` changes to a standalone PR https://github.com/oxidecomputer/crucible/pull/1085, which must be merged before this one.

(this PR is currently rebased onto that branch so that tests pass, but we'll re-rebase it onto `main` once #1085 is merged)

upstairs/src/deferred.rs
```diff
@@ -0,0 +1,186 @@
+// Copyright 2023 Oxide Computer Company
```
leftwo, line 1: 2024!!!!!!!!!!

leftwo (approved):

upstairs/src/block_req.rs
```diff
@@ -38,7 +38,10 @@ impl Drop for BlockRes {
     fn drop(&mut self) {
         // Dropping a BlockRes without issuing a completion would mean the
         // associated waiter would be stuck waiting forever for a result.
-        assert!(self.0.is_none(), "result should be sent for BlockRes");
+        if let Some(t) = self.0.take() {
+            // Ignore errors, since we don't care if the receiver is dropped
```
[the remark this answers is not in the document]
mkeeter: Yup, done in #1085

upstairs/src/deferred.rs
mkeeter: Fixed in 85b6994675fbe104475376eee6f21e2616e1ef16, WELCOME TO THE FUTURE

PR #1068, opened 2023-12-20
Avoid unnecessary allocations during frame decode

jmpesp:
Implementations of Decoder::decode should ensure that the buffer has appropriate capacity for the rest of the bytes of a frame. Our code was previously reserving the frame's total length, not space for the rest of the frame we're currently reading. This commit fixes that to avoid allocating unnecessary space.

Also, bail out during Encoder::encode if the frame is larger than MAX_FRM_LEN: the decode step will fail anyway.

### DIFF
--- a/protocol/src/lib.rs
+++ b/protocol/src/lib.rs
@@ -834,10 +834,17 @@ impl Encoder<Message> for CrucibleEncoder {
         dst: &mut BytesMut,
     ) -> Result<(), Self::Error> {
         let len = CrucibleEncoder::serialized_size(&m)?;
+        if len > MAX_FRM_LEN {
+            // Bail out before creating a frame that the decoder will refuse to
+            // deserialize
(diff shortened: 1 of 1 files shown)


protocol/src/lib.rs
```diff
@@ -834,10 +834,17 @@ impl Encoder<Message> for CrucibleEncoder {
         dst: &mut BytesMut,
     ) -> Result<(), Self::Error> {
         let len = CrucibleEncoder::serialized_size(&m)?;
+        if len > MAX_FRM_LEN {
```
mkeeter, line 837: Nit: can you add a test for this behavior?

mkeeter (approved):

protocol/src/lib.rs
jmpesp: done in ae64683, thanks!

Issue #1078, opened 2024-01-02
Panic during slow deactivation


mkeeter:
We spotted a panic during a [CI run of `test-crudd-benchmark`](https://buildomat.eng.oxide.computer/wg/0/details/01HK5K88PMBYB6809P36ETMABF/a7sOFPtRFA6rwn190QRJWsK0kQs6IO0Bf0T6c6ujKgmT1Eu8/01HK5K8ZJYQFMFGBY16J810JQN#S762).

The relevant portion of the logs are

```
{"msg":"[1] check deactivate YES","v":0,"name":"crucible","level":30,"time":"2024-01-02T17:55:30.459809148Z","hostname":"ip-10-150-1-178.us-west-2.compute.internal","pid":1192,"":"downstairs","":"downstairs","session_id":"56b3037b-bfb0-457f-a99f-dfd12d6c83e7"}
{"msg":"ds_transition from Active to Deactivated","v":0,"name":"crucible","level":30,"time":"2024-01-02T17:55:30.459820354Z","hostname":"ip-10-150-1-178.us-west-2.compute.internal","pid":1192,"client":"1","":"downstairs","session_id":"56b3037b-bfb0-457f-a99f-dfd12d6c83e7"}
{"msg":"[1] Transition from Active to Deactivated","v":0,"name":"crucible","level":30,"time":"2024-01-02T17:55:30.459832154Z","hostname":"ip-10-150-1-178.us-west-2.compute.internal","pid":1192,"client":"1","":"downstairs","session_id":"56b3037b-bfb0-457f-a99f-dfd12d6c83e7"}
{"msg":"deactivated client 1","v":0,"name":"crucible","level":30,"time":"2024-01-02T17:55:30.45984367Z","hostname":"ip-10-150-1-178.us-west-2.compute.internal","pid":1192,"session_id":"56b3037b-bfb0-457f-a99f-dfd12d6c83e7"}
{"msg":"client task is sending Done(RequestedStop(Deactivated))","v":0,"name":"crucible","level":40,"time":"2024-01-02T17:55:30.459890864Z","hostname":"ip-10-150-1-178.us-west-2.compute.internal","pid":1192,"":"io task","client":"1","":"downstairs","session_id":"56b3037b-bfb0-457f-a99f-dfd12d6c83e7"}
[...]

PR #1080, opened 2024-01-02
Wait to reconnect if `auto_promote` is false

mkeeter:
Fixes https://github.com/oxidecomputer/crucible/issues/1078

This PR adds a unit test for deactivation where one Downstairs finishes faster than the others (replicating the issue above), then fixes the resulting panic.

It does so by holding the IO task until a oneshot is fired.  If `auto_promote` is true (e.g. because we're reconnecting to a Downstairs), then this happens right away and there's no impact.  Otherwise, we wait for an activation request before connecting, meaning the downstairs state can't make it to `WaitForActive`.

I also squashed a few other bugs discovered implementing this unit test:
- Panic during shutdown if we were in `GoActive` or `Deactivating`, because a `BlockRes` was held.  This was fixed by implementing `Drop` on `Upstairs` and sending back an error reply.
- Panic if `Upstairs::set_active_request` was called repeatedly, because we didn't `return` after sending an error on the channel (so `DownstairsClient::set_active_request` was called repeatedly, which is invalid)

### DIFF
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -44,6 +44,12 @@ struct ClientTaskHandle {
     /// replies, but also things like "the I/O task has stopped"
     client_response_rx: mpsc::Receiver<ClientResponse>,
 
+    /// One-shot sender to ask the client to open its connection
+    ///
+    /// This is used to hold the client (without connecting) in cases where we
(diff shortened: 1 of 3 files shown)


leftwo (approved):
👍

PR #1083, opened 2024-01-03
Remove unused fields in match statements and consolidate

mkeeter:
This is minor house-keeping that I extracted from a pending branch, since it's standalone:

- `match` statements don't need to explicitly ignore all members; you can use `..` as "ignore the rest"
- `match` statements with equivalent arms can be consolidated to reduce duplication
- Replacing a few `for` loops with `iter().all(..)`


leftwo (approved):

leftwo:
Thanks for breaking out changes like this.
It makes it easier when going back through history to know what changes are associated with what PR.

PR #1085, opened 2024-01-04
Remove panic-on-drop logic

mkeeter:
https://github.com/oxidecomputer/crucible/pull/1064 added new logic to the `BlockReq`: if the notification channel was dropped without a reply having been sent, Crucible would panic.

After a bunch of discussion, I no longer think this extra complexity is worth it, and would like to remove it.

## Why did we do this in the first place?
In general, we expect a `BlockReq` to _always_ be answered by Crucible.  If one is dropped without a reply being sent, it indicates a programmer error.  This behavior was inspired by [similar logic in Propolis](https://github.com/oxidecomputer/propolis/blob/85dc6db36709981c1d35b0651a18e20caee0b73f/lib/propolis/src/block/mod.rs#L176-L182).

## Propolis and Crucible are different
The `Request` in Propolis is built around a `device::TrackingMarker`.  If it is dropped without a reply, then the listening side will **wait forever**; this is a potential liveness issue, so turning it into a panic is sensible.

The `BlockReq` in Crucible is built around a `tokio::sync::oneshot` channel.  If it is dropped without a reply, then the `oneshot::Receiver` will return immediately with an error.  In fact, we handle that case in our `BlockReqWaiter`, translating this error into `CrucibleError::RecvDisconnected` ([here](https://github.com/oxidecomputer/crucible/blob/e813da3f/upstairs/src/block_req.rs#L67)).  In other words, we're _already_ handling the case of a `BlockReq` being dropped without a reply (by reporting an error).  There is no potential liveness issue.

(This also means that [the block comment in `BlockRes::drop`](https://github.com/oxidecomputer/crucible/blob/e813da3f/upstairs/src/block_req.rs#L39-L40) is incorrect)

It's arguably better to panic upon programmer error, versus going through normal error-handling paths (which defer handling to the caller of the `Guest` API).  However...

## False positives during program exit
There is one time when a `BlockReq` can be dropped without a reply: during program exit (or scope exit, for unit tests).

As far as I can tell, this panic **hasn't caught any true programmer errors**.  Instead, it's found a bunch of cases where `BlockReq` are dropped during exit, and we've had to add `Drop` implementations to prevent it from firing.  The original PR added [`impl Drop for Guest`](https://github.com/oxidecomputer/crucible/blob/e813da3f/upstairs/src/lib.rs#L2424-L2446), but it missed a sneaky case where `BlockReq` were stored in the `UpstairsState`; a later PR added [`impl Drop for Upstairs`](https://github.com/oxidecomputer/crucible/blob/e813da3f/upstairs/src/upstairs.rs#L194-L207) to avoid panics during certain unit tests.

## Future false positives
This check also makes it easy to accidentally create false positives when writing new code.  The [`offload-write-encryption` branch](https://github.com/oxidecomputer/crucible/pull/1066) moves some `BlockReq`s into a thread pool to perform encryption outside of the main task.  However, because we didn't add special handling for program exit, this caused a panic at shutdown.

## This is hard to do by accident
Remember what we're actually trying to protect against: someone dropping a `BlockReq` without replying.  Because we've annotated relevant functions with `#[must_use]`, this would require either ignoring warnings or explicitly discarding the `BlockReq`.  It's hard to do by accident!

## Summary
- This `panic!` guards against something that's hard to do by accident
- If we **do** make this mistake, it's reported as a distinct error to the caller
  - Unlike Propolis, there's no liveness concern
- So far, 100% of issues it discovered are false positives at program exit
  - It is easy to introduce similar false positives when writing new code
  - New false positives will manifest as panics during program exit
  - If we introduce new false positives, we will not discover the issue until runtime (and it may be rare, because Tokio does not destroys tasks in a deterministic order)

## Proposal
- Remove the `panic!`
- Remove the `Drop` implementations which work around it at exit
- Add logging for this case:
  - Fire a DTrace probe if a `BlockReq` is dropped without a reply
  - Log when we see a `CrucibleError::RecvError`

### DIFF
--- a/upstairs/src/block_req.rs
+++ b/upstairs/src/block_req.rs
@@ -36,9 +36,11 @@ impl BlockRes {
 }
 impl Drop for BlockRes {
     fn drop(&mut self) {
-        // Dropping a BlockRes without issuing a completion would mean the
-        // associated waiter would be stuck waiting forever for a result.
-        assert!(self.0.is_none(), "result should be sent for BlockRes");
+        if self.0.is_some() {
+            // During normal operation, we expect to reply to every BlockReq, so
+            // we'll fire a DTrace probe here.
+            cdt::up__block__req__dropped!();
+        }
     }
 }
 
@@ -61,10 +63,19 @@ impl BlockReqWaiter {
     }
 
     /// Consume this BlockReqWaiter and wait on the message
-    pub async fn wait(self) -> Result<(), CrucibleError> {
+    ///
+    /// If the other side of the oneshot drops without a reply, log an error
+    pub async fn wait(self, log: &Logger) -> Result<(), CrucibleError> {
         match self.recv.await {
             Ok(v) => v,
-            Err(_) => crucible_bail!(RecvDisconnected),
+            Err(_) => {
+                warn!(
+                    log,
+                    "BlockReqWaiter disconnected; \
+                     this should only happen at exit"
+                );
+                Err(CrucibleError::RecvDisconnected)
+            }
         }
     }
 
@@ -92,7 +103,7 @@ mod test {
 
         res.send_ok();
 
-        brw.wait().await.unwrap();
+        brw.wait(&crucible_common::build_logger()).await.unwrap();
     }
 
     #[tokio::test]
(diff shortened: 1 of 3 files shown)


leftwo (approved):

pfmooney:
Since this is more than just cleaning up that `drop` logic, I'd suggest the commit message reflect that.  Something more along the lines of better handling runtime shutdown or whatever.

pfmooney (approved):

PR #1086, opened 2024-01-05
Make guest work queue async

mkeeter:
The upstairs `struct Guest` is currently used in at least two different tasks:
- The guest uses the `trait BlockIO` to push requests into a queue (protected by a `tokio::sync::Mutex`).  There can be multiple tasks sharing an `Arc<Guest>` here, e.g. 8x in Propolis.
- The `struct Upstairs` pulls requests from that queue in the `up_main` event loop

This PR splits the `Guest` into two separate `structs`, which are **owned** (not shared) by their respective tasks:
- `Guest` (same name as before) implements `BlockIO` and is used by the guest task to send new jobs
- `GuestIoHandle` (new!) is used by the `struct Upstairs` and receives those jobs

The two halves communicate via a `tokio::sync::mpsc` queue, removing the previous combination of `tokio::sync::Mutex<VecDeque<BlockReq>> + tokio::sync::Notify`

It's still possible to share a `Guest` by wrapping it in an `Arc`, just like before.

As a side effect of removing the `tokio::sync::Mutex`, many functions become synchronous; don't be alarmed by the LOC  in this PR, since they're mostly mechanical.

I don't see any performance changes here (alas), but I think it's still worthwhile: it simplifies our code and helps set up a future consolidation of backpressure logic.

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -4,7 +4,7 @@
 #![allow(clippy::mutex_atomic)]
 
 use std::clone::Clone;
-use std::collections::{HashMap, VecDeque};
+use std::collections::HashMap;
 use std::fmt;
 use std::fmt::{Debug, Formatter};
 use std::io::{Read as _, Result as IOResult, Seek, SeekFrom, Write as _};
@@ -25,7 +25,7 @@ use ringbuffer::{AllocRingBuffer, RingBuffer};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use slog::{error, info, o, warn, Logger};
-use tokio::sync::{mpsc, oneshot, Mutex, Notify, RwLock};
+use tokio::sync::{mpsc, oneshot, Mutex, RwLock};
 use tokio::time::Instant;
 use tracing::{instrument, span, Level};
 use usdt::register_probes;
@@ -73,7 +73,7 @@ use async_trait::async_trait;
 mod client;
 mod downstairs;
 mod upstairs;
-use crate::upstairs::UpCounters;
+use upstairs::{UpCounters, UpstairsAction};
 
 // Max number of outstanding IOs between the upstairs and the downstairs
 // before we give up and mark that downstairs faulted.
@@ -2076,6 +2076,10 @@ struct GuestWork {
 }
 
 impl GuestWork {
+    fn is_empty(&self) -> bool {
+        self.active.is_empty()
+    }
+
     fn next_gw_id(&mut self) -> GuestWorkId {
         let id = self.next_gw_id;
         self.next_gw_id += 1;
@@ -2134,76 +2138,18 @@ impl Default for GuestWork {
     }
 }
 
-/**
- * This is the structure we use to keep track of work passed into crucible
- * from the "Guest".
- *
- * Requests from the guest are put into the reqs VecDeque initially.
- *
- * A task on the Crucible side will receive a notification that a new
- * operation has landed on the reqs queue and will take action:
- *
- * * Pop the request off the reqs queue.
- *
- * * Copy (and optionally encrypt) any data buffers provided to us by the
- *   Guest.
- *
- * * Create one or more downstairs DownstairsIO structures.
- *
- * * Create a GtoS tracking structure with the id's for each downstairs task
- *   and the read result buffer if required.
- *
- * * Add the GtoS struct to the in GuestWork active work hashmap.
- *
- * * Put all the DownstairsIO structures on the downstairs work queue.
- *
- * * Send notification to the upstairs tasks that there is new work.
- *
- * Work here will be added to storage side queues and the responses will
- * be waited on and processed when they arrive.
- *
- * This structure and operations on in handle the translation between
- * outside requests and internal upstairs structures and work queues.
- */
+/// IO handles used by the guest uses to pass work into Crucible proper
+///
+/// This data structure is the counterpart to the [`GuestIoHandle`], which
+/// receives work from the guest and is exclusively owned by the
+/// [`upstairs::Upstairs`]
+///
+/// Requests from the guest are put into the `req_tx` queue by the guest, and
+/// received by the [`GuestIoHandle::req_rx`] side.
 #[derive(Debug)]
 pub struct Guest {
-    /*
-     * New requests from outside go onto this VecDeque. The notify is how
-     * the submission task tells the listening task that new work has been
-     * added.
-     */
-    reqs: Mutex<VecDeque<BlockReq>>,
-    notify: Notify,
-
-    /*
-     * When the crucible listening task has noticed a new IO request, it
-     * will pull it from the reqs queue and create an GuestWork struct
-     * as well as convert the new IO request into the matching
-     * downstairs request(s). Each new GuestWork request will get a
-     * unique gw_id, which is also the index for that operation into the
-     * hashmap.
-     *
-     * It is during this process that data will encrypted. For a read, the
-     * data is decrypted back to the guest provided buffer after all the
-     * required downstairs operations are completed.
-     */
-    guest_work: Mutex<GuestWork>,
-
-    /*
-     * Setting an IOP limit means that the rate at which block reqs are
-     * pulled off will be limited. No setting means they are sent right
-     * away.
-     */
-    iop_tokens: std::sync::Mutex<usize>,
-    bytes_per_iop: Option<usize>,
-    iop_limit: Option<usize>,
-
-    /*
-     * Setting a bandwidth limit will also limit the rate at which block
-     * reqs are pulled off the queue.
-     */
-    bw_tokens: std::sync::Mutex<usize>, // bytes
-    bw_limit: Option<usize>,            // bytes per second
+    /// New requests from outside go into this queue
+    req_tx: mpsc::Sender<BlockReq>,
 
     /// Local cache for block size
     ///
@@ -2213,10 +2159,10 @@ pub struct Guest {
     block_size: AtomicU64,
 
     /// Backpressure is implemented as a delay on host write operations
-    backpressure_us: AtomicU64,
-
-    /// Backpressure configuration, as a starting point and max delay
-    backpressure_config: BackpressureConfig,
+    ///
+    /// It is stored in an `Arc` so that the `GuestIoHandle` can update it from
+    /// the IO task.
+    backpressure_us: Arc<AtomicU64>,
 
     /// Lock held during backpressure delay
     ///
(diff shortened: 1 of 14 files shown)


upstairs/src/lib.rs
```diff
@@ -2607,6 +2411,186 @@ impl BlockIO for Guest {
...
+        } else if let Some(req) = self.req_rx.recv().await {
+            // NOTE: once we take this req from the queue, we must be cancel
+            // safe!  In other words, we cannot yield until either (1) returning
+            // the req or (2) storing it in self.req_head for safe-keeping.
+            req
+        } else {
+            futures::future::pending().await
+        };
+
+        // Check if we can consume right away
```
leftwo, line 2508: The meat of this method appears to be pretty much the same as it was before.
Thanks github for not making that easy to determine.

upstairs/src/test.rs
```diff
@@ -711,27 +700,32 @@ pub(crate) mod up_test {
...
-        ($guest:expr) => {{
-            assert!($guest.consume_req().await.is_none());
-        }};
+
+    async fn assert_none_consumed(io: &mut GuestIoHandle) {
+        tokio::select! {
+            _ = io.recv() => {
+                panic!("got message when expecting nothing")
+            },
+            _ = tokio::time::sleep(std::time::Duration::from_millis(1)) => {
```
leftwo, line 719: Whats the purpose of the sleep here (and, really the above one also)?

upstairs/src/upstairs.rs
```diff
@@ -1119,8 +1115,7 @@ impl Upstairs {
          * ID and the next_id are connected here, in that all future writes
          * should be flushed at the next flush ID.
          */
-        let mut gw = self.guest.guest_work.lock().await;
-        let gw_id = gw.next_gw_id();
+        let gw_id = self.guest.guest_work.next_gw_id();
```
leftwo, line 1118: The guest work lock (along with the former downstairs lock) provided protection
during IO submission that prevented any interleaving of IO requests with each other.

In the new "one task" world, there will only ever be one task that calls
`submit_flush`
`submit_read`
`submit_write`
Correct? 

Is this also enforced by the way the new `Guest` struct is split, such that only one owner can exist and attempt to handle incoming IOs?

upstairs/src/upstairs.rs
```diff
@@ -2109,22 +2093,18 @@ pub(crate) mod test {
         upstairs.force_active().unwrap();
 
         // op 0
-        upstairs
```
leftwo, line 2112: Did `cargo fmt` do this?
Maybe it decided to change things because the `.await;` went away?

upstairs/src/lib.rs
```diff
@@ -2140,76 +2108,75 @@ struct BackpressureConfig {
...
-        self.notify.notify_one();
-
+        if let Err(e) = self.req_tx.send(BlockReq { op, res }).await {
+            // This could happen during shutdown, if the up_main task is
+            // destroyed while the Guest is still trying to do work.
+            //
+            // If this happens, then the BlockReqWaiter will immediately return
+            // with CrucibleError::RecvDisconnected (since the oneshot::Sender
+            // will have been dropped into the void).
+            warn!(self.log, "failed to send op to guest: {e}");
```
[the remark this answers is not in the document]
mkeeter: `req_tx` is bounded to 500 requests (value chosen at random).

However, the Crucible upstairs on the other side is constantly processing data from the queue (running `up_main`).  If the queue is temporarily full, we'll just wait here at the `send(..).await`.  To the caller, waiting for things to go into the queue is indistinguishable from waiting for the `BlockReqWaiter`.

upstairs/src/lib.rs
```diff
@@ -2607,6 +2411,186 @@ impl BlockIO for Guest {
...
+        self.req_limited = false;
+    }
+
+    /// Listen for new work
+    ///
+    /// This will wait forever if we are currently IOP / BW limited; otherwise,
+    /// it will return the next value from the `BlockReq` queue.
+    async fn recv(&mut self) -> BlockReq {
+        let req = if self.req_limited {
+            futures::future::pending().await
```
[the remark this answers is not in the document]
mkeeter: This is always called in a `select!` statement with a timeout as one of the other branches, which will leak IOP / BW tokens.

upstairs/src/lib.rs
mkeeter: Yes, that's correct – once we get the request off the queue, the logic should be the same.

upstairs/src/lib.rs
```diff
@@ -2607,6 +2411,186 @@ impl BlockIO for Guest {
...
+            req
+        } else {
+            assert!(self.req_head.is_none());
+            self.req_head = Some(req);
+            futures::future::pending().await
+        }
+    }
+
+    /// Set `self.backpressure_us` based on outstanding IO ratio
+    fn set_backpressure(&self, bytes: u64, ratio: f64) {
```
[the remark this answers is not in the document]
mkeeter: Yup; it moved because ownership of the `backpressure_config` moved into the `GuestIoHandle`

upstairs/src/test.rs
mkeeter: `io.recv()` will either
1) return an item (if available)
2) wait for an available item (if nothing is available but we're not BW / IOP limited), or
3) wait forever (if we're BW / IOP limited)

This makes sense in production use, because it's always used as part of a `select!` statement (so if it waits forever, a timeout branch in the `select!` will be taken).  In the test suite, we're always either doing (1) or (3), so this `select!` distinguishes between the "return right away" and "wait forever" cases.

upstairs/src/upstairs.rs
mkeeter: It has always been possible to clone a `Guest` and submit many jobs simultaneously, e.g. Propolis uses 8x Crucible worker tasks.

In the old code, jobs were serialized through the `VecDeque` in the `Guest`, which was locked by both sides (8x worker tasks and one Upstairs task).  Now, jobs are serialized through an `mpsc` queue between the `Guest` and the `GuestIoHandle`, which should be equivalent.

`Upstairs::submit_flush/read/write` all take `&mut self`, so they are guaranteed to only be called by the single owner of the `struct Upstairs`.

upstairs/src/upstairs.rs
mkeeter: Yup, thanks `cargo fmt`

upstairs/src/lib.rs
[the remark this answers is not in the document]
leftwo: As long as we can't deadlock ourselves here, where Crucible Upstairs is stuck trying to send something
over the `BlockReqWaiter`.

I bring it up because we have done this to ourselves in the past :)

upstairs/src/upstairs.rs
leftwo: > Upstairs::submit_flush/read/write all take &mut self, so they are guaranteed to only be called by the single owner of the struct Upstairs.

Perfect, this is what I wanted to see/hear

upstairs/src/lib.rs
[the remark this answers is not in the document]
leftwo: Ah, is this done in the `upstairs/src/upstairs.rs` one task select:
```
            _ = sleep_until(self.leak_deadline) => {
                UpstairsAction::LeakCheck
            }
```

leftwo (approved):
I have nothing more to add, thanks for the work here.

upstairs/src/lib.rs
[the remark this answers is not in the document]
hawkw: it might be worth mentioning that explicitly in the doc comment on this method, so that it's a bit more obvious?

upstairs/src/lib.rs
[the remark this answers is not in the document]
mkeeter: I think we're safe – replying to a `BlockReqWaiter` is a synchronous operation that doesn't require cooperation from the other side, since it's a `tokio::sync::oneshot`.

upstairs/src/lib.rs
```diff
@@ -2049,43 +2049,11 @@ impl Default for GuestWork {
  */
 #[derive(Debug)]
 pub struct Guest {
```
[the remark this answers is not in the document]
mkeeter: Yeah, done in 972143930ac6ba6ea3270d5ce81184bab16741f2 (by rearranging some other comments and adding more details)

upstairs/src/lib.rs
```diff
@@ -2140,76 +2085,75 @@ struct BackpressureConfig {
...
-            notify: Notify::new(),
-            /*
-             * The active hashmap is for in-flight I/O operations
-             * that we have taken off the incoming queue, but we have not
-             * received the response from downstairs.
-             * Note that a single IO from outside may have multiple I/O
-             * requests that need to finish before we can complete that IO.
-             */
-            guest_work: Mutex::new(GuestWork {
+        let (req_tx, req_rx) = mpsc::channel(500);
```
jmpesp, line 2090: It'd be worth adding a comment for how this number was chosen, even if it's arbitrary

upstairs/src/lib.rs
```diff
@@ -2607,6 +2388,212 @@ impl BlockIO for Guest {
...
+            futures::future::pending().await
+        } else if let Some(req) = self.req_head.take() {
+            req
+        } else if let Some(req) = self.req_rx.recv().await {
+            // NOTE: once we take this req from the queue, we must be cancel
+            // safe!  In other words, we cannot yield until either (1) returning
+            // the req or (2) storing it in self.req_head for safe-keeping.
+            req
+        } else {
+            futures::future::pending().await
```
jmpesp, line 2508: If we enter this branch, it means that the `recv` above returned `None` right? In that case the `req` channel was closed: it seems like `crucible_bail!(RecvDisconnected)` (or an equivalent for this new channel) may apply here?

jmpesp (changes requested):

upstairs/src/upstairs.rs
```diff
@@ -449,12 +441,13 @@ impl Upstairs {
                 const LEAK_MS: usize = 1000;
                 let leak_tick =
                     tokio::time::Duration::from_millis(LEAK_MS as u64);
-                if let Some(iop_limit) = self.guest.get_iop_limit() {
-                    let tokens = iop_limit / (1000 / LEAK_MS);
+                let limits = *self.guest.limits.lock().unwrap();
```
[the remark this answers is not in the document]
mkeeter: Sure, done in 84669db1de07f2a27ec8797f80bc5f22546011c2

It's starting to feel a _little_ silly: right now, they only exist for the sake of unit tests in `crucible-upstairs`, since no one is setting them to anything other than `None` in production code.  Still, further refactoring / cleanup is probably out of scope for this PR.

upstairs/src/lib.rs
mkeeter: Added a longer comment in 1be982b7173d7bdf9d3485e02dfc7cad08bbd46e

upstairs/src/lib.rs
mkeeter: I'm not sure!

In the current code, there's nothing to acknowledge when the last clone of the `Arc<Guest>` (that _isn't_ owned by the upstairs) is destroyed; we just stop seeing messages come through from the one remaining clone.  This means that the `Upstairs` will pick a different action in its event loop, since we have plenty of remaining valid `select!` branches.

This code matches that existing behavior: if the `Guest` is dropped, we'll never see any more messages from it, and the `Upstairs` will go about its business by picking one of the other `select!` branches.

My inclination is to keep the current logic, but add a logged message to that effect.  Maybe in the future we'd want to quit the Upstairs event loop if the `Guest` is dropped, since no one would be able to control it at that point, but I think that could come in a later PR.

What do you think?

upstairs/src/lib.rs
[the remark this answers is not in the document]
mkeeter: Yup, added in 9d6e0b108377c419444d86552963bb97ae77c0b3

upstairs/src/lib.rs
jmpesp: > In the current code, there's nothing to acknowledge when the last clone of the Arc<Guest> (that isn't owned by the upstairs) is destroyed; we just stop seeing messages come through from the one remaining clone.

With this PR, it may be ok to abort if we see None here: `Upstairs` now holds `GuestIoHandle`, not `Guest`, so I think the last drop means the channel will be closed?

mkeeter: I'm not sure I understand what you mean by "ok to abort".

The `Upstairs` holds the `GuestIoHandle`'s `mpsc::Receiver` end, but the `Guest` holds the `mpsc::Sender`.  If the `Guest` is dropped, then the receiver will see `None` here.  As such, if we panicked upon seeing `None`, we'd panic whenever the guest is dropped, which we don't have control over (since that's managed by whoever has embedded the `crucible-upstairs` library).

Here's a diff to double-check:
```diff
diff --git a/integration_tests/src/lib.rs b/integration_tests/src/lib.rs
index c767476..61967ab 100644
--- a/integration_tests/src/lib.rs
+++ b/integration_tests/src/lib.rs
@@ -3102,6 +3102,24 @@ mod test {
         Ok(())
     }

+    #[tokio::test]
+    async fn integration_test_guest_drop() -> Result<()> {
+        // Spin off three downstairs, build our Crucible struct.
+        let tds = TestDownstairsSet::small(false).await?;
+        let opts = tds.opts();
+
+        let (guest, io) = Guest::new(None);
+
+        let join_handle = up_main(opts, 1, None, io, None)?;
+
+        drop(guest);
+        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+
+        let r = join_handle.await;
+        assert!(r.is_ok());
+        Ok(())
+    }
+
     #[tokio::test]
     async fn integration_test_guest_zero_length_io() -> Result<()> {
         // Test the guest layer with a write and read of zero length
diff --git a/upstairs/src/lib.rs b/upstairs/src/lib.rs
index 8644305..f208408 100644
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -2505,7 +2505,7 @@ impl GuestIoHandle {
         } else {
             if !self.guest_dropped {
                 self.guest_dropped = true;
-                warn!(self.log, "Guest handle has been dropped");
+                panic!("Guest handle has been dropped");
             }
             futures::future::pending().await
         };
```

This unit test fails because this panic fires when the `Guest` is dropped:
```
thread 'test::integration_test_guest_drop' panicked at upstairs/src/lib.rs:2508:17:
Guest handle has been dropped
```

...and then the join handle returns an error: `Err(JoinError::Panic(Id(13), ...))`

(This would be even more dramatic with `panic=abort`, which isn't enabled for unit tests)

mkeeter: Got it, please take a look at 6874eb3b89f8d64f577714ba3bc094fc53ac5e0e (while we wait for CI to test everything)

jmpesp: Nice :+1:

jmpesp (approved):

PR #1089, opened 2024-01-05
Implement read decryption offloading

mkeeter:
Analogous to #1066 , this PR moves read decryption into the rayon thread pool.  It uses exactly the same infrastructure (particularly `DeferredQueue`), and is implemented in a very similar way.

I need to do some benchmarking before this goes in, but wanted to open it for review early.

### DIFF
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -2782,7 +2782,7 @@ pub(crate) fn validate_unencrypted_read_response(
             Ok(None)
         } else {
             error!(log, "got empty block context with non-blank block");
-            Err(CrucibleError::HashMismatch)
+            Err(CrucibleError::MissingBlockContext)
         }
     }
 }
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -1170,10 +1170,15 @@ impl DownstairsClient {
     /// Handles a single IO operation
     ///
     /// Returns `true` if the job is now ackable, `false` otherwise
+    ///
+    /// If this is a read response, then the values in `responses` must
+    /// _already_ be decrypted (with corresponding hashes stored in
+    /// `read_response_hashes`).
     pub(crate) fn process_io_completion(
         &mut self,
         job: &mut DownstairsIO,
-        mut responses: Result<Vec<ReadResponse>, CrucibleError>,
+        responses: Result<Vec<ReadResponse>, CrucibleError>,
+        read_response_hashes: Vec<Option<u64>>,
         deactivate: bool,
         extent_info: Option<ExtentInfo>,
     ) -> bool {
@@ -1189,26 +1194,9 @@ impl DownstairsClient {
         let mut jobs_completed_ok = job.state_count().completed_ok();
         let mut ackable = false;
 
-        // Validate integrity hashes and optionally authenticated decryption.
-        //
-        // With AE, responses can come back that are invalid given an encryption
-        // context. Test this here. If it fails, then something has gone
-        // irrecoverably wrong and we should panic.
-        let mut read_response_hashes = Vec::new();
-        let new_state = match &mut responses {
-            Ok(responses) => {
-                responses.iter_mut().for_each(|x| {
-                    let mh =
-                        if let Some(context) = &self.cfg.encryption_context {
-                            validate_encrypted_read_response(
-                                x, context, &self.log,
-                            )
-                        } else {
-                            validate_unencrypted_read_response(x, &self.log)
-                        }
-                        .expect("decryption failed");
-                    read_response_hashes.push(mh);
-                });
+        let new_state = match &responses {
+            Ok(..) => {
+                // Messages have already been decrypted out-of-band
                 jobs_completed_ok += 1;
                 IOState::Done
             }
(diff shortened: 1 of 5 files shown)


mkeeter:
I did some benchmarking on `c12`, using the following `fio` script:
```
[global]
filename=/dev/nvme0n1
iodepth=25
ioengine=aio
time_based
runtime=60
numjobs=1
direct=1
stonewall=1

[randwrite-4M]
bs=4M
rw=randwrite

[randread-4K]
bs=4K
rw=randread

[randread-1M]
bs=1M
rw=randread

[randread-4M]
bs=4M
rw=randread
```

(the 4M writes are to put some data on the virtual disk, which is otherwise blank)

I see the following performance (in MiB/s)

| Read size | `main` | `offload-read-decryption` |
|-|-|-|
| 4K | 50.5 | 42.1 |
| 1M | 164 | 216 |
| 4M | 156 | 218 |

It looks like the overhead of moving decryption off-thread isn't worth it for small reads, which isn't totally surprising.

Doing a more detailed sweep, I see this:
Screenshot 2024-01-08 at 11 02 05 AM

The crossover point looks like about 8 KiB, so I'm going to tweak the code to decrypt immediately below that size.

mkeeter:
Running with the change above (8f3ba46bd1477df82ba668c8b07b783d50215e66) to only defer > 8 KiB read decryption, I'm only seeing a small improvement (42.7 -> 44.2 MiB/s).  This is still below the performance on `main` (50.6 MiB/s), so I think more investigation is needed.

mkeeter:
Re-running on Crucible `main` (099c201ec0a2710472417d0d711fa870ec65fe85), I see 44.5 MiB/s, so I suspect the 50.6 MiB/s reading was a fluke.  This means that there's no performance regression, so we can move forward with the PR.

faithanalog:
I am not getting the exact same numbers you are on my test setup, but I am getting a similar ratio of improvement. about 40-45% better for 16k reads, 75-80% better for 4M reads.

upstairs/src/client.rs
```diff
@@ -2623,8 +2611,12 @@ pub(crate) fn validate_encrypted_read_response(
         // expect to see this case unless this is a blank block.
         //
         // XXX if it's not a blank block, we may be under attack?
-        assert!(response.data[..].iter().all(|&x| x == 0));
-        return Ok(None);
+        if response.data[..].iter().all(|&x| x == 0) {
+            return Ok(None);
+        } else {
+            error!(log, "got empty block context with non-blank block");
+            return Err(CrucibleError::DecryptionError);
```
jmpesp, line 2618: We should have a different error variant for this

upstairs/src/downstairs.rs
```diff
@@ -3185,6 +3187,7 @@ impl Downstairs {
             ds_id,
             client_id,
             responses,
+            vec![], // XXX this could lead to false matches; does it matter?
```
jmpesp, line 3190: I'm surprised this works!

upstairs/src/upstairs.rs
```diff
@@ -1420,7 +1453,62 @@ impl Upstairs {
...
+                    const MIN_DEFER_SIZE_BYTES: u64 = 8192;
+                    let should_defer = !self.deferred_msgs.is_empty()
+                        || match responses {
+                            Ok(rs) => {
+                                // Find the number of bytes being decrypted
+                                let response_size = rs.len() as u64
+                                    * self
+                                        .ddef
+                                        .get_def()
+                                        .map(|b| b.block_size())
```
jmpesp, line 1476: This multiplication assumes that we only send block sized chunks - we currently do, but why not sum up the ReadResponse data lengths?

jmpesp (changes requested):

upstairs/src/client.rs
mkeeter: Sure, done in 264cbeff4edc63aa13b70721dd6fae65de66df88

upstairs/src/downstairs.rs
mkeeter: Yup, this takes post-decryption data, so it's not actually checking the hashes (other than seeing that each of the 3x reads returns the same thing).  It's a little awkward, but this is a test-only function, so I think we can let it slide.

(I also want to remove `process_ds_completion` entirely at some point; it's only still around because I didn't want to port a zillion unit tests while also refactoring the Upstairs)

upstairs/src/upstairs.rs
mkeeter: No particular reason, other than it's O(1) instead of O(n).  I can switch if you have a preference, although I think if we switched away from block-size chunks, **many other** places would break too.

upstairs/src/downstairs.rs
jmpesp: > so it's not actually checking the hashes (other than seeing that each of the 3x reads returns the same thing)

Yeah, this is why I was surprised, but I guess that makes sense: all three reads did return the same thing (`vec![]`). Is it worth a check to see that _some_ hash was returned?

upstairs/src/upstairs.rs
jmpesp: Agreed, if we decide to support non-block-sized chunks, the existing tests will all happily tell us what to fix :)

upstairs/src/downstairs.rs
mkeeter: I think it makes sense to assert that the number of hashes == number of blocks in `process_io_completion`; if that's not the case, then it's definitely a logic error in the offloaded decryption!

Implemented in 9d9f15fe498344b81778ec72bdbcea8dc8f13f03, with an explanatory comment 😄

PR #1091, opened 2024-01-09
Make Buffer track per-block (not per-byte) ownership

mkeeter:
This reduces storage (and operations) in the `struct Buffer` by the block size, so a factor of 512× or 4096×.

The downside is that we need to know the block size at construction time, so there's a bunch of mechanical changes in this PR.

I see significant speedups for all sizes of random reads except 4K, which become slightly slower (50.6 MiB/s → 46.9 MiB/s, not sure why).

Screenshot 2024-01-09 at 3 01 13 PM

### DIFF
--- a/crudd/src/main.rs
+++ b/crudd/src/main.rs
@@ -258,7 +258,7 @@ async fn cmd_read<T: BlockIO>(
         // let block_remainder = remainder % native_block_size;
         // round up
         let blocks = (remainder + native_block_size - 1) / native_block_size;
-        let buffer = Buffer::new(blocks, native_block_size as usize);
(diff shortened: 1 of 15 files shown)


mkeeter:
Re-running on Crucible `main` (099c201ec0a2710472417d0d711fa870ec65fe85), I see 44.5 MiB/s, so I suspect the 50.6 MiB/s reading was a fluke.  This means that there's no performance regression, so we can move forward with the PR.

mkeeter:
(let's put this on hold until #1094, which touches a bunch of similar code)

mkeeter:
Closing in favor of #1105, since this has diverged from `main` and I want to split into two bite-sized PRs.

Issue #1092, opened 2024-01-09
Don't check the read_only_parent if the scrub is done


leftwo:
If a scrub is completed, there is no reason to check the read only parent.

The current code can, if an IO spans the end of the ROP region, be allowed to
be sent to the ROP.  The correct data is still returned as the read to the Sub Volume
will come back and we use that, but we should not even be asking the ROP for
anything once we have finished the scrub.

Mostly harmless, unless the ROP downstiars are offline.  In that case the
IO that is sent to the ROP (that we will eventually discard) will get stuck forever.

The check in the volume layer:
```
if start + length <= scrub_point {
```
Is where the problem lies.  While this IO is, in part, "below" the scrub point, you would
think it should be sent to the ROP.  In this case however, the ROP ends at the scrub point, so
we don't need to ask the ROP for anything.

PR #1093, opened 2024-01-10
Don't check ROP if the scrub is done

leftwo:
**Don't ask the read only parent for anything once the scrub has completed.**

When asking a Volume with a ROP to see if we need to read from the ROP
or not, also check to see if the scrub is done and don't read from the ROP if so.

There was a problem where a read that started inside the ROP range, but extended
beyond it would still request data from the ROP (which we would have later discarded).

This is not a terrible problem, unless the ROP has gone away.  If the ROP is gone, then the
IO will sit in the upstairs waiting for at least one downstairs to answer.


leftwo:
Fixes https://github.com/oxidecomputer/crucible/issues/1092

PR #1094, opened 2024-01-10
Remove Arc + Mutex from Buffer

jmpesp:
By transfering Buffer's ownership to the BlockIO implementation instead of passing by reference, the Arc + Mutex that guards Buffer's internals can be removed. Return the Buffer when done so that ownership is transferred back to the caller. Callers will have to change like so:

    -        crucible.read(offset, buffer.clone()).await?;
    +        let buffer = crucible.read(offset, buffer).await?;

Non-read operations are not affected.

Additionally, this commit removes the ability for one Guest operation to be composed of multiple Downstairs operations: this support could be added back, but enough was changing in this commit as is :) We don't currently use this "bulk" functionality except in tests.

### DIFF
--- a/crudd/src/main.rs
+++ b/crudd/src/main.rs
@@ -242,8 +242,8 @@ async fn cmd_read<T: BlockIO>(
     // Drain the outstanding commands
     if !futures.is_empty() {
         // drain the buffer to the output file
-        while !futures.is_empty() {
-            let r_buf = futures.pop_front().unwrap().await?;
+        for f in futures.drain(..) {
+            let r_buf = f.await?;
             output.write_all(&r_buf)?;
         }
     }
--- a/crudd/src/main.rs
+++ b/crudd/src/main.rs
@@ -222,13 +222,12 @@ async fn cmd_read<T: BlockIO + std::marker::Send + 'static>(
 
             // Send the read command with whichever buffer is at the back of the
             // queue. We re-use the buffers to avoid lots of allocations
-            for mut buffer in buffers.drain(..) {
-                let crucible = crucible.clone();
-                futures.push_back(tokio::spawn(async move {
-                    crucible.read(offset, &mut buffer).await?;
-                    Ok(buffer)
-                }));
-            }
+            let mut buffer = buffers.pop().unwrap();
+            let crucible = crucible.clone();
+            futures.push_back(tokio::spawn(async move {
+                crucible.read(offset, &mut buffer).await?;
+                Ok(buffer)
+            }));
 
             total_bytes_read +=
                 (opt.iocmd_block_count * native_block_size) as usize;
@@ -246,6 +245,8 @@ async fn cmd_read<T: BlockIO + std::marker::Send + 'static>(
             return Ok(total_bytes_read);
         }
     }
+
+    assert!(futures.is_empty());
 
     // Issue our final read command, if any. This could be interleaved with
     // draining the outstanding commands but it's more complicated and
--- a/crudd/src/main.rs
+++ b/crudd/src/main.rs
@@ -154,7 +154,6 @@ async fn cmd_read<T: BlockIO>(
     //
     // TODO this is no longer true after the Upstairs is async, multiple
     // requests can be submitted and await'ed on at the same time.
-    let mut buffers = VecDeque::with_capacity(opt.pipeline_length);
     let mut futures = VecDeque::with_capacity(opt.pipeline_length);
 
     // First, align our offset to the underlying blocks with an initial read
@@ -174,13 +173,13 @@ async fn cmd_read<T: BlockIO>(
         let buffer = Buffer::new(native_block_size as usize);
         let block_idx = opt.byte_offset / native_block_size;
         let offset = Block::new(block_idx, native_block_size.trailing_zeros());
-        crucible.read(offset, buffer.clone()).await?;
+        let buffer = crucible.read(offset, buffer).await?;
 
         // write only (block size - misalignment) bytes
         // So say we have an offset of 5. we're misaligned by 5 bytes, so we
         // read 5 bytes we don't need. we skip those 5 bytes then write
         // the rest to the output
-        let bytes = buffer.into_vec().unwrap();
+        let bytes = buffer.into_vec();
         output.write_all(
             &bytes[offset_misalignment as usize
                 ..(offset_misalignment + alignment_bytes) as usize],
@@ -215,37 +214,37 @@ async fn cmd_read<T: BlockIO>(
 
         // Send the read command with whichever buffer is at the back of the
         // queue. We re-use the buffers to avoid lots of allocations
-        let w_buf =
+        let buf =
             Buffer::new((opt.iocmd_block_count * native_block_size) as usize);
-        let w_future = crucible.read(offset, w_buf.clone());
-        buffers.push_back(w_buf);
-        futures.push_back(w_future);
+        let r_future = crucible.read(offset, buf);
+        futures.push_back(r_future);
         total_bytes_read +=
             (opt.iocmd_block_count * native_block_size) as usize;
 
         // If we have a queue of futures, drain the oldest one to output.
         if futures.len() == opt.pipeline_length {
-            futures.pop_front().unwrap().await?;
-            let r_buf = buffers.pop_front().unwrap();
-            output.write_all(&r_buf.as_vec().await)?;
+            let r_buf = futures.pop_front().unwrap().await?;
+            output.write_all(&r_buf)?;
         }
 
         if early_shutdown.try_recv().is_ok() {
             eprintln!("shutting down early in response to SIGUSR1");
-            join_all(futures).await?;
+
+            // Drain all read buffers
+            for future in futures {
+                let _r_buf = future.await?;
+            }
+
             return Ok(total_bytes_read);
         }
     }
 
     // Drain the outstanding commands
     if !futures.is_empty() {
-        crucible::join_all(futures).await?;
-
         // drain the buffer to the output file
-        while !buffers.is_empty() {
-            // unwrapping is safe because of the length check
-            let r_buf = buffers.pop_front().unwrap();
-            output.write_all(&r_buf.as_vec().await)?;
+        while !futures.is_empty() {
+            let r_buf = futures.pop_front().unwrap().await?;
+            output.write_all(&r_buf)?;
         }
     }
 
@@ -259,9 +258,9 @@ async fn cmd_read<T: BlockIO>(
         let buffer = Buffer::new((blocks * native_block_size) as usize);
         let block_idx = (cmd_count * opt.iocmd_block_count) + block_offset;
         let offset = Block::new(block_idx, native_block_size.trailing_zeros());
-        crucible.read(offset, buffer.clone()).await?;
+        let buffer = crucible.read(offset, buffer).await?;
         total_bytes_read += remainder as usize;
-        output.write_all(&buffer.as_vec().await[0..remainder as usize])?;
+        output.write_all(&buffer[0..remainder as usize])?;
     }
 
     Ok(total_bytes_read)
@@ -280,7 +279,7 @@ async fn write_remainder_and_finalize<'a, T: BlockIO>(
     offset: Block,
     n_read: usize,
     native_block_size: u64,
-    mut futures: VecDeque<crucible::CrucibleBlockIOFuture<'a>>,
+    mut futures: VecDeque<CrucibleBlockIOFuture<'a>>,
 ) -> Result<()> {
     // the input stream ended,
     // - read/mod/write for alignment
@@ -311,10 +310,10 @@ async fn write_remainder_and_finalize<'a, T: BlockIO>(
             native_block_size.trailing_zeros(),
         );
         let uflow_r_buf = Buffer::new(native_block_size as usize);
-        crucible.read(uflow_offset, uflow_r_buf.clone()).await?;
+        let uflow_r_buf = crucible.read(uflow_offset, uflow_r_buf).await?;
 
         // Copy it into w_buf
-        let r_bytes = uflow_r_buf.into_vec().unwrap();
+        let r_bytes = uflow_r_buf.into_vec();
         w_buf[n_read..n_read + uflow_backfill]
             .copy_from_slice(&r_bytes[uflow_remainder as usize..]);
 
@@ -323,12 +322,12 @@ async fn write_remainder_and_finalize<'a, T: BlockIO>(
         futures.push_back(w_future);
     }
 
-    // Flush
-    let flush_future = crucible.flush(None);
-    futures.push_back(flush_future);
+    // Wait for all the writes first, then flush
+    for future in futures {
+        future.await?;
+    }
 
-    // Wait for all the writes
-    join_all(futures).await?;
+    crucible.flush(None).await?;
 
     Ok(())
 }
@@ -398,9 +397,9 @@ async fn cmd_write<T: BlockIO>(
         let buffer = Buffer::new(native_block_size as usize);
         let block_idx = opt.byte_offset / native_block_size;
         let offset = Block::new(block_idx, native_block_size.trailing_zeros());
-        crucible.read(offset, buffer.clone()).await?;
+        let buffer = crucible.read(offset, buffer).await?;
 
-        let mut w_vec = buffer.into_vec().unwrap();
+        let mut w_vec = buffer.into_vec();
         // Write our data into the buffer
         let bytes_read = input.read(
             &mut w_vec[offset_misalignment as usize
@@ -485,7 +484,9 @@ async fn cmd_write<T: BlockIO>(
 
         if early_shutdown.try_recv().is_ok() {
             eprintln!("shutting down early in response to SIGUSR1");
-            join_all(futures).await?;
+            for future in futures {
+                future.await?;
+            }
             crucible.flush(None).await?;
             return Ok(total_bytes_written);
         }
@@ -523,7 +524,9 @@ async fn cmd_write<T: BlockIO>(
         )
         .await?;
     } else {
-        join_all(futures).await?;
+        for future in futures {
+            future.await?;
+        }
         crucible.flush(None).await?;
     }
 
--- a/crudd/src/main.rs
+++ b/crudd/src/main.rs
@@ -11,6 +11,7 @@ use std::{cmp, io};
 use anyhow::{bail, Result};
 use clap::Parser;
 use futures::stream::StreamExt;
+use itertools::Itertools;
 use signal_hook::consts::signal::*;
 use signal_hook_tokio::Signals;
 use tokio::sync::oneshot;
@@ -206,58 +207,43 @@ async fn cmd_read<T: BlockIO + std::marker::Send + 'static>(
     let cmd_count = num_bytes / (opt.iocmd_block_count * native_block_size);
     let remainder = num_bytes % (opt.iocmd_block_count * native_block_size);
 
+    let buffer_size = (opt.iocmd_block_count * native_block_size) as usize;
+    let mut buffers: Vec<Buffer> = (0..opt.pipeline_length)
+        .map(|_| Buffer::new(buffer_size))
+        .collect();
+
     // Issue all of our read commands
-    for i in 0..cmd_count {
-        // XXX issue opt.pipeline_length tokio::spawns at a time, reuse the
-        // buffers here!
+    for cmd_range in &(0..cmd_count as usize).chunks(opt.pipeline_length) {
+        for i in cmd_range {
+            // which blocks in the underlying store are we accessing?
+            let block_idx = block_offset + (i as u64 * opt.iocmd_block_count);
+            let offset =
+                Block::new(block_idx, native_block_size.trailing_zeros());
 
-        // which blocks in the underlying store are we accessing?
-        let block_idx = block_offset + (i * opt.iocmd_block_count);
-        let offset = Block::new(block_idx, native_block_size.trailing_zeros());
+            // Send the read command with whichever buffer is at the back of the
+            // queue. We re-use the buffers to avoid lots of allocations
+            for mut buffer in buffers.drain(..) {
+                let crucible = crucible.clone();
+                futures.push_back(tokio::spawn(async move {
+                    crucible.read(offset, &mut buffer).await?;
+                    Ok(buffer)
+                }));
+            }
 
-        // Send the read command with whichever buffer is at the back of the
-        // queue. We re-use the buffers to avoid lots of allocations
-        {
-            let crucible = crucible.clone();
-            let iocmd_block_count = opt.iocmd_block_count;
-            futures.push_back(tokio::spawn(async move {
-                let mut buf = Buffer::new(
-                    (iocmd_block_count * native_block_size) as usize,
-                );
-                crucible.read(offset, &mut buf).await?;
-                Ok(buf)
-            }));
+            total_bytes_read +=
+                (opt.iocmd_block_count * native_block_size) as usize;
         }
 
-        total_bytes_read +=
-            (opt.iocmd_block_count * native_block_size) as usize;
-
-        // If we have a queue of futures, drain the oldest one to output.
-        if futures.len() == opt.pipeline_length {
-            let result: Result<Buffer, CrucibleError> =
-                futures.pop_front().unwrap().await?;
+        for future in futures.drain(..) {
+            let result: Result<Buffer, CrucibleError> = future.await?;
             let r_buf = result?;
             output.write_all(&r_buf)?;
+            buffers.push(r_buf);
         }
 
         if early_shutdown.try_recv().is_ok() {
             eprintln!("shutting down early in response to SIGUSR1");
-
-            // Drain all read buffers
-            for future in futures {
-                let _r_buf = future.await?;
-            }
-
             return Ok(total_bytes_read);
-        }
-    }
-
-    // Drain the outstanding commands
-    if !futures.is_empty() {
-        // drain the buffer to the output file
-        for f in futures.drain(..) {
-            let r_buf = f.await??;
-            output.write_all(&r_buf)?;
         }
     }
 
--- a/crudd/src/main.rs
+++ b/crudd/src/main.rs
@@ -102,7 +102,7 @@ pub fn opts() -> Result<Opt> {
     Ok(opt)
 }
 
-async fn cmd_read<T: BlockIO>(
+async fn cmd_read<T: BlockIO + std::marker::Send + 'static>(
     opt: &Opt,
     crucible: Arc<T>,
     mut early_shutdown: oneshot::Receiver<()>,
@@ -170,10 +170,10 @@ async fn cmd_read<T: BlockIO>(
         cmp::min(num_bytes, native_block_size - offset_misalignment);
     if offset_misalignment != 0 {
         // Read the full block
-        let buffer = Buffer::new(native_block_size as usize);
+        let mut buffer = Buffer::new(native_block_size as usize);
         let block_idx = opt.byte_offset / native_block_size;
         let offset = Block::new(block_idx, native_block_size.trailing_zeros());
-        let buffer = crucible.read(offset, buffer).await?;
+        crucible.read(offset, &mut buffer).await?;
 
         // write only (block size - misalignment) bytes
         // So say we have an offset of 5. we're misaligned by 5 bytes, so we
@@ -208,22 +208,35 @@ async fn cmd_read<T: BlockIO>(
 
     // Issue all of our read commands
     for i in 0..cmd_count {
+        // XXX issue opt.pipeline_length tokio::spawns at a time, reuse the
+        // buffers here!
+
         // which blocks in the underlying store are we accessing?
         let block_idx = block_offset + (i * opt.iocmd_block_count);
         let offset = Block::new(block_idx, native_block_size.trailing_zeros());
 
         // Send the read command with whichever buffer is at the back of the
         // queue. We re-use the buffers to avoid lots of allocations
-        let buf =
-            Buffer::new((opt.iocmd_block_count * native_block_size) as usize);
-        let r_future = crucible.read(offset, buf);
-        futures.push_back(r_future);
+        {
+            let crucible = crucible.clone();
+            let iocmd_block_count = opt.iocmd_block_count;
+            futures.push_back(tokio::spawn(async move {
+                let mut buf = Buffer::new(
+                    (iocmd_block_count * native_block_size) as usize,
+                );
+                crucible.read(offset, &mut buf).await?;
+                Ok(buf)
+            }));
+        }
+
         total_bytes_read +=
             (opt.iocmd_block_count * native_block_size) as usize;
 
         // If we have a queue of futures, drain the oldest one to output.
         if futures.len() == opt.pipeline_length {
-            let r_buf = futures.pop_front().unwrap().await?;
+            let result: Result<Buffer, CrucibleError> =
+                futures.pop_front().unwrap().await?;
+            let r_buf = result?;
             output.write_all(&r_buf)?;
         }
 
@@ -243,7 +256,7 @@ async fn cmd_read<T: BlockIO>(
     if !futures.is_empty() {
         // drain the buffer to the output file
         for f in futures.drain(..) {
-            let r_buf = f.await?;
+            let r_buf = f.await??;
             output.write_all(&r_buf)?;
         }
     }
@@ -255,10 +268,10 @@ async fn cmd_read<T: BlockIO>(
         // let block_remainder = remainder % native_block_size;
         // round up
         let blocks = (remainder + native_block_size - 1) / native_block_size;
-        let buffer = Buffer::new((blocks * native_block_size) as usize);
+        let mut buffer = Buffer::new((blocks * native_block_size) as usize);
         let block_idx = (cmd_count * opt.iocmd_block_count) + block_offset;
         let offset = Block::new(block_idx, native_block_size.trailing_zeros());
-        let buffer = crucible.read(offset, buffer).await?;
+        crucible.read(offset, &mut buffer).await?;
         total_bytes_read += remainder as usize;
         output.write_all(&buffer[0..remainder as usize])?;
     }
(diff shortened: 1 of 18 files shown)


crudd/src/main.rs
```diff
@@ -215,37 +214,37 @@ async fn cmd_read<T: BlockIO>(
...
     if !futures.is_empty() {
-        crucible::join_all(futures).await?;
-
         // drain the buffer to the output file
-        while !buffers.is_empty() {
-            // unwrapping is safe because of the length check
-            let r_buf = buffers.pop_front().unwrap();
-            output.write_all(&r_buf.as_vec().await)?;
+        while !futures.is_empty() {
+            let r_buf = futures.pop_front().unwrap().await?;
```
mkeeter, line 246: Nit:
```suggestion
        while let Some(f) = futures.pop_front() {
            let r_buf = f.await?;
```

jmpesp:
Here's the patch required in propolis to test this:

```patch
diff --git a/lib/propolis/src/block/crucible.rs b/lib/propolis/src/block/crucible.rs
index c3bad705..46539265 100644
--- a/lib/propolis/src/block/crucible.rs
+++ b/lib/propolis/src/block/crucible.rs
@@ -308,13 +308,12 @@ async fn process_request(
             // Perform one large read from crucible, and write from data into
             // mappings
             let data = Buffer::new(len);
-            let _ = block.read(offset, data.clone()).await?;
+            let data = block.read(offset, data).await?;
 
-            let source = data.as_vec().await;
             let mut nwritten = 0;
             for mapping in maps {
                 nwritten += mapping.write_bytes(
-                    &source[nwritten..(nwritten + mapping.len())],
+                    &data[nwritten..(nwritten + mapping.len())],
                 )?;
             }
 ```

File: crutest/src/main.rs [hunk 2]
```diff
@@ -2132,18 +2130,23 @@ async fn perf_workload(
     for _ in 0..read_loop {
         let big_start = Instant::now();
         for _ in 0..count {
+            let read_buffers: Vec<Buffer> =
```
mkeeter, line 2133: This will have a measurement impact, since we're now doing allocation in the inner loop.

Maybe allocate `read_buffers` once outside (like before), then re-collect them when joining futures?

e.g. around L2145

```rust
let mut done_buffers = Vec::with_capacity(io_depth);
for future in read_futures {
    done_buffers.push(future.await?);
}
read_buffers = done_buffers;
```

integration_tests/src/lib.rs
```diff
@@ -398,11 +398,11 @@ mod test {
...
         // Verify contents are zero on init
         let buffer = Buffer::new(BLOCK_SIZE * 10);
-        volume
-            .read(Block::new(0, BLOCK_SIZE.trailing_zeros()), buffer.clone())
+        let buffer = volume
+            .read(Block::new(0, BLOCK_SIZE.trailing_zeros()), buffer)
             .await?;
 
-        assert_eq!(vec![0x00_u8; BLOCK_SIZE * 10], *buffer.as_vec().await);
+        assert_eq!(vec![0x00_u8; BLOCK_SIZE * 10], buffer.into_vec());
```
mkeeter, line 405: Nit: use `&buffer[..]` here to match below?

upstairs/src/block_req.rs
```diff
@@ -65,7 +74,10 @@ impl BlockReqWaiter {
     /// Consume this BlockReqWaiter and wait on the message
     ///
     /// If the other side of the oneshot drops without a reply, log an error
-    pub async fn wait(self, log: &Logger) -> Result<(), CrucibleError> {
+    pub async fn wait(
```
mkeeter, line 77: Annotate with `#[must_use]`, to force people to think about the buffer (or assert that it is none)?

upstairs/src/in_memory.rs
```diff
@@ -57,28 +57,29 @@ impl BlockIO for InMemoryBlockIO {
     async fn read(
         &self,
         offset: Block,
-        data: Buffer,
-    ) -> Result<(), CrucibleError> {
+        mut data: Buffer,
+    ) -> Result<Buffer, CrucibleError> {
```
mkeeter, line 61: Annotate this with `#[must_use]` as well?  I'm not sure whether the compiler will warn you if you do `read(..).await?` without using the `Ok(..)` branch.

upstairs/src/in_memory.rs
```diff
@@ -57,28 +57,29 @@ impl BlockIO for InMemoryBlockIO {
...
-        Ok(())
+        Ok(data)
     }
 
     async fn write(
         &self,
         offset: Block,
         data: Bytes,
     ) -> Result<(), CrucibleError> {
+        // Write from Buffer into self
```
mkeeter, line 81: Want to elevate this to a docstring, e.g.
```
/// Write from `data` to `self`, setting all owned bits to `true`
```

upstairs/src/lib.rs
```diff
@@ -1486,16 +1494,9 @@ impl Buffer {
...
-    /// Will succeed if no other references (clones) to the Buffer exist,
-    /// otherwise will return the Buffer as it still exists.
-    pub fn into_vec(self) -> Result<Vec<u8>, Self> {
-        let Buffer { len, data, owned } = self;
-        match Arc::try_unwrap(data) {
-            Ok(buf) => Ok(buf.into_inner()),
-            Err(data) => Err(Buffer { len, data, owned }),
-        }
+    /// Extract the underlying `Vec<u8>` bearing buffered data.
+    pub fn into_vec(self) -> Vec<u8> {
```
mkeeter, line 1498: `#[must_use]`?

upstairs/src/lib.rs
```diff
@@ -1506,56 +1507,140 @@ impl Buffer {
...
     }
 
-    pub async fn as_vec(&self) -> MutexGuard<'_, Vec<u8>> {
-        self.data.lock().await
+    pub fn write(&mut self, offset: usize, data: &[u8]) {
+        assert!(offset + data.len() <= self.data.len());
+        for i in 0..data.len() {
+            self.data[offset + i] = data[i];
+            self.owned[offset + i] = true;
+        }
```
mkeeter, line 1515:
```suggestion
        self.data[offset + i][..data.len()].copy_from_slice(data);
        self.owned[offset + i][..data.len()].fill(true);
```

upstairs/src/lib.rs
```diff
@@ -1506,56 +1507,140 @@ impl Buffer {
...
+
+    pub fn write_read_response(
+        &mut self,
+        offset: usize,
+        response: &ReadResponse,
+    ) {
+        assert!(offset + response.data.len() <= self.data.len());
+        if !response.block_contexts.is_empty() {
+            for i in 0..response.data.len() {
+                self.data[offset + i] = response.data[i];
```
mkeeter, line 1541: Same here (using `copy_from_slice` / `fill` instead of explicit loop)

upstairs/src/lib.rs
```diff
@@ -2432,7 +2426,8 @@ impl BlockIO for Guest {
 
     /// Disable any more IO from this guest and deactivate the downstairs.
     async fn deactivate(&self) -> Result<(), CrucibleError> {
-        self.send_and_wait(BlockOp::Deactivate).await
+        let _ = self.send_and_wait(BlockOp::Deactivate).await?;
```
mkeeter, line 2429: Nit: should we assert that the result is `None` here?  It's paranoid, but could catch very weird mixups.

upstairs/src/pseudo_file.rs
```diff
@@ -71,51 +71,47 @@ impl IOSpan {
...
     }
 
     #[instrument]
     pub async fn read_from_blocks_into_buffer(&self, data: &mut [u8]) {
         assert_eq!(data.len(), self.sz as usize);
-
-        for (i, item) in data.iter_mut().enumerate() {
-            *item = self.buffer.as_vec().await[self.phase as usize + i];
-        }
+        self.buffer.read(self.phase as usize, data);
```
mkeeter, line 108: The `async` is no longer required here; if you run `cargo clippy`, it should warn you about it (and then the cascading fixes may let you make other functions non-async as well!)

upstairs/src/volume.rs
```diff
@@ -2694,15 +2615,17 @@ mod test {
 
         assert_eq!(volume.total_size().await?, block_size as u64 * 10);
 
-        // Verify parent contents in one read
-        let buffer = Buffer::new(block_size * 10);
-        volume
-            .read(Block::new(0, block_size.trailing_zeros()), buffer.clone())
+        // Verify volume contents in one read
+        let buffer = Buffer::new(block_size * 1);
```
mkeeter, line 2619: This seems like you're testing a different thing now?

upstairs/src/lib.rs
```diff
@@ -1506,56 +1507,140 @@ impl Buffer {
...
+    pub fn read(&self, offset: usize, data: &mut [u8]) {
+        assert!(offset + data.len() <= self.data.len());
+        for i in 0..data.len() {
+            if self.owned[offset + i] {
+                data[i] = self.data[offset + i];
+            }
+        }
+    }
+
+    pub fn as_bytes(&self) -> Bytes {
```
mkeeter, line 1556: This could be sneakily expensive, since it's doing a full copy.  It could be more efficient if it was consuming (`fn as_bytes(self) -> Bytes`, since `Bytes` implements `From<Vec<u8>>` and reuses the allocation (we'd probably rename as `into_bytes` in that case).

Looking at the code, `Buffer::as_bytes` is only used in two places:
- In `Volume::scrub`, no issues with it being consuming
- In `IOSpan::write_affected_blocks_to_volume`, it _looks_ bad; however, it turns out that we could **also** make `IOSpan::write_affected_blocks_to_volume` consuming.

upstairs/src/volume.rs
```diff
@@ -684,100 +678,35 @@ impl BlockIO for Volume {
...
+                    .read_only_parent
                     .as_ref()
                     .unwrap()
-                    .read(parent_offset, parent_buffer.clone())
+                    .read(parent_offset, Buffer::new(sub_sz))
                     .await?;
 
-                let mut data_vec = data.as_vec().await;
-                let mut owned_vec = data.owned_vec().await;
+                data.eat(data_index, parent_buffer);
```
mkeeter, line 688: Nit: `eat` could take the buffer by reference, and then we could keep the parent buffer around and try to reuse its allocations in this loop.  We'd need a `Buffer::reset` function or similar to resize + clear, which would reuse the allocations as long as it's not growing.

upstairs/src/volume.rs
```diff
@@ -684,100 +678,35 @@ impl BlockIO for Volume {
...
-
-                            data_vec[data_index + inside_block] =
-                                parent_data_vec[inside_block];
-
-                            // this layer of the volume doesn't own this
-                            // block, it came from the read only parent. Note
-                            // this doesn't depend if the block was owned by
-                            // the read only parent, just that the subvolume
-                            // doesn't own it.
-                            owned_vec[data_index + inside_block] = false;
```
mkeeter, line 738: I believe this behavior has changed: because we're using `Buffer::eat` to load the parent buffer (at L688), the new implementation will set `owned = true` for bytes coming from the parent.

mkeeter:
The API could be a bit less verbose if the `read` function had the signature
```rust
async fn read(&self, offset: Block, buf: &mut Buffer) -> Result<(), CrucibleError>
```

Under the hood, this could be done by deriving `Default` for `Buffer` then doing something like
```rust
async fn read(&self, offset: Block, buf: &mut Buffer) -> Result<(), CrucibleError> {
    let b = std::mem::take(buf);
    
    // previous code here, writing to `b` or sending it into the `Upstairs`
    // (since it's now owned)
    
    *buf = b;
    Ok(())
}
```

This wouldn't incur any new allocations, since constructing an empty `Buffer` is allocation-free.

upstairs/src/pseudo_file.rs
```diff
@@ -71,51 +71,47 @@ impl IOSpan {
...
             )
             .await
     }
 
     #[instrument]
     pub async fn read_from_blocks_into_buffer(&self, data: &mut [u8]) {
         assert_eq!(data.len(), self.sz as usize);
-
-        for (i, item) in data.iter_mut().enumerate() {
-            *item = self.buffer.as_vec().await[self.phase as usize + i];
```
mkeeter, line 108: TIL that we were acquiring an async lock for each individual byte 😬

crudd/src/main.rs
hawkw: could also be:
```suggestion
        for f in futures.drain(..) {
            let r_buf = f.await?;
```

upstairs/src/in_memory.rs
hawkw: @mkeeter `Result` is already `#[must_use]`, so adding `#[must_use]` to this function won't do anything (and, i think, will result in a compiler warning).

crudd/src/main.rs
jmpesp: I went with the drain in e6f3e38

integration_tests/src/lib.rs
jmpesp: done in 919841b

upstairs/src/block_req.rs
jmpesp: I like this suggestion, but this didn't seem to have any effect. However, adding the `#[must_use]` to `Buffer` in 115510f did :)

upstairs/src/in_memory.rs
jmpesp: covered by 115510f (I think)

upstairs/src/in_memory.rs
```diff
@@ -57,28 +57,29 @@ impl BlockIO for InMemoryBlockIO {
...
-
         let start = offset.value as usize * self.block_size as usize;
 
-        for i in 0..data_vec.len() {
-            data_vec[i] = inner.bytes[start + i];
-            owned_vec[i] = inner.owned[start + i];
-        }
+        data.write_with_ownership(
+            0,
+            &inner.bytes[start..(start + data.len())],
```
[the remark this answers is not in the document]
jmpesp: I like it, done in 4b89d18

upstairs/src/in_memory.rs
jmpesp: done in 4b89d18

upstairs/src/lib.rs
jmpesp: cool, done in cfeb274

upstairs/src/lib.rs
jmpesp: nice, done in 3822e92

upstairs/src/lib.rs
jmpesp: same in 9cd5f29

upstairs/src/pseudo_file.rs
jmpesp: it didn't seem to cascade, unfortunately: d197011

upstairs/src/volume.rs
jmpesp: excellent catch, reverted in 31048fa

upstairs/src/lib.rs
jmpesp: changed in a5ac359

pfmooney (commented):
Whoops, I meant to submit these comments yesterday

upstairs/src/volume.rs
jmpesp: You're not wrong, but I believe (after sleeping on it for a few nights) that the original behaviour might have been a bug! Ownership should come from the downstairs: has this block been written to or not? This referenced line *does not* take ownership information from `parent_buffer`, it overwrites it to false.

This test fails on main:

```patch
index e388767..bf21fac 100644
--- a/upstairs/src/volume.rs
+++ b/upstairs/src/volume.rs
@@ -2150,6 +2150,76 @@ mod test {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn is_it_a_bug_or_not() -> Result<()> {
+        const BLOCK_SIZE: u64 = 512;
+
+        let parent = Arc::new(InMemoryBlockIO::new(
+            Uuid::new_v4(),
+            BLOCK_SIZE,
+            2048,
+        ));
+
+        // Ownership starts off false
+
+        let buffer = Buffer::new(1024);
+        parent
+            .read(Block::new(0, BLOCK_SIZE.trailing_zeros()), buffer.clone())
+            .await?;
+
+        assert_eq!(&*buffer.owned_vec().await, &[false; 1024]);
+
+        // Ownership is set by writing to blocks
+
+        parent
+            .write(
+                Block::new(0, BLOCK_SIZE.trailing_zeros()),
+                Bytes::from(vec![9; 512]),
+            )
+            .await?;
+
+        // Ownership is returned by the downstairs
+
+        let buffer = Buffer::new(1024);
+        parent
+            .read(Block::new(0, BLOCK_SIZE.trailing_zeros()), buffer.clone())
+            .await?;
+
+        let mut expected = vec![9u8; 512];
+        expected.extend(vec![0u8; 512]);
+
+        assert_eq!(&*buffer.as_vec().await, &expected);
+
+        let mut expected_ownership = vec![true; 512];
+        expected_ownership.extend(vec![false; 512]);
+
+        assert_eq!(&*buffer.owned_vec().await, &expected_ownership);
+
+        // Ownership through a volume should be the same!!
+
+        let disk = Arc::new(InMemoryBlockIO::new(
+            Uuid::new_v4(),
+            BLOCK_SIZE,
+            4096,
+        ));
+
+        let mut volume = Volume::new(BLOCK_SIZE, csl());
+        volume.add_subvolume(disk).await?;
+        volume.add_read_only_parent(parent.clone()).await?;
+
+        // So is it?
+
+        let buffer = Buffer::new(1024);
+        volume
+            .read(Block::new(0, BLOCK_SIZE.trailing_zeros()), buffer.clone())
+            .await?;
+
+        assert_eq!(&*buffer.as_vec().await, &expected);
+        assert_eq!(&*buffer.owned_vec().await, &expected_ownership);
+
+        Ok(())
+    }
+
     #[tokio::test]
     async fn test_parent_uninitialized_read_only_region_one_subvolume(
     ) -> Result<()> {
```

This passes on this branch:

```patch
index d88b8ff..848cc24 100644
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -1568,6 +1568,10 @@ impl Buffer {
             }
         }
     }
+
+    pub fn owned_ref(&self) -> &[bool] {
+        &self.owned
+    }
 }
 
 impl std::ops::Deref for Buffer {
index 4051f40..4a60ca9 100644
--- a/upstairs/src/volume.rs
+++ b/upstairs/src/volume.rs
@@ -2077,6 +2077,76 @@ mod test {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn is_it_a_bug_or_not() -> Result<()> {
+        const BLOCK_SIZE: u64 = 512;
+
+        let parent = Arc::new(InMemoryBlockIO::new(
+            Uuid::new_v4(),
+            BLOCK_SIZE,
+            2048,
+        ));
+
+        // Ownership starts off false
+
+        let buffer = Buffer::new(1024);
+        let buffer = parent
+            .read(Block::new(0, BLOCK_SIZE.trailing_zeros()), buffer)
+            .await?;
+
+        assert_eq!(buffer.owned_ref(), &[false; 1024]);
+
+        // Ownership is set by writing to blocks
+
+        parent
+            .write(
+                Block::new(0, BLOCK_SIZE.trailing_zeros()),
+                Bytes::from(vec![9; 512]),
+            )
+            .await?;
+
+        // Ownership is returned by the downstairs
+
+        let buffer = Buffer::new(1024);
+        let buffer = parent
+            .read(Block::new(0, BLOCK_SIZE.trailing_zeros()), buffer)
+            .await?;
+
+        let mut expected = vec![9u8; 512];
+        expected.extend(vec![0u8; 512]);
+
+        assert_eq!(&*buffer, &expected);
+
+        let mut expected_ownership = vec![true; 512];
+        expected_ownership.extend(vec![false; 512]);
+
+        assert_eq!(buffer.owned_ref(), &expected_ownership);
+
+        // Ownership through a volume should be the same!!
+
+        let disk = Arc::new(InMemoryBlockIO::new(
+            Uuid::new_v4(),
+            BLOCK_SIZE,
+            4096,
+        ));
+
+        let mut volume = Volume::new(BLOCK_SIZE, csl());
+        volume.add_subvolume(disk).await?;
+        volume.add_read_only_parent(parent.clone()).await?;
+
+        // So is it?
+
+        let buffer = Buffer::new(1024);
+        let buffer = volume
+            .read(Block::new(0, BLOCK_SIZE.trailing_zeros()), buffer)
+            .await?;
+
+        assert_eq!(&*buffer, &expected);
+        assert_eq!(&*buffer.owned_ref(), &expected_ownership);
+
+        Ok(())
+    }
+
     #[tokio::test]
     async fn test_parent_uninitialized_read_only_region_one_subvolume(
     ) -> Result<()> {
```

upstairs/src/lib.rs
```diff
@@ -222,6 +222,14 @@ pub async fn join_all<'a>(
         .map(|_| ())
 }
 
+pub type CrucibleBlockIOReadFuture<'a> = Pin<
+    Box<
+        dyn futures::Future<Output = Result<Buffer, CrucibleError>>
+            + std::marker::Send
+            + 'a,
+    >,
+>;
```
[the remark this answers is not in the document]
jmpesp: The `Future` output of `CrucibleBlockIOFuture` is `Result<(), CrucibleError>`, which isn't the same as what `read` returns now. All the other stuff is to match the output of what the `async_trait` macro does to the `BlockIO` trait.

upstairs/src/block_io.rs
```diff
@@ -246,20 +235,17 @@ impl BlockIO for ReqwestBlockIO {
...
         let bytes = response
             .bytes()
             .await
             .map_err(|e| CrucibleError::GenericError(e.to_string()))?;
 
-        for i in 0..data_vec.len() {
-            data_vec[i] = bytes[i];
-            owned_vec[i] = true;
-        }
+        data.write(0, &bytes);
```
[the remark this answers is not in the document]
jmpesp: You're not wrong, but in the case of `ReqwestBlockIO` I think it makes sense to always return owned bytes, as this impl of BlockIO does not support writes.

upstairs/src/block_io.rs
```diff
@@ -63,22 +63,18 @@ impl BlockIO for FileBlockIO {
...
-        file.seek(SeekFrom::Start(start))?;
-        file.read_exact(&mut data_vec[..])?;
+        file.seek(SeekFrom::Start(start as u64))?;
 
-        for i in 0..data_vec.len() {
-            owned_vec[i] = true;
-        }
+        let mut buf = vec![0u8; data.len()];
+        file.read_exact(&mut buf)?;
+        data.write(0, &buf);
```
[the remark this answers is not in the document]
jmpesp: I added ownership to `FileBlockIO` in 6a0983ad0409700df67cf072b2b2fa672e9a10c0.

For the allocation question: I don't see a good way for us to do something like

```rust
file.read_exact(&mut *data)?;
```

It would make sense to have an `impl AsMut<u8>` for Buffer _except_ that ownership information would not appropriately be filled in.

upstairs/src/block_io.rs
```diff
@@ -246,20 +235,17 @@ impl BlockIO for ReqwestBlockIO {
         )
         .map_err(|e| CrucibleError::GenericError(e.to_string()))?;
 
-        assert_eq!(total_size, data_vec.len() as u64);
+        assert_eq!(total_size, data.len() as u64);
```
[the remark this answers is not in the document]
jmpesp: Done! c2ed80b

upstairs/src/pseudo_file.rs
jmpesp: :car:

upstairs/src/volume.rs
mkeeter: Okay, I agree that the new behavior seems correct!

jmpesp: I've added this as a test in 9bd6061

upstairs/src/block_io.rs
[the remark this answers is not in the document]
jmpesp: I reverted the ownership addition - it didn't make sense because it wasn't stored in the backing `File`, and if the `FileBlockIO` was dropped and recreated the ownership would be lost. Without any sort of persistent ownership storage, it makes sense to me for it to always return `true`.

upstairs/src/lib.rs
```diff
@@ -1506,56 +1509,152 @@ impl Buffer {
...
+        data: &[u8],
+        owned: &[bool],
+    ) {
+        assert!(offset + data.len() <= self.data.len());
+        for i in 0..data.len() {
+            if owned[i] {
+                self.data[offset + i] = data[i];
+                self.owned[offset + i] = true;
+            }
+        }
```
pfmooney, line 1547: Use copy_from_slice and fill?

upstairs/src/block_io.rs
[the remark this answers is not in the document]
jmpesp: I added a block comment about ownership in fd61bb6

upstairs/src/lib.rs
```diff
@@ -1506,56 +1509,152 @@ impl Buffer {
...
+            self.owned[offset..][..response.data.len()].fill(true);
+        }
+    }
+
+    pub fn read(&self, offset: usize, data: &mut [u8]) {
+        assert!(offset + data.len() <= self.data.len());
+        for i in 0..data.len() {
+            if self.owned[offset + i] {
+                data[i] = self.data[offset + i];
+            }
```
pfmooney, line 1568: This conditional read seems extremely fraught.  (I do a read, but no data is actually copied, but I don't actually know that)  Are we required to keep this around?  It appears to be only used in a test and some test-adjacent (nbd) stuff.

upstairs/src/lib.rs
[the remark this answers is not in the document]
mkeeter: Would [switching the external API to taking the buffer as a `&mut`](https://github.com/oxidecomputer/crucible/pull/1094#issuecomment-1885285513) make this separate future type unnecessary?

upstairs/src/volume.rs
jmpesp: I think 22b8efa covers this, let me know!

upstairs/src/lib.rs
jmpesp: I'm not sure I can with the conditional there, though https://github.com/oxidecomputer/crucible/pull/1091 will be able to per-block.

upstairs/src/lib.rs
jmpesp: There's two reasons we need to keep conditional reads / ownership around: one is that it's essential for the Volume hierarchy layering to work, and two to support `write_unwritten`, which `InMemoryBlockIO` supports for tests (as you noted :)).

In order for Volume hierarchy layering to work, we have to be able to distinguish when the subvolume has been written to in order to know when to layer the subvolume read results on top of what came from the read-only parent. The read-only parent will almost always start out with owned blocks, as the result of a bulk import or snapshot or otherwise, and the subvolumes will start out blank. Ownership is our way of knowing that the zeroes coming from blocks in a subvolume that have never been written to should not override the owned blocks coming from the read-only parent. If there was a write of zeroes to the subvolume, then those _should_ overwrite the blocks from the read-only parent.

It's technically possible to have a read-only parent that was never written to: in this case, `Volume::read` would read no owned blocks from both sides, and the conditional linked to in this comment would never be true.

The second reason is to support `write_unwritten`, but that's not something that occurs in the Upstairs. Receivers of the `write_unwritten` command (i.e. a Downstairs, or `InMemoryBlockIO`) have to determine based on their particular ownership persistence whether or not to write the data: if the block was never written to, then the write goes through. If it was, then the write does not. In the `Downstairs`' case, this is done by querying for existing block context. In `InMemoryBlockIO`'s case, this is done with a `Vec<bool>`.

With authenticated encryption in the mix, we also have the added benefit of only ever returning either all zeroes with no associated authentication, or an authenticated block of data. `owned` in that case means that the read was validated. See `validate_encrypted_read_response` for more on this.

jmpesp:
> The API could be a bit less verbose if the `read` function had the signature
> 
> ```rust
> async fn read(&self, offset: Block, buf: &mut Buffer) -> Result<(), CrucibleError>
> ```
> 
> Under the hood, this could be done by deriving `Default` for `Buffer` then doing something like
> 
> ```rust
> async fn read(&self, offset: Block, buf: &mut Buffer) -> Result<(), CrucibleError> {
>     let b = std::mem::take(buf);
>     
>     // previous code here, writing to `b` or sending it into the `Upstairs`
>     // (since it's now owned)
>     
>     *buf = b;
>     Ok(())
> }
> ```
> 
> This wouldn't incur any new allocations, since constructing an empty `Buffer` is allocation-free.

I'm not sure how to make this work - if `read` took a `&mut Buffer`:

```rust
let mut futures = Vec::with_capacity(8);

// Submit 8 reads
for i in 0..8 {
  let mut buffer = Buffer::new(512);
  futures.push(block_io.read(Block::new_512(i), &mut buffer));
}

// Wait for all the futures
crucible::join_all(futures)?;
```

I tried it and saw messages similar to:

```
error[E0597]: `w_buf` does not live long enough
   --> crudd/src/main.rs:220:46
    |
218 |         let mut w_buf =
    |             --------- binding `w_buf` declared here
219 |             Buffer::new((opt.iocmd_block_count * native_block_size) as usize);
220 |         let r_future = crucible.read(offset, &mut w_buf);
    |                                              ^^^^^^^^^^ borrowed value does not live long enough
...
238 |     }
    |     - `w_buf` dropped here while still borrowed
...
241 |     if !futures.is_empty() {
    |         ------- borrow later used here


```

We can't store the Buffer objects either, the future has a `&mut`:

```rust
let mut futures = Vec::with_capacity(8);
let mut buffers = Vec::with_capacity(8);

// Submit 8 reads
for i in 0..8 {
  let mut buffer = Buffer::new(512);
  futures.push(block_io.read(Block::new_512(i), &mut buffer));
  buffers.push(buffer);
}

// Wait for all the futures
crucible::join_all(futures)?;

// Read the results
for i in 0..8 {
  // access buffers[i]
}
```

Something like that gives me

```
error[E0505]: cannot move out of `w_buf` because it is borrowed
   --> crudd/src/main.rs:221:27
    |
218 |         let mut w_buf =
    |             --------- binding `w_buf` declared here
219 |             Buffer::new((opt.iocmd_block_count * native_block_size) as usize);
220 |         let r_future = crucible.read(offset, &mut w_buf);
    |                                              ---------- borrow of `w_buf` occurs here
221 |         buffers.push_back(w_buf);
    |                           ^^^^^ move out of `w_buf` occurs here
222 |         futures.push_back(r_future);
    |                           -------- borrow later used here


```

crutest/src/main.rs
jmpesp: Done in 7819007, nice catch :)

upstairs/src/volume.rs
```diff
@@ -372,28 +372,32 @@ impl Volume {
...
                 // on. If that initial connection fails, we can retry.
-                let mut retry_needed = true;
                 let mut retry_count = 0;
-                while retry_needed {
-                    match read_only_parent.read(block, buffer.clone()).await {
-                        Ok(_) => {
-                            retry_needed = false;
+
+                let buffer = loop {
+                    let buffer = Buffer::new(block_count * bs);
```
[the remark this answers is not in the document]
jmpesp: I don't see how to make this work, because `Buffer` isn't returned for the `Err` case. The only way I can see this working is if `read` had the signature:

```rust
    async fn read(
        &self,
        offset: Block,
        mut data: Buffer,
    ) -> Result<Buffer, (Buffer, CrucibleError)>;
```

which maybe it should? Just because `read` returned `Err` doesn't mean it should eat the Buffer...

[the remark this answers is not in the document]
jmpesp: Alternatively,

```rust
    async fn read(
        &self,
        offset: Block,
        mut data: Buffer,
    ) -> (Buffer, Result<(), CrucibleError>);
```

mkeeter:
Would something like this work?
```
let mut futures = Vec::with_capacity(8);
let mut buffers = Vec::with_capacity(8);

for i in 0..8 {
  let mut buffer = Buffer::new(512);
  buffers.push(buffer);
}

// Submit 8 reads
for (i, b) in buffers.iter_mut().enumerate() {
  futures.push(block_io.read(Block::new_512(i), b));
}

// Wait for all the futures
crucible::join_all(futures)?;

// Read the results
for i in 0..8 {
  // access buffers[i]
}
```

jmpesp:
I think so - 5b060cf changes `read` to accept `&mut Buffer`, and 4153795 makes sure that it is always returned so the caller does not have to reallocate.

upstairs/src/lib.rs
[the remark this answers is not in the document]
jmpesp: Yeah, it was removed in 5b060cf

upstairs/src/volume.rs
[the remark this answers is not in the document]
jmpesp: It moved out in 5b060cf

mkeeter (approved):

jmpesp:
@mkeeter 7d05554 has the change we were talking about in DMs

mkeeter:
This looks like a significant improvement on large random reads!

Testing on the Madrid cluster with a 128 GiB disk, 4K blocks, 16K blocks/extent, encrypted:

Before:
```
4K READ: bw=44.6MiB/s (46.7MB/s), 44.6MiB/s-44.6MiB/s (46.7MB/s-46.7MB/s), io=2674MiB (2804MB), run=60002-60002msec
1M READ: bw=158MiB/s (166MB/s), 158MiB/s-158MiB/s (166MB/s-166MB/s), io=9513MiB (9975MB), run=60152-60152msec
4M READ: bw=159MiB/s (167MB/s), 159MiB/s-159MiB/s (167MB/s-167MB/s), io=9664MiB (10.1GB), run=60631-60631msec
```

After:
```
4K READ: bw=52.1MiB/s (54.6MB/s), 52.1MiB/s-52.1MiB/s (54.6MB/s-54.6MB/s), io=3126MiB (3278MB), run=60002-60002msec
1M READ: bw=355MiB/s (372MB/s), 355MiB/s-355MiB/s (372MB/s-372MB/s), io=20.8GiB (22.4GB), run=60080-60080msec
4M READ: bw=352MiB/s (369MB/s), 352MiB/s-352MiB/s (369MB/s-369MB/s), io=20.7GiB (22.2GB), run=60273-60273msec
```

```toml
[global]
filename=/dev/nvme0n1
iodepth=25
ioengine=aio
time_based
runtime=60
numjobs=1
direct=1
stonewall=1

[randwrite-4M]
bs=4M
rw=randwrite

[randread-4K]
bs=4K
rw=randread

[randread-1M]
bs=1M
rw=randread

[randread-4M]
bs=4M
rw=randread
```

PR #1096, opened 2024-01-11
[smf] Use new zone network config service

karencfv:
https://github.com/oxidecomputer/omicron/pull/4677 Will implement a new zone network configuration setup service so control plane services don't have to set this up themselves. 

Once that PR is merged, I'll open up a separate one in the omicron repo that will include [these](https://github.com/oxidecomputer/omicron/pull/4677/files#diff-b8a6f13742cae29f44d095f6b9e8c2febc712e0ff86f01f3c8ec9d4e5d2db396R1516-R1523) changes in the `sled-agent/src/services.rs` file for crucible and crucible pantry which I'll have to somehow coordinate with this PR.

PR #1097, opened 2024-01-12
Order packages topologically

smklein:
This doesn't make much of a difference for `main`, but it matters for an upcoming PR which adds more complexity to the Crucible build process: https://github.com/oxidecomputer/crucible/pull/1096 (reasonably so! the network service will be great to have)

Previously, Crucible packages were ordered according to the parsed order of the TOML file (which appears to have been alphabetical). With this PR, packages are now sorted topologically, so that "dependencies are built first".

Also, improved logging for error cases where dependencies cannot be found.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1035,6 +1035,7 @@ dependencies = [
  "crucible-workspace-hack",
  "omicron-zone-package",
  "tokio",
+ "topological-sort",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


smklein:
@karencfv : This should hopefully unblock you, I filed https://github.com/oxidecomputer/omicron-package/issues/56 to pull this out as a more general-purpose utility in the `omicron-package` crate, so other repos don't get hit here too

karencfv (approved):
Ha! That's what was missing, mystery solved. Thanks a million for looking into this :bowing_woman: !

leftwo (approved):

PR #1098, opened 2024-01-12
Allow dsc to start the downstairs in read-only mode.

leftwo:
Add the ability for dsc to start the three downstairs in read only mode.


mkeeter (approved):
This seems fine for now, although `DscInfo::new` is approaching the point where we may want to switch to the [builder pattern](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html) (due to sheer number of unnamed arguments).

PR #1099, opened 2024-01-15
Use the omicron-zone-package methods for topo sorting

smklein:
I moved this logic (and tested it) in `omicron-zone-package`: https://github.com/oxidecomputer/omicron-package/pull/57

This PR uses that version, to simplify how much logic Crucible needs to know


karencfv (approved):
Looks good, thanks!

PR #1102, opened 2024-01-18
Comment cleanup and rename of DsState::Repair -> Reconcile

leftwo:
Some misc comment cleanup and a rename of the Downstairs state Repair to Reconcile.  The term "repair" was often used in both the reconcile case (what happens at start up) and during live repair (what happens when the upstairs is active).  This cleans up the comments and renames the original DsState::Repair to be called DsState::Reconcile.

There should be no functional change from this.


openapi/crucible-control.json
```diff
@@ -96,8 +96,8 @@
           "wait_quorum",
           "bad_region",
           "disconnected",
-          "repair",
-          "failed_repair",
+          "reconcile",
+          "failed_reconcile",
```
mkeeter, line 100: Is anyone using these APIs that would need to change?

upstairs/src/downstairs.rs
```diff
@@ -3548,7 +3548,8 @@ impl Downstairs {
     }
 
     #[cfg(test)]
-    /// Create a test downstiars which has one downstairs client transitioned to LiveRepair
+    /// Create a test downstiars which has one downstairs client transitioned to
```
mkeeter, line 3551:
```suggestion
    /// Create a test downstairs which has one downstairs client transitioned to
```

mkeeter: (not a new typo, but I saw it while reviewing)

mkeeter (approved):
This is a much-desired cleanup!

jmpesp (approved):

upstairs/src/downstairs.rs
leftwo: I went and found and replaced all the downstiars cases.

PR #1105, opened 2024-01-23
Block size buffers

mkeeter:
This PR changes the `struct Buffer` to enforce that it is always an exact multiple of blocks.

The property is enforced at construction: the only way to build or resize a `Buffer` is to provide a tuple of `(block_count, block_size)`.  Various implementors of `BlockIO` also now check that their reads and writes are block-aligned, returning the existing `CrucibleError::DataLenUnaligned` if that's not the case.

These changes are a step towards per-block ownership, but make sense as a standalone PR:

- This PR touches a bunch of lines, but is mostly mechanical changes
- The upcoming per-block ownership PR will touch fewer lines, but will be making trickier logical changes

### DIFF
--- a/pantry/src/pantry.rs
+++ b/pantry/src/pantry.rs
@@ -87,6 +87,18 @@ where
     }
 }
 
+// Static assertions to ensure that MAX_CHUNK_SIZE is divisible into blocks.
+//
+// Block size is always a power of two, so if we're divisible by the largest
+// possible block, then we're also divisible by all others.
+static_assertions::const_assert_eq!(
+    PantryEntry::MAX_CHUNK_SIZE % crucible::MAX_BLOCK_SIZE,
+    0
+);
+static_assertions::const_assert!(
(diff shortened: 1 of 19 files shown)


pantry/src/pantry.rs
```diff
@@ -326,17 +332,34 @@ impl PantryEntry {
                 block_size,
             );
         }
+        if Self::MAX_CHUNK_SIZE % block_size as usize != 0 {
+            crucible_bail!(
+                InvalidNumberOfBlocks,
```
jmpesp, line 337: Maybe should be

```rust
    #[error("Data length not block size multiple")]
    DataLenUnaligned,
```

?

jmpesp (approved):
:rocket:

pantry/src/pantry.rs
mkeeter: I added static assertions relating `MAX_CHUNK_SIZE` and `MAX_BLOCK_SIZE`, so this should always be true and I switched it to an assertion (4c2c6575e5528aa4974ebc4dd8d0237e357e9594)

PR #1106, opened 2024-01-23
Enforce buffer alignment

mkeeter:
This is a follow-up to #1105, so only look at the last commit (b78c98771f9a2d39b2c9058b57eece8461e99aba).  I'm marking it as a draft until #1105 is merged; then I'll rebase it onto `main`.

This PR makes `Buffer` operations stricter: reads and writes now panic if we try to access a `Buffer` without block alignment.  To make this less dangerous, the relevant functions are changed from `pub` to `pub(crate)`, so they can't be called from the public API.  Propolis only uses `Buffer::new` and `Buffer::as_vec`, so that shouldn't be an issue.

Within `crucible-upstairs`, I had to update the implementation of `IOSpan` to do read-modify-write operations; previously, three unit tests would panic under the strict regime.

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -1475,7 +1475,7 @@ impl fmt::Display for AckStatus {
 }
 
 /*
- * Provides a shared Buffer that Read operations will write into.
+ * Provides a strongly-owned Buffer that Read operations will write into.
  *
  * Originally BytesMut was used here, but it didn't guarantee that memory was
  * shared between cloned BytesMut objects. Additionally, we added the idea of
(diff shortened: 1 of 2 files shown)


mkeeter:
This is now rebased and ready for review!

upstairs/src/lib.rs
```diff
@@ -1551,20 +1556,42 @@ impl Buffer {
         self.data.is_empty()
     }
 
-    pub fn write(&mut self, offset: usize, data: &[u8]) {
+    /// Writes data to the buffer, setting `owned` to `true`
+    ///
+    /// # Panics
+    /// - The offset must be block-aligned
+    /// - The data length must be divisible by block size
+    /// - Data cannot exceed the buffer's lengt
```
jmpesp, line 1564: typo: lengt

leftwo, line 1564: I think my other comment (on the specific git rev before the rebase) made it out, but
I'm not sure.  Just in case, there is a typo here.

leftwo (approved):

jmpesp (approved):

Issue #1103, opened 2024-01-22
Crucible panics with `caller must change state from LiveRepair when requesting a stop` during stress testing.


leftwo:
@faithanalog reported the following: 
i ran 100 instances at the same time (in parallel) on madrid. each 1cpu, 1gig of ram, with my standard set of fio jobs. 93 of them had nvme errors (of varying description). 5 of them panicked (we are getting logs and core dumps for investigation of this). 2 of them succeeded running all the tests

propolis logs/core dumps from the 5 panicks are in /home/iliana/pylons-2024-01-21-001 on jeeves (which you should be able to access, xie set it 755 and i can read it)

The five instances that reported a panic were:
```
3153e243-e706-42ef-82f0-f96dbdd42577 - worker-19
9e636ea6-4285-41f5-8f1c-55202a9993b4 - worker-34
b36cad58-7662-4130-ad6a-fc49239c2e63 - worker-43
c3859f3e-1b21-406b-af21-b821f96c2534 - worker-50
7314b0a3-0d39-4480-9068-067ba494683a - worker-80
```

The panic messages looked like this:
```
thread 'tokio-runtime-worker' panicked at /home/build/.cargo/git/checkouts/crucible-f3b5bdecdc6486d6/e71b10d/upstairs/src/downstairs.rs:689:17:
caller must change state from LiveRepair when requesting a stop
```

leftwo:
Looking at: `c3859f3e-1b21-406b-af21-b821f96c2534 - worker-50`

`omdb db instances` gives us:

```
ID                                   NAME      STATE   PROPOLIS_ID                          SLED_ID                              HOST_SERIAL 
c3859f3e-1b21-406b-af21-b821f96c2534 worker-50 running 8be04e29-7bc7-41f4-b4ac-f7a880bf642f 55157b83-4f9e-4e81-9171-6ad2786a88e4 BRM44220001 
```

This leads us to the log file:
`oxz_propolis-server_8be04e29-7bc7-41f4-b4ac-f7a880bf642f.system-illumos-propolis-server:default.log`

From that, we see we have kicked out a downstairs because too many jobs have built up:

```
04:31:01.353Z WARN propolis-server (crucible-36ab48d8-cefc-48b4-b0a1-f9fff604bfbd): downstairs failed, too many outstanding jobs 57001
[...]

PR #1109, opened 2024-01-24
Handle timeout in the client IO task

mkeeter:
This PR moves the timeout into the client IO task, for consistency with other forms of disconnection.

Previously, the main Upstairs event loop would notice the timeout, then tell the client task to stop; this was a roundabout way of doing things led to stuff like the panic seen in #1103.

For symmetry, we also move the ping task into the client IO task, since there's no reason to bother the main event loop about it: we always want to be pinging periodically, as long as the client IO task is running.

### DIFF
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -17,7 +17,7 @@ use slog::{debug, error, info, o, warn, Logger};
 use tokio::{
     net::{TcpSocket, TcpStream},
     sync::{mpsc, oneshot},
-    time::{sleep_until, Instant},
+    time::sleep_until,
 };
 use tokio_util::codec::{FramedRead, FramedWrite};
 use uuid::Uuid;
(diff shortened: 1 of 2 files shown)


upstairs/src/client.rs
```diff
@@ -2478,43 +2432,34 @@ async fn proc_stream(
...
-async fn cmd_loop<R, W>(
-    rx: &mut mpsc::Receiver<Message>,
-    stop: &mut oneshot::Receiver<ClientStopReason>,
-    tx: &mpsc::Sender<ClientResponse>,
+async fn rx_loop<R>(
+    tx: mpsc::Sender<ClientResponse>,
     mut fr: FramedRead<R, crucible_protocol::CrucibleDecoder>,
-    mut fw: FramedWrite<W, crucible_protocol::CrucibleEncoder>,
-    log: &Logger,
+    log: Logger,
```
mkeeter, line 2451: Github is butchering the diff here, sorry about that!

I moved the code from within the `tokio::spawn` into a separate function named `rx_loop`, which now includes the timeout as well as reading from the `FramedRead` (so it's got a `select!` in there)

leftwo: Ah yeah, thanks for the comment otherwise I would have puzzled about this for a while.

leftwo (approved):
Yeah, inside the client task is the right place for this stuff.

PR #1110, opened 2024-01-24
Bump nix to 0.27.1

mkeeter:
This replaces the Renovate PR (#914) for bumping `nix` to 0.27.1

It switches to using the new `AsFd` trait, instead of `AsRawFd`


leftwo (approved):
Did some reading about raw fds:

https://rust-lang.github.io/rfcs/3128-io-safety.html
And,
https://blog.sunfishcode.online/rustix-and-io-safety/

PR #1114, opened 2024-01-24
Do not panic on reinitialize of a downstairs client.

leftwo:
Remove the panic during task reinitialize because of a timeout decided by the upstairs.  
We also don't need to send the reason to reinit any longer.

This along with the timeout fix will prevent https://github.com/oxidecomputer/crucible/issues/1103

### DIFF
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -258,7 +258,7 @@ impl DownstairsClient {
         .await;
     }
 
-    pub(crate) fn halt_io_task(&mut self, r: ClientStopReason) {
+    fn halt_io_task(&mut self, r: ClientStopReason) {
         if let Some(t) = self.client_task.client_stop_tx.take() {
             if let Err(_e) = t.send(r) {
                 warn!(self.log, "failed to send stop request")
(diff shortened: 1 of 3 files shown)


mkeeter:
I have subtle feelings about this.

The panic is a check that
1) If the Upstairs has asked the client task to stop, then
2) The Upstairs has also changed the client state out of `Active` / `LiveRepair`

#1103 was caused by violating this rule: we asked the client task to stop, but didn't change the client state.  I believe that #1109 **alone** will fix the panic, because the client task is deciding (on its own) to stop, so condition (1) is no longer met.

It's possible for future us to **make this mistake again** in the future, which would suggest leaving the panic in place.

However, we could make the mistake impossible by making `Client::halt_io_task` private, so the only way to halt the IO task is by going through a function that _also_ changes state (e.g. `Client::fault`).  There are no users`halt_io_task` outside of `Client` at this point; the last one was the incorrect handler for `ClientAction::Timeout`, now removed.

tl;dr: make `Client::halt_io_action` private and I'll approve the PR 😄

leftwo:
> tl;dr: make `Client::halt_io_action` private and I'll approve the PR 😄

Done in 89cbbda

mkeeter (approved):

PR #1116, opened 2024-01-25
Update usdt to v0.5.0

bnaecker:


leftwo (approved):

leftwo:
There is no _ vs. - issue with dtrace probe names?

bnaecker:
That's right. The old behavior was to rewrite double-underscores, `__`, into dashes, `-`, both in probe names and provider names. Version 0.5.0 updated the behavior to only do that rewrite for probe names, not providers. That was done to match DTrace's renaming for probes in traditional C code.

bnaecker:
Also, holy shit Crucible has a lot of probes!

PR #1126, opened 2024-01-30
Ignore client messages after stopping the IO task

mkeeter:
This protects us from a subtle race condition:

- The main task decides to stop the client task (e.g. because we've hit `IO_OUTSTANDING_MAX`)
  - It sends a stop request down the `tokio::sync::oneshot` channel
  - Then, it marks all jobs as skipped, which may result in discarding some of them
- The IO task receives a message from the downstairs
  - It passes this message to the main task
  - This message is a reply to one of the jobs that we skipped
  - The main task gets mad
  
I think this is may be the failure seen here:
```
{"msg":"cea05089-f330-4222-81d2-9c03095dfb2b WARNING finish job 125586 when downstairs state: Offline","v":0,"name":"propolis-server","level":40,"time":"2024-01-30T01:48:01.555693836Z","hostname":"oxz_propolis-server_edb3b1a5-7a60-441f-bfb0-a8cafae99226","pid":22206,"client":"2","":"downstairs","session_id":"9be1d7cc-2bf0-48cd-a496-d34f9c8245ec","component":"crucible-cea05089-f330-4222-81d2-9c03095dfb2b"}
thread 'tokio-runtime-worker' panicked at /home/build/.cargo/git/checkouts/crucible-f3b5bdecdc6486d6/2d4bc11/upstairs/src/client.rs:1237:13:
[2] Job 125586 completed while not InProgress: New
```

(from `/staff/core/crucible-unsorted/oxz_propolis-server_edb3b1a5.log`)

### DIFF
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -285,13 +285,21 @@ impl DownstairsClient {
     /// must the select expressions be cancel safe, but the **bodies** must also
     /// be cancel-safe.  This is why we simply return a single value in the body
     /// of each statement.
+    ///
+    /// This function will wait forever if we have asked for the client task to
+    /// stop, so it should only be called in a higher-level `select!`.
     pub(crate) async fn select(&mut self) -> ClientAction {
         tokio::select! {
-            d = self.client_task.client_response_rx.recv() => {
+            d = self.client_task.client_response_rx.recv(),
+                if self.client_task.client_stop_tx.is_some() =>
+            {
                 match d {
                     Some(c) => c.into(),
                     None => ClientAction::ChannelClosed,
                 }
+            }
+            _ = futures::future::pending() => {
+                unreachable!()
             }
         }
     }
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -289,17 +289,17 @@ impl DownstairsClient {
     /// This function will wait forever if we have asked for the client task to
     /// stop, so it should only be called in a higher-level `select!`.
     pub(crate) async fn select(&mut self) -> ClientAction {
-        tokio::select! {
-            d = self.client_task.client_response_rx.recv(),
-                if self.client_task.client_stop_tx.is_some() =>
+        loop {
+            let out = match self.client_task.client_response_rx.recv().await {
+                Some(c) => c.into(),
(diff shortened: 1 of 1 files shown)


leftwo (commented):
Could this still race with the downstairs where we have just taken a message
off the socket, but that message has not made it into the upstairs?

I ask because there are some panic's I'm updating to not panic if we receive
a message from a job we already finished and/or no longer exists.
(See `process_io_completion()`)

mkeeter:
> Could this still race with the downstairs where we have just taken a message off the socket, but that message has not made it into the upstairs?

I don't think so, but I'm also not sure I understand the question.

The IO task is responsible for taking messages off the socket and sending them via an `mpsc` queue to the main task.  The main task reads + processes those messages, and is _also_ responsible for stopping the IO task under certain circumstances.  When the IO task is requested to stop (by the main task), any pending jobs are handled, e.g. skipping them (also in the main task).

This PR means that once we've done those two things (stopping the IO task and marking pending jobs as skipped), we no longer see any messages from the IO task – even if they have arrived over the socket or were _just about_ to be pushed into the `mpsc` queue.

I think that if we do this correctly, we could keep those panics in place, because they're checking a real thing (and this PR fixes a real issue).

leftwo:
> > Could this still race with the downstairs where we have just taken a message off the socket, but that message has not made it into the upstairs?
> 
> This PR means that once we've done those two things (stopping the IO task and marking pending jobs as skipped), we no longer see any messages from the IO task – even if they have arrived over the socket or were _just about_ to be pushed into the `mpsc` queue. The IO task is dead to us; the only `ClientAction` that we care about is the actual notification that the task has stopped.
> 
> I think that if we do this correctly, we could (and **should!**) keep those panics in place, because they're checking a real thing (and this PR fixes a real issue).

Yes, perfect!
And, for LIveRepair operations, those also come throught the same client IO task, correct?

mkeeter:
The client IO task is the owner of the actual socket, so we shouldn't see any LiveRepair replies after stopping it.  This PR doesn't change _sending_ LiveRepair requests (which is managed by the main task).

leftwo (approved):
Thanks for answering my questions, let's get this in!

PR #1127, opened 2024-01-30
Minor Downstairs cleanup

mkeeter:
- `tokio::sync::mpsc::Sender` is `Clone`, so there's no need to wrap it in an `Arc` or `Mutex`
- We only ever send one message on the `terminate_sender` queue, so use a `tokio::sync::oneshot`
- Since we now mix `mpsc` and `oneshot`, use qualified names for `channel(..)`, `Sender`, and `Receiver`
- Small style tweaks in `downstairs/src/main.rs`


faithanalog (approved):
makes sense to me

leftwo (approved):
Yeah, looks good.

I was a little worried about the interactions between multiple upstairs connecting
at the same time, but the tests cover those cases so it should all be good there.

PR #1128, opened 2024-01-31
Update test_fail_live_repair to support pstop

leftwo:
Updated the tools/test_fail_live_repair.sh test script to support either stopping a 
downstairs process, or pausing it via pstop/prun

This test helped find a smattering of issues, so it's worth keeping around.

### DIFF
--- a/tools/test_fail_live_repair.sh
+++ b/tools/test_fail_live_repair.sh
@@ -3,10 +3,15 @@
 # Create and start up the downstairs.
 # In a loop:
 #  Send IO through crutest
-#  Fault a downstairs
+#  `pstop` or kill the downstairs process
+#  wait for missing downstairs to be faulted
+#  `pstart` or restart the downstairs
 #  Let the repair start.
-#  Fault the downstairs again.
+#  `pstop` or kill the same downstairs again.
+#  wait for missing downstairs to be faulted again
+#  `pstart` or restart the downstairs again
 #  Let the upstairs repair start over and finish.
+#  Stop crutest, then restart and verify the whole disk.
 
 err=0
 total=0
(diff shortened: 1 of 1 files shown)


tools/test_fail_live_repair.sh
```diff
@@ -120,67 +133,100 @@ count=1
...
     echo "" > "$test_log"
     echo "New loop starts now $(date) faulting: $choice" | tee -a "$test_log"
+    echo "Downstairs client $choice has pid: $ds_pid" | tee -a "$test_log"
+
     # Start sending IO.
     "$crucible_test" generic "${args[@]}" --continuous \
         -q -g "$gen" --verify-out "$verify_file" \
         --verify-in "$verify_file" \
-        --control 127.0.0.1:7777 \
+        --control 127.0.0.1:7890 \
```
mkeeter, line 147: Nit: why did the control port change?

tools/test_fail_live_repair.sh
```diff
@@ -120,67 +133,100 @@ count=1
...
     sleep $rand_sleep
     echo "After $rand_sleep seconds, Stop $choice again" | tee -a "$test_log"
-    ${dsc} cmd stop -c "$choice" >> "$dsc_test_log" 2>&1 &
+    if [[ $pstop -eq 0 ]]; then
+        ${dsc} cmd stop -c "$choice" >> "$dsc_test_log" 2>&1 &
+    else
+        pstop "$ds_pid"
+    fi
 
+    echo "Wait for downstairs $choice go go back to faulted"
```
mkeeter, line 203:
```suggestion
    echo "Wait for downstairs $choice to go back to faulted"
```

mkeeter (approved):
A few small comments, but seems good otherwise!

tools/test_fail_live_repair.sh
leftwo: I was getting port collisions with something else on atrium, so I just moved this to
pick a different port.

tools/test_fail_live_repair.sh
leftwo: Fixed in debcfb7

tools/test_fail_live_repair.sh
mkeeter: 👍🏻  ship it

Issue #1018, opened 2023-11-06
How to replace downstairs of a ReadOnly parent.


leftwo:
Tracking issue for:
We don't yet have a solution for the situation where there is a read only volume that we need to "repair" but in this case the repair is for downstairs replacement.

_Originally posted by @leftwo in https://github.com/oxidecomputer/crucible/pull/991#discussion_r1350763142_

leftwo:
I have an idea how this could work.
We have a new downstairs mode that will recover itself from a given repair address.

This recovery will connect to another downstairs that is read-only.
A few new endpoints to the current repair server will allow a connecting downstairs to verify
that the regions match, then the "new downstairs" will use the repair framework to copy the
existing downstairs extent and metadata files over.

Once completed, the downstairs repair mode will exit, and a new RO downstairs can be started using
the region built by the repair.

leftwo:
There is also work that will need to happen in Omicron to support this, but the crucible side will need
to come first.  The idea is that whatever the final work done to support regular RW region replacement
will be re-used here, but just done a little differently under the covers.

leftwo:
The now merged downstairs clone command is the path forward here.

PR #1129, opened 2024-02-01
    New downstairs clone subcommand.

leftwo:
Add support for a new downstairs clone subcommand.
    
This enables a downstairs to "clone" another downstairs using the
repair endpoint.  It requires the source downstairs to be read only
and will destroy whatever data exists on the destination downstairs.
    
Support for cloning SQLite backend downstairs is supported, and I had
to bring back some of the old repair code that supported the additional
files for that.  The resulting clone'd downstairs will still have a SQLite
backend.
    
Found an fixed a bug where the incorrect file types were returned
during a repair that contained SQLite files.

A bunch more tests were added to cover the clone process, and new
API endpoints were added to the downstairs repair server.

The is the base layer we need to support replacement of read only parents.

For that the workflow up in omicron will be to:
* Identify a good working downstairs we can clone from.
* Get the repair port for that good downstairs.
* Create a new downstairs region on a good disk with the same region size, but
don't run the downstairs once the region is created.
* Call the downstairs clone using the region created and the repair port from
the good downstairs.
* Wait for the clone to finish.
* Restart the new downstairs in RO mode.
* Replace the failed donwstairs with the cloned downstairs.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -159,8 +159,8 @@ pub enum CrucibleError {
     #[error("missing block context for non-empty block")]
     MissingBlockContext,
 
-    #[error("Incompatable RegionDefinition {0}")]
-    RegionIncompatable(String),
+    #[error("Incompatible RegionDefinition {0}")]
+    RegionIncompatible(String),
 }
 
 impl From<std::io::Error> for CrucibleError {
@@ -366,7 +366,7 @@ impl From<CrucibleError> for dropshot::HttpError {
             | CrucibleError::ModifyingReadOnlyRegion
             | CrucibleError::OffsetInvalid
             | CrucibleError::OffsetUnaligned
-            | CrucibleError::RegionIncompatable(_)
+            | CrucibleError::RegionIncompatible(_)
             | CrucibleError::ReplaceRequestInvalid(_)
             | CrucibleError::SnapshotExistsAlready(_)
             | CrucibleError::Unsupported(_) => {
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -158,6 +158,9 @@ pub enum CrucibleError {
 
     #[error("missing block context for non-empty block")]
     MissingBlockContext,
+
+    #[error("Incompatable RegionDefinition {0}")]
+    RegionIncompatable(String),
 }
 
 impl From<std::io::Error> for CrucibleError {
@@ -363,6 +366,7 @@ impl From<CrucibleError> for dropshot::HttpError {
             | CrucibleError::ModifyingReadOnlyRegion
             | CrucibleError::OffsetInvalid
             | CrucibleError::OffsetUnaligned
+            | CrucibleError::RegionIncompatable(_)
             | CrucibleError::ReplaceRequestInvalid(_)
             | CrucibleError::SnapshotExistsAlready(_)
             | CrucibleError::Unsupported(_) => {
--- a/common/src/region.rs
+++ b/common/src/region.rs
@@ -1,5 +1,6 @@
 // Copyright 2021 Oxide Computer Company
 use anyhow::{bail, Result};
+use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
@@ -17,7 +18,16 @@ use super::*;
  * downstairs expects Block { 2, 12 }.
  */
 #[derive(
-    Deserialize, Serialize, Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord,
+    Deserialize,
+    Serialize,
+    Copy,
+    Clone,
+    Debug,
+    PartialEq,
+    Eq,
+    JsonSchema,
+    PartialOrd,
+    Ord,
 )]
 pub struct Block {
     // Value could mean a size or offset
@@ -118,7 +128,7 @@ impl Block {
     }
 }
 
-#[derive(Deserialize, Serialize, Copy, Clone, Debug, PartialEq)]
+#[derive(Deserialize, Serialize, Copy, Clone, Debug, JsonSchema, PartialEq)]
 pub struct RegionDefinition {
     /**
      * The size of each block in bytes. Must be a power of 2, minimum 512.
@@ -168,6 +178,55 @@ impl RegionDefinition {
             database_read_version: DATABASE_READ_VERSION,
             database_write_version: DATABASE_WRITE_VERSION,
         })
+    }
+
+    // Compare two RegionDefinitions and verify they are compatable.
+    // Compatable is valid if all fields are the same, expect for the
+    // UUID. The UUID should be different.
+    pub fn compatable(
+        self,
+        other: RegionDefinition,
+    ) -> Result<(), CrucibleError> {
+        // These fields should be the same.
+        if self.block_size != other.block_size {
+            return Err(CrucibleError::RegionIncompatable(
+                "block_size".to_string(),
+            ));
+        }
+        if self.extent_size != other.extent_size {
+            return Err(CrucibleError::RegionIncompatable(
+                "extent_size".to_string(),
+            ));
+        }
+        if self.extent_count != other.extent_count {
+            return Err(CrucibleError::RegionIncompatable(
+                "extent_count".to_string(),
+            ));
+        }
+        if self.encrypted != other.encrypted {
+            return Err(CrucibleError::RegionIncompatable(
+                "encrypted".to_string(),
+            ));
+        }
+        if self.database_read_version != other.database_read_version {
+            return Err(CrucibleError::RegionIncompatable(
+                "database_read_version".to_string(),
+            ));
+        }
+        if self.database_write_version != other.database_write_version {
+            return Err(CrucibleError::RegionIncompatable(
+                "database_write_version".to_string(),
+            ));
+        }
+
+        // If the UUIDs are the same, this is invalid.
+        if self.uuid == other.uuid {
+            return Err(CrucibleError::RegionIncompatable(
+                "UUIDs are the same".to_string(),
+            ));
+        }
+
+        Ok(())
     }
 
     pub fn database_read_version(&self) -> usize {
@@ -488,5 +547,114 @@ mod test {
          *   IO         |---|---|---|---|
          */
         assert!(rd.validate_io(Block::new(1, 9), 2048).is_err());
+    }
+
+    fn test_rd() -> RegionDefinition {
+        RegionDefinition {
+            block_size: 512,
+            extent_size: Block::new(10, 9),
+            extent_count: 8,
+            uuid: Uuid::new_v4(),
+            encrypted: false,
+            database_read_version: DATABASE_READ_VERSION,
+            database_write_version: DATABASE_WRITE_VERSION,
+        }
+    }
+
+    #[test]
+    fn test_region_compare_block() {
+        let mut rd1 = test_rd();
+        let rd2 = test_rd();
+
+        // Basic positive test first.
+        assert_eq!(rd1.compatable(rd2), Ok(()));
+
+        rd1.block_size = 4096;
+        assert!(rd1.compatable(rd2).is_err());
+
+        let rd1 = test_rd();
+        let mut rd2 = test_rd();
+        rd2.block_size = 4096;
+        assert!(rd1.compatable(rd2).is_err());
+    }
+
+    #[test]
+    fn test_region_compare_extent_size() {
+        let mut rd1 = test_rd();
+        let rd2 = test_rd();
+
+        rd1.extent_size = Block::new(2, 9);
+        assert!(rd1.compatable(rd2).is_err());
+
+        let rd1 = test_rd();
+        let mut rd2 = test_rd();
+        rd2.extent_size = Block::new(2, 9);
+        assert!(rd1.compatable(rd2).is_err());
+    }
+
+    #[test]
+    fn test_region_compare_extent_count() {
+        let mut rd1 = test_rd();
+        let rd2 = test_rd();
+
+        rd1.extent_count = 9;
+        assert!(rd1.compatable(rd2).is_err());
+
+        let rd1 = test_rd();
+        let mut rd2 = test_rd();
+        rd2.extent_count = 9;
+        assert!(rd1.compatable(rd2).is_err());
+    }
+
+    #[test]
+    fn test_region_compare_uuid() {
+        // Verify region compare, UUIDs must be different
+        let mut rd1 = test_rd();
+        let rd2 = test_rd();
+
+        rd1.uuid = rd2.uuid;
+        assert!(rd1.compatable(rd2).is_err());
+    }
+
+    #[test]
+    fn test_region_compare_encrypted() {
+        let mut rd1 = test_rd();
+        let rd2 = test_rd();
+
+        rd1.encrypted = true;
+        assert!(rd1.compatable(rd2).is_err());
+
+        let rd1 = test_rd();
+        let mut rd2 = test_rd();
+        rd2.encrypted = true;
+        assert!(rd1.compatable(rd2).is_err());
+    }
+
+    #[test]
+    fn test_region_compare_db_read_version() {
+        let mut rd1 = test_rd();
+        let rd2 = test_rd();
+
+        rd1.database_read_version = DATABASE_READ_VERSION + 1;
+        assert!(rd1.compatable(rd2).is_err());
+
+        let rd1 = test_rd();
+        let mut rd2 = test_rd();
+        rd2.database_read_version = DATABASE_READ_VERSION + 1;
+        assert!(rd1.compatable(rd2).is_err());
+    }
+
+    #[test]
+    fn test_region_compare_db_write_version() {
+        let mut rd1 = test_rd();
+        let rd2 = test_rd();
+
+        rd1.database_write_version = DATABASE_WRITE_VERSION + 1;
+        assert!(rd1.compatable(rd2).is_err());
+
+        let rd1 = test_rd();
+        let mut rd2 = test_rd();
+        rd2.database_write_version = DATABASE_WRITE_VERSION + 1;
+        assert!(rd1.compatable(rd2).is_err());
     }
 }
--- a/common/src/region.rs
+++ b/common/src/region.rs
@@ -180,10 +180,10 @@ impl RegionDefinition {
         })
     }
 
-    // Compare two RegionDefinitions and verify they are compatable.
-    // Compatable is valid if all fields are the same, expect for the
+    // Compare two RegionDefinitions and verify they are compatible.
+    // compatible is valid if all fields are the same, expect for the
     // UUID. The UUID should be different.
-    pub fn compatable(
+    pub fn compatible(
         self,
         other: RegionDefinition,
     ) -> Result<(), CrucibleError> {
@@ -567,15 +567,15 @@ mod test {
         let rd2 = test_rd();
 
         // Basic positive test first.
-        assert_eq!(rd1.compatable(rd2), Ok(()));
+        assert_eq!(rd1.compatible(rd2), Ok(()));
 
         rd1.block_size = 4096;
-        assert!(rd1.compatable(rd2).is_err());
+        assert!(rd1.compatible(rd2).is_err());
 
         let rd1 = test_rd();
         let mut rd2 = test_rd();
         rd2.block_size = 4096;
-        assert!(rd1.compatable(rd2).is_err());
+        assert!(rd1.compatible(rd2).is_err());
     }
 
     #[test]
@@ -584,12 +584,12 @@ mod test {
         let rd2 = test_rd();
 
         rd1.extent_size = Block::new(2, 9);
-        assert!(rd1.compatable(rd2).is_err());
+        assert!(rd1.compatible(rd2).is_err());
 
         let rd1 = test_rd();
         let mut rd2 = test_rd();
         rd2.extent_size = Block::new(2, 9);
-        assert!(rd1.compatable(rd2).is_err());
+        assert!(rd1.compatible(rd2).is_err());
     }
 
     #[test]
@@ -598,12 +598,12 @@ mod test {
         let rd2 = test_rd();
 
         rd1.extent_count = 9;
-        assert!(rd1.compatable(rd2).is_err());
+        assert!(rd1.compatible(rd2).is_err());
 
         let rd1 = test_rd();
         let mut rd2 = test_rd();
         rd2.extent_count = 9;
-        assert!(rd1.compatable(rd2).is_err());
+        assert!(rd1.compatible(rd2).is_err());
     }
 
     #[test]
@@ -613,7 +613,7 @@ mod test {
         let rd2 = test_rd();
 
         rd1.uuid = rd2.uuid;
-        assert!(rd1.compatable(rd2).is_err());
+        assert!(rd1.compatible(rd2).is_err());
     }
 
     #[test]
@@ -622,12 +622,12 @@ mod test {
         let rd2 = test_rd();
 
         rd1.encrypted = true;
-        assert!(rd1.compatable(rd2).is_err());
+        assert!(rd1.compatible(rd2).is_err());
 
         let rd1 = test_rd();
         let mut rd2 = test_rd();
         rd2.encrypted = true;
-        assert!(rd1.compatable(rd2).is_err());
+        assert!(rd1.compatible(rd2).is_err());
     }
 
     #[test]
@@ -636,12 +636,12 @@ mod test {
         let rd2 = test_rd();
 
         rd1.database_read_version = DATABASE_READ_VERSION + 1;
-        assert!(rd1.compatable(rd2).is_err());
+        assert!(rd1.compatible(rd2).is_err());
 
         let rd1 = test_rd();
         let mut rd2 = test_rd();
         rd2.database_read_version = DATABASE_READ_VERSION + 1;
-        assert!(rd1.compatable(rd2).is_err());
+        assert!(rd1.compatible(rd2).is_err());
     }
 
     #[test]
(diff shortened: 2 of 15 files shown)


common/src/lib.rs
```diff
@@ -158,6 +158,9 @@ pub enum CrucibleError {
 
     #[error("missing block context for non-empty block")]
     MissingBlockContext,
+
+    #[error("Incompatable RegionDefinition {0}")]
```
mkeeter, line 162: `s/Incompatable/Incompatible/g`

common/src/region.rs
```diff
@@ -118,7 +128,7 @@ impl Block {
     }
 }
 
-#[derive(Deserialize, Serialize, Copy, Clone, Debug, PartialEq)]
+#[derive(Deserialize, Serialize, Copy, Clone, Debug, JsonSchema, PartialEq)]
 pub struct RegionDefinition {
```
mkeeter, line 132: Are there any consumers of this API right now?

If so, I'd suggest taking this opportunity to fix the duplicated data (#923) before they get too used to it.

If it's only used for debugging, then it's fine to leave as-is.

common/src/region.rs
```diff
@@ -170,6 +180,55 @@ impl RegionDefinition {
         })
     }
 
+    // Compare two RegionDefinitions and verify they are compatable.
+    // Compatable is valid if all fields are the same, expect for the
+    // UUID. The UUID should be different.
```
mkeeter, line 185: Out of curiosity, why can't the UUIDs match?  It seems like a region should be (tautologically) compatible with itself.

downstairs/src/extent.rs
```diff
@@ -653,13 +654,19 @@ pub(crate) fn move_replacement_extent<P: AsRef<Path>>(
...
-    // unfinished migrations.
+    // presence of the `.db` file.  Under normal conditions, we should never
+    // do extent repair across different extent formats; in fact, we should
+    // never extent repair SQLite-backed extents at all, but must still
+    // handle the case of unfinished migrations.
+    // If we are replacing a downstairs, and have created the region empty
+    // with the plan to replace it, then we can have a mismatch between
+    // what files we have and what files we want to replace them with.  When
+    // repairing a downstairs, we don't know in advance if the source of our
+    // repair will be SQLite-backend or not.
```
mkeeter, line 665: I found this confusing, how about something like
```
    // We distinguish between SQLite-backend and raw-file extents based on the
    // presence of the `.db` file.  We should never do extent repair across
    // different extent formats; it must be SQLite-to-SQLite or raw-to-raw.
    //
    // It is uncommon to perform extent repair on SQLite-backed extents at all,
    // because they are automatically migrated into raw file extents or
    // read-only.  However, it must be supported for two cases:
    //
    // - If there was an unfinished replacement, we must finish that replacement
    //   before migrating from SQLite -> raw file backend, which happens
    //   automatically later in startup.
    // - We use this same code path to perform clones of read-only regions,
    //   which may be SQLite-backed (and will never migrate to raw files,
    //   because they are read only).  This is only the case when the `clone`
    //   argument is `true`.
    //
    // In the first case, we know that we are repairing an SQLite-based extent
    // because the target (original) extent includes a `.db` file.
    //
    // In the second case, the target (original) extent is not present, so we
    // check whether the new files include a `.db` file.
```

downstairs/src/extent.rs
```diff
@@ -669,6 +676,12 @@ pub(crate) fn move_replacement_extent<P: AsRef<Path>>(
                 e
             );
         }
+        info!(
```
mkeeter, line 679: Do you want to keep these new log messages around?

downstairs/src/extent.rs
```diff
@@ -688,8 +701,15 @@ pub(crate) fn move_replacement_extent<P: AsRef<Path>>(
...
             }
+            info!(
+                log,
+                "Moved db-shm file {:?} to {:?}",
+                new_file.clone(),
+                original_file.clone()
+            );
             sync_path(&original_file, log)?;
         } else if original_file.exists() {
+            assert!(!clone);
```
mkeeter, line 712: Add a comment explaining why this is impossible (I think it's because the "original" region is empty?)

downstairs/src/main.rs
```diff
@@ -239,6 +255,51 @@ async fn main() -> Result<()> {
     let log = build_logger();
 
     match args {
+        Args::Clone {
+            data,
+            source,
+            trace_endpoint,
+        } => {
+            /*
```
mkeeter, line 263: Is this necessary, since we compile with `panic = abort`?

downstairs/src/region.rs
```diff
@@ -32,8 +32,30 @@ use crate::extent::{
  * There is only one repair file: the raw file itself (which also contains
  * structured context and metadata at the end).
  */
-pub fn validate_repair_files(eid: usize, files: &[String]) -> bool {
-    files == [extent_file_name(eid as u32, ExtentType::Data)]
+pub fn validate_repair_files(
```
mkeeter, line 35: Nit: how about splitting this into `validate_repair_files` and `validate_replacement_files`, since they're totally independent paths (and the docstring is wrong if this function does both)?

downstairs/src/region.rs
```diff
@@ -32,8 +32,30 @@ use crate::extent::{
...
+        let one = vec![extent_file_name(eid as u32, ExtentType::Data)];
+
+        let mut some = one.clone();
+        some.extend(vec![extent_file_name(eid as u32, ExtentType::Db)]);
+
+        let mut all = some.clone();
+        all.extend(vec![
+            extent_file_name(eid as u32, ExtentType::DbShm),
+            extent_file_name(eid as u32, ExtentType::DbWal),
+        ]);
```
mkeeter, line 50: Are these always guaranteed to be in the same order?  Using a `BTreeSet<String>` or similar would make it more obvious that the comparison is meant to be order-agnostic.

downstairs/src/region.rs
```diff
@@ -605,11 +649,46 @@ impl Region {
...
+                    .await?;
+                count += 1;
+            }
+        }
 
         // After we have all files: move the repair dir.
         info!(
             self.log,
-            "Repair files downloaded, move directory {:?} to {:?}",
+            "{count} Repair files downloaded, move directory {:?} to {:?}",
```
mkeeter, line 691:
```suggestion
            "{count} repair files downloaded, move directory {:?} to {:?}",
```

downstairs/src/region.rs
```diff
@@ -605,11 +649,46 @@ impl Region {
             }
         };
         save_stream_to_file(extent_copy, repair_stream.into_inner()).await?;
+        let mut count = 1;
+
+        // If we are replacing our region with one from an older version
+        // that contained the SQLite, then look for and copy the database
+        // files.
+        for opt_file in &[ExtentType::Db, ExtentType::DbShm, ExtentType::DbWal]
```
mkeeter, line 657: I believe we could do a loop over `[ExtentType::Data, ExtentType::Db, ExtentType::DbShm, ExtentType::DbWal]` and remove the specific handling of the data file above (L635-651).

downstairs/src/region.rs
```diff
@@ -605,11 +649,46 @@ impl Region {
...
+        {
+            let filename = extent_file_name(eid as u32, opt_file.clone());
+
+            if repair_files.contains(&filename) {
```
mkeeter, line 661: Nit: to prevent the rightward creep of indentation, perhaps
```rust
            if !repair_files.contains(&filename) {
                continue;
            }
```

downstairs/src/region.rs
```diff
@@ -605,11 +649,46 @@ impl Region {
...
         };
         save_stream_to_file(extent_copy, repair_stream.into_inner()).await?;
+        let mut count = 1;
+
+        // If we are replacing our region with one from an older version
+        // that contained the SQLite, then look for and copy the database
+        // files.
+        for opt_file in &[ExtentType::Db, ExtentType::DbShm, ExtentType::DbWal]
+        {
+            let filename = extent_file_name(eid as u32, opt_file.clone());
```
mkeeter, line 659: Nit: `ExtentType` should derive `Copy`, letting you remove this clone:
```suggestion
            let filename = extent_file_name(eid as u32, opt_file);
```

downstairs/src/region.rs
```diff
@@ -605,11 +649,46 @@ impl Region {
...
+                    eid,
+                    Some(opt_file.clone()),
+                )?;
+                let repair_stream = match repair_server
+                    .get_extent_file(eid as u32, opt_file.to_file_type())
+                    .await
+                {
+                    Ok(rs) => rs,
+                    Err(e) => {
+                        crucible_bail!(
```
mkeeter, line 673: There's a race condition here if the repair server cleans up the shm / wal files before you get to them, but that's probably fine...

downstairs/src/repair.rs
```diff
@@ -126,10 +133,10 @@ async fn get_extent_file(
             extent_path.set_extension("db");
         }
         FileType::DatabaseSharedMemory => {
-            extent_path.set_extension("db-wal");
+            extent_path.set_extension("db-shm");
```
mkeeter, line 136: Good catch!

downstairs/src/repair.rs
```diff
@@ -226,7 +234,12 @@ fn extent_file_list(
     eid: u32,
 ) -> Result<Vec<String>, HttpError> {
     let mut files = Vec::new();
-    let possible_files = vec![(extent_file_name(eid, ExtentType::Data), true)];
+    let possible_files = vec![
+        (extent_file_name(eid, ExtentType::Data), true),
+        (extent_file_name(eid, ExtentType::Db), false),
+        (extent_file_name(eid, ExtentType::DbShm), false),
+        (extent_file_name(eid, ExtentType::DbWal), false),
+    ];
```
mkeeter, line 242: Two nits:
- This could be an array instead of a `Vec`
- No need to call `extent_file_name(eid, ..)` here, just do it in the loop instead

```suggestion
    let possible_files = [
        (ExtentType::Data, true),
        (ExtentType::Db, false),
        (ExtentType::DbShm, false),
        (ExtentType::DbWal, false),
    ];
```

downstairs/src/repair.rs
```diff
@@ -240,6 +253,39 @@ fn extent_file_list(
...
+    method = GET,
+    path = "/region-info",
+}]
+async fn get_region_info(
+    rqctx: RequestContext<Arc<FileServerContext>>,
+) -> Result<HttpResponseOk<crucible_common::RegionDefinition>, HttpError> {
+    let region_info = rqctx.context().region_dir.clone();
+    let cp = region::config_path(region_info);
+
+    let def: crucible_common::RegionDefinition = match read_json(&cp) {
```
mkeeter, line 267: Could you instead pass the region definition into the `FileServerContext` at construction?  We're not worried about it changing on-disk, since that would mean we have bigger problems.

downstairs/src/lib.rs
```diff
@@ -3350,6 +3351,72 @@ pub async fn start_downstairs(
     Ok(join_handle)
 }
 
+/// Clone the extent files in a region from another running downstairs.
+///
+/// Use the reconcile/repair extent methods to copy another downstairs.
+/// The source downstairs must have the same RegionDefinition as we do,
+/// and both downstairs must be running in read only mode.
+pub async fn start_clone(
```
mkeeter, line 3359: This function appears to complete the clone, so `start_clone` is a bit misleading.

downstairs/src/lib.rs
```diff
@@ -3350,6 +3351,72 @@ pub async fn start_downstairs(
...
+    d: Arc<Mutex<Downstairs>>,
+    source: SocketAddr,
+) -> Result<()> {
+    let info = crucible_common::BuildInfo::default();
+    let log = d.lock().await.log.new(o!("task" => "main".to_string()));
```
mkeeter, line 3364: Why `"main"` instead of `"clone"`?

common/src/region.rs
leftwo: The only consumer is the downstairs itself.  One downstairs calls anothers repair endpoint.

If we want to fix #923 then I think it might be better to roll that in its own PR as I suspect it will
be a bunch of boilerplate changes (and an upstairs/downstairs version change too possibly).  
I would not want these changes to also include that.

common/src/region.rs
leftwo: From a strict comparison point of view, yes it would be compatable.

In this case, if you were trying to clone a downstairs, then your new downstairs should not
have the same UUID as the original.  So, it's a bit of implementation detail that has leaked
into the method here.  It also prevents you from trying to clone from yourself :)

downstairs/src/extent.rs
leftwo: Yeah, the original region in this case is the "new" region we just created, and that new region better not have
any SQLite files in it.  If it does, then something terrible is about to happen.

downstairs/src/region.rs
leftwo: We always produce them in the same order, so assuming `extent_file_list()`'s `into_iter()` call
will walk the `vec` provided to it in order, then I suppose it's safe to expect that order never
changes?  
We also (and I'm not sure why any longer) call `repair_files.sort()` over in
`downstairs/src/region.rs`, so even if the remote side does not produce the files in order, we
sort them on the requesting side.

Now that I think about it, I wonder if that `repair_files.sort()` in downstairs/src/region.rs was
attempting to work around the bug in the file type translation, where we had mixed `db-shm`
and `db-wal`.

downstairs/src/region.rs
leftwo: Yes, with the validate_*_files()` function, we know we have a correct set of files to stream over.

downstairs/src/repair.rs
```diff
@@ -159,6 +166,7 @@ async fn get_a_file(path: PathBuf) -> Result<Response<Body>, HttpError> {
             "Expected a file, found a directory".to_string(),
         ))
     } else {
+        println!("Get file at: {:?}", path);
```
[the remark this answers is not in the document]
leftwo: It was intentional when I wrote it:)
I was debugging the FileType bug and this got left behind.  I'll remove.

downstairs/src/lib.rs
leftwo: It is the "main" task for the clone command, but I'm not attached to the name really.
I'll change it to clone.

integration_tests/src/lib.rs
```diff
@@ -2662,6 +3052,142 @@ mod test {
...
+            )
+            .await
+            .unwrap();
+
+        assert_eq!(res, ReplaceResult::Started);
+
+        // We can use the result from calling replace_downstairs to
+        // intuit status on progress of the replacement.
+        loop {
+            tokio::time::sleep(tokio::time::Duration::from_secs(2)).await;
```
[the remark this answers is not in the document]
leftwo: It will do the replacement of one downstairs target with another.  But, since it's a RO volume, we don't
try to do any LiveRepair. I'm sure it would be less than a second to do the replacement, but I did
not want to spam the logs with as many times as we would check for it being done in that second.

common/src/lib.rs
leftwo: Done in [7bef555](https://github.com/oxidecomputer/crucible/commit/7bef55560bd1a345855ca77b6c296616ee66e0fe)

downstairs/src/extent.rs
leftwo: Done in [7bef555](https://github.com/oxidecomputer/crucible/commit/7bef55560bd1a345855ca77b6c296616ee66e0fe)

downstairs/src/extent.rs
leftwo: Probably not.  Now that I know where the bug was, I can remove them.
Done in [7bef555](https://github.com/oxidecomputer/crucible/commit/7bef55560bd1a345855ca77b6c296616ee66e0fe)

downstairs/src/region.rs
leftwo: See what I did in [7bef555](https://github.com/oxidecomputer/crucible/commit/7bef55560bd1a345855ca77b6c296616ee66e0fe)

downstairs/src/region.rs
```diff
@@ -389,6 +411,16 @@ impl Region {
         Ok(())
     }
 
+    pub async fn close_all_extents(&mut self) -> Result<()> {
```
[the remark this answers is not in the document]
leftwo: Done in [7bef555](https://github.com/oxidecomputer/crucible/commit/7bef55560bd1a345855ca77b6c296616ee66e0fe)

downstairs/src/region.rs
leftwo: Done in [7bef555](https://github.com/oxidecomputer/crucible/commit/7bef55560bd1a345855ca77b6c296616ee66e0fe)

downstairs/src/region.rs
leftwo: Done in [7bef555](https://github.com/oxidecomputer/crucible/commit/7bef55560bd1a345855ca77b6c296616ee66e0fe)

downstairs/src/region.rs
leftwo: Done in [7bef555](https://github.com/oxidecomputer/crucible/commit/7bef55560bd1a345855ca77b6c296616ee66e0fe)

downstairs/src/region.rs
leftwo: Done in [7bef555](https://github.com/oxidecomputer/crucible/commit/7bef55560bd1a345855ca77b6c296616ee66e0fe)

downstairs/src/region.rs
```diff
@@ -605,11 +649,46 @@ impl Region {
...
+
+        // If we are replacing our region with one from an older version
+        // that contained the SQLite, then look for and copy the database
+        // files.
+        for opt_file in &[ExtentType::Db, ExtentType::DbShm, ExtentType::DbWal]
+        {
+            let filename = extent_file_name(eid as u32, opt_file.clone());
+
+            if repair_files.contains(&filename) {
+                let extent_shm = Self::create_copy_file(
```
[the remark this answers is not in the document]
leftwo: Yeah, not the best name.
Done in [7bef555](https://github.com/oxidecomputer/crucible/commit/7bef55560bd1a345855ca77b6c296616ee66e0fe)

downstairs/src/repair.rs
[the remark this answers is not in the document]
leftwo: Removed in [7bef555](https://github.com/oxidecomputer/crucible/commit/7bef55560bd1a345855ca77b6c296616ee66e0fe)

downstairs/src/repair.rs
leftwo: Done in [7bef555](https://github.com/oxidecomputer/crucible/commit/7bef55560bd1a345855ca77b6c296616ee66e0fe)

downstairs/src/repair.rs
leftwo: Yeah, that's a much better idea.
Done in [7bef555](https://github.com/oxidecomputer/crucible/commit/7bef55560bd1a345855ca77b6c296616ee66e0fe)

downstairs/src/lib.rs
leftwo: Updated in [7bef555](https://github.com/oxidecomputer/crucible/commit/7bef55560bd1a345855ca77b6c296616ee66e0fe)

downstairs/src/lib.rs
leftwo: Done in [7bef555](https://github.com/oxidecomputer/crucible/commit/7bef55560bd1a345855ca77b6c296616ee66e0fe)

common/src/region.rs
```diff
@@ -170,6 +180,55 @@ impl RegionDefinition {
         })
     }
 
+    // Compare two RegionDefinitions and verify they are compatable.
+    // Compatable is valid if all fields are the same, expect for the
```
mkeeter, line 184:
```suggestion
    // Compare two RegionDefinitions and verify they are compatible.
    // Compatible is valid if all fields are the same, expect for the
```

common/src/region.rs
```diff
@@ -170,6 +180,55 @@ impl RegionDefinition {
         })
     }
 
+    // Compare two RegionDefinitions and verify they are compatable.
```
mkeeter, line 183: More instances of "compatable" in this comment + function name

downstairs/src/extent.rs
mkeeter: The new comment is confusing: if we're cloning a region, then the comment suggests that the region is created using `Backend::RawFile`, but then we (possibly) copy SQLite-formated extents into it?

Thinking about it a little more, I guess it's fine because we close the extent after creating it, then reopen it after cloning (and decide dynamically upon reopening whether it's raw or SQLite).  Still, it's a little odd.

I wonder if it would be worth creating a `Backend::Empty`, which creates the directory but no data files (and returns an error for every `ExtentInner` function)?

downstairs/src/region.rs
mkeeter: 👍🏻 not a concern, because "the other Downstairs decides to be adversarial" isn't really in our system model

downstairs/src/repair.rs
```diff
@@ -65,11 +69,26 @@ pub async fn repair_main(
...
+    let rd = region::config_path(region_dir.clone());
+    let region_definition = match read_json(&rd) {
+        Ok(def) => def,
+        Err(e) => {
+            return Err(format!(
+                "Can't get region definition from {:?}: {:?}",
+                rd, e
+            ));
+        }
+    };
```
mkeeter, line 82:
```suggestion
    let region_definition = ds.region.def();
```

leftwo:
Fixes https://github.com/oxidecomputer/crucible/issues/1018

common/src/region.rs
mkeeter: Yeah, I realized #923 is going to be more complicated to fix because it's already baked into our `region.json` files, so I'm fine leaving it alone.

downstairs/src/extent.rs
leftwo: > The new comment is confusing: if we're cloning a region, then the comment suggests that the region is created using Backend::RawFile, but then we (possibly) copy SQLite-formated extents into it?

Yes, the idea here is that we use the regular path (in Omicron) to create a region using whatever the
default standard region backend is, which would be fewer changes on the Omicron side.  Then, the clone
command just ignores whatever backend version the current region is and replaces it with the source.

I think `Backend::Empty` idea is interesting, but it would require updates on the omicron side to create
a region with a different set of options than default.  Not terrible, just more work.

downstairs/src/main.rs
leftwo: I dragged it from the other paths in main, but, yeah, given the top level, I don't think
we need this anywhere.

I'll file another issue to remove it from the 11 other places, no need to dirty up this PR with
those changes.

mkeeter:
This is _almost_ good to go; my only remaining comment is fixing the remaining instances of "compatable"

mkeeter (approved):

downstairs/src/lib.rs
```diff
@@ -3350,6 +3351,72 @@ pub async fn start_downstairs(
...
+    let my_def = ds.region.def();
+    info!(log, "my def is {:?}", my_def);
+
+    if let Err(e) = my_def.compatible(source_def) {
+        bail!("Incompatible region definitions: {e}");
+    }
+
+    if let Err(e) = ds.region.close_all_extents().await {
+        bail!("Failed to close all extents: {e}");
+    }
```
jmpesp, line 3406: At this point, do we need to reopen them? If we try to run the clone again, will it fail?

downstairs/src/main.rs
```diff
@@ -239,6 +255,41 @@ async fn main() -> Result<()> {
...
+                    .expect("Error init tracing subscriber");
+            }
+
+            let d = build_downstairs_for_region(
+                &data,
+                false,
+                false,
+                false,
+                false,
+                true, // read_only
```
jmpesp, line 286: Maybe it's time for the builder pattern... :upside_down_face:

downstairs/src/repair.rs
```diff
@@ -226,20 +236,51 @@ fn extent_file_list(
...
+}]
+async fn get_region_info(
+    rqctx: RequestContext<Arc<FileServerContext>>,
+) -> Result<HttpResponseOk<crucible_common::RegionDefinition>, HttpError> {
+    let region_definition = rqctx.context().region_definition;
+
+    Ok(HttpResponseOk(region_definition))
+}
+
+/// Return the RegionDefinition describing our region.
```
jmpesp, line 272: s/RegionDefinition/region-mode/

integration_tests/src/lib.rs
```diff
@@ -2570,6 +2596,370 @@ mod test {
...
+        .unwrap();
+
+        assert!(ds_two.reboot_clone(clone_source).await.is_err());
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn integration_test_clone_not_ro() -> Result<()> {
+        // Test downstairs replacement.
+        // Verify you can't replace from a RW downstairs
```
jmpesp, line 2896: s/replace/clone/

jmpesp (changes requested):

downstairs/src/lib.rs
leftwo: The clone region subcommand either runs to completion, or exits with an error.
So if we fail here, we are exiting the program.  If we run  the `clone` again, it will
start from extent 0 replace everything, even if a previous run only made it partially
through the list.

downstairs/src/main.rs
leftwo: Made a new PR for just that:
https://github.com/oxidecomputer/crucible/pull/1152

downstairs/src/repair.rs
leftwo: Fixed in 198b0ca

integration_tests/src/lib.rs
leftwo: Fixed in 198b0ca
Also, cleaned up a few more comments for a similar violation.

downstairs/src/lib.rs
jmpesp: I was worried that reissuing `clone` would hit

https://github.com/oxidecomputer/crucible/blob/2d7e499a9dfe5a6c287f70d63beb676c30bfbb3d/downstairs/src/region.rs#L467

but if failing will exit the program, then rerunning it again will start fresh. Makes sense :)

jmpesp (approved):

PR #1130, opened 2024-02-01
Add client id to error message

leftwo:
I saw a panic again on latest.
It would help if the error message confirmed what client the error came from.

### DIFF
--- a/upstairs/src/downstairs.rs
+++ b/upstairs/src/downstairs.rs
@@ -3151,7 +3151,7 @@ impl Downstairs {
         let Some(job) = self.ds_active.get_mut(&ds_id) else {
             error!(
                 self.clients[client_id].log,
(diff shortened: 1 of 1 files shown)


upstairs/src/downstairs.rs
```diff
@@ -3151,7 +3151,7 @@ impl Downstairs {
         let Some(job) = self.ds_active.get_mut(&ds_id) else {
             error!(
                 self.clients[client_id].log,
-                "IO completion error: missing {ds_id} "
+                "[{client_id}] IO completion error: missing {ds_id} "
```
mkeeter, line 3154: This should already be in the logs: we're writing to `self.clients[client_id].log`, which has a `client => id` tag.

leftwo:
Yeah, I can live without this

PR #1148, opened 2024-02-06
Use FIOFFS on illumos instead of fsync

jmpesp:
illumos supports sending a FIOFFS ioctl to a ZFS dataset in order to sync _all_ outstanding IO. Use this ioctl in `region_flush` for the whole region dataset instead of calling `sync_all` for each extent file. This is hidden behind a new `omicron-build` feature, which is true for our production builds.

This necessitated separating flush (which commits bits to durable storage and is a no-op if `omicron-build` is used) and `post_flush` (which performs clean up or any other kind of accounting). Splitting this up exposed a bug in the sqlite implementation of ExtentInner: it wasn't checking to see if the extent was dirty and wasn't calling `set_flush_number` before calling `sync_all` on the user data file.

### DIFF
--- a/downstairs/src/extent_inner_raw.rs
+++ b/downstairs/src/extent_inner_raw.rs
@@ -325,24 +325,42 @@ impl ExtentInner for RawInner {
         // operation atomic.
         self.set_flush_number(new_flush, new_gen)?;
 
-        // Now, we fsync to ensure data is flushed to disk.  It's okay to crash
-        // before this point, because setting the flush number is atomic.
-        cdt::extent__flush__file__start!(|| {
-            (job_id.get(), self.extent_number, 0)
-        });
-        if let Err(e) = self.file.sync_all() {
-            /*
-             * XXX Retry?  Mark extent as broken?
-             */
-            return Err(CrucibleError::IoError(format!(
-                "extent {}: fsync 1 failure: {e:?}",
-                self.extent_number,
-            )));
+        if cfg!(feature = "omicron-build") {
+            // no-op, FIOFFS will be called in region_flush
+        } else {
+            // Now, we fsync to ensure data is flushed to disk.  It's okay to crash
+            // before this point, because setting the flush number is atomic.
+            cdt::extent__flush__file__start!(|| {
+                (job_id.get(), self.extent_number, 0)
+            });
+
+            if let Err(e) = self.file.sync_all() {
+                /*
+                 * XXX Retry?  Mark extent as broken?
+                 */
+                return Err(CrucibleError::IoError(format!(
+                    "extent {}: fsync 1 failure: {e:?}",
+                    self.extent_number,
+                )));
+            }
+
+            cdt::extent__flush__file__done!(|| {
+                (job_id.get(), self.extent_number, 0)
+            });
         }
+
+        cdt::extent__flush__done!(|| { (job_id.get(), self.extent_number, 0) });
+
+        Ok(())
+    }
+
+    fn post_flush(
+        &mut self,
+        _new_flush: u64,
+        _new_gen: u64,
+        _job_id: JobOrReconciliationId,
+    ) -> Result<(), CrucibleError> {
         self.context_slot_dirty.fill(0);
-        cdt::extent__flush__file__done!(|| {
-            (job_id.get(), self.extent_number, 0)
-        });
 
         // Check for fragmentation in the context slots leading to worse
         // performance, and defragment if that's the case.
@@ -352,15 +370,11 @@ impl ExtentInner for RawInner {
             .unwrap_or(0);
         self.extra_syscall_count = 0;
         self.extra_syscall_denominator = 0;
-        let r = if extra_syscalls_per_rw > DEFRAGMENT_THRESHOLD {
+        if extra_syscalls_per_rw > DEFRAGMENT_THRESHOLD {
             self.defragment()
         } else {
             Ok(())
-        };
-
-        cdt::extent__flush__done!(|| { (job_id.get(), self.extent_number, 0) });
-
-        r
+        }
     }
 
     #[cfg(test)]
@@ -1852,6 +1866,7 @@ mod test {
 
         // Flush!  This will mark all slots as synched
         inner.flush(12, 12, JobId(11).into())?;
+        inner.post_flush(12, 12, JobId(11).into())?;
         assert_eq!(inner.active_context[0], ContextSlot::B);
         assert_eq!(inner.context_slot_dirty[0], 0b00);
 
@@ -1904,6 +1919,7 @@ mod test {
 
         // Flush!  This will mark all slots as synched
         inner.flush(12, 12, JobId(11).into())?;
+        inner.post_flush(12, 12, JobId(11).into())?;
         assert_eq!(inner.active_context[0], ContextSlot::A);
         assert_eq!(inner.context_slot_dirty[0], 0b00);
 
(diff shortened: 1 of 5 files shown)


jmpesp:
For posterity, the behaviour of `FIOFFS` was confirmed by the following test:

```c
#include <errno.h>
#include <fcntl.h>
#include <stdio.h>
#include <string.h>
#include <sys/filio.h>
#include <unistd.h>

int main() {
        int fd = open("/oxp_3d7b3f53-6487-4f0b-8a73-ba7f95a66d76/testpost", O_DIRECTORY | O_RDONLY, 0);

        if (fd == -1) {
                int e = errno;
                printf("could not open: %s\n", strerror(e));
                return 1;
        }

        printf("submitting %i %x %x:%x\n", _FIOFFS, _FIOFFS, 'f', 66);

        if (ioctl(fd, _FIOFFS, NULL) == -1) {
                int e = errno;
                printf("could not ioctl: %s\n", strerror(e));
        }

        if (close(fd) == -1) {
                int e = errno;
                printf("could not close: %s\n", strerror(e));
                return 1;
        }

        return 0;
}
```

Running this program to submit a `FIOFFS` for `/oxp_3d7b3f53-6487-4f0b-8a73-ba7f95a66d76/testpost` (the mountpoint of a dataset) and running dtrace shows that both `zfs_sync` and `zil_commit` are called:

```
james@dinnerbone:~$ gcc -o ioctl ioctl.c
james@dinnerbone:~$ pfexec dtrace -c 'pfexec ./ioctl' -n 'fbt::zfs_sync:entry' -n 'fbt::zfs_sync:return { print(arg1); }' -n 'fbt::zil_commit:entry'
dtrace: description 'fbt::zfs_sync:entry' matched 1 probe
dtrace: description 'fbt::zfs_sync:return ' matched 1 probe
dtrace: description 'fbt::zil_commit:entry' matched 1 probe
submitting 536897090 20006642 66:42
dtrace: pid 20032 has exited
CPU     ID                    FUNCTION:NAME
  0  49695                   zfs_sync:entry 
  0  47450                 zil_commit:entry 
  0  49696                  zfs_sync:return int64_t 0
```

jmpesp:
> This is hidden behind a new omicron-build feature, which is true for our production builds.

Also noting: this is true because we currently build with `--all-features` in our `build-release` buildomat job.

downstairs/src/extent_inner_raw.rs
```diff
@@ -325,24 +325,42 @@ impl ExtentInner for RawInner {
...
+                (job_id.get(), self.extent_number, 0)
+            });
         }
+
+        cdt::extent__flush__done!(|| { (job_id.get(), self.extent_number, 0) });
+
+        Ok(())
+    }
+
+    fn post_flush(
```
mkeeter, line 357: It weirds me out that `Extent::flush` doesn't actually flush, depending on `omicron-build`.  I understand why that's the case, but would like to propose an alternative API.  I think we should break flushing behavior into three functions:

- `pre_flush` contains everything through `self.set_flush_number(..)` in the current implementation
- `flush_inner` implements the `else` branch, i.e. calling `self.file.sync_all` (unconditionally!)
- `post_flush` is the same as in this PR

Then, we can have a default implementation of `flush` on `trait ExtentInner` which calls these functions one after the other, meaning unit tests that want to flush a single extent can just call `flush()`.  Right now, those unit tests are no longer actually syncing the file to disk!

More importantly, the decision to skip calling `flush_inner` (i.e. just calling `pre_flush` + `post_flush`) is moved into the `Region::region_flush` implementation, which is right next to the FIOFFS call, which makes it more obvious what's going on.

downstairs/src/extent_inner_raw.rs
```diff
@@ -325,24 +325,42 @@ impl ExtentInner for RawInner {
...
+                    self.extent_number,
+                )));
+            }
+
+            cdt::extent__flush__file__done!(|| {
+                (job_id.get(), self.extent_number, 0)
+            });
         }
+
+        cdt::extent__flush__done!(|| { (job_id.get(), self.extent_number, 0) });
```
mkeeter, line 352: I feel like this should be at the beginning of `post_flush`, so that it accounts for the actual flush time (even though that's amortized across many extents)

downstairs/src/extent_inner_sqlite.rs
```diff
@@ -75,33 +75,63 @@ impl ExtentInner for SqliteInner {
...
+            /*
+             * If we have made no writes to this extent since the last flush,
+             * we do not need to update the extent on disk
+             */
+            return Ok(());
+        }
+
+        // We put all of our metadata updates into a single write to make this
+        // operation atomic.
+        self.set_flush_number(new_flush, new_gen)?;
```
mkeeter, line 88: I'm suspicious about this change.

If we lose power right after this line, the metadata DB will tell us that we're at a particular flush (`new_flush`) and `dirty = 0`, but the data in the extent file has not necessarily been persisted to disk!

I think the existing implementation – which updates flush number in the DB **after** syncing the extent file – is correct.

downstairs/src/extent_inner_sqlite.rs
```diff
@@ -75,33 +75,63 @@ impl ExtentInner for SqliteInner {
...
-        cdt::extent__flush__file__start!(|| {
-            (job_id.get(), self.extent_number, n_dirty_blocks)
-        });
-        if let Err(e) = self.file.sync_all() {
+        if cfg!(feature = "omicron-build") {
+            // no-op, FIOFFS will be called in region_flush
+        } else {
             /*
-             * XXX Retry?  Mark extent as broken?
+             * We must first fsync to get any outstanding data written to disk.
```
mkeeter, line 101: (Same vibes here about changing to `pre_flush` / `flush_inner` / `post_flush`; you'll obviously have to fix both extent formats if you change the trait)

downstairs/src/region.rs
```diff
@@ -905,6 +908,43 @@ impl Region {
             result??;
         }
 
+        if cfg!(feature = "omicron-build") {
```
mkeeter, line 911: Using  `if cfg!(..)` weirds me out because the code is still compiled even if the feature is disabled; I guess the `ioctl` code builds just fine on MacOS, but I'd rather not!

What about something like this?

```rust
#[cfg(feature = "omicron-build")]
{
    #[cfg(not(target_os = "illumos"))]
    compile_error!("cannot use FIOFFS on non-illumos systems");
    
    // .. normal code continues below
```

leftwo (commented):
For `LiveRepair`, we required the ability to flush some extents (and their meta-data and
their dirty bit) and not flush other extents. 

Will this change still allow that ability?

jclulow:
I'm pretty sure that this ioctl is extremely private, FWIW.  How did we arrive here?  Is there some analysis and some description of why this is definitely safe?

jmpesp:
> For `LiveRepair`, we required the ability to flush some extents (and their meta-data and their dirty bit) and not flush other extents.
> 
> Will this change still allow that ability?

It shouldn't - `extent_limit` should still be honoured. What this PR changes is instead of iterating through the dirty extents (where we noted that writes occurred) and calling `fsync` on those file handles, we now issue a `zfs_sync` for the whole region dataset. I believe this is equivalent: writes should not have been occurring on extents higher than the extent limit independent of which flush technique is used.

leftwo:
This PR is obsolete now due to: https://github.com/oxidecomputer/crucible/pull/1199
Correct?

PR #1150, opened 2024-02-07
Simplify the `do_work_task` loop

mkeeter:
- Use a `VecDeque` to remove the inner / outer loops
- Clarify that the `is_lossy` conditional helps exercise reordering / dependency tracking
- Use the `let Some(v) = v else { continue; };` pattern to reduce rightward indentation creep
- Move the logic from `check_message_for_abort` next to the code that actually handles errors + replying

This is _mostly_ indentation changes by volume; sorry that the Github UI is Very Mad about it.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -4,7 +4,7 @@
 
 use futures::lock::{Mutex, MutexGuard};
 use std::cmp::Ordering;
-use std::collections::HashMap;
+use std::collections::{HashMap, VecDeque};
 use std::fmt;
 use std::fs::File;
 use std::io::{Read, Write};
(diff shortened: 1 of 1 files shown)


downstairs/src/lib.rs
```diff
@@ -1018,118 +1018,100 @@ async fn do_work_task(
...
-
-                            ads.lock()
-                                .await
-                                .complete_work_inner(
-                                    upstairs_connection,
-                                    job_id,
-                                    is_flush,
-                                )
-                                .await?;
+            cdt::work__process!(|| job_id.0);
```
leftwo, line 1047: I thought maybe this probe was in a bad spot, as we could hit this, then
fal out of the loop if some other downstairs was promoted at line 1057,
but, if that really is the case, then the whole accounting is meaningless 
as we should be shutting this task down with the other upstairs takes over.

Also, this issue was here before you refactored.

leftwo (approved):
Yeah, looks good.  Just a comment but no changes.

PR #1151, opened 2024-02-07
Move UUID checking to `resp_loop`

mkeeter:
This PR moves the checking of `upstairs_id` and `session_id` into `resp_loop`, immediately upon receiving a message from the `FramedRead`.  Doing the check early simplifies the code in `pf_task`, which was previously doing the equivalent check in a bunch of individual branches.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -533,39 +533,14 @@ pub mod cdt {
     fn work__done(_: u64) {}
 }
 
-// Check if a Message is valid on this downstairs or not.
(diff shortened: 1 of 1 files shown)


mkeeter:
On second thought, this is a bad idea because it means that pings don't reset the timeout (which is managed elsewhere).  Converting to draft while I think about a clean way to fix that...

mkeeter:
#1160 cleans up the call to `is_message_valid` in a similar way as this PR (but without the timeout-related shenanigans), so I'm going to close this one.

PR #1152, opened 2024-02-08
A builder for the Downstairs Downstairs struct.

leftwo:
Made a builder for  the Downstairs `Downstairs` struct.

Removed the `build_downstairs_for_region_with_backend()` and `build_downstairs_for_region()`
functions as they are no longer needed.


downstairs/src/lib.rs
```diff
@@ -1886,34 +1979,16 @@ pub struct Downstairs {
...
 #[allow(clippy::too_many_arguments)]
 impl Downstairs {
-    fn new(
-        region: Region,
-        lossy: bool,
-        read_errors: bool,
-        write_errors: bool,
-        flush_errors: bool,
-        read_only: bool,
-        encrypted: bool,
```
jmpesp, line 1896: how do we configure whether or not encryption is used?

leftwo: Hmm, I guess you can't any more when starting a downstairs.

The region creation itself is where the decision to have or not have encryption is made.  When we start a downstairs, the opening of the region will auto-select the encryption state.

If we allow start of a downstairs to pick what encryption it has, all that will end up doing is:
1: Match what the region open finds and move forward.
2: Not match the region open, and exit with error or panic.

Do we need to have this ability in the downstairs run path?

jmpesp: Definitely not, ignore my comment!

jmpesp (approved):

PR #1154, opened 2024-02-09
Remove `ExtentFlushClose::source_downstairs` from upstairs `IOop`

mkeeter:
Analogous to #1149, it turns out this field isn't used **anywhere**


leftwo (approved):
I just wanted to be sure we did not need this to tell the difference between a
downstairs under repair, which should not get a flush before close, and
the source of a repair, which should get a flush and close.

That logic is handled in `upstairs/src/downstairs.rs` in `io_send()`.

Issue #1155, opened 2024-02-10
Upstairs tries to process job after downstairs disconnects


leftwo:
The `tools/test_restart_repair.sh` test can trigger a panic in the upstairs when
a job sneaks through just after we have disconnected from a downstairs.

After running for a short time, the test fails as follows:

```
00:04:00.507Z INFO crucible: Gone missing, transition from Active to Offline client = 2
00:04:00.507Z INFO crucible: [2] client re-new 13 jobs since flush 1233
00:04:00.507Z INFO crucible: client task is exiting client = 2
00:04:00.507Z WARN crucible: c1a86d46-ae11-4752-9430-93245b6cc5f5 WARNING finish job 1246 when downstairs state: Offline  client = 2
thread 'tokio-runtime-worker' panicked at upstairs/src/client.rs:1249:13:
[2] Job 1246 completed while not InProgress: New   
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

leftwo:
I think a fix here might be to check in `/upstairs/src/downstairs.rs`'s `process_io_completion()`
(Where the warning message comes from). And, if the current state is Offline and the job has
moved to New, we discard the processing of it at that point.
We know that the job will be replayed if the downstairs comes back.

leftwo:
In testing, I seem to be able to hit this on slower machines easier, FWIW.  On the bench gimlet, I can't get it.

PR #1157, opened 2024-02-12
Drop jobs from Offline downstairs.

leftwo:
If we receive a job ACK from an offline downstairs, and that job has already been moved
back to New, then discard it as we know the replay will resend it when the downstairs 
comes back.

This is to fix https://github.com/oxidecomputer/crucible/issues/1155

### DIFF
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -112,6 +112,24 @@ struct ClientTaskHandle {
     client_stop_tx: Option<oneshot::Sender<ClientStopReason>>,
 }
 
+#[derive(Debug)]
+pub struct ConnectionId(pub u64);
+
+impl std::fmt::Display for ConnectionId {
+    fn fmt(
+        &self,
+        f: &mut std::fmt::Formatter<'_>,
+    ) -> Result<(), std::fmt::Error> {
+        self.0.fmt(f)
+    }
+}
+
+impl ConnectionId {
+    fn update(&mut self) {
+        self.0 += 1;
+    }
+}
+
 /// Per-client data
 ///
 /// This data structure contains client-specific state and manages communication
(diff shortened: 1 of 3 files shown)


upstairs/src/client.rs
```diff
@@ -195,6 +195,9 @@ pub(crate) struct DownstairsClient {
 
     /// State for startup negotiation
     negotiation_state: NegotiationState,
+
+    /// Session ID for a clients connection to a downstairs.
```
mkeeter, line 199:
```suggestion
    /// Session ID for a clients connection to a downstairs.
    ///
    /// This value is incremented whenever we restart the client IO task.
```

mkeeter:
This looks much better, thanks!

My one remaining thought is wondering whether we should define a `struct ConnectionId(u64)`, instead of the current `usize` (in `DownstairsClient`) → `u64` (in `get_connection_id`)

leftwo:
> This looks much better, thanks!
> 
> My one remaining thought is wondering whether we should define a `struct ConnectionId(u64)`, instead of the current `usize` (in `DownstairsClient`) → `u64` (in `get_connection_id`)

I've done that in the last commit.  I think this should be ready to go now.

upstairs/src/client.rs
```diff
@@ -2197,7 +2215,7 @@ impl DownstairsClient {
     /// different connections to the same Downstairs.
     pub(crate) fn get_connection_id(&self) -> Option<u64> {
         if self.client_task.client_stop_tx.is_some() {
-            Some(self.connection_id as u64)
+            Some(self.connection_id.0)
```
mkeeter, line 2218: Sorry to be so pedantic, but now that you've added the new type, we should actually use it – this function should be changed to return an `Option<ConnectionId>`, and then fix all of the compiler errors and we should be good :)

Issue #1146, opened 2024-02-05
Checksum does not match the data after LiveRepair


leftwo:
On a bench gimlet running `tools/test_fail_live_repair.sh` we failed during the verification phase.

On the 22nd loop, we saw the test fail:

```
[021][0] 4:54  ds_err:0 ave:5:00 total:105:00 last_run:294
New loop starts now Sat Feb  3 17:20:48 UTC 2024 faulting: 1
Downstairs client 1 has pid: 21537
Wait for our downstairs to fault
Wait for our downstairs to begin live_repair
After 6 seconds, Stop 1 again
Wait for downstairs 1 to go back to faulted
Start 1 for a second time
Now wait for all downstairs to be active
All downstairs active, now stop IO test and wait for it to finish
verify failed on loop 22
Final results:
[022] 4:54  ave:5:00  total:105:00 errors:1 last_run_seconds:294
Sat Feb  3 17:25:31 UTC 2024 Test ends with 1
```

The test logs show this complaint:
```
Region: es:300 ec:400 bs:4096  ts:491520000  tb:120000  max_io:256 or 1048576                                     
Loading write count information from file "/tmp/test_fail_live_verify"                                            
Read and Verify all blocks (0..120000 range:false)                                                                
17:25:30.198Z ERRO crucible: No match computed hash:0x9d8150b03debe3ca                          
17:25:30.198Z ERRO crucible: No match          hash:0x1e04f80286818d35                                            
17:25:30.198Z ERRO crucible: Data from hash:                                                                         
17:25:30.198Z ERRO crucible: [0]:75                                                                                 
17:25:30.198Z ERRO crucible: [1]:36                                                                                  
17:25:30.198Z ERRO crucible: [2]:36                                                                                  
17:25:30.198Z ERRO crucible: [3]:36                                                                                
17:25:30.198Z ERRO crucible: [4]:36                                                                                  
17:25:30.198Z ERRO crucible: [5]:36                                                                                                             
17:25:30.198Z ERRO crucible: decryption failure: HashMismatch
thread 'tokio-runtime-worker' panicked at upstairs/src/client.rs:1296:37:   
```

Full test output as well as region directories here: `/staff/alan/sn05-mismatch`

leftwo:
region.json for these regions looks like this:
```
{
  "block_size": 4096,
  "extent_size": {
    "value": 300,
    "shift": 12
  },
  "extent_count": 400,
[...]

PR #1159, opened 2024-02-12
Verify extent under repair is valid after copying files

leftwo:
Added a call to the downstairs repair API that will verify that an extent is closed (or the region is read only), 
meaning the extent is okay to copy.

This prevents a rare but possible condition where an extent is closed when a remote side contacts it for repair, but is re-opened before the extent file copy starts.  This leaves the resulting copy as invalid.

During the pstop/prun tests, I saw this once where the pstop to a downstairs happened just after we had
pulled over the list of files to verify, but before the actual files were copied over.  The details of that
are described here: https://github.com/oxidecomputer/crucible/issues/1146

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -39,7 +39,9 @@ mod stats;
 mod extent_inner_raw;
 mod extent_inner_sqlite;
 
+use extent::ExtentState;
 use region::Region;
+use repair::RepairReadyRequest;
 
 pub use admin::run_dropshot;
 pub use dump::dump_region;
@@ -1593,6 +1595,40 @@ where
 
     resp_loop(ads, fr, fw, another_upstairs_active_rx, upstairs_connection)
         .await
+}
+
+// Repair info task
+// listen for the repair API to request if a specific extent is closed
+// or if the entire region is read only.
+async fn repair_info(
+    ads: &Arc<Mutex<Downstairs>>,
+    mut repair_request_rx: mpsc::Receiver<RepairReadyRequest>,
+) -> Result<()> {
+    info!(ads.lock().await.log, "Started repair info task");
+    while let Some(repair_request) = repair_request_rx.recv().await {
+        let ds = ads.lock().await;
+        let state = if ds.read_only {
+            true
+        } else {
+            matches!(ds.region.extents[repair_request.eid], ExtentState::Closed)
+        };
+        drop(ds);
+
+        if let Err(e) = repair_request.tx.send(state) {
+            info!(
+                ads.lock().await.log,
+                "Error {e} sending repair info for extent {}",
+                repair_request.eid,
+            );
+        } else {
+            info!(
+                ads.lock().await.log,
+                "Repair info sent {state} for extent {}", repair_request.eid,
+            );
+        }
+    }
+    warn!(ads.lock().await.log, "Repair info task ends");
+    Ok(())
 }
 
 async fn reply_task<WT>(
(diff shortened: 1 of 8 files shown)


downstairs/src/lib.rs
```diff
@@ -3255,23 +3291,42 @@ pub async fn start_downstairs(
...
 
-    let repair_listener =
-        match repair::repair_main(&dss, repair_address, &repair_log).await {
-            Err(e) => {
-                // TODO tear down other things if repair server can't be
-                // started?
-                bail!("got {:?} from repair main", e);
-            }
+    let (repair_request_tx, repair_request_rx) = mpsc::channel(10);
+    let repair_listener = match repair::repair_main(
```
mkeeter, line 3295: Instead of spawning a separate task, could we handle this new functionality in `repair::repair_main`?  We're already passing in an `Arc<Mutex<Downstairs>>` (as the `&dss` argument), so I'm not sure what the extra task buys us.

mkeeter:
The compexity here of "checking before, sending files, checking after" makes me nervous – could we instead hold the `Downstairs` lock for the entire time we're sending files to the requester, ensuring that no one else could change them?

leftwo:
> The complexity here of "checking before, sending files, checking after" makes me nervous – could we instead hold the `Downstairs` lock for the entire time we're sending files to the requester, ensuring that no one else could change them?

I thought about that idea originally.  The repair API task does not have access to the downstairs lock currently, but it
could be updated to have it.  However, I still want to be able to recover from the case where a downstairs under
repair starts streaming, then goes out to lunch.  I would be concerned the source downstairs side is holding a lock, 
then never letting it go because the remote side is stuck/crashed/unavailable.  This would eventually result in the 
upstairs giving up and the `LiveRepair` needing to abort.

I was also not sure how to tell in dropshot when know when the stream has ended.

mkeeter:
> I thought about that idea originally. The repair API task does not have access to the downstairs lock currently, but it
could be updated to have it. However, I still want to be able to recover from the case where a downstairs under
repair starts streaming, then goes out to lunch. I would be concerned the source downstairs side is holding a lock,
then never letting it go because the remote side is stuck/crashed/unavailable. This would eventually result in the
upstairs giving up and the LiveRepair needing to abort.
>
> I was also not sure how to tell in dropshot when know when the stream has ended.

Okay, that seems reasonable as a justification for the three-step process.

Somewhat orthogonally, I still think that moving this new API into `repair_main` is sensible (per [this comment](https://github.com/oxidecomputer/crucible/pull/1159#discussion_r1486601982)); what do you think?

downstairs/src/lib.rs
leftwo: We do pass in the downstairs mutex, but we don't keep it.  We only use it to pull out a few fields.
I was actually moddeling this after what you did in `upstairs/src/control.rs` for the single task
refactoring (#1058).

mkeeter: I think it's 6 of one, half-dozen of the other.

Having _any task_ holding on to an `Arc<Mutex<Downstairs>>` is something that will have to change if we switch to a single-task implementation (with a dedicated task owning a `Downstairs`).  It doesn't much matter whether the `Arc<Mutex<Downstairs>>` is held by `repair_main` or the new `repair_info`.

leftwo: My hope here was to make an eventual transition to a similar one-task implementation easier, and make the
downstairs similar to the upstairs in that regard.  Is this at least a step in that direction?

mkeeter: Previously, there were only two categories of tasks using the `Arc<Mutex<Downstairs>>`:
- The main listener task
- Any tasks spawned by it via `proc_stream`, i.e. `pf_task` and `dw_task`

After this PR, the important change is that we'll have a **third category of task** using the `Arc<Mutex<Downstairs>>`.

This will either be `repair_info` (in the current implementation) or `repair_main` (in my suggestion).  Which one isn't important, the important part (architecturally) is that it's a **third thing** sharing that mutex; and that's what we'll have to deal with in a future single-task refactoring.

From that perspective, either implementation is making a single-task refactoring harder.

That being said, it's not going to make such a refactoring **significantly** harder; we can just make requests to the `Downstairs` task through a queue, instead of calling into its APIs directly.

Since both implementations are architecturally equivalent (w.r.t. a single-task refactoring), I'd vote for the one that doesn't proliferate tasks further (i.e. adding this API to `repair_main`).  Does that make sense?  I don't feel _too_ strongly about it, so I'd defer to your judgement if you still think a separate task is the Right Choice.

leftwo: I put it all in `repair_main()`,  no more repair task.

mkeeter (approved):

jmpesp (approved):

PR #1160, opened 2024-02-13
Move a few methods into downstairs

mkeeter:
At some point, we may want the `Downstairs` to own its data.  This PR moves `proc_frame` and the core of `do_work_task` into methods on the `Downstairs`, though they still take a `Mutex<Downstairs>` as their argument.

The purpose of this PR is to make later PRs more readable: this is mostly mechanical moving + indenting of code, so that a later PR that changes logic won't also have 500 LOC of reorganization.

(I did sneak in _one_ small change, which is [consolidating calls to `is_message_valid`](https://github.com/mkeeter/crucible/blob/29185f5538632a13039785e911d1aaa0146bfa14/downstairs/src/lib.rs#L2254-L2312))

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -1182,18 +1182,27 @@ where
         let resp_channel_tx = resp_channel_tx.clone();
         tokio::spawn(async move {
             while let Some(m) = message_channel_rx.recv().await {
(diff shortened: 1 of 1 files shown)


leftwo (approved):
Github just can't figure out what to do with your PRs.
I just did a old/new files side by side and there really is not much to it.

PR #1162, opened 2024-02-14
Do not 500 on snapshot delete for deleted region

jmpesp:
The code path that validates a snapshot delete request checks that a region exists, and returns 404 if it does not:

    match rc.context().get(&p.id) {
        Some(_) => (),
        None => {
            return Err(HttpError::for_not_found(
                None,
                format!("region {:?} not found", p.id),
            ));
        }
    }

However there's a flaw in this check: if the region existed but was deleted, this will not return a 404, and proceed into `delete_snapshot`.

`delete_snapshot` will check if a running snapshot exists, and if it does will bail out: it's an invalid request to delete a snapshot for which there's a read-only downstairs running for.

However, the very next thing the code does is try to make a ZFSDataset object out of the region's data path. If the region was deleted this will lead to `zfs list` exiting with a non-zero return code, and will bubble up to the user as a 500.

Catch ZFSDataset returning Err, and if the region existed but was deleted, we can assume the snapshot was deleted prior to this (as the agent would not allow otherwise). If the region existed and isn't deleted, and `zfs list` for that region path failed, then the agent's datafile doesn't match reality.

### DIFF
--- a/agent/src/datafile.rs
+++ b/agent/src/datafile.rs
@@ -464,7 +464,7 @@ impl DataFile {
             path.into_os_string().into_string().unwrap(),
         ) {
             Ok(dataset) => dataset,
-            Err(_e) => {
+            Err(e) => {
                 // This branch can only be entered if `zfs list` for that
                 // dataset path failed to return anything.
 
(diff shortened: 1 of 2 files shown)


agent/src/datafile.rs
```diff
@@ -460,8 +460,43 @@ impl DataFile {
...
+                        }
+
+                        State::Failed => {
+                            // Something has set the region to state failed, so
+                            // we can't delete this snapshot.
+                            bail!("Region {} is in state failed", request.id.0);
+                        }
+                    }
+                } else {
+                    // In here, the region never existed!
```
leftwo, line 495: Does this mean we need to keep the list of deleted regions forever?
If we don't keep regions forever, then this will fail.  This requirement may already
be baked into the way the agent works, idk.

jmpesp: That's already a requirement for the agent, yeah. See section 4 of RFD 419:

> From this, daemons that are contacted by sagas must themselves also soft-delete resources so that requests for these resources can return the same results. This must extend to nested paths as well: if trip 123 was soft deleted, then all sub-resources must first have been soft-deleted, and subsequent delete requests should return 204.

agent/src/main.rs
```diff
@@ -95,7 +98,9 @@ impl ZFSDataset {
             .output()?;
 
         if !cmd.status.success() {
-            bail!("zfs list failed!");
+            let stderr =
+                String::from_utf8_lossy(&cmd.stderr).trim_end().to_string();
```
leftwo, line 102: Yes!

leftwo (approved):

PR #1165, opened 2024-02-16
Make `crucible-downstairs` not depend on upstairs

mkeeter:
It always weirded me out that `crucible-downstairs` depended on `crucible` proper, instead of `crucible-protocol` + `crucible-common`.  For example, one awkward side effect was that the `crucible-downstairs` binary had all of the upstairs' DTrace probes.

It turns out that removing the dependency was (almost) trivial: it required moving one module (`impacted_blocks`) into `crucible-common`, but no other changes.

This may also be an incremental speedup for building / linking.


leftwo (approved):
Yeah, this is probably left over from before there was even a common or protocol
as a separate thing.

Issue #1133, opened 2024-02-02
remove custom panic setup


leftwo:
_Originally posted by @leftwo in https://github.com/oxidecomputer/crucible/pull/1129#discussion_r1476505152_
            

Now that we set `panic = abort` at the top level, we
don't need to set it specifically for each subsystem.

PR #1174, opened 2024-02-20
Remove individual panic setup, use global panic settings

leftwo:
Fix for https://github.com/oxidecomputer/crucible/issues/1133


mkeeter (approved):
🚀

PR #1175, opened 2024-02-21
Make `gw_ds_complete` less public

mkeeter:
I'm a little confused by this – `crucible` compiles just fine, but using it in `propolis-standalone` gives a compile-time error message:

```
error[E0446]: crate-private type `GuestWorkId` in public interface
   --> /home/matt/crucible/upstairs/src/guest.rs:246:5
    |
150 | pub(crate) struct GuestWorkId(pub u64);
    | ----------------------------- `GuestWorkId` declared as crate-private
...
246 |     #[instrument]
    |     ^^^^^^^^^^^^^ can't leak crate-private type
    |
    = note: this error originates in the attribute macro `instrument` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0446`.
```

### DIFF
--- a/upstairs/src/guest.rs
+++ b/upstairs/src/guest.rs
@@ -147,7 +147,7 @@ impl GtoS {
 
 /// Strongly-typed ID for guest work (stored in the [`GuestWork`] map)
 #[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, Ord, PartialOrd)]
-pub struct GuestWorkId(pub u64);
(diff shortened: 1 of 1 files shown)


jmpesp (approved):

mkeeter:
Update: it turns out that changing `gw_ds_complete` from `pub` to `pub(crate)` also resolves the issue, so let's do that instead!

Issue #1167, opened 2024-02-16
Revenge of the backpressure


mkeeter:
[RFD 445](https://rfd.shared.oxide.computer/rfd/445) is barely released, and I've found another interesting corner case in our current backpressure implementation.

Right now, we don't apply backpressure on reads.  This is because reads don't have the fast-ack optimization: we actually need to wait for a read to finish before acking it to the guest, so there's no way for jobs to pile up...

_...or is there?_

I tested a 1 MiB random read workload in Crucible branch with [a bunch of bonus DTrace probes](https://github.com/mkeeter/crucible/tree/tracing-upstairs-jobs).  The guest is seeing about 500-600 IOPS, with 8 worker tasks.

Measuring rx-to-tx time in each of the 3x Downstairs, the result is surprising:

![upper](https://github.com/oxidecomputer/crucible/assets/745333/9139834f-b4e1-4a19-bfb1-9b8378445ec7)

`ds0` is serving all of our reads, with a turnaround of about 10 ms (this would normally limit us to 100 IOPS, but I think the 8x worker threads help saturate it more fully).

Meanwhile, the other two downstairs get slower and slower over the course of the run.  We can see this in the `IP` columns of `upstairs_info.d`:
```
    UPW   DSW BAKPR   NEW0  NEW1  NEW2    IP0   IP1   IP2     D0    D1    D2
      8   308     0      0     0     0      8    74    76    300   234   232
      7   302     0      0     0     0      7    92    73    295   210   229
      8   311     0      0     0     0      8   122   103    303   189   208
      7   302     0      0     0     0      7   121    72    295   181   230
      7   309     0      0     0     0      7   120    69    302   189   240
      7   304     0      0     0     0      7   113    76    297   191   228
      7   309     0      0     0     0      7   116    76    302   193   233
      7   308     0      0     0     0      7   126    89    301   182   219
      7   308     0      0     0     0      7   109    64    301   199   244
      8   301     0      0     0     0      8   123    88    293   178   213
      7   307     0      0     0     0      7   147    99    300   160   208
      7   298     0      0     0     0      7   150   100    291   148   198
      8   298     0      0     0     0      8   130   102    290   168   196
      8   301     0      0     0     0      8   140   103    293   161   198
```

## The mistaken assumption

Previously, we declared that we "wait for a read to finish before acking it to the guest, so there's no way for jobs to pile up".

[...]

PR #1176, opened 2024-02-21
Backpressure consolidation

mkeeter:
This PR changes the way that backpressure is tracked.  Previously, we had to manually increment / decrement counters whenever jobs were created or destroyed (which could happen in multiple places).  In this PR, I introduce two things:

- An `Arc<BackpressureCounters>`, which is shared between the guest and upstairs and can be updated atomically
- A new `struct BackpressureGuard`, which does three things:
    - Stores an `Arc<BackpressureCounters>`, i.e. a handle to the global counter object
    - Upon construction, increments those counters
    - When dropped, decrements those counters

In other words, as long as the `BackpressureGuard` lives alongside the IO that it represents, we get updates to our counters for free.  To make this happen, I added a `backpressure_guard` field to the `struct DownstairsIO`, which lives as long as the IO operation is active.

One side effect of this change is that the "write bytes in flight" counter is now updated **immediately** upon receiving the job, rather than when encryption finishes.  This means that we can do our fast-ack before encryption as well, which wasn't possible before*.  In other words, we're removing the last known source of Accidental Backpressure in the system, and rely solely on Intentional Backpressure to delay writes.

This PR doesn't address #1167 and doesn't make the full set of changes recommended in [RFD 445](https://rfd.shared.oxide.computer/rfd/445); in particular, we still need to add the "mark Downstairs as failed if it has too much data in flight" change.

I still need to get some performance numbers, but this is ready for initial review.

*acking before encryption + updating the backpressure counters _after_ encryption is effectively adding lag to the backpressure control loop, and makes it go unstable.

### DIFF
--- a/upstairs/src/backpressure.rs
+++ b/upstairs/src/backpressure.rs
@@ -1,4 +1,4 @@
-// Copyright 2023 Oxide Computer Company
+// Copyright 2024 Oxide Computer Company
 use std::{
     sync::{
         atomic::{AtomicU64, Ordering},
--- a/upstairs/src/backpressure.rs
+++ b/upstairs/src/backpressure.rs
@@ -1,0 +1,121 @@
+// Copyright 2023 Oxide Computer Company
+use std::{
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc,
+    },
+    time::Duration,
+};
+
+#[derive(Debug, Default)]
+pub(crate) struct BackpressureCounters {
+    /// The number of write bytes that haven't finished yet
+    ///
+    /// This is used to configure backpressure to the host, because writes
+    /// (uniquely) will return before actually being completed by a Downstairs
+    /// and can clog up the queues.
+    write_bytes_outstanding: AtomicU64,
+
+    /// Number of IO jobs in the downstairs queue
+    io_jobs: AtomicU64,
+
+    /// Cached value for backpressure
+    ///
+    /// This is written whenever we recompute backpressure, and read by the
+    /// `GuestIoHandle` when it wants to print updated stats.
+    backpressure_us: AtomicU64,
+}
+
+impl BackpressureCounters {
+    pub fn new() -> Arc<Self> {
+        Arc::new(Self::default())
+    }
+
+    pub fn write_bytes_outstanding(&self) -> u64 {
+        self.write_bytes_outstanding.load(Ordering::Relaxed)
+    }
+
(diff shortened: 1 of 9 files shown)


upstairs/src/backpressure.rs
```diff
@@ -0,0 +1,121 @@
+// Copyright 2023 Oxide Computer Company
```
jmpesp, line 1: 2024 :)

upstairs/src/block_req.rs
```diff
@@ -57,16 +101,43 @@ impl BlockRes {
...
+    /// The backpressure counters are stored locally and decremented when this
+    /// `BlockRes` is dropped, so we automatically keep track of outstanding
+    /// work without human intervention.
+    fn bind(&mut self, op: &BlockOp, bp: Arc<BackpressureCounters>) {
+        // Count this job as 1 job and some number of bytes
+        //
+        // (we only count write bytes because only writes return immediately)
+        assert!(self.backpressure_guard.is_none());
+        let bytes = match op {
+            BlockOp::Write { data, .. } => data.len() as u64,
```
jmpesp, line 121: does it make sense to include `WriteUnwritten` here? That should never come from the Guest, except through scrubbing.

jmpesp (approved):

upstairs/src/backpressure.rs
mkeeter: hah, I literally started this PR back in December

fixed in 5c1dc1c4b8a30fcd1ef620114b41b62e0473a376

upstairs/src/block_req.rs
mkeeter: My intuition is "no", because `WriteUnwritten` operations don't get to use the fast-ack optimization, so they should get "natural" backpressure (in that we'll only reply to the guest once the operation is done*).  That's how it's handled in the current code, so my (weakly-held) inclination is to leave it be.

*with some caveats related to #1167, but that's neither here nor there

leftwo (approved):
I should have looked at this backpressure PR first.

mkeeter:
Status: I'm still thinking about whether this is the right approach, so will hold off on merging for now.

mkeeter:
This no longer sparks joy, so I'm going to close it.

- I don't think the fool-proof-ness of updating backpressure on `Drop` is worth the extra complexity,
- The design somewhat conflicts with #1181 and #1192, which are more important

PR #1181, opened 2024-02-26
Add per-client backpressure

### LINKED ISSUE #1167: Revenge of the backpressure

mkeeter:
Fixes #1167 

This PR adds per-client delays to prevent queues from becoming arbitrarily long during read-heavy workflows.

In our current code, because we only need one reply to ack a read, the fastest Downstairs will consistently beat the others.  This means that the queues for the shorter two Downstairs will keep getting longer and longer (so the fastest Downstairs will keep winning), and those queues could eventually hit the `IO_OUTSTANDING_MAX` limit (which means the Upstairs would mark the slower Downstairs as faulted).

This PR adds monitoring of the difference between Downstairs reply times with a new `reply_times: ClientData<Instant>` in the `struct DownstairsIO`.  If the difference between fastest and slowest is significant (> 10 ms), then we add an artificial delay to the fastest downstairs.

Compared to `main`, this slows down reads, but keeps the queues relatively bounded.  Here's performance before:
```
1M WRITE: bw=692MiB/s (726MB/s), 692MiB/s-692MiB/s (726MB/s-726MB/s), io=40.7GiB (43.7GB), run=60146-60146msec
4K WRITE: bw=28.4MiB/s (29.8MB/s), 28.4MiB/s-28.4MiB/s (29.8MB/s-29.8MB/s), io=1707MiB (1789MB), run=60005-60005msec
1M WRITE: bw=713MiB/s (748MB/s), 713MiB/s-713MiB/s (748MB/s-748MB/s), io=41.8GiB (44.9GB), run=60051-60051msec
4M WRITE: bw=713MiB/s (748MB/s), 713MiB/s-713MiB/s (748MB/s-748MB/s), io=41.9GiB (45.0GB), run=60181-60181msec
4K READ: bw=29.0MiB/s (30.4MB/s), 29.0MiB/s-29.0MiB/s (30.4MB/s-30.4MB/s), io=1741MiB (1826MB), run=60004-60004msec
1M READ: bw=603MiB/s (632MB/s), 603MiB/s-603MiB/s (632MB/s-632MB/s), io=35.3GiB (37.9GB), run=60038-60038msec
4M READ: bw=738MiB/s (774MB/s), 738MiB/s-738MiB/s (774MB/s-774MB/s), io=43.3GiB (46.5GB), run=60135-60135msec
```
upstairs_info_prev

(notice pending jobs creeping up in the 1M and 4M read sections of the graph)

...and after:
```
1M WRITE: bw=687MiB/s (721MB/s), 687MiB/s-687MiB/s (721MB/s-721MB/s), io=40.4GiB (43.4GB), run=60198-60198msec
4K WRITE: bw=29.4MiB/s (30.8MB/s), 29.4MiB/s-29.4MiB/s (30.8MB/s-30.8MB/s), io=1765MiB (1850MB), run=60008-60008msec
1M WRITE: bw=714MiB/s (749MB/s), 714MiB/s-714MiB/s (749MB/s-749MB/s), io=41.9GiB (44.9GB), run=60029-60029msec
4M WRITE: bw=697MiB/s (731MB/s), 697MiB/s-697MiB/s (731MB/s-731MB/s), io=40.9GiB (44.0GB), run=60134-60134msec
4K READ: bw=28.8MiB/s (30.2MB/s), 28.8MiB/s-28.8MiB/s (30.2MB/s-30.2MB/s), io=1731MiB (1815MB), run=60003-60003msec
1M READ: bw=562MiB/s (589MB/s), 562MiB/s-562MiB/s (589MB/s-589MB/s), io=32.9GiB (35.4GB), run=60038-60038msec
4M READ: bw=561MiB/s (588MB/s), 561MiB/s-561MiB/s (588MB/s-588MB/s), io=32.9GiB (35.4GB), run=60162-60162msec
```
upstairs_info_pr

I don't think we can avoid slowing down reads in the steady-state: the whole point is to keep the queues bounded, which necessarily means matching the speed of the slowest Downstairs.  However, burst performance should be unaffected, because the delays only kick in at > 10 ms of discrepancy (which takes some amount of time to wind up).

Note that we **don't** delay writes!  This took a little convincing, but I believe it's correct: during write-heavy workloads, we're already limited by global backpressure (both jobs and bytes-in-flight), so it's already impossible for a queue to grow without bounds.

### DIFF
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -643,6 +643,10 @@ impl DownstairsClient {
         }
 
         self.connection_id.update();
+
+        // Clear per-client delay, because we're starting a new session
+        self.set_delay_us(0);
+
         // Restart with a short delay
         self.start_task(true, auto_promote);
     }
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -644,9 +644,6 @@ impl DownstairsClient {
 
         self.connection_id.update();
 
-        // Clear per-client delay, because we're starting a new session
-        self.set_delay_us(0);
-
         // Restart with a short delay
         self.start_task(true, auto_promote);
     }
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -10,7 +10,14 @@ use crate::{
 use crucible_common::x509::TLSContext;
 use crucible_protocol::{ReconciliationId, CRUCIBLE_MESSAGE_VERSION};
 
-use std::{collections::BTreeSet, net::SocketAddr, sync::Arc};
+use std::{
+    collections::BTreeSet,
+    net::SocketAddr,
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc,
+    },
+};
 
 use futures::StreamExt;
 use slog::{debug, error, info, o, warn, Logger};
@@ -18,7 +25,7 @@ use tokio::{
     io::AsyncWriteExt,
     net::{TcpSocket, TcpStream},
     sync::{mpsc, oneshot},
-    time::sleep_until,
+    time::{sleep_until, Duration},
 };
 use tokio_util::codec::{Encoder, FramedRead};
 use uuid::Uuid;
@@ -216,6 +223,9 @@ pub(crate) struct DownstairsClient {
 
     /// Session ID for a clients connection to a downstairs.
     connection_id: ConnectionId,
+
+    /// Per-client delay, shared with the [`DownstairsClient`]
+    client_delay_us: Arc<AtomicU64>,
 }
 
 impl DownstairsClient {
@@ -226,6 +236,7 @@ impl DownstairsClient {
         log: Logger,
         tls_context: Option<Arc<crucible_common::x509::TLSContext>>,
     ) -> Self {
+        let client_delay_us = Arc::new(AtomicU64::new(0));
         Self {
             cfg,
             client_task: Self::new_io_task(
@@ -234,6 +245,7 @@ impl DownstairsClient {
                 false, // do not start the task until GoActive
                 client_id,
                 tls_context.clone(),
+                client_delay_us.clone(),
                 &log,
             ),
             client_id,
@@ -253,6 +265,7 @@ impl DownstairsClient {
             repair_info: None,
             io_state_count: ClientIOStateCount::new(),
             connection_id: ConnectionId(0),
+            client_delay_us,
         }
     }
 
@@ -262,6 +275,7 @@ impl DownstairsClient {
     /// client will disappear into the void.
     #[cfg(test)]
     fn test_default() -> Self {
+        let client_delay_us = Arc::new(AtomicU64::new(0));
         let cfg = Arc::new(UpstairsConfig {
             encryption_context: None,
             upstairs_id: Uuid::new_v4(),
@@ -290,6 +304,7 @@ impl DownstairsClient {
             repair_info: None,
             io_state_count: ClientIOStateCount::new(),
             connection_id: ConnectionId(0),
+            client_delay_us,
         }
     }
 
@@ -652,6 +667,7 @@ impl DownstairsClient {
             connect,
             self.client_id,
             self.tls_context.clone(),
+            self.client_delay_us.clone(),
             &self.log,
         );
     }
@@ -662,6 +678,7 @@ impl DownstairsClient {
         connect: bool,
         client_id: ClientId,
         tls_context: Option<Arc<TLSContext>>,
+        client_delay_us: Arc<AtomicU64>,
         log: &Logger,
     ) -> ClientTaskHandle {
         #[cfg(test)]
@@ -672,6 +689,7 @@ impl DownstairsClient {
                 connect,
                 client_id,
                 tls_context,
+                client_delay_us,
                 log,
             )
         } else {
@@ -685,6 +703,7 @@ impl DownstairsClient {
             connect,
             client_id,
             tls_context,
+            client_delay_us,
             log,
         )
     }
@@ -695,6 +714,7 @@ impl DownstairsClient {
         connect: bool,
         client_id: ClientId,
         tls_context: Option<Arc<TLSContext>>,
+        client_delay_us: Arc<AtomicU64>,
         log: &Logger,
     ) -> ClientTaskHandle {
         // These channels must support at least MAX_ACTIVE_COUNT messages;
@@ -730,6 +750,7 @@ impl DownstairsClient {
                     log: log.clone(),
                 },
                 delay,
+                client_delay_us,
                 log,
             };
             c.run().await
@@ -1241,6 +1262,8 @@ impl DownstairsClient {
             warn!(self.log, "Dropping already skipped job {}", ds_id);
             return false;
         }
+
+        job.reply_time[self.client_id] = Some(std::time::Instant::now());
 
         let mut jobs_completed_ok = job.state_count().completed_ok();
         let mut ackable = false;
@@ -2220,6 +2243,11 @@ impl DownstairsClient {
             None
         }
     }
+
+    /// Sets the per-client delay
+    pub(crate) fn set_delay_us(&self, delay: u64) {
+        self.client_delay_us.store(delay, Ordering::Relaxed);
+    }
 }
 
 /// How to handle "promote to active" requests
@@ -2419,6 +2447,9 @@ struct ClientIoTask {
 
     /// Handle for the rx task
     recv_task: ClientRxTask,
+
+    /// Shared handle to receive per-client backpressure delay
+    client_delay_us: Arc<AtomicU64>,
 
     log: Logger,
 }
@@ -2692,6 +2723,24 @@ impl ClientIoTask {
             + std::marker::Send
             + 'static,
     {
+        // Delay communication with this client based on backpressure, to keep
+        // the three clients relatively in sync with each other.
+        //
+        // We don't need to delay writes, because they're already constrained by
+        // the global backpressure system and cannot build up an unbounded
+        // queue.  This is admittedly quite subtle; see crucible#1167 for
+        // discussions and graphs.
+        if !matches!(
+            m,
+            ClientRequest::Message(Message::Write { .. })
+                | ClientRequest::RawMessage(RawMessage::Write { .. }, ..)
+        ) {
+            let d = self.client_delay_us.load(Ordering::Relaxed);
+            if d > 0 {
+                tokio::time::sleep(Duration::from_micros(d)).await;
+            }
+        }
+
         // There's some duplication between this function and `cmd_loop` above,
         // but it's not obvious whether there's a cleaner way to organize stuff.
         tokio::select! {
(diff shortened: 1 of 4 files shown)


upstairs/src/client.rs
```diff
@@ -2692,6 +2723,24 @@ impl ClientIoTask {
...
             + 'static,
     {
+        // Delay communication with this client based on backpressure, to keep
+        // the three clients relatively in sync with each other.
+        //
+        // We don't need to delay writes, because they're already constrained by
+        // the global backpressure system and cannot build up an unbounded
+        // queue.  This is admittedly quite subtle; see crucible#1167 for
+        // discussions and graphs.
+        if !matches!(
```
leftwo, line 2733: reads and flushes, those we should respect the backpressure delay for.

What about `WriteUnwritten`, and/or any of the repair operations?
If we block those here, will that have any impact elsewhere in the system?
There should never be more than one `WriteUnwritten`, or four of the
repair jobs outstanding at a time, so I don't really think there could be
too much of an impact if we slow them down.

Edit: It looks like we don't set timing data for repair operations, but they
will still be subject to delays if reads have backed up enough such that
we have set a delay.

upstairs/src/downstairs.rs
```diff
@@ -3166,6 +3174,30 @@ impl Downstairs {
...
+                    // numbers!
+                    let dt = slowest_time - job.reply_time[i].unwrap();
+                    let lead_time_us: u64 = dt.as_micros().try_into().unwrap();
+                    (lead_time_us.saturating_sub(10_000) / 100).min(10_000)
+                } else {
+                    0
+                };
+
+                self.clients[i].set_delay_us(delay_time_us);
+            }
```
leftwo, line 3198: Is there any way we can see this from the outside?  It would be interesting to be able
to track this and watch how it changes, but I'm not sure how to bubble it out to
where we could grab it with the rest of the `collect_stats()`.

leftwo (approved):
My only real question is if there is a way to know from the outside what the
delay settings are currently set to?  

I could see in the future we could do even more with that info, but I don't think 
we need to hold up this PR to figure out the best way to figure out how to
archive it, but it would be cool to know what the settings are at some sample
frequency or point in time.

upstairs/src/downstairs.rs
mkeeter: I added this to the existing `up-status` DTrace probe in 597b9500b147b344dfdcb1d65190bc505a12c11b

(The string is now 768 bytes long with all zeros, so still well below the `strsize=1k` limit)

upstairs/src/client.rs
mkeeter: `WriteUnwritten` don't go through the normal backpressure system, so I think they qualify for delays here (but am not totally sure).

I'm also not too worried about it, since the max per-client delay is 10ms, and we only hit that if one client is **a full second** behind the others.  The intent for this delay is to serve as a small nudge, not "blocking" in any significant sense.

jmpesp (changes requested):
Some questions:

- should we change `IO_OUTSTANDING_MAX`?
- what if jobs arrive out of order? technically the downstairs is free to do non-dependent jobs in any order, and`reply_time` isn't the same as some notion of how long a job took to do by a particular downstairs.

upstairs/src/downstairs.rs
```diff
@@ -3166,6 +3174,30 @@ impl Downstairs {
...
+        }
```
jmpesp, line 3199: If we're not setting the per-client time, should we clear it? What if one Downstairs is `Faulted`, or in replay - the previously set delays will still apply.

jmpesp:
> I'm not sure how to fix this – if end-to-end time varies based on things other than true execution time + queue delays, then it becomes tricky to control (e.g. what if every downstairs randomly delayed jobs by 1-10 ms?).
>
> My inclination is to punt on it until we actually start reordering jobs in the Downstairs, and then do so with caution.

I'm curious if doing per-client backpressure based on the queue lengths makes sense? It doesn't take into account the actual time difference like `reply_time` does though, and I don't know how that would affect how we decide to kick out a very slow Downstairs.

For what it's worth, starting a Downstairs with `--lossy` will randomly insert pauses and skip jobs, and lead to this sort of out of order execution.

upstairs/src/downstairs.rs
mkeeter: > If we're not setting the per-client time, should we clear it?

I don't think so.  If we did that, then the first two replies to a job would clear the delay, then the third reply would set it (before it's cleared again by the first reply to the next job).

> What if one Downstairs is Faulted, or in replay - the previously set delays will still apply.

I added logic to clear the per-client delay in `Client::reinitialize` (1190fa30237afd358ff98f397d78ea82e8297014).

Maybe we should clear it if _any_ of the clients restart?  Otherwise, I could imagine a situation where 

- DS1 is temporarily slower than DS2, so we delay DS2
- Then DS3 returns an error, so we restart it
- We're still delaying DS2, and that won't change until DS3 comes back online

jmpesp: > I don't think so. If we did that, then the first two replies to a job would clear the delay, then the third reply would set it (before it's cleared again by the first reply to the next job).

:+1:

> Maybe we should clear it if any of the clients restart?

That sounds like a good idea, yeah.

mkeeter:
> I'm curious if doing per-client backpressure based on the queue lengths makes sense? It doesn't take into account the actual time difference like reply_time does though, and I don't know how that would affect how we decide to kick out a very slow Downstairs.

I considered this, but initially decided against it because queue length is a pretty crude measurement of expected delay: a queue with 500x 4K jobs is very different from a queue with 500x 1 MiB jobs.

If we use the existing backpressure system as an analogy, we could instead delay individual clients based on the maximum of relative jobs-in-flight and bytes-in-flight (versus the slowest Downstairs).  Do you want me to throw that together?

Kicking out a very slow Downstairs should take the same amount of time as before; it will just delay the _other two_ Downstairs before it gets kicked out.  (In the current code, the max delay is 10 ms/job)

jmpesp:
> Do you want me to throw that together?

Summarizing a long DM chain, I do think this is worth testing! It may solve a bunch of questions I/we had.

upstairs/src/downstairs.rs
mkeeter: Okay, done in 80b6f734175c2bc8e222c8d6f7875e53cbd5c9ca

mkeeter:
Done in #1186

PR #1182, opened 2024-02-27
Add generic 'message or raw message' code

mkeeter:
This PR adds a generic implementation of the 'Message or raw message' serializer used for reduced-mempcy serialization (#1087), then updates `crucible-upstairs` to use the new implementation.

It also more generally reduces allocations in the socket write path by reusing the same `BytesMut` (for `Message` serialization) or a `Vec` (for raw message headers).  This may have a small performance impact; I didn't measure it.

These changes are a building block for faster serialization of Downstairs `ReadResponse`, but I pulled them into a standalone PR for ease of review.

### DIFF
--- a/protocol/src/lib.rs
+++ b/protocol/src/lib.rs
@@ -709,6 +709,9 @@ pub trait RawMessageDiscriminant {
 pub struct WireMessageWriter<W, T> {
     writer: W,
 
+    /// Scratch space for `Message` encoding
+    scratch: BytesMut,
+
     /// Scratch space for the raw header
     header: Vec<u8>,
     _phantom: std::marker::PhantomData<T>,
@@ -724,6 +727,7 @@ where
     pub fn new(writer: W) -> Self {
         Self {
             writer,
+            scratch: BytesMut::new(),
             header: vec![],
             _phantom: std::marker::PhantomData,
         }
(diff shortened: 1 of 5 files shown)


protocol/src/lib.rs
```diff
@@ -674,6 +674,125 @@ impl Message {
...
+                let len: u32 =
+                    (self.header.len() + data.len()).try_into().unwrap();
+                self.header[0..4].copy_from_slice(&len.to_le_bytes());
+
+                // Patch the discriminant in the header
+                bincode::serialize_into(
+                    &mut self.header[4..8],
+                    &m.discriminant(),
+                )
+                .unwrap();
```
leftwo, line 785: I realize these were here in the old `ClientRequest`, but what would cause these
`unwraps()` to trigger?  The `data.len()).try_into().unwrap()` seems pretty
unlikely, but what about this `serialize_into()`?

leftwo (approved):
This is not actually changing what is going over the wire, right?
like an older downstairs could still read something sent by this upstairs?

Assuming so, then let it go (back to main).

mkeeter:
> This is not actually changing what is going over the wire, right?
> like an older downstairs could still read something sent by this upstairs?

That is correct, the wire format is staying the same.

protocol/src/lib.rs
mkeeter: It's hard to trace through `bincode`, but I think the only error conditions are
- Size limits (which we aren't enabling)
- A failure of the `impl std::io::Write`

In this case, we're serializing a tag-only enum into a `u32`, so it should be infallible.

PR #1183, opened 2024-02-27
Reduce serialization during Downstairs reads

mkeeter:
(This PR is staged on top of #1182, and is therefore marked as a draft; you should ignore all changes in `crucible-protocol` and `crucible-upstairs`)

#1087 changed how we handle writes Upstairs to avoid memcpy'ing bulk data in `Message::Write/WriteUnwritten`.  This PR does the same thing for `Message::ReadResponse` in the Downstairs.

Just like #1087, it's Unfortunately Messy: we have to do the `bincode` encoding by hand at the same time that we're reading data from the file, to avoid an extra copy operation.

The benefits are what you'd expect: the 11% of our runtime spent in `memcpy` disappears, and things get faster.

Here's the flamegraph before these changes...
ds_main
([interactive version](https://github.com/oxidecomputer/crucible/assets/745333/8c012922-1180-4009-9d44-d71bd6ca525d))

...and after
ds_zc
([interactive version](https://github.com/oxidecomputer/crucible/assets/745333/45fb7415-9789-4a76-9751-7020e6fe3dc7))

Running my usual `fio` sweep, I see noticeable improvements in large read performance.

### Before
```
4K READ: bw=27.8MiB/s (29.1MB/s), 27.8MiB/s-27.8MiB/s (29.1MB/s-29.1MB/s), io=1667MiB (1748MB), run=60003-60003msec
1M READ: bw=525MiB/s (550MB/s), 525MiB/s-525MiB/s (550MB/s-550MB/s), io=30.8GiB (33.0GB), run=60048-60048msec
4M READ: bw=596MiB/s (625MB/s), 596MiB/s-596MiB/s (625MB/s-625MB/s), io=35.0GiB (37.6GB), run=60138-60138msec
```
### After
```
4K READ: bw=27.8MiB/s (29.2MB/s), 27.8MiB/s-27.8MiB/s (29.2MB/s-29.2MB/s), io=1670MiB (1751MB), run=60027-60027msec
1M READ: bw=675MiB/s (708MB/s), 675MiB/s-675MiB/s (708MB/s-708MB/s), io=39.6GiB (42.5GB), run=60035-60035msec
4M READ: bw=756MiB/s (793MB/s), 756MiB/s-756MiB/s (793MB/s-793MB/s), io=44.4GiB (47.7GB), run=60127-60127msec
```
(The Upstairs includes the changes from #1181, which is why initial performance is slower than you may expect)

### DIFF
--- a/downstairs/src/extent.rs
+++ b/downstairs/src/extent.rs
@@ -5,7 +5,6 @@ use std::fmt::Debug;
 use std::fs::{File, OpenOptions};
 
 use anyhow::{anyhow, bail, Context, Result};
-use bytes::BufMut;
 use nix::unistd::{sysconf, SysconfVar};
 
 use serde::{Deserialize, Serialize};
@@ -54,24 +53,28 @@ pub(crate) trait ExtentInner: Send + Sync + Debug {
         requests: &[crucible_protocol::ReadRequest],
         iov_max: usize,
         out: &mut BytesMut,
-    ) -> Result<(), CrucibleError> {
-        // Call the existing read implementation, then do serialization by hand
-        // here.  Note that we serialize the individual ReadResponse values, but
-        // not the leading size, because that's already placed into the buffer.
-        let data = self.read(job_id, requests, iov_max)?;
-        let mut w = out.writer();
-        for d in data {
-            bincode::serialize_into(&mut w, &d).unwrap();
-        }
-        Ok(())
-    }
+    ) -> Result<(), CrucibleError>;
 
+    /// Read the given requests into a `Vec<ReadResponse>`
+    ///
+    /// The default implementation defers to `read_raw`, then deserializes the
+    /// resulting buffer.
+    ///
+    /// This function is only built during unit tests, because it's less
+    /// efficient than `read_raw`.
+    #[cfg(test)]
     fn read(
         &mut self,
         job_id: JobId,
         requests: &[crucible_protocol::ReadRequest],
         iov_max: usize,
-    ) -> Result<Vec<crucible_protocol::ReadResponse>, CrucibleError>;
+    ) -> Result<Vec<crucible_protocol::ReadResponse>, CrucibleError> {
+        use bytes::BufMut;
+        let mut b = BytesMut::new();
+        b.put_u64_le(requests.len() as u64);
+        self.read_raw(job_id, requests, iov_max, &mut b)?;
+        Ok(bincode::deserialize(&b).unwrap())
+    }
 
     fn write(
         &mut self,
--- a/downstairs/src/extent.rs
+++ b/downstairs/src/extent.rs
@@ -540,7 +540,7 @@ impl Extent {
 
     /// Read the real data off underlying storage, and get block metadata. If
     /// an error occurs while processing any of the requests, the state of
-    /// `responses` is undefined.
+    /// `out` is undefined.
     #[instrument]
     pub async fn read(
         &mut self,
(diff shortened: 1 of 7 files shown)


downstairs/src/extent.rs
```diff
@@ -518,17 +546,18 @@ impl Extent {
         &mut self,
         job_id: JobId,
         requests: &[crucible_protocol::ReadRequest],
-    ) -> Result<Vec<crucible_protocol::ReadResponse>, CrucibleError> {
+        out: &mut BytesMut,
```
leftwo, line 549: We need to update the comment above this method as there is no more `responses`

downstairs/src/extent_inner_raw.rs
```diff
@@ -220,24 +225,73 @@ impl ExtentInner for RawInner {
...
             let mut expected_bytes = 0;
-            for resp in
-                &mut responses[resp_run_start..][..n_contiguous_requests]
-            {
-                expected_bytes += resp.data.len();
-                iovecs.push(IoSliceMut::new(&mut resp.data[..]));
+            for chunk in &mut chunks {
+                expected_bytes += block_size as usize;
+                iovecs.push(IoSliceMut::new(
+                    &mut chunk[28..][..block_size as usize],
```
leftwo, line 293: The 28 here, that's `eid` + `offset.value` + `offset.shift` + `block_size`?
Is there any way to.. hmm.  this is really hand rolled serialization.
Maybe comment what the 28 is for.

If we ever change how this is rolled up, I guess it just has to be done carefully.

downstairs/src/region.rs
```diff
@@ -841,13 +842,54 @@ impl Region {
...
+                    + core::mem::size_of::<u64>()  // block.value
+                    + core::mem::size_of::<u32>()  // block.shift
+                    + core::mem::size_of::<usize>()  // data size
+                    + self.def.block_size() as usize // read data
+                    + core::mem::size_of::<usize>() // context count
+                    + core::mem::size_of::<u64>() // hash
+                    + core::mem::size_of::<u8>() // option tag
+                    + core::mem::size_of::<[u8; 12]>() // nonce
+                    + core::mem::size_of::<[u8; 16]>()); // tag
+
```
leftwo, line 889: Is the above what exactly the data looks like with bincode?
If so, and in some future bincode changes, would this fail to compile?
Or is bincode changing it's format not a realistic scenario?

leftwo (commented):
Some questions to be sure I'm following along correctly here.

downstairs/src/extent.rs
mkeeter: Yup, fixed in 6c12d1b7944333f9a0a6d686c84ae0ae56e1bb5c

downstairs/src/extent_inner_raw.rs
mkeeter: Yes, that's correct.  I've added a named constant and some internal consistency checks in a184fe0e745f27076ef068cf43bf19a0d8e52c03

downstairs/src/region.rs
mkeeter: Yes, this is exactly how the data is serialized.

`bincode`'s binary format is [specified](https://github.com/bincode-org/bincode/blob/trunk/docs/spec.md) and should not change (although I did open a ticket to [clarify `Option`'s behavior](https://github.com/bincode-org/bincode/issues/700))

I moved this blob into a standalone function and added a unit test in 8f04737bc10f4cb7036b8f318b0a84f89a1bb7c3

leftwo: >  moved this blob into a standalone function and added a unit test in

I'm I little disappointed you (correctly) branch predicted my follow on comment, robbing
me the chance to do so, but don't worry, I'll recover.

leftwo (approved):
Thanks for the updates

mkeeter:
#1206 does this but better

PR #1193, opened 2024-03-07
Fix outdated log message

mkeeter:
This used to be a counter, but is now a boolean, so the log message is silly:
```
Waiting for true more clients to be ready
```


leftwo (approved):
This seems like a pretty big change, we should get 6 or 7 more people to verify it.

PR #1192, opened 2024-03-06
Fault if too many bytes are in flight

mkeeter:
As recommended in [RFD 445](https://rfd.shared.oxide.computer/rfd/445#_problematic_bytes_in_flight_delay), this PR adds a fault condition if too many bytes pile up in flight for any particular downstairs.

This adds a nice symmetry to the system: both backpressure and per-downstairs faults consider both jobs and bytes in flight.

PR #1198, opened 2024-03-08
Automatically split > 1 MiB writes

mkeeter:
In #1192 , we noticed that arbitrarily-large writes can overwhelm our backpressure implementation, because they take longer than our maximum backpressure.

This PR automatically splits large writes into 1 MiB chunks.  Propolis already does this, so I don't see any performance impacts, but it means that we'll be robust against stuff like "`crudd` tries to send unreasonably large write jobs".

### DIFF
--- a/upstairs/src/guest.rs
+++ b/upstairs/src/guest.rs
@@ -613,8 +613,8 @@ impl BlockIO for Guest {
 
     async fn write(
         &self,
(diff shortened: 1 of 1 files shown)


jmpesp (approved):

faithanalog (approved):
makes sense to me

PR #1203, opened 2024-03-12
Automatically split > 1 MiB read jobs

mkeeter:
Analogous to #1198 , but for reads instead.

This improves the interaction between large reads and per-queue backpressure.

I also added a unit test for large (> 1 MiB) IOs in `crucible-integration-tests`

### DIFF
--- a/upstairs/src/guest.rs
+++ b/upstairs/src/guest.rs
@@ -580,7 +580,7 @@ impl BlockIO for Guest {
 
     async fn read(
         &self,
-        offset: Block,
+        mut offset: Block,
         data: &mut Buffer,
     ) -> Result<(), CrucibleError> {
         let bs = self.check_data_size(data.len()).await?;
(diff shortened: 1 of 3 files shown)


upstairs/src/lib.rs
```diff
@@ -1781,6 +1781,65 @@ impl Buffer {
...
+    ///
+    /// Afterwards `self` contains elements `[0, at)`, and the returned `Buffer`
+    /// contains elements `[at, capacity)`.
+    ///
+    /// This is an `O(1)` operation that just increases the reference count and
+    /// sets a few indices.
+    ///
+    /// # Panics
+    /// The `index` must be an even multiple of block size
+    pub fn split_off(&mut self, index: usize) -> Self {
```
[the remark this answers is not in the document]
leftwo: I would lean a little toward index being block offset over byte offset, but don't feel strongly either way.  We don't ever plan to operate on a byte offset here.  I would not care if you left it like it is here either.

upstairs/src/lib.rs
```diff
@@ -1781,6 +1781,65 @@ impl Buffer {
...
+        assert_eq!(index % self.block_size, 0);
+        let data = self.data.split_off(index);
+        let owned = self.owned.split_off(index / self.block_size);
+        Self {
+            block_size: self.block_size,
+            data,
+            owned,
```
leftwo, line 1802: I believe using `owned` like this is okay, but we should probably update the comment in `upstairs/src/lib.rs`
for `pub struct Buffer`, as it describes ownership in terms of a block being written to.
As I read that comment though, I wonder if this buffer splitting works with the scrubber correctly?

If the scrubber is set to 131072, then that should fall below the level of breaking up IOs.

mkeeter: Fixed the comment in a2e5c68eeeb3a8199a852ca40aba1e10f956a57b (that should have been changed in #1200, oops)

Can you say more about interactions with the scrubber?  I'm not as familiar with how it works.

upstairs/src/guest.rs
```diff
@@ -593,19 +593,50 @@ impl BlockIO for Guest {
...
+            assert_eq!(chunk.len() as u64 % bs, 0);
+
+            let offset_change = chunk.len() as u64 / bs;
+            let rio = BlockOp::Read {
+                offset,
+                data: chunk,
+            };
+
+            let reply = self.send_and_wait(rio).await;
+
```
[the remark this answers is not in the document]
mkeeter: Yes, and I think is it's the correct behavior.

Right now, there's no backpressure on reads, so forcing large reads to happen serially seems safer than running them in parallel.

(In fact, there's a corner case where someone could overwhelm the system by spawning a zillion tasks that do simultaneous reads, but we haven't hit that issue in practice yet)

upstairs/src/lib.rs
leftwo: I'm not sure myself if this buffer layer does interact, but I'd like to call @jmpesp to the witness
stand and have his thoughts here.  The [scrubber](https://github.com/oxidecomputer/crucible/blob/6f7884ae4dcf21ee7a92664b83d9b4ae89130223/upstairs/src/volume.rs#L301) just reads from the ROP and writes
to the region, so on the surface I don't see the buffer's owner being used, but I suspect there
might be something under the covers, or in other backends that make use of it?

mkeeter: As far as I can tell, I haven't changed any behavior of `owned`.  It sounds like you think something has changed there?

leftwo (commented):
Assuming my understanding of `owned` is correct, I'm good to go here.
(If I'm incorrect, then, sorry, help me understand it :) )

upstairs/src/lib.rs
mkeeter: Your understanding is correct – we're splitting the buffer into two smaller buffers, but `data` and `owned` remain consistent w.r.t. each other.

leftwo (approved):

upstairs/src/lib.rs
jmpesp: > The scrubber just reads from the ROP and writes to the region, so on the surface I don't see the buffer's owner being used, but I suspect there might be something under the covers, or in other backends that make use of it?

No, you're right - the scrubber does a `read` (which returns a `Buffer`), then calls `write_unwritten` and passes it `buffer.into_bytes()`, tossing away ownership information. But as I understand it this PR doesn't change the Buffer that would be returned, it simply does the splitting and joining under the covers. The caller of `read` won't notice a difference. If it did, that'd be a bug :)

jmpesp (approved):

PR #1207, opened 2024-03-14
Make three smaller timeouts instead of one large one.

leftwo:
Updated the Upstairs <-> Downstairs message timeout to have three smaller timeouts 
before triggering a disconnect due to inactivity.  Log a message on each smaller timeout.

The timeout was changed from 50 seconds to 45 (3x15), but I don't expect that to make
much of a difference.

Moved deadline_secs() to a common location for both upstairs and downstairs to use.  
This resulted in a little shuffling of included libraries as well.

The messages the upstairs logs for a downstairs timeout now look like this:
```
15:41:34.545Z WARN crucible: Nothing received from downstairs, timeout 1/3                                                             
     = downstairs                                                                                                                      
    client = 0                                                                                                                         
    session_id = bfe20188-ad31-4cbd-897c-2083976c7ac4                                                                                  
15:41:49.545Z WARN crucible: Nothing received from downstairs, timeout 2/3                                                             
     = downstairs                                                                                                                      
    client = 0                                                                                                                         
    session_id = bfe20188-ad31-4cbd-897c-2083976c7ac4                                                                                  
15:42:04.547Z WARN crucible: Nothing received from downstairs, timeout 3/3                                                             
     = downstairs                                                                                                                      
    client = 0                                                                                                                         
    session_id = bfe20188-ad31-4cbd-897c-2083976c7ac4                                                                                  
15:42:04.547Z WARN crucible: client task is sending Done(Timeout)                                                                      
     = downstairs                                                                                                                          
     client = 0                                                                                                                         
    session_id = bfe20188-ad31-4cbd-897c-2083976c7ac4
```

From the downstairs, when the upstairs goes quiet, we see this:
```
15:42:55.556Z WARN crucible: Nothing received from upstairs, timeout 1/3
    task = main
15:43:10.558Z WARN crucible: Nothing received from upstairs, timeout 2/3
    task = main
15:43:25.559Z WARN crucible: Nothing received from upstairs, timeout 3/3
    task = main
15:43:25.559Z ERRO crucible: connection (127.0.0.1:41938) Exits with error: inactivity timeout
```

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -15,8 +15,9 @@ use std::time::Duration;
 
 use crucible_common::{
     build_logger, crucible_bail, deadline_secs,
-    impacted_blocks::extent_from_offset, integrity_hash, mkdir_for_file, Block,
-    CrucibleError, RegionDefinition, MAX_ACTIVE_COUNT, MAX_BLOCK_SIZE,
+    impacted_blocks::extent_from_offset, integrity_hash, mkdir_for_file,
+    verbose_timeout, Block, CrucibleError, RegionDefinition, MAX_ACTIVE_COUNT,
+    MAX_BLOCK_SIZE,
 };
 use crucible_protocol::{
     BlockContext, CrucibleDecoder, CrucibleEncoder, JobId, Message,
@@ -1112,7 +1113,6 @@ where
         + std::marker::Send
         + 'static,
 {
-    let mut lossy_interval = deadline_secs(5.0);
     // Create the log for this task to use.
     let log = ads.lock().await.log.new(o!("task" => "main".to_string()));
 
@@ -1212,16 +1212,11 @@ where
         })
     };
 
-    // The lossy attribute currently does not change at runtime. To avoid
-    // continually locking the downstairs, cache the result here.
-    let lossy = ads.lock().await.lossy;
-
     // How long we wait before logging a message that we have not heard from
     // the upstairs.
     const TIMEOUT_SECS: f32 = 15.0;
-    // How many timeouts will we tolerate before we disconnect from the upstairs.
-    const TIMEOUT_LIMIT: u8 = 3;
-    let mut timeout_trip = 0;
+    // How many timeouts will tolerate before we disconnect from the upstairs.
+    const TIMEOUT_LIMIT: usize = 3;
     loop {
         tokio::select! {
             e = &mut dw_task => {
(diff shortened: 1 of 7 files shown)


downstairs/src/lib.rs
```diff
@@ -1241,15 +1241,25 @@ where
...
             /*
-             * Don't wait more than 50 seconds to hear from the other side.
+             * Don't wait more than TIMEOUT_SECS * TIMEOUT_LIMIT seconds to hear
+             * from the other side.
              * XXX Timeouts, timeouts: always wrong!  Some too short and
              * some too long.
              */
-            _ = sleep_until(deadline_secs(50)) => {
-                bail!("inactivity timeout");
+            _ = sleep_until(deadline_secs(TIMEOUT_SECS)) => {
```
jmpesp, line 1252: I'm wondering if this is the same behaviour as what we have in the Upstairs: we have a variable `timeout = deadline_secs(TIMEOUT_SECS);`, which is bumped when we receive a new message, where her it's on the tokio select expression, recomputed each loop.

mkeeter:
Implementing this manually in each place seems awkward; what about doing it in a single future instead?

For example, we could add this to `crucible-common`:
```rust
pub async fn verbose_timeout(secs: f32, n: usize, log: slog::Logger) {
    let d = Duration::from_secs_f32(secs);
    for i in 0..n {
        tokio::time::sleep(d).await;
        slog::warn!(log, "timeout {}/{n}", i + 1,);
    }
}
```

Then use it in `select!` statements like so:
```rust
tokio::select! {
    // other branches...
    _ = verbose_timeout(TIMEOUT_SECS, TIMEOUT_LIMIT, log.clone()) => {
        bail!("inactivity timeout");
    }
```

downstairs/src/lib.rs
mkeeter: I had the same thought, but believe the behavior is identical in both places: in the Upstairs, all branches either reset the timeout or break out of the loop.  That's equivalent to starting the timeout from 0 on each iteration through the loop.

(also, there's no reason to use `sleep_until(deadline_secs(..))` versus plain `sleep(..)`)

mkeeter: However, thinking about it again, the **downstairs** behavior may be wrong!  Right now, if `lossy` is set, then we'll never timeout, because `lossy_interval < TIMEOUT_SECS`

leftwo: Oh, if hitting the lossy timeout reset the whole select, then that issue has been.. here.. for a while now.
I guess because we never really test both lossy and timeouts, we never noticed.

leftwo:
> Implementing this manually in each place seems awkward; what about doing it in a single future instead?
> 
> For example, we could add this to `crucible-common`:
> 
> ```rust
> pub async fn verbose_timeout(secs: f32, n: usize, log: slog::Logger) {
>     let d = Duration::from_secs_f32(secs);
>     for i in 0..n {
>         tokio::time::sleep(d).await;
>         slog::warn!(log, "timeout {}/{n}", i + 1,);
>     }
> }
> ```
> 
> Then use it in `select!` statements like so:
> 
> ```rust
> tokio::select! {
>     // other branches...
>     _ = verbose_timeout(TIMEOUT_SECS, TIMEOUT_LIMIT, log.clone()) => {
>         bail!("inactivity timeout");
>     }
> ```

I can roll it up that way.
I was considering if we wanted different behavior between the upstairs and
the downstairs, but this does not preclude allowing a difference between
the two side.

downstairs/src/lib.rs
leftwo: Turn out the lossy in this select was not even needed.  The way the `do_work_for()` loop is structured,
it will always run to completion (skipped jobs are put on the end).  So, it's no longer needed to have
lossy make the call it was doing here.

I've removed it in a76fbc1

leftwo:
PR comments (I believe) have been addressed.

How `lossy` works could probably use some attention, but that's not for this PR.

mkeeter (approved):

PR #1199, opened 2024-03-11
Use FIOFFS on illumos instead of fsync (again)

mkeeter:
This is a cleaned-up version of #1148, which has fallen well behind `main` (rebasing was awkward, so I just copy-pasted the relevant code)

It splits the per-extent flush operation into `pre_flush`, `flush_inner`, and `post_flush`, then skips `flush_inner` in favor of the `ioctl` if we're running with the `omicron-build` feature.

PR #1208, opened 2024-03-15
Don't count `Done` bytes for backpressure

mkeeter:
Stemming from [this comment](https://github.com/oxidecomputer/crucible/pull/1199#issuecomment-1998467951), I investigated why #1192 decreased write speeds (from > 700 MiB/s → 550ish for 1 MiB random writes).

It turns out there was a problem with our byte-based backpressure implementation: we were counting all un-retired bytes, including bytes in jobs that were `Done`.

This is nonsensical: the point of backpressure is to normalize write time (as experienced by the Guest) with the actual time that a write takes.  Jobs remain in the `Done` bucket until a flush, but that's unrelated to how long the write itself took (i.e. in our current implementation, changing the flush timeout would change backpressure, which is silly).

In other words, backpressure looks like a sawtooth graph: it drops to a minimal value after each flush, then climbs up as more jobs are issued, completed, and **not** retired (until the next flush).

#1192 makes this worse because it increases the backpressure gain, so the slope of the sawtooth is steeper.  In other words, our backpressure was _always_ unnecessarily high, but this change made the average higher.  Here's the aforementioned sawtooth under slow and fast builds:

Screenshot 2024-03-14 at 6 00 30 PM

This PR changes the byte-based backpressure to only track bytes in flight.  This matches the existing job-based backpressure, which uses the sum of `self.io_state_count.new + self.io_state_count.in_progress` (note the lack of `io_state_count.done` in that summation!).

Sure enough, this improves performance back to > 700 MiB/s.

It's not perfect – doing fine-grained sampling of backpressure, I see that it looks slightly unstable:

Screenshot 2024-03-15 at 10 21 03 AM

Still, this is a noticeable performance improvement and seems philosophically correct.

### DIFF
--- a/upstairs/src/downstairs.rs
+++ b/upstairs/src/downstairs.rs
@@ -2728,15 +2728,7 @@ impl Downstairs {
             }
             // Now that we've collected jobs to retire, remove them from the map
             for &id in &retired {
-                let job = self.ds_active.remove(&id);
-                // Update pending bytes when this job is retired
-                let change = match &job.work {
-                    IOop::Write { data, .. }
-                    | IOop::WriteUnwritten { data, .. } => data.len() as u64,
-                    _ => 0,
-                };
-                self.write_bytes_outstanding =
-                    self.write_bytes_outstanding.checked_sub(change).unwrap();
+                let _ = self.ds_active.remove(&id);
             }
 
             debug!(self.log, "[rc] retire {} clears {:?}", ds_id, retired);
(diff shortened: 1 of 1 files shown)


leftwo (approved):
So, I think this is fine.
We do have to have some way of capping the total amount of memory an upstairs
will require.  But trying to use guest backpressure as a way to do that might
not be the right way.  Especially since the amount of memory we need is more
about how quickly we are flushing ACK'd jobs and that is opaque to the guest.

Maybe in the future we adjust the flush timeout to accomplish that, but that's
not for this PR.

PR #1209, opened 2024-03-15
Activate the NBD server's psuedo file

char:
I don't know if any of you are using nbd_server but this corrects the reported `cpf.sz()` (i.e. not zero) and has let me use a filesystem on crucible via nbd for my own shenanigans

### DIFF
--- a/nbd_server/src/main.rs
+++ b/nbd_server/src/main.rs
@@ -97,8 +97,8 @@ async fn main() -> Result<()> {
 
     // NBD server
 
-    guest.activate().await?;
     let mut cpf = crucible::CruciblePseudoFile::from(guest)?;
+    cpf.activate().await?;
 
     let listener = TcpListener::bind("127.0.0.1:10809").unwrap();
 


leftwo:
Hi @char, 
I tried this change, but it won't compile for me:

```
error[E0382]: borrow of moved value: `guest`
   --> nbd_server/src/main.rs:101:5
    |
93  |     let (guest, io) = Guest::new(None);
    |          ----- move occurs because `guest` has type `crucible::Guest`, which does not implement the `Copy` trait
...
100 |     let mut cpf = crucible::CruciblePseudoFile::from(guest)?;
    |                                                      ----- value moved here
101 |     guest.activate().await?;
    |     ^^^^^ value borrowed here after move
```
Did it work for you?

And, yeah, `nbd_server` may have gotten a little stale.

char:
that's super weird! line 101 should not call `guest.activate()` but `cpf.activate()`.

compiles for me just fine:
screenshot of crucible-nbd-server successfully compiling

leftwo:
Ah, that's my mistake,  I just manually made the change in a local repo and tried it, but I made the wrong
change.  It is building now and I'll run it through the CI test pipeline.

Thanks for opening the PR!

leftwo (approved):

PR #1211, opened 2024-03-18
Various buffer optimizations

mkeeter:
This PR is a set of small cleanups to reduce `memcpy` and `memset` calls in `Buffer` wrangling.

- `Buffer::eat` just swaps pointers if it's a complete (fully owned) overwrite, instead of using `memcpy`
- `Buffer::eat` no longer sets the size of the eaten buffer to 0, to avoid `memset` when scaling it up later
- Added a new `Buffer:reset_owned`, which resizes and clears the `owned` flags but **does not** clear buffer data (for the case where we're about to overwrite buffer data anyways)
    - Use this new function in `Volume::read`
- In `Volume::read`, read directly into the provided buffer if there's only one subvolume (instead of reading into a scratch buffer then copying)
    - The scratch buffer is no longer always allocated; it's resized if needed


upstairs/src/lib.rs
```diff
@@ -1652,15 +1652,23 @@ impl Buffer {
...
+            for (b, (owned, chunk)) in buffer.blocks().enumerate() {
+                if owned {
+                    let block = start_block + b;
+                    self.owned[block] = 1;
+                    self.block_mut(block).copy_from_slice(chunk);
+                }
             }
         }
-
-        buffer.reset(0, self.block_size);
```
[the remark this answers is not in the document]
mkeeter: The function is `pub(crate)`, so it can't be used by anyone outside of the `crucible` crate.  Within the crate, it's only used in `Volume::read` (the target of this optimization!) and a few unit tests, which I audited by hand.

pfmooney (approved):

upstairs/src/lib.rs
```diff
@@ -1639,8 +1639,8 @@ impl Buffer {
 
     /// Consume and layer buffer contents on top of this one
     ///
-    /// The `buffer` argument will be reset to zero size, but preserves its
-    /// allocations for reuse.
+    /// The `buffer` argument will be left in an unknown state, and should be
+    /// reset before reuse.
```
jmpesp, line 1643: Could `eat` consume the argument `Buffer` then? This makes me uncomfortable, as it depends on the caller remembering to do something that isn't checked by the compiler.

mkeeter: I was thinking about that, but I think we want to return the buffer so that it can be reused.  If you look at [`Volume::read`](https://github.com/oxidecomputer/crucible/blob/90b734824a07de2fcf654813bb39bee52d94116e/upstairs/src/volume.rs#L656-L700), we reuse `read_buffer` repeatedly.

I also convinced myself that resetting versus not-resetting doesn't make a difference, because `eat` was resetting the buffer to **size 0** (meaning we'd have to reset it _again_ anyways):

- If `eat` resets the buffer to 0 items, we have to call `Buffer::reset` or `Buffer::reset_owned` before using it again.  Since we're resizing from 0, we have to `memset` every item in both cases.
- If `eat` doesn't reset the buffer, we _still_ have to call `Buffer::reset` or `Buffer::reset_owned` before using it again.  However, `reset_owned` is a little cheaper, because we can skip calling `memset` on bytes in the range `0..previous size`.

mkeeter: Addressed in 9907977ed5bec952f03af0dc977efa8b84d9d52a

upstairs/src/lib.rs
```diff
@@ -1639,15 +1639,19 @@ impl Buffer {
...
     /// - Both buffers must have the same block size
     ///
     /// If either of these conditions is not met, the function will panic
-    pub(crate) fn eat(&mut self, offset: usize, buffer: &mut Buffer) {
+    #[must_use]
+    pub(crate) fn eat(
+        &mut self,
+        offset: usize,
+        mut buffer: Buffer,
+    ) -> UninitializedBuffer {
```
pfmooney, line 1654: I like this explicit pattern much more than what came before.

upstairs/src/lib.rs
```diff
@@ -1784,6 +1772,31 @@ impl std::ops::Deref for Buffer {
...
+#[derive(Debug, Default)]
+pub(crate) struct UninitializedBuffer(Buffer);
+
+impl UninitializedBuffer {
+    /// Resets to the given size, clearing the `owned` array
+    ///
+    /// The data array is resized (padding with 0s) but is not cleared; it is
+    /// expected that this function will be called right before reading into it,
+    /// so clearing it is unnecessary.
+    pub(crate) fn reset_owned(
```
pfmooney, line 1785: naming nit: You could just call this `reset`, considering how it's leaving practically nothing about the buffer untouched

pfmooney (approved):

upstairs/src/lib.rs
mkeeter: I wanted to make it distinct from `Buffer::reset`, which sets all data values to zero; this function leaves data values (below the resizing point) set to their previous values.

pfmooney: Fair enough.

jmpesp (approved):

PR #1217, opened 2024-03-20
Remove `MAX_ACTIVE_COUNT` flow control system

### LINKED ISSUE #1191: A downstairs deadlock is possible during LiveRepair if enough IOs are in flight.
The upstairs starts flow control for a downstairs if MAX_ACTIVE_COUNT IOs have been submitted to
a downstairs.  This means that no more IO will be sent to a downstairs until that downstairs starts
acking some of the outstanding IO we have sent it.

For LR, when we repair an extent, the repair jobs are "barrier" in that they make a dependency that 
could cascade into any job arriving after the repair has started will depend on that repair
completing before the new job can move forward.  In addition, during LR, we will reserve job
IDs, but we won't send all the repair jobs (specifically the NoOP) till we have gotten back the
ack from all three downstairs that the repair IO has finished.  The NoOP is a required part of
the repair and prevents downstairs from being out of sync with each other.

Now, if fwe have a repair, and we have the conditions that new IO will depend on the repair, and
we have enough new IO + current IO in flight to get us to MAX_ACTIVE_COUNT, we have the
potential for deadlock.  All it takes is for us to hit flow control before we send out that repair
NoOp IO.

This will result in the downstairs having unmet dependencies with the current set of jobs it
has to do, but not receiving any new work.

Another way to think about it is that there is a disconnect between job dependencies and
flow control (channel depth). If the dependency list can be greater than the number of 
outstanding IOs, then deadlock can happen.

mkeeter:
Fixes #1191 by removing `MAX_ACTIVE_COUNT`, pushing all work into queues / sockets right away.

We are limited by our total limit on jobs and bytes in flight (`IO_OUTSTANDING_MAX_JOBS/BYTES`), so we can't run away to infinitely long queues.  In addition, backpressure tries to limit queues and bytes to reasonable values (although they must _eventually_ be able to hit `IO_OUTSTANDING_MAX_*`, so that we can detect and fault that Downstairs).

### DIFF
--- a/aws_benchmark/driver.sh
+++ b/aws_benchmark/driver.sh
@@ -34,9 +34,6 @@ cleanup
 # show results
 set -x
 
-# was job based flow control triggered?
-grep "flow control" results.txt || true
-
 # show IOPS measurements
 grep "IOPS mean " results.txt || true
 
--- a/cmon/src/main.rs
+++ b/cmon/src/main.rs
@@ -36,7 +36,6 @@ enum DtraceDisplay {
     LiveRepair,
     Connected,
     Replaced,
-    FlowControl,
     ExtentRepair,
 }
 
@@ -51,7 +50,6 @@ impl fmt::Display for DtraceDisplay {
             DtraceDisplay::LiveRepair => write!(f, "live_repair"),
             DtraceDisplay::Connected => write!(f, "connected"),
             DtraceDisplay::Replaced => write!(f, "replaced"),
-            DtraceDisplay::FlowControl => write!(f, "flow_control"),
             DtraceDisplay::ExtentRepair => write!(f, "extent_repair"),
         }
     }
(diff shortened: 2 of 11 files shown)


leftwo:
I'm seemingly hitting this downstairs deadlock much more often than before.
And I don't think I ever saw it before release 6 landed.  I'm going to bubble this up to
the top of the review pile.

leftwo (approved):
I'm curious as to what happens to memory sizes of the downstairs after this
change, but we know there is an upper limit to jobs, and having a deadlock is
bad and we can't have that.

jmpesp (approved):

mkeeter:
Both upstairs and downstairs should still be limited by our maximum jobs and bytes-in-flight values (57000 and 1 GiB respectively); if we exceed those values, we'll kick out the downstairs (and therefore stop sending it jobs).

PR #1218, opened 2024-03-21
Simplify `BlockReq`

mkeeter:
Our handling of return values from the `BlockOp` has grown rather confusing:

- Every `BlockReq` has a `mpsc::Sender<BlockReqReply>` to notify the caller when it's done, which sends a `BlockReqReply`
  - The `BlockReqReply` contains both a `Result<(), CrucibleError>` and an `Option<Buffer>`, which is only valid for read replies
    - We have a bunch of assertions checking that replies for non-read operations do not contain a buffer
- Any `BlockOp` which wants to return a value (other than a `Buffer`) also contains a `tokio::sync::Mutex<T>` for that type, and uses the `BlockReqReply` as a doorbell instead
- The `struct GtoS` contains an `Option<Buffer>`
  - This field is only valid for reads

In general, there are a bunch of invariants which have to be enforced by user code (and not by the type system).

This PR is an attempt to clean things up, and ends up being a decent LOC decrease.

Here's the idea:

- `BlockRes` is made generic across a `Result<T, E>`
- Every `BlockOp` contains a `BlockRes<..>` for **specifically the type it wants to return**
  - This is often a `Result<(), CrucibleError>`
  - For `BlockOp` that previously contained a `Mutex<T>`, it's a `Result<T, CrucibleError>`
  - For reads, is a `Result<Buffer, (Buffer, CrucibleError)>` because we always want to return the buffer
- The `GtoS` stores a new `enum GuestBlockRes`, which is either a plain `BlockRes<()>` or a buffer plus a `BlockRes` that returns that buffer

This means that our replies from `BlockOp`s are strongly typed and impossible to screw up; if the `BlockOp` is expecting a `Buffer`, you can't forget to populate the `Option<Buffer>` field.


upstairs/src/guest.rs
```diff
@@ -127,20 +90,45 @@ impl GtoS {
...
+                    /*
+                     * Should this panic?  If the caller is requesting a
+                     * transfer, the guest_buffer should exist. If it does not
+                     * exist, then either there is a real problem, or the
+                     * operation was a write or flush and why are we requesting
+                     * a transfer for those.
+                     *
+                     * However, dropping a Guest before receiving a downstairs
+                     * response will trigger this, so eat it for now.
+                     */
```
pfmooney, line 120: style nit: convert this to `//`-style comment, since it's being altered/moved?

upstairs/src/block_req.rs
```diff
@@ -2,69 +2,40 @@
...
     }
 
     /// Consume this BlockRes and send a Result to the receiver
-    fn send_result(
-        mut self,
-        buffer: Option<Buffer>,
-        result: Result<(), CrucibleError>,
-    ) {
+    pub(crate) fn send_result(mut self, result: Result<T, E>) {
         // XXX this eats the result!
```
pfmooney, line 33: Should this XXX comment remain?

upstairs/src/block_req.rs
```diff
@@ -78,53 +49,58 @@ impl Drop for BlockRes {
...
         let (send, recv) = oneshot::channel();
         (Self { recv }, BlockRes(Some(send)))
     }
 
     /// Consume this BlockReqWaiter and wait on the message
     ///
     /// If the other side of the oneshot drops without a reply, log an error
-    pub async fn wait(self, log: &Logger) -> BlockReqReply {
+    /// and return `None`.
+    pub async fn wait_raw(self, log: &Logger) -> Option<Result<T, E>> {
```
pfmooney, line 70: Should this be `pub (crate)`?

upstairs/src/block_req.rs
```diff
@@ -78,53 +49,58 @@ impl Drop for BlockRes {
...
         match self.recv.await {
-            Ok(reply) => reply,
+            Ok(reply) => Some(reply),
             Err(_) => {
                 warn!(
                     log,
                     "BlockReqWaiter disconnected; \
                      this should only happen at exit"
                 );
```
pfmooney, line 78: Is it worth doing this logging so deeply in the `BlockReqWaiter`, or leaving it to something higher up to do it inferred from the resulting `RecvDisconnected` error?  It seems a little weird to me to keep a side effect like logging in a result type like this.  *shrug*

upstairs/src/guest.rs
mkeeter: Sure, done in 515ae69e5d0251bc5062e9a9411f7739a7f85a19

upstairs/src/block_req.rs
mkeeter: It hasn't changed from before, so I decided to leave it.

upstairs/src/block_req.rs
mkeeter: The `BlockReqWaiter` is already `pub(crate)`, so I think `pub` and `pub(crate)` are equivalent here; I don't have a a strong preference for which to use.  `pub(crate)` is more explicit, but redundant + longer.

upstairs/src/block_req.rs
mkeeter: IIRC this was the compromise position in https://github.com/oxidecomputer/crucible/pull/1085 when we removed "panic if dropped without a reply".  I'd be happy to remove it, though!

pfmooney: I guess what I'm asking is: If we choose to continue logging this condition, should it be done this deep in the call stack, considering it has a specific error code (`RecvDisconnected`) dedicated to it, which could be printed later.

Either way, it's not something to get wrapped around the axle over.

upstairs/src/block_req.rs
pfmooney: I asked because `BlockRes` seemed to have a mix, despite the struct itself being `pub(crate)`.

Whatever makes sense to you works for me.

upstairs/src/guest.rs
```diff
@@ -615,22 +549,36 @@ impl BlockIO for Guest {
...
                 data: chunk,
+                done,
             };
 
-            let reply = self.send_and_wait(rio).await;
-
-            // Reattach the chunk to `data`
-            //
-            // [---data---][--------buffer-------]
-            data.unsplit(reply.buffer.unwrap());
```
leftwo, line 628: Seeing an `unwrap()` go away makes me happy.

leftwo (approved):
Thanks for the clarifying comments in chat as well.

upstairs/src/block_req.rs
```diff
@@ -2,69 +2,40 @@
...
-/**
- * Couple a BlockOp with a notifier for calling code. This uses a single-use
- * channel to send the result of a particular operation, and is meant to be
- * paired with a BlockReqWaiter.
- */
-#[must_use]
+/// Wrapper to contain a `BlockOp`
+///
+/// The `BlockReq` used to have more fields, but now it's down to just this one;
+/// we can remove it at a later point.
```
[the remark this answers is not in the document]
mkeeter: It made the diff terrible to review 😄 

I can do the removal in a mostly mechanical follow-up PR, but didn't want to mix it with logic changes.

upstairs/src/guest.rs
```diff
@@ -127,20 +90,43 @@ impl GtoS {
...
+                                    .entered();
+
+                            buffer.write_read_response(offset, response);
+                            offset += response.data.len();
+                        }
                     }
-                    None => res.send_err(e),
-                },
+                } else {
+                    // Should this panic?  If the caller is requesting a
```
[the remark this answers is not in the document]
mkeeter: This is a pre-existing comment, so I didn't think too hard about it:
https://github.com/oxidecomputer/crucible/blob/89662ead77a0ad4c278677253e2d46ffe8f94830/upstairs/src/guest.rs#L106-L114

upstairs/src/block_req.rs
mkeeter: Good point, I switched the `BlockRes::send_result` to `pub` in b63943358734ae03bbd7f6963f96ad3275d0f3cd

upstairs/src/block_req.rs
mkeeter: I think returning the error code is sufficient, and removed the logging in 3a54ada1d1aed63e293555af29097cfae3d0f7ea.

upstairs/src/block_req.rs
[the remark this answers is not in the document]
leftwo: Yeah, makes sense.

upstairs/src/lib.rs
```diff
@@ -1850,57 +1850,66 @@ fn test_buffer_eats() {
  * tell the upstairs to behave in specific ways.
  */
 #[derive(Debug)]
-pub enum BlockOp {
+pub(crate) enum BlockOp {
```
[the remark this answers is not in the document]
mkeeter: Updated in ce0473672719e2261d3dc5143d82a6a88f8380bd

upstairs/src/guest.rs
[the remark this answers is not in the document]
leftwo: It may have been out of date before this change came along.
In this path, the buffer has to exist, so the comment can't be
true with regards to that.  I also think even before that, there
were problems here.  I can come in later an fix this to reflect reality.

I think the `else` here is only possible if we had no downstairs_responses
then the IO has failed in some way where we did not get a response
at all. 

I suspect if we failed all three reads we could not have any
downstairs_responses.  In that case we could verify that the result
was an error.  I'll ponder it a bit more to see if there is any condition
where we are not returning an error, but don't have any downstairs_response.

upstairs/src/block_req.rs
```diff
@@ -78,53 +49,51 @@ impl Drop for BlockRes {
...
     /// Create associated `BlockReqWaiter`/`BlockRes` pair
-    pub fn pair() -> (BlockReqWaiter, BlockRes) {
+    pub fn pair() -> (Self, BlockRes<T, E>) {
         let (send, recv) = oneshot::channel();
         (Self { recv }, BlockRes(Some(send)))
     }
 
     /// Consume this BlockReqWaiter and wait on the message
     ///
     /// If the other side of the oneshot drops without a reply, log an error
```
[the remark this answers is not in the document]
mkeeter: Good catch, fixed in ce16698e0b2dfb8b4b1aa4f8a8c6785c4e0950e2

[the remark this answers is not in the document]
jmpesp: It could be useful to define another probe, like we do for `impl Drop for BlockRes`, for when this case occurs?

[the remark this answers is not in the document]
mkeeter: Oops, I didn't get this notification until after hitting merge.  I'm not too worried about this, and would be fine building more debug infrastructure only when needed.

PR #1233, opened 2024-03-26
`crutest` cleaning; adding rand-read/write tests

mkeeter:
This is a grab-bag of `crutest` improvements and features, with the goal of making benchmarking and performance evaluation easier.

# New features
Adding `rand-read` and `rand-write` workloads.  This is meant to mimic the `fio` workload, but running outside of the Propolis VM to remove one layer of abstraction.

# Medium optimization
Refactor `fill_workload` and `verify_volume` to use 8x worker tasks and run in parallel.  This is a significant speedup: for a 32 GiB region, running fill + verify goes from 7:47 down to 1:35.

# Minor tweaks
- Use `&Guest` instead of `Arc<&Guest>` in function arguments where possible
- Use `default_value_t` instead of `default_value` so that default arguments are checked at compile-time
- Make some functions take `&self` instead of `&mut self`
- Delete unused functions (`RegionInfo::set_wc_min`)
- Replace `Vec<Future<..>>` with `FuturesOrdered` in a few places, and remove our custom `join_all` function

### DIFF
--- a/crutest/src/main.rs
+++ b/crutest/src/main.rs
@@ -1,5 +1,6 @@
 // Copyright 2023 Oxide Computer Company
 use std::fs::File;
+use std::io::Write;
 use std::net::{IpAddr, SocketAddr};
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
@@ -10,6 +11,7 @@ use clap::Parser;
 use csv::WriterBuilder;
 use futures::stream::FuturesOrdered;
 use futures::StreamExt;
+use human_bytes::human_bytes;
 use indicatif::{ProgressBar, ProgressStyle};
 use rand::prelude::*;
 use rand_chacha::rand_core::SeedableRng;
@@ -86,6 +88,20 @@ enum Workload {
         /// Number of write test loops to do.
         #[clap(long, default_value_t = 2, action)]
         write_loops: usize,
+    },
+    RandRead {
+        /// Size in blocks of each IO
+        #[clap(long, default_value_t = 1, action)]
+        io_size: usize,
+        /// Number of outstanding IOs at the same time.
+        #[clap(long, default_value_t = 1, action)]
+        io_depth: usize,
+        /// Number of seconds to run
+        #[clap(long, default_value_t = 60, action)]
+        time: u64,
+        /// Fill the region with random data first
+        #[clap(long)]
+        fill: bool,
     },
     Repair,
     /// Test the downstairs replay path.
@@ -590,6 +606,7 @@ async fn handle_signals(
 #[tokio::main]
 async fn main() -> Result<()> {
     let opt = opts()?;
+    let is_encrypted = opt.key.is_some();
 
     // If we just want the version, print that and exit.
     if let Workload::Version = opt.workload {
@@ -903,6 +920,26 @@ async fn main() -> Result<()> {
                 io_size,
                 write_loops,
                 read_loops,
+            )
+            .await?;
+            if opt.quit {
+                return Ok(());
+            }
+        }
+        Workload::RandRead {
+            io_size,
+            io_depth,
+            time,
+            fill,
+        } => {
+            rand_read_workload(
+                &guest,
+                &mut region_info,
+                is_encrypted,
+                io_depth,
+                io_size,
+                time,
+                fill,
             )
             .await?;
             if opt.quit {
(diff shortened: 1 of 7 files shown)


crutest/src/main.rs
```diff
@@ -2185,11 +2351,149 @@ async fn perf_workload(
...
+            human_bytes(delta as f64),
+            delta / (cfg.blocks_per_io * block_size),
+            cfg.time_secs - i,
+        );
+        std::io::stdout().lock().flush().unwrap();
+    }
+    stop.store(true, Ordering::Release);
+
+    let bytes = byte_count.load(Ordering::Acquire);
+    println!(
```
leftwo, line 2483: This final print can sometimes get lost if there is still IO outstanding when the
timer elapses.  Might be a little nicer if the test paused for the queues to
empty before printing out a summary?  I ran this with an `io-depth` of 16 and I
got a bunch of `WARN exiting client task is ignoring message` message which
buried my results.

crutest/src/main.rs
```diff
@@ -2185,11 +2351,149 @@ async fn perf_workload(
...
+                            .unwrap();
+                        byte_count.fetch_add(buf.len(), Ordering::Relaxed);
+                    }
+                }
+            }
+            Ok::<(), CrucibleError>(())
+        });
+    }
+
+    let mut prev = 0;
```
leftwo, line 2465: I don't think it matters really for the ways we will be using this test, but the first
second of time elapsed, if I'm reading this correctly, will count not only all the IO
that happened in that first second, but also include all the IO generated duing the setup
loop as well.  Basically all the IO done by tasks `0..cfg.io_depth` as we spun them out.

It probably only going to make a difference if you had a large number of tasks with a
short test duration.  Given the default of 60 seconds, it's probably not going to move
the needle much.

crutest/src/main.rs
```diff
@@ -2185,11 +2351,149 @@ async fn perf_workload(
...
+    if cfg.fill {
+        println!("filling region before {desc}");
+        fill_workload(guest, ri, true).await?;
+    }
+
+    if cfg.blocks_per_io > ri.total_blocks {
+        bail!("too many blocks per IO; can't exceed {}", ri.total_blocks);
+    }
+
+    println!(
```
leftwo, line 2404: I notice that often the upstairs `slog` messages can clobber this preamble and the
work you did to display the region info:
```
Transition from Reconcile to Active, client: 1, : downstairs, session_id: 04466381-d5e9-4b36-9cea-7d96cbd2de71
Mar 27 00:07:15.698 INFO Transition from Reconcile to Active, client: 2, : downstairs, session_id
-----------------------------------
: random write with 512 B chunks (1 block)
04466381-d5e9-4b36-9cea-7d96cbd2de71
Mar 27 00:07:15.698region info:
  block size:      512 bytes
   blocks / extent: 256
  extent size:     128 KiB
INFO  extent count:    2
  total blocks:    512
   total size:      256 KiB
All required reconciliation work is completed,   encryption:      no
session_id: 04466381-d5e9-4b36-9cea-7d96cbd2de71
Mar 27 00:07:15.698 INFO Set Downstairs and Upstairs active after reconciliatio, session_id: 04466381-d5e9-4b36-9cea-7d96cbd2de71
```

I had added at some point a wait for `guest.query_is_active()` to come back, but it seems like even that
is not waiting long enough to avoid `slog` clobbering things.  I'm loathe to suggest a sleep..

crutest/src/main.rs
```diff
@@ -2185,11 +2351,149 @@ async fn perf_workload(
...
+        let stop = stop.clone();
+        let guest = guest.clone();
+        let byte_count = byte_count.clone();
+        tokio::spawn(async move {
+            match cfg.mode {
+                RandReadWriteMode::Write => {
+                    let mut buf = BytesMut::new();
+                    buf.resize(cfg.blocks_per_io * block_size, 0u8);
+                    let mut rng = rand_chacha::ChaCha8Rng::from_entropy();
+                    rng.fill_bytes(&mut buf);
```
leftwo, line 2428: If you generate and fill the buffer inside the loop, won't that get counted in the
performance results?  I don't know if we care really that much, or how much time we
actually spend in the allocator.

leftwo (approved):
Nice, and it seems like we could update this without too much work to do
some mixed workloads too (like 50/50 RW or something like that)

Just some thoughts, but nothing really required, take them and do  with
them what you want.

crutest/src/main.rs
leftwo: Oh, I think we might want to require this.

If I have a bunch of IO outstanding, it's possible to exit the test early
with IO inflight such that some downstairs did get the IO and some did
not.  Also, if any extents are dirty, we don't actually know what they
have. All this means the next time you try to use that region set, it will have to
reconcile the three downstairs.

I did it with a region like this:
```
./target/release/dsc start --create --cleanup --extent-size=16 --extent-count=1024 --block-size=4096
```

Then, running a write workload to it.

leftwo (commented):
Oh, I did find something that we should fix

crutest/src/main.rs
```diff
@@ -71,21 +77,49 @@ enum Workload {
...
         /// Number of read test loops to do.
-        #[clap(long, default_value = "2", action)]
+        #[clap(long, default_value_t = 2, action)]
         read_loops: usize,
         /// Number of write test loops to do.
-        #[clap(long, default_value = "2", action)]
+        #[clap(long, default_value_t = 2, action)]
         write_loops: usize,
     },
+    RandRead {
```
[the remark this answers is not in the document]
mkeeter: good catch, done in b8fb362088f30dc5f3fa126f364ea09a2bcca838

crutest/src/main.rs
mkeeter: I decided to mute the guest log by default for these new commands, with a `--verbose` CLI flag if you really want to see it; updated in 45c3c243df599036d239889a4e1d2dfa2e921174

(I left the existing commands unchanged, because I'm not sure if anyone is relying on their current behavior)

crutest/src/main.rs
mkeeter: This is outside the IO loop, which begins at `while !stop.load(..)` 😉

crutest/src/main.rs
```diff
@@ -297,6 +279,25 @@ fn history_file<P: AsRef<Path>>(file: P) -> PathBuf {
...
+    /// Number of outstanding IOs at the same time.
+    #[clap(long, default_value_t = 1, action)]
+    io_depth: usize,
+    /// Number of seconds to run
+    #[clap(long, default_value_t = 60, action)]
+    time: u64,
+    /// Completely fill the region with random data first
+    #[clap(long)]
+    fill: bool,
+    /// Print the guest log to `stderr`
```
leftwo, line 296: This changes the default logging for the upstairs to log only errors?  Or does
it still print all messages but only to stderr?  It suggests the latter, but the code
below makes me think we are only printing error level logs?

crutest/src/main.rs
```diff
@@ -1194,40 +1198,31 @@ async fn verify_volume(
...
         let guest = guest.clone();
         let write_log = write_log.clone();
-        let bytes_done = bytes_done.clone();
+        let blocks_done = blocks_done.clone();
         let total_blocks = ri.total_blocks;
         let block_size = ri.block_size;
         let pb = pb.clone();
         tasks.push(tokio::task::spawn(async move {
             let mut result = Ok(());
+            let mut data = crucible::Buffer::new(0, block_size as usize);
```
leftwo, line 1245: Does this mean the initial buffer will be filled with 0's?
I believe I had 255 there for some sort of canary, but I no longer
remember what the reason was I put that fence up..

crutest/src/main.rs
mkeeter: Yeah, I added some code to skip the first data sample (af4b3aa2d33e445425d1e899f59998ebf0c7a75a)

crutest/src/main.rs
mkeeter: No, this means that the initial buffer has a size of 0 blocks; it's always reset before use.

(Admittedly, the reset buffer _will_ be filled with 0s, but after the read, we validate its data against the extremely-not-zero expected pattern, so I don't think we need to guard against it)

crutest/src/main.rs
leftwo: I think being able to debugging the other tests when they fail in CI or elsewhere probably wants the
upstairs output, so that's good.  I may add a "log the upstairs to this file" option as well, as I really
don't need to see the upstairs logs 90% of the time.

crutest/src/main.rs
leftwo: Ah yeah, got it.

crutest/src/main.rs
leftwo: A blank region will be all zeros, but really we need a few things wrong here to ever have
that situation.

PR #1235, opened 2024-03-28
test-replay.sh fix name of DTrace script

leftwo:
This was failing with:

```
dtrace: failed to open /input/build/work/scripts/upstairs-info.d: No such file or directory
```

### DIFF
--- a/.github/buildomat/jobs/test-replay.sh
+++ b/.github/buildomat/jobs/test-replay.sh
@@ -58,7 +58,7 @@ iostat -T d -xn 1 > /tmp/debug/iostat.txt 2>&1 &
 mpstat -T d 1 > /tmp/debug/mpstat.txt 2>&1 &
 vmstat -T d -p 1 < /dev/null > /tmp/debug/paging.txt 2>&1 &
 pfexec dtrace -Z -s $input/scripts/perf-downstairs-tick.d > /tmp/debug/perf.txt 2>&1 &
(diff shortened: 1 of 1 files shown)


mkeeter:
The fix seems reasonable, but why was this failure not showing up in the Github Actions report?

mkeeter (approved):

leftwo:
> The fix seems reasonable, but why was this failure not showing up in the Github Actions report?

I think some combo of the redirection and backgrounding of it hid the failure from the calling script.
I can reproduce it locally as well:

```
alan@atrium:crucible$ cat fff.sh
#!/bin/bash
#

set -o errexit
set -o pipefail
set -o xtrace

echo "Starting fake test"
mkdir /tmp/debug
pfexec dtrace -Z -s /scripts/upstairs-info.d > /tmp/debug/upinfo.txt 2>&1 &

echo "Run fake test now"
sleep 5
echo "all done"
```

Which "works" even though it fails.

```
alan@atrium:crucible$ ./fff.sh
+ echo 'Starting fake test'
Starting fake test
+ mkdir /tmp/debug
+ echo 'Run fake test now'
Run fake test now
+ sleep 5
+ pfexec dtrace -Z -s /scripts/upstairs-info.d
+ echo 'all done'
all done
alan@atrium:crucible$ more /tmp/debug/upinfo.txt 
dtrace: failed to open /scripts/upstairs-info.d: No such file or directory
```

PR #1238, opened 2024-03-30
[meta] update omicron-common and related deps

sunshowers:
Picks up the fix for https://github.com/oxidecomputer/security-operations/issues/43.


Cargo.lock
```diff
@@ -4174,7 +4189,8 @@ dependencies = [
  "schemars_derive",
  "serde",
  "serde_json",
- "uuid",
+ "uuid 0.8.2",
+ "uuid 1.8.0",
```
mkeeter, line 4193: Any idea why `schemars` is depending on two different `uuid` versions?

mkeeter: [lol here we are](https://github.com/GREsau/schemars/blob/master/schemars/Cargo.toml#L26C1-L27)

bnaecker (approved):

workspace-hack/Cargo.toml
```diff
@@ -77,15 +74,13 @@ phf_shared = { version = "0.11" }
 rand = { version = "0.8", features = ["min_const_gen", "small_rng"] }
 rand_chacha = { version = "0.3" }
 rand_core = { version = "0.6", default-features = false, features = ["std"] }
-schemars = { version = "0.8", features = ["bytes", "chrono", "uuid1"] }
+schemars = { version = "0.8", features = ["bytes", "chrono", "uuid", "uuid1"] }
```
mkeeter, line 77: Is there an easy way to use `cargo hakari` to figure out where the `uuid` feature is being added?  We only enable `uuid1` in Crucible's `Cargo.toml`.

bnaecker: I think you can use `cargo tree -e features` to see which features are enabled on each dependency.

sunshowers: Oh I fixed this in maghemite: https://github.com/oxidecomputer/maghemite/pull/191

PR #1244, opened 2024-04-01
Update oximeter dependency

bnaecker:


bnaecker:
This is to pull in https://github.com/oxidecomputer/omicron/pull/5340, which is a pre-req to making larger changes to the producer internals. See https://github.com/oxidecomputer/omicron/issues/5284#issuecomment-2024112759 for background.

bnaecker:
It feels like every time I update any of {omicron,crucible,propolis}, I get mired in a circular dependency around the client interfaces. I think this is partially due to still using Git deps for Progenitor, which pull in slightly different versions of transitive dependencies like `reqwest`, but I'm not positive.

workspace-hack/Cargo.toml
```diff
@@ -43,7 +42,7 @@ rand = { version = "0.8", features = ["min_const_gen", "small_rng"] }
 rand_chacha = { version = "0.3" }
 rand_core = { version = "0.6", default-features = false, features = ["std"] }
 reqwest = { version = "0.11", features = ["blocking", "json", "rustls-tls", "stream"] }
-schemars = { version = "0.8", features = ["bytes", "chrono", "uuid1"] }
+schemars = { version = "0.8", features = ["bytes", "chrono", "uuid", "uuid1"] }
```
ahl, line 45: I don't think you should specify both of these feature flags


... drives off....

bnaecker: 😑

leftwo (approved):

PR #1247, opened 2024-04-03
Dummy downstairs cleanup

mkeeter:
I want `dummy_downstairs.rs` to be easier to use, since it's a great way to do end-to-end testing.

In the current code, the `FramedRead/Write` are shared by both the main (test) task and IO worker task; this shared state makes the code harder to follow, because there are `lock().await`s everywhere.  The division of labor is also unclear: in some cases, the test task will write to those streams directly; in other cases, it will use `mpsc` channels to the worker task.

This PR refactors the IO worker task to **own** the `FramedRead/Write`; the test task now **always** communicates to it via `mpsc` channels, and cannot read or write to the streams directly.

This removes a bunch of locks, for a significant LOC decrease.

It also removes `bail_assert!` and `bail_assert_eq!` in favor of their normal counterparts.

### DIFF
--- a/upstairs/src/dummy_downstairs_tests.rs
+++ b/upstairs/src/dummy_downstairs_tests.rs
@@ -137,79 +137,71 @@ pub(crate) mod protocol_test {
 
         pub async fn negotiate_start(&mut self) -> Result<()> {
             let packet = self.recv().await.unwrap();
-            match &packet {
-                Message::HereIAm {
-                    version,
-                    upstairs_id: _,
-                    session_id: _,
-                    gen: _,
-                    read_only,
-                    encrypted: _,
-                    alternate_versions: _,
-                } => {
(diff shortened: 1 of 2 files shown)


upstairs/src/dummy_downstairs_tests.rs
```diff
@@ -1,9 +1,9 @@
 // Copyright 2023 Oxide Computer Company
 
-#[cfg(test)]
+#[cfg(not(test))]
+compile_error!("dummy_downstairs should only be used in unit tests");
+
 pub(crate) mod protocol_test {
```
mkeeter, line 6: After the PR is approved, I'll push a final (purely mechanical) commit which removes `mod protocol_test`, saving a level of indentation.

jmpesp (approved):

PR #1249, opened 2024-04-05
Add a clone option to downstairs create

leftwo:
When creating a region, add the ability to optionally populate that region from an existing and running downstairs using the clone ability.  

This makes use of the already written clone subcommand functions, but calls them directly after creating the region.

The current `clone` subcommand still exists and we can still use it directly if so desired.

### DIFF
--- a/integration_tests/src/lib.rs
+++ b/integration_tests/src/lib.rs
@@ -38,6 +38,7 @@ mod test {
     }
 
     impl TestDownstairs {
+        #[allow(clippy::too_many_arguments)]
         pub async fn new(
             address: IpAddr,
             encrypted: bool,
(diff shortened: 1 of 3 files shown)


leftwo:
Ahh, let me fix the `clap` arg option that is making everything fall apart.

integration_tests/src/lib.rs
```diff
@@ -2657,6 +2667,9 @@ mod test {
         // Clone a read only downstairs to the new downstairs
         // Make a new volume with two old and the one new downstairs and verify
         // the original data we wrote.
+        // Create a new downstairs directly from the clone.
```
jmpesp, line 2670: pasted from somewhere else?

jmpesp (approved):

leftwo:
Also updated all the `NAME` to `VALUE_NAME` for clap.

integration_tests/src/lib.rs
leftwo: No, that I really added.
This test did it the old way, where it made the region first, then did the clone.
I added this section that does the create+clone in one step.
(then does the same test we did for the "create then clone" test.)

jmpesp: Ah, I get it now - these are steps the test takes

Issue #502, opened 2022-11-05
Update some tests to not return `Result<()>` and use unwrap() instead of `?`


leftwo:
With tests that return `Result<()>`, a failure can result in a non specific message like this:

```
thread 'test::integration_test_region' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/65c55bf931a55e6b1e5ed14ad8623814a7386424/library/test/src/lib.rs:195:5
```

That may be fine for a simple test that only does one thing, but many crucible tests do a bunch
of stuff, include calling out to places that can fail.  If we change these tests to `.unwrap()` instead
of `?` and don't return `Resutl<()>` from the test, we can get more meaningful errors, like this:

```
thread 'test::integration_test_region' panicked at 'called `Result::unwrap()` on an `Err` value: Block size mismatch', integration_tests/src/lib.rs:232:66
```

PR #1251, opened 2024-04-08
Panic instead of returning errors in unit tests

mkeeter:
This is both shorter and gives a backtrace to the exact line of the error.

### DIFF
--- a/upstairs/src/dummy_downstairs_tests.rs
+++ b/upstairs/src/dummy_downstairs_tests.rs
@@ -3,6 +3,7 @@
 #[cfg(not(test))]
 compile_error!("dummy_downstairs should only be used in unit tests");
 
+use std::collections::VecDeque;
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


upstairs/src/dummy_downstairs_tests.rs
```diff
@@ -128,7 +126,7 @@ impl DownstairsHandle {
         self.tx.send(m)
     }
 
-    pub async fn negotiate_start(&mut self) -> Result<()> {
+    pub async fn negotiate_start(&mut self) {
```
leftwo, line 129: Since this is a sub-function, and not a test itself, I thought would it better for this to 
return error and have the calling funciton panic.  If it panics here, then I get a 
`line 149: read only mismatch`, but then I have to figure out where in my test I called `
negotiate_start()`.  However, after going through the places that call
`negotiate_start()`, it does seem pretty obious where they come from so I think
it's okay that we just panic in this method.

upstairs/src/dummy_downstairs_tests.rs
```diff
@@ -457,29 +448,28 @@ impl TestHarness {
...
+                ds
             });
         }
 
-        let ds1 = fut.next().await.unwrap().unwrap();
-        let ds2 = fut.next().await.unwrap().unwrap();
-        let ds3 = fut.next().await.unwrap().unwrap();
+        let ds1 = fut.next().await.unwrap();
+        let ds2 = fut.next().await.unwrap();
+        let ds3 = fut.next().await.unwrap();
```
leftwo, line 472: If we take out the `new_` returning result, is it still obvious from the stack which
of these ones fails?  (assuming I'm correct in that the change to `new_` is called
here.

leftwo (commented):
Fixes: https://github.com/oxidecomputer/crucible/issues/502 ?

upstairs/src/dummy_downstairs_tests.rs
mkeeter: Also, we get a full backtrace with `RUST_BACKTRACE=1` 😉

upstairs/src/dummy_downstairs_tests.rs
mkeeter: It looks like the `FuturesOrdered` will panic if any of the members panic, so the line number won't necessarily be meaningful.

It's debatable whether it matters, but c11aaf98b2b5d3500a9079133f622aa7441426a5 switches to checking the futures one by one, which _does_ give meaningful line numbers.

leftwo (approved):

Issue #1252, opened 2024-04-09
Only kick a Downstairs on timeout


mkeeter:
Right now, there are several ways a Downstairs can be kicked out of the `Active` state

- If there are more than 57K jobs or 1 GiB in flight, it's moved to `Faulted`, which means it goes through live-repair when reconnecting
- If it doesn't reply for 45 seconds, it's moved to `Offline`, which means it goes through replay when reconnecting
  - Note that it can _later_ be moved from `Offline` to `Faulted` if it hits either 57K jobs / 1 GiB bytes-in-flight criteria

(there are other paths, e.g. returning an IO error or having a connection close, but those aren't relevant to the issue)

The jobs / bytes-in-flight criteria is uncomfortably coupled to backpressure: if the maximum backpressure is less than the true job completion time, then jobs will inevitably accumulate and will eventually cause the Downstairs to be faulted.  In other words, we have to **guess** a value for maximum backpressure such that the system will never exceed it during normal operation.

We've gone through several rounds of tuning maximum byte-based backpressure (#1240, #1243), and are now seeing (https://github.com/oxidecomputer/customer-support/issues/121) that job-based backpressure may also be insufficient.

After talking about this in the **2024-04-01 Crucible Huddle**, our new plan is to remove the jobs / bytes-in-flight criteria for the `Active` → `Faulted` transition.  In other words, as long as a Downstairs is replying (however slowly), we won't mark it as faulted.  The `Offline` → `Faulted` transition will remain in place, so that we don't accumulate jobs forever.

This _theoretically_ means that the Upstairs can buffer arbitrary amounts of data on behalf of a pathologically-slow Downstairs; however, with quadratic backpressure, we can do the math and satisfy ourselves that it won't ever hit worrying values.

PR #1253, opened 2024-04-09
Reorganize dummy downstairs tests

mkeeter:
This is more prep for addressing #1252 

Previously, our live-repair tests were checking two different things in a single test:

- Sending 57K jobs caused a Downstairs to be marked as faulted
- Once faulted, various live-repair operations occurred

There's also a separate test, `test_byte_fault_condition` (#1192), to confirm that sending > 1 GiB caused a Downstairs to be marked as faulted.

This PR splits those two checks into separate unit tests, adding new (test-only) APIs to (1) check Downstairs state and (2) manually fault a Downstairs.  We want these checks to be separate because – once we make the changes for #1252 – sending 57K jobs will _no longer_ cause an active downstairs to be marked as faulted.

In addition, common `TestHarness` code is consolidated into helper functions:

- Clone the `Guest` and `tokio::spawn` → `TestHarness::spawn`
- Expect a particular message type and reply to it →`DownstairsHandle::ack_read`, `ack_write`, `ack_flush`

### DIFF
--- a/upstairs/src/dummy_downstairs_tests.rs
+++ b/upstairs/src/dummy_downstairs_tests.rs
@@ -2142,9 +2142,6 @@ async fn test_deactivate_slow() {
 
     // At this point, the upstairs sends YouAreNoLongerActive, but then
     // drops the sender end, so we can't actually see it.
-    //
-    // TODO is this correct?  Shouldn't values in the mpsc queue be read out,
-    // even if the sender has been dropped?
     assert!(harness.ds1().recv().await.is_none());
 
     // Reconnect ds1, since the Downstairs always tries to reconnect
--- a/upstairs/src/dummy_downstairs_tests.rs
+++ b/upstairs/src/dummy_downstairs_tests.rs
@@ -13,6 +13,7 @@ use crate::BlockContext;
 use crate::BlockIO;
 use crate::Buffer;
 use crate::CrucibleError;
+use crate::DsState;
 use crate::{IO_OUTSTANDING_MAX_BYTES, IO_OUTSTANDING_MAX_JOBS};
 use crucible_client_types::CrucibleOpts;
 use crucible_common::Block;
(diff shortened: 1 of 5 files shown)


upstairs/src/dummy_downstairs_tests.rs
```diff
@@ -2380,75 +2122,30 @@ async fn test_deactivate_slow() {
...
-        result: Ok(()),
-    };
-    harness.ds1().send(response.clone()).unwrap();
+    harness.ds1().ack_flush().await;
 
     // At this point, the upstairs sends YouAreNoLongerActive, but then
     // drops the sender end, so we can't actually see it.
+    //
+    // TODO is this correct?  Shouldn't values in the mpsc queue be read out,
+    // even if the sender has been dropped?
```
jmpesp, line 2148: I don't think our `ClientIoTask` drains the requests before stopping: a quick glace at client.rs and it looks like landing in a `self.stop` branch breaks out of the task?

mkeeter: That makes sense, I forgot that the `ClientIoTask` drops values while exiting.

mkeeter: (removed the comment in a64ce0c5124e7668d876634fbc2b2d675665fbcd)

jmpesp (approved):

leftwo (approved):

upstairs/src/dummy_downstairs_tests.rs
```diff
@@ -259,6 +260,81 @@ impl DownstairsHandle {
...
     }
+
+    /// Awaits a `Message::Flush { .. }` and sends a `FlushAck`
+    ///
+    /// Returns the flush number for further checks.
+    ///
+    /// # Panics
+    /// If a non-flush message arrives
+    pub async fn ack_flush(&mut self) -> u64 {
+        let Message::Flush {
```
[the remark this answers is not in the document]
mkeeter: Take a look at the [implementation of `recv`](https://github.com/oxidecomputer/crucible/blob/a64ce0c5124e7668d876634fbc2b2d675665fbcd/upstairs/src/dummy_downstairs_tests.rs#L85-L101) 😉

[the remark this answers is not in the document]
leftwo: thats cheating

Issue #1241, opened 2024-03-31
upstairs DTrace probes need unique identifier for each upstairs.


leftwo:
If a propolis-server process has multiple virtual disks, it will have a different upstairs
running for each one.  It's difficult to determine which set of dtrace probes apply
to which upstairs.  We need some method to tell which probes are for which volume.

leftwo:
Done in #1254

PR #1254, opened 2024-04-09
Add Upstairs session ID to dtrace stat probe, cleanup closure

leftwo:
The one task refactor opened the door to lockless stat gathering for the
upstairs.  Since we no longer need to grap locks to gather this data, we can
stuff it all inside the closure.  This should mean we only pay the stat collecting
tax when the probe is actually enabled.

Added another field to the stat probe that records the session ID.  This will be
unique not only per PID, but also for each Upstairs that could be running inside
a single PID.  This is specifically useful when propolis has more than one virtual
disk attached to the same instance.

### DIFF
--- a/upstairs/src/lib.rs
+++ b/upstairs/src/lib.rs
@@ -1595,6 +1595,8 @@ async fn test_return_iops() {
  */
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Arg {
+    /// Upstairs session UUID
+    pub session_id: String,
     /// Jobs on the upstairs guest work queue.
     pub up_count: u32,
     /// Apply loop counter
--- a/upstairs/src/upstairs.rs
+++ b/upstairs/src/upstairs.rs
@@ -721,63 +721,51 @@ impl Upstairs {
 
     /// Fires the `up-status` DTrace probe
     fn on_stat_update(&self) {
-        let up_count = self.guest.guest_work.len() as u32;
-        let up_counters = self.counters;
-        let next_job_id = self.downstairs.peek_next_id();
-        let ds_count = self.downstairs.active_count() as u32;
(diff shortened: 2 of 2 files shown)


leftwo:
A possible fix for https://github.com/oxidecomputer/crucible/issues/1241

upstairs/src/lib.rs
```diff
@@ -1595,6 +1595,8 @@ async fn test_return_iops() {
  */
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Arg {
+    /// Upstairs session UUID
+    pub session_id: String,
```
bnaecker, line 1599: It's not really important, but you can keep this a UUID if you'd prefer. It'll be `serde`-serialized in that case.

upstairs/src/upstairs.rs
```diff
@@ -721,63 +721,51 @@ impl Upstairs {
...
-                ds_live_repair_completed,
-                ds_live_repair_aborted,
-                ds_connected,
-                ds_replaced,
-                ds_extents_repaired,
-                ds_extents_confirmed,
-                ds_extent_limit,
-                ds_delay_us,
-                ds_ro_lr_skipped,
+                session_id: self.cfg.session_id.to_string(),
```
bnaecker, line 726: Yay for is-enabled probes!

bnaecker (approved):
Looks great!

upstairs/src/upstairs.rs
ahl: hah. indeed!

ahl: And agfreed, this can stay a uuid, but whatever..

ahl (commented):
ship it

PR #1256, opened 2024-04-10
Switch to unbounded queues

mkeeter:
This is another standalone prep PR for #1252

We are planning to use backpressure alone to limit queue lengths for an `Active` downstairs, removing the `IO_MAX_ACTIVE_JOBS/BYTES` fault condition; using bounded queues doesn't make sense any more (and gives a false sense of security).

Using unbounded queues is more honest about our strategy, and also removes `async` from large chunks of the codebase.  The latter isn't done in this PR (to avoid churn), but once `DownstairsClient::send` is made sync, it will ripple through a bunch of places!


jmpesp (approved):

leftwo (approved):
Release the queues!

PR #1257, opened 2024-04-11
Add Active -> Offline -> Faulted tests

mkeeter:
This is (hopefully) the final prep PR before addressing #1252.

There was an untested path of `Active` -(timeout)-> `Offline` -(too many jobs)-> `Faulted`; this PR adds tests for both the job and byte-based fault conditions.  In addition, it confirms that live-repair works after all fault-inducing tests, moving that logic to a new `async fn run_live_repair(mut harness: TestHarness)`.

Making these tests pass required fixing a (probably innocuous) race condition:

- Downstairs is marked as offline, begins to reconnect (with a 10s pause)
- Too many jobs accumulate, and we try to stop the downstairs (to restart it again)
- `ClientIoTask::run_inner` ignores the `stop` message during the initial 10s pause
- After the pause completes, the client IO task connects then immediately disconnects, because it finally looks at the stop message

I think this was only an issue in unit tests, where we manage reconnections by hand; in the real system, the second reconnection should work fine.  Anyhow, I fixed it by adding the `stop` condition to the initial client sleep and connection calls, so that it can interrupt the client at any point.

### DIFF
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -2573,7 +2573,11 @@ impl ClientIoTask {
             tcp = sock.connect(self.target) => {
                 match tcp {
                     Ok(tcp) => {
-                        info!(self.log, "ds_connection connected");
+                        info!(
+                            self.log,
+                            "ds_connection connected from {:?}",
+                            tcp.local_addr()
+                        );
                         tcp
                     }
                     Err(e) => {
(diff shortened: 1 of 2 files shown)


leftwo (approved):
I thought I had tested this path, but I believe I did so before I then
later changed the timeout before reconnect to 10 seconds!

PR #1262, opened 2024-04-18
Agent update to support cloning.

leftwo:
This updates the crucible agent with new fields for the agent's Region structure as well as the CreateRegion structure.  These new fields tell the agent that it should create a region by cloning another region at the provided IP and Port, which involves a slightly different set of steps during region create.

The Region source and read_only field are also persisted in the agents database file and a region created from a source must be a read only region.  Supporting a read only region requires updating the SMF service to add a 'mode' option that is passed to the downstairs.

I also added a bunch of agent-antagonist updates to test the clone behavior.  Now, on the random match, we will also:
- Start the snapshot running that we just created.
- create a new region using clone where the source is our new snapshot.
- delete the new clone region we just created.

In addition, agent was updated to listen for a SIGUSR1 signal, and when it is received, we shutdown all the tasks cleanly. Agent will now watch for any task to panic/stop and if detected, it will stop the rest of the tasks.

Refactor the agent-antagonist into reusable functions.

Updated the agent endpoint with new fields and the ability to create a region as a clone if a source endpoint is provided.


leftwo:
Some further thoughts:

Region clone should use a global repair offset instead of hard coded 4000

Do we want to add additional fields to the agents data file at this time? Like perhaps a time stamp when updated, or a new format that will identify which version we are running?

Can we make a test to update the crucible.json file from old to new? Better names for the old structure version?

agent-antagonist/src/main.rs
```diff
@@ -91,227 +110,546 @@ async fn main_thread(log: Logger, agent: SocketAddr, dataset: String) {
...
     }
+
+    // Our snapshot is ready, and we have populated the source_addr with
+    // the IP (our agent IP) and the repair port ID of the snapshot.
+    region_request.source = Some(source_addr.to_string());
+    region_request.id = RegionId(clone_region_id.to_string());
+
+    info!(
+        log,
+        "Ucreatedse {:?} for source clone {:?}", source_addr, region_request
```
jmpesp, line 567: hysteresis induced paste!

agent/src/model.rs
```diff
@@ -40,6 +68,11 @@ pub struct Region {
     pub key_pem: Option<String>,
 
     pub root_pem: Option<String>,
+    // If this region was created as part of a clone.
+    pub source: Option<SocketAddr>,
+
+    // If this region is read only
+    pub read_only: bool,
```
jmpesp, line 75: We may be able to get away with using `#[serde(default)]` for these two fields, and not have to split into RegionV1 vs Region. That should deserialize the existing data files without errors.

jmpesp (changes requested):

agent/src/model.rs
leftwo: Yeah, I got something to work there.

agent-antagonist/src/main.rs
leftwo: Fixed in latest

downstairs/src/lib.rs
```diff
@@ -3327,12 +3327,32 @@ pub async fn clone_region(
...
+
     let source_def = match repair_server.get_region_info().await {
         Ok(def) => def.into_inner(),
         Err(e) => {
             bail!("Failed to get source region definition: {e}");
         }
     };
+
+    // For the first contact with the remote side, we will retry forever.
+    // We do this in case the source for our snapshot is not ready yet.
```
jmpesp, line 3339: Should this be left in or deleted?

jmpesp (approved):

PR #1263, opened 2024-04-18
Make Region test suite generic across backends

mkeeter:
I'm considering a new raw file backend format, and want to validate it against the same suite of unit tests that we use for our existing extent formats.

This PR is a mostly-mechanical change to convert our test suite into something that can be run against any `Backend`:

- Convert all region test functions to panic on errors; previously, some returned `Result<()>`
- Make the test functions take `backend: Backend` and use `create_region_with_backend(.., backend, ..)` to build the region
- Combine tests that were manually written against both backends, e.g. `reopen_extent_cleanup_replay_sqlite` and `reopen_extent_cleanup_replay`
- Add a macro to generate per-backend test functions in bulk (this is needed because we want to generate a function with a `#[tokio::test]` annotation)

The region tests are now _all_ run against both `Backend::SQLite` and `Backend::RawFile`; previously, only a few hand-crafted tests checked `Backend::SQLite`, because `Backend::RawFile` was the default.  Improving test coverage for the SQLite backend wasn't an explicit goal, but it's nice to have.

### DIFF
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -1749,60 +1749,6 @@ pub(crate) mod test {
         assert!(!Path::new(&cd).exists());
     }
 
-    #[tokio::test]
-    async fn reopen_extent_no_replay_readonly_sqlite() -> Result<()> {
-        // Verify on a read-only region a replacement directory will
-        // be ignored.  This is required by the dump command, as it would
-        // be tragic if the command to inspect a region changed that
-        // region's contents in the act of inspecting.
-
-        // Create the region, make three extents
-        let dir = tempdir()?;
-        let mut region = Region::create_with_backend(
-            &dir,
-            new_region_options(),
-            Backend::SQLite,
-            csl(),
-        )
-        .await?;
-        region.extend(3).await?;
-
-        // Make copy directory for this extent
-        let _ext_one = region.get_opened_extent(1);
-        let cp = Region::create_copy_dir(&dir, 1)?;
-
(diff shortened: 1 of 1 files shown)


leftwo:
I think we may be able to get rid of the `Backend::SQLite` type completely.
Last I checked at customer locations, I did not find any `000.db` files, which suggests to me
that there are no remaining snapshots left with the old format.

However, having the ability to test new backends without all the work to re-implement tests for
them is probably still worth having (and more so now that you have already done the work :) ).

downstairs/src/region.rs
```diff
@@ -4418,10 +4409,83 @@ pub(crate) mod test {
...
-        region.region_write(&writes, JobId(0), true).await?;
+        region.region_write(&writes, JobId(0), true).await.unwrap();
 
         // Validate written data by reading everything back and comparing with
         // data buffer
-        validate_whole_region(&mut region, &data).await
+        validate_whole_region(&mut region, &data).await;
+    }
+
+    macro_rules! region_test_suite {
```
leftwo, line 4419: We should put some comments here to instruct future us that if we wish to add a region test, it should be added to this list here (as, I assume that is what we need to do).

leftwo (approved):

downstairs/src/region.rs
mkeeter: Good idea, added in d06451a93a90ecb656faea5a5cd809c473c60dad

PR #1264, opened 2024-04-18
Remove async from now-synchronous functions

mkeeter:
#1256 allowed `DownstairsClient::send` to become synchronous, but left it `async` to avoid touching too many LOC.

This PR makes that single change, then turns the crank until `rustc` and `clippy` stop complaining.

The result is pretty nice: waiting for events in `Upstairs::select` remains async, but applying them becomes synch.


jmpesp (approved):
there's no crank emoji, so you'll have to imagine one :)

PR #1268, opened 2024-04-22
Raw extent cleanup

mkeeter:
This is standalone cleanup extracted from the upcoming `raw-file-v2` branch:

- Pass `iov_max` to `trait ExtentInner` functions
- Move `OnDiskMeta`, `pread_all`, `pwrite_all` into a new `mod extent_raw_inner_common`
    - Move the helper logic of getting the version tag into `OnDiskMeta::get_version_tag`
- Add a `# Panics` comment explaining that some extent implementations may not be able to manipulate contexts separately from data blocks
- In unit tests that need block contexts, use `ExtentInner::read_into` instead of `ExtentInner::get_block_contexts`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -975,7 +975,6 @@ dependencies = [
  "usdt",
  "uuid 1.8.0",
  "version_check",
- "zerocopy 0.7.32",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


leftwo:
This does not change the on disk format yet, correct?

Issue #871, opened 2023-08-17
Crucible should allow downstairs replacement before it is active.


leftwo:
New title for this issue,
Currently the Upstairs panics when trying to replace a downstairs before it is active:

To reproduce.

start three downstairs with dsc,

start cli server:
```
alan@atrium:crucible$ ./target/release/crutest cli-server -t 127.0.0.1:8810 -t 127.0.0.1:8820 -t 127.0.0.1:8830
Aug 17 01:23:41.182 INFO Upstairs starts
Aug 17 01:23:41.183 INFO Crucible Version: BuildInfo {
    version: "0.0.1",
    git_sha: "175a8cee9d08a3791a89a32a22844f04ad025c31",
    git_commit_timestamp: "2023-08-14T22:40:27.000000000Z",
    git_branch: "main",
    rustc_semver: "1.70.0",
    rustc_channel: "stable",
    rustc_host_triple: "x86_64-unknown-illumos",
    rustc_commit_sha: "90c541806f23a127002de5b4038be731ba1458ca",
    cargo_triple: "x86_64-unknown-illumos",
    debug: false,
    opt_level: 3,
}
Crucible runtime is spawned
Aug 17 01:23:41.183 INFO Listening for a CLI connection on: 0.0.0.0:5050
Upstairs <-> Downstairs Message Version: 4
Aug 17 01:23:41.183 INFO Crucible stats registered with UUID: b242d529-fe6f-4afc-b99d-04a734628f27
Aug 17 01:23:41.183 INFO Crucible b242d529-fe6f-4afc-b99d-04a734628f27 has session id: 17fe4903-c653-41df-8a85-6882d91d3156
Aug 17 01:23:41.183 INFO [1] connecting to 127.0.0.1:8820, looper: 1
Aug 17 01:23:41.183 INFO [0] connecting to 127.0.0.1:8810, looper: 0
Aug 17 01:23:41.183 INFO [2] connecting to 127.0.0.1:8830, looper: 2
Aug 17 01:23:41.183 INFO up_listen starts, task: up_listen
Aug 17 01:23:41.183 INFO Wait for all three downstairs to come online
Aug 17 01:23:41.183 INFO Flush timeout: 0.5
Aug 17 01:23:41.183 INFO [1] b242d529-fe6f-4afc-b99d-04a734628f27 looper connected, looper: 1
[...]

PR #1269, opened 2024-04-22
More Pantry updates for Region replacement

jmpesp:
Add some status endpoints so that Nexus can ask the Pantry some questions:

- how many volumes and job handles are you currently holding on to?

- for a given volume, is it active, and how many job handles refer to it?

Also:

- add an `attach_activate_background` endpoint to the Pantry, so activation can occur in a job instead of blocking the activate endpoint. this will be used by Nexus when it knows that a volume needs reconcilation.

- add a `replace` endoint to the Pantry, so that Nexus can call `target_replace` for volumes attached to the antry

Additionally, move `ReplacementResult` to crucible-client-types, and have `target_replace` return that instead of (). This will be returned by Propolis so that Nexus can tell what the status of a requested replacement is.

### DIFF
--- a/pantry/src/pantry.rs
+++ b/pantry/src/pantry.rs
@@ -396,7 +396,6 @@ impl PantryEntry {
     }
 
     pub async fn detach(&self) -> Result<(), CrucibleError> {
-        self.volume.flush(None).await?;
         self.volume.deactivate().await?;
         Ok(())
     }
--- a/pantry/src/pantry.rs
+++ b/pantry/src/pantry.rs
@@ -1,6 +1,7 @@
 // Copyright 2022 Oxide Computer Company
 
 use std::collections::BTreeMap;
+use std::collections::BTreeSet;
 use std::future::Future;
 use std::str::FromStr;
 use std::sync::Arc;
@@ -20,6 +21,7 @@ use tokio::task::JoinHandle;
 use uuid::Uuid;
 
 use crucible::BlockIO;
+use crucible::ReplaceResult;
 use crucible::SnapshotDetails;
 use crucible::Volume;
 use crucible::VolumeConstructionRequest;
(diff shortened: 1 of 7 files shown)


pantry/src/pantry.rs
```diff
@@ -396,6 +400,116 @@ impl PantryEntry {
...
+                .filter(|(_, job_ids)| job_ids.contains(job_id))
+                .map(|(volume_id, _)| volume_id.clone())
+                .collect();
+
+            assert_eq!(job_volume_id.len(), 1);
+
+            let volume_id = job_volume_id.pop().unwrap();
+
+            let existing = self
+                .volume_id_to_job_ids
```
leftwo, line 503: What is this doing here?  Removing a volume_id from the `volume_id_to_job_ids` list?

jmpesp: right - remove the job from the volume's list of job ids.

leftwo:
I think the case that we saw during debugging that was wrong was a
new pantry was started with a VCR that had one new (port #4) downstairs in it.
Pantry starts up with new connection to the three downstairs.
Pantry begins reconciliation, making the one new downstairs  (port #4) the same as the two original downstairs.

Request then comes in to replace a downstairsi, port #4 -> port #5 

Now, what do we do?  
We have two "good" and one "new" downstairs and are in the process of making them consistent.
But, A downstairs can fail at any time, so we can't refuse it, really..
But, we also can't replace all three downstairs at the same time.

So, I think what we need to try to do is:

1) If an ongoing reconciliation is making progress, allow it to "finish" first.
2) If an ongoing reconciliation is not making progress, then perhaps
allow the replacement to happen before allowing another?

3) But.. If the reconciliation is blocked/stopped because a downstairs
is not available, and that downstairs is the one we want to replace, then maybe
allow the replacement.

4) But, if the downstairs we want to replace is different than the downstairs
that is preventing reconciliation from making progress, then dont allow it.

This is messy and requires more thought.

leftwo:
For the replacement while reconciling issue, I'm going to use this existing
issue as it really is the same situation: https://github.com/oxidecomputer/crucible/issues/871

leftwo (approved):

PR #1270, opened 2024-04-22
Raw extent v2

mkeeter:
(Note: this PR is staged on top of #1268 ; only the last two commits are relevant)

The current raw extent implementation lays out on-disk data like this:

```
[ --- block --- | --- block --- | ... | ctx a0 | ctx a1 | ... | ctx b0 | ctx b1 | ... | metadata ]
```

In other words,
- _N_ Blocks are tightly packed at the front of the file
- Then, there are **two** arrays of block contexts (one slot per block per array, so _2N_ slots in total)
- Finally, there's a trailing metadata chunk

(The layout is also discussed in [this block comment](https://github.com/oxidecomputer/crucible/blob/1ef72f3c935e7dc936bf43310c04668fb60d7a20/downstairs/src/extent_inner_raw.rs#L58-L85)).

This layout prioritizes keeping blocks contiguous / aligned / tightly packed.  In doing so, it makes sacrifices:

- Context slots cannot be written atomically with their block data
  - This means that we have to have **two** context slot arrays, and ping-pong between then
- It's possible for context slot arrays to get fragmented (some blocks using array A, and some using array B), which means reads have to be split between the two arrays.
  - We have extra logic to defragment if this effect becomes significant, which is subtle
- Reading or writing a single block requires two syscalls
  - Those IO operations touch very different parts of the file!

The latter is the original sin of this layout, and is clearly visible in flamegraphs: reading the block data (4 KiB) and context slot (48 bytes) take basically the same amount of time, because (spoilers) each one has to read a full ZFS record (128 KiB).

315651005-3912780b-fde3-4fd8-b4ca-1e3a1f94e483

(graph from https://github.com/oxidecomputer/stlouis/issues/541)

Having blocks be aligned to 512 / 4K boundaries makes sense in terms of mechanical sympathy with the drive, but we're not controlling the drive directly; we've got all of ZFS between our filesystem operations and bytes actually landing in an SSD.

I decided to experiment with an alternate layout that tightly packs blocks and context slots:
```
[ --- block --- | ctx 0 | ctx 1 | --- block --- | --- block --- | ctx 2 | ctx 3 | ... | metadata ]
```

(Naively, you could do `[block | ctx | block | ctx | ...`, but interleaving them uses fewer iovecs)

Blocks and contexts are organized so that they always live in the same ZFS record (i.e. a 128 KiB chunk of the file), and are written in a single `pwritev` call.  Based on [discussion in #oxide-q&a](https://matrix.to/#/!oSMXTSrgambaNFEemN:oxide.computer/$tWSDvIthknMyE09DXo7gz4JeRZPF6rYu45T_Gw25u6k?via=oxide.computer&via=unix.house&via=matrix.org), this means that block + context should always be written together (or not at all).

The new format looks like a significant performance improvement.  Doing random writes to a completely full 128 GiB disk, here's the numbers:

| Write size (blocks) | V1                           | V2                           |
| -                   | -                            | -                            |
| 1                   | 12.7 MiB/sec ± 328.8 KiB/sec | 17.7 MiB/sec ± 269.9 KiB/sec |
| 2                   | 24.6 MiB/sec ± 764.4 KiB/sec | 34 MiB/sec ± 334.8 KiB/sec   |
| 4                   | 47.1 MiB/sec ± 1 MiB/sec     | 65.4 MiB/sec ± 870.7 KiB/sec |
| 8                   | 81 MiB/sec ± 658.1 KiB/sec   | 112.1 MiB/sec ± 1.8 MiB/sec  |
| 16                  | 135.8 MiB/sec ± 2.7 MiB/sec  | 192.5 MiB/sec ± 3.8 MiB/sec  |
| 32                  | 220.5 MiB/sec ± 5.1 MiB/sec  | 291.3 MiB/sec ± 6.9 MiB/sec  |
| 64                  | 290.2 MiB/sec ± 8 MiB/sec    | 368.8 MiB/sec ± 12.4 MiB/sec |
| 128                 | 455.5 MiB/sec ± 5.2 MiB/sec  | 549.9 MiB/sec ± 7.6 MiB/sec  |
| 192                 | 561.6 MiB/sec ± 6.8 MiB/sec  | 651.8 MiB/sec ± 11 MiB/sec   |
| 256                 | 580.6 MiB/sec ± 19.2 MiB/sec | 746.2 MiB/sec ± 16.5 MiB/sec |
| 320                 | 552.8 MiB/sec ± 17.4 MiB/sec | 679.9 MiB/sec ± 9.5 MiB/sec  |
| 384                 | 580.6 MiB/sec ± 36.2 MiB/sec | 680.1 MiB/sec ± 15.8 MiB/sec |
| 448                 | 611.3 MiB/sec ± 18.6 MiB/sec | 738.4 MiB/sec ± 26 MiB/sec   |
| 512                 | 652 MiB/sec ± 47.6 MiB/sec   | 708.8 MiB/sec ± 29.6 MiB/sec |
| 576                 | 605.8 MiB/sec ± 28.7 MiB/sec | 712.3 MiB/sec ± 26.6 MiB/sec |
| 640                 | 641.5 MiB/sec ± 13.4 MiB/sec | 720.8 MiB/sec ± 26 MiB/sec   |
| 704                 | 653.8 MiB/sec ± 8.2 MiB/sec  | 747.6 MiB/sec ± 6.2 MiB/sec  |
| 768                 | 666.2 MiB/sec ± 13.7 MiB/sec | 756 MiB/sec ± 27.1 MiB/sec   |
| 832                 | 654.3 MiB/sec ± 7.6 MiB/sec  | 700.9 MiB/sec ± 21.8 MiB/sec |
| 896                 | 673.2 MiB/sec ± 14.1 MiB/sec | 746.7 MiB/sec ± 21.9 MiB/sec |
| 960                 | 676.7 MiB/sec ± 11 MiB/sec   | 776 MiB/sec ± 12.9 MiB/sec   |
| 1024                | 707.7 MiB/sec ± 32.3 MiB/sec | 764.8 MiB/sec ± 24.9 MiB/sec |

In graph form:

Screenshot 2024-04-19 at 4 40 12 PM

**Todo**

- [ ] Write a big comment about `ZFS_RECORDSIZE` shenanigans / crash consistency
- [ ] Retest after a870b1e634a18417c8c58e6dcb0362ff8413f7b6
- [ ] Test read performance
- [ ] Make integration tests generic across extent types and run them for both `RawFile` and `RawFileV2`?


downstairs/src/extent_inner_raw_v2.rs
```diff
@@ -0,0 +1,1181 @@
...
+    extent_number: u32,
+
+    /// Extent size, in blocks
+    extent_size: Block,
+
+    /// Helper `struct` controlling layout within the file
+    layout: RawLayout,
+
+    /// Has this block been written?
+    block_written: Vec<bool>,
```
aapoalas, line 149: Excuse my random outsider comment:

**suggestion:** This seems to never change in size after creation, so it could maybe be a boxed slice?

mkeeter:
This is now updated onto `main` and ready for review.

It's a little simpler than the previous description: I decided to remove the fancy interleaving of blocks and contexts, and went back to the straight-forward
```
[block | ctx | block | ctx | ... ]
```

(this may change if I see compelling benchmarks, to come soon)

downstairs/src/extent_inner_raw_v2.rs
```diff
@@ -0,0 +1,1058 @@
...
+use std::io::{BufReader, Read};
+use std::os::fd::{AsFd, AsRawFd};
+use std::path::Path;
+use std::{
+    fs::{File, OpenOptions},
+    io::IoSlice,
+};
+use zerocopy::AsBytes;
+
+pub(crate) const DEFAULT_ZFS_RECORDSIZE: u64 = 128 * 1024;
```
jmpesp, line 27: both cases in `get_record_size` where this is used, it doesn't seem like the path is a ZFS file system, so maybe this should be `NON_ZFS_RECORDSIZE` or something?

downstairs/src/extent_inner_raw_v2.rs
```diff
@@ -0,0 +1,1058 @@
...
+///
+/// where `R` is the recordsize / (block + context size)
+///
+/// We don't want to split blocks or block-context pairs across ZFS records,
+/// because that could make writes non-atomic.
+///
+/// ## Expected recordsize
+/// After the block data, we store a single `u64` representing the expected
+/// recordsize when the file was written.  When the file is reopened, we detect
+/// if its recordsize has changed, which would be surprising!
```
jmpesp, line 58: TODO-jwm this may happen after migration

downstairs/src/extent_inner_raw_v2.rs
```diff
@@ -0,0 +1,1058 @@
...
+                start_new_chunk = false
+            }
+            let vs = &mut iovecs.last_mut().unwrap().iovecs;
+            vs.push(IoSlice::new(data));
+            vs.push(IoSlice::new(ctx));
+            expected_bytes += data.len() + ctx.len();
+            if self.layout.has_padding_after(block) {
+                vs.push(IoSlice::new(&padding));
+                expected_bytes += padding.len();
+            }
```
jmpesp, line 168: TODO-jwm can pwritev partially fail?

downstairs/src/extent_inner_raw_v2.rs
```diff
@@ -0,0 +1,1058 @@
...
+            return Ok(());
+        }
+
+        self.set_dirty()?;
+
+        cdt::extent__write__file__start!(|| {
+            (job_id.0, self.extent_number.0, n_blocks as u64)
+        });
+
+        // Now execute each chunk in a separate `pwritev` call
```
jmpesp, line 188: TODO-jwm this means the Crucible level write can partially succeed

downstairs/src/extent_inner_raw_v2.rs
```diff
@@ -0,0 +1,1058 @@
...
+        });
+
+        r
+    }
+
+    fn read(
+        &mut self,
+        job_id: JobId,
+        req: ExtentReadRequest,
+        iov_max: usize, // unused by raw backend
```
jmpesp, line 247: not true here :)

downstairs/src/extent_inner_raw_v2.rs
```diff
@@ -0,0 +1,1058 @@
...
+                iov_base: buf_ptr,
+                iov_len: block_size,
+            });
+            iovecs.push(libc::iovec {
+                iov_base: ctx as *mut _ as *mut _,
+                iov_len: BLOCK_CONTEXT_SIZE_BYTES as usize,
+            });
+            if self.layout.has_padding_after(block) {
+                iovecs.push(libc::iovec {
+                    iov_base: padding.as_mut_ptr() as *mut _,
```
jmpesp, line 277: What is preadv's behaviour when it's passed multiple iovecs with the same address?

downstairs/src/extent_inner_raw_v2.rs
```diff
@@ -0,0 +1,1058 @@
...
+
+        let mut block = start_block;
+        let mut padding_count = 0;
+        for ctx in &mut ctxs {
+            iovecs.push(libc::iovec {
+                iov_base: buf_ptr,
+                iov_len: block_size,
+            });
+            iovecs.push(libc::iovec {
+                iov_base: ctx as *mut _ as *mut _,
```
jmpesp, line 272: what's going on here haha?

downstairs/src/extent_inner_raw_v2.rs
```diff
@@ -0,0 +1,1058 @@
...
+                                "zfs returned non-integer for recordsize: \
+                                  {out:?} ({e}), stderr: {err}"
+                            ))
+                        })?
+                    }
+                }
+                Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
+                    // If the `zfs` executable isn't present, then we're
+                    // presumably on a non-ZFS filesystem and will use a default
+                    // recordsize
```
jmpesp, line 657: There's a remote possibility that PATH isn't set correctly but we are on a ZFS filesystem. I'm wondering if the downstairs should somehow expect that this binary is present if we're running on illumos (we already expect it for taking snapshots!)

downstairs/src/extent_inner_raw_v2.rs
```diff
@@ -0,0 +1,1058 @@
...
+/// Data structure that implements the on-disk layout of a raw extent file
+struct RawLayout {
+    extent_size: Block,
+    recordsize: u64,
+}
+
+impl std::fmt::Debug for RawLayout {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("RawLayout")
+            .field("extent_size", &self.extent_size)
```
jmpesp, line 680: Why only this?

downstairs/src/extent_inner_raw_v2.rs
jmpesp: To be clear, after region / region snapshot replacement.

jmpesp:
Looks like I clicked "Comment" too early, so there's a few `TODO-jwm` there! :)

downstairs/src/extent_inner_raw_v2.rs
jmpesp: To be clear! I think this was always true but was TODO for myself to 1) verify this and 2) suggest we add this as documentation somewhere?

downstairs/src/extent_inner_raw_v2.rs
jmpesp: I added this as a comment for myself to verify, but my thinking here was

1) of course pwritev can partially fail, it can return a different number of bytes written out than what was added
2) are there any zfs specific considerations to think about

downstairs/src/extent_inner_raw_v2.rs
mkeeter: Good point, renamed to `DUMMY_RECORDSIZE` in f7c13d7d4369

downstairs/src/extent_inner_raw_v2.rs
mkeeter: I believe they're processed in array order, so the last one wins.  It doesn't really matter here, because we're not using the padding data for anything.

downstairs/src/extent_inner_raw_v2.rs
mkeeter: lol it's casting from `&mut [u8; 32]` → `*mut [u8; 32]` → `*mut c_void`

downstairs/src/extent_inner_raw_v2.rs
mkeeter: yup, fixed in 19589ed19254

downstairs/src/extent_inner_raw_v2.rs
mkeeter: This was copy-pasta from `extent_inner_raw.rs`, but I'm not sure why they're not just deriving Debug.

Fixed in 5b2a9425ea09 for both

downstairs/src/extent_inner_raw_v2.rs
mkeeter: Good idea, changed in 5c446c791dac to return an error if `zfs` isn't present on an illumos system.

downstairs/src/extent_inner_raw_v2.rs
```diff
@@ -0,0 +1,1013 @@
...
+/// Max of serialized block context (which is a [`ReadBlockContext`])
+pub(crate) const BLOCK_CONTEXT_SIZE_BYTES: u64 = 32;
+
+/// `RawInnerV2` is a wrapper around a [`std::fs::File`] representing an extent
+///
+/// The file contains [`extent_size`](RawInnerV2::extent_size) blocks, which I'm
+/// writing as `N` for simplicity here.
+///
+/// # File organization
+/// The file is structured in three parts.  Getting specific offsets within the
```
leftwo, line 39: I would think this is 4 parts?
`[Block data and context]`
`[Expected recordsize]`
`[Written block array]`
`[File metadata]`

Ah, maybe `[Written block array]` and `[File metadata]` are considered the same part because
we write them in a single write?

downstairs/src/extent_inner_raw_v2.rs
```diff
@@ -0,0 +1,1013 @@
...
+    /// Returns the number of blocks that fit into a ZFS recordsize
+    fn blocks_per_record(&self) -> u64 {
+        // Each block contains data and a single context slot
+        let bytes_per_block = self.block_size() + BLOCK_CONTEXT_SIZE_BYTES;
+        self.recordsize / bytes_per_block
+    }
+
+    /// Checks whether there is padding after the given block
+    fn has_padding_after(&self, block: BlockOffset) -> bool {
+        // No padding at the end of the file
```
leftwo, line 851: Is this `end of the file` the actual end of the file, or just the end of the data/context section?

downstairs/src/lib.rs
```diff
@@ -3416,9 +3417,12 @@ enum WrappedStream {
 /// tests, it can be useful to create volumes using older backends.
 #[derive(Copy, Clone, Default, Debug, PartialEq)]
 pub enum Backend {
-    #[default]
+    #[cfg(any(test, feature = "integration-tests"))]
```
leftwo, line 3420: I wonder if there is some value in allowing `crucible-downstairs` to pick what backend it
wants at creation time?  I think there could be value in making "old" versions for various kinds
of tests and not hiding them behind a feature would allow that.

It's also not something to worry about for this PR though.  Why do I even bring it up then?
Good question.  For which I have no answer..

leftwo (commented):
I think we might need some additional testing, or some updates to the testing
done in `integration_tests/src/lib.rs`.  That file by default will only create `RAW`
and `SQLite` backend types, and we should go through it and figure out where
we need to also do the same tests for `RAW_V2` (or if all the `RAW` tests become
`RAW_V2`

We have snapshots with SQLite
We have snapshots with RAW.
So, we need to support the downstairs being able to "clone" either of those.
`integration_test_clone_raw()`

Should we add a test (or do we have a test already) of downstairs migration from 
SQLite to RAW_V2?

downstairs/src/extent_inner_raw_v2.rs
```diff
@@ -0,0 +1,1013 @@
...
+    }
+
+    /// Update the flush number, generation number, and clear the dirty bit
+    fn set_flush_number(
+        &mut self,
+        new_flush: u64,
+        new_gen: u64,
+    ) -> Result<(), CrucibleError> {
+        self.layout.write_block_written_and_metadata(
+            &self.file,
```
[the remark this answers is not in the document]
mkeeter: Yup, there's a `pwrite_all` in `write_block_written_and_metadata`

downstairs/src/extent_inner_raw_v2.rs
mkeeter: This was outdated, fixed!

downstairs/src/extent_inner_raw_v2.rs
mkeeter: The end of the data blocks section; fixed

downstairs/src/extent_inner_raw_v2.rs
```diff
@@ -0,0 +1,1013 @@
...
+        }
+
+        if total_bytes != expected_bytes as usize {
+            return Err(CrucibleError::IoError(format!(
+                "extent {}: incomplete read \
+                 (expected {expected_bytes}, got {total_bytes})",
+                self.extent_number
+            )));
+        }
+        cdt::extent__read__file__done!(|| {
```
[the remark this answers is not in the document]
mkeeter: Fixed (both here and in `extent_inner_raw.rs`, which had the same behavior)

downstairs/src/extent_inner_raw_v2.rs
```diff
@@ -0,0 +1,1058 @@
...
+/// writing as `N` for simplicity here.
+///
+/// # File organization
+/// The file is structured in three parts.  Getting specific offsets within the
+/// file is implemented in the [`RawLayout`] helper class.
+///
+/// ## Block data and contexts
+/// Block data and per-block contexts are paired up as follows:
+/// ```text
+/// [ ----- block  ----- | context | ----- block  ----- | context] x R
```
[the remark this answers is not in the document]
mkeeter: Yup, fixed

downstairs/src/extent_inner_raw_v2.rs
```diff
@@ -0,0 +1,1058 @@
+// Copyright 2023 Oxide Computer Company
```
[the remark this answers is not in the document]
mkeeter: Fixed!

downstairs/src/extent_inner_raw_v2.rs
```diff
@@ -0,0 +1,1013 @@
...
+        };
+
+        if r.is_err() {
+            for i in 0..n_blocks {
+                let block = BlockOffset(write.offset.0 + i as u64);
+
+                // Try to recompute the context slot from the file.  If this
+                // fails, then we _really_ can't recover, so bail out
+                // unceremoniously.
+                self.recompute_block_written_from_file(block).unwrap();
```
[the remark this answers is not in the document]
mkeeter: This comment was out of date; fixed!

downstairs/src/extent_inner_raw_v2.rs
```diff
@@ -0,0 +1,1058 @@
...
+/// written or not.
+///
+/// ## File metadata
+/// The last [`BLOCK_META_SIZE_BYTES`] in the file contain an [`OnDiskMeta`]
+/// serialized using `bincode`.  The first byte of this range is `dirty`,
+/// serialized as a `u8` (where `1` is dirty and `0` is clean).
+///
+/// There are a few considerations that led to this particular ordering:
+/// - The written blocks array and metadata must be contiguous, because we want
+///   to write them atomically when clearing the `dirty` flag.
```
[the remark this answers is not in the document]
mkeeter: Yes, and `[written blocks array] + [file metadata]` must be in the same record as well.  This is checked in `test_metadata_position`.

mkeeter:
> I think we might need some additional testing, or some updates to the testing done in `integration_tests/src/lib.rs`. That file by default will only create `RAW` and `SQLite` backend types, and we should go through it and figure out where we need to also do the same tests for `RAW_V2` (or if all the `RAW` tests become `RAW_V2`
> 
> We have snapshots with SQLite We have snapshots with RAW. So, we need to support the downstairs being able to "clone" either of those. `integration_test_clone_raw()`

I've gone through and made `integration_tests` generic across different backends (361584843ef1).

> Should we add a test (or do we have a test already) of downstairs migration from SQLite to RAW_V2?

Right now, the only migration is the automatic SQLite → RAW_V1, which happens if a read-write SQLite extent is opened.  Since we don't have any read-write SQLite extents in the field (only read-only snapshots), I didn't bother implementing SQLite → RAW_V2.

leftwo:
> > Should we add a test (or do we have a test already) of downstairs migration from SQLite to RAW_V2?
> 
> Right now, the only migration is the automatic SQLite → RAW_V1, which happens if a read-write SQLite extent is opened. Since we don't have any read-write SQLite extents in the field (only read-only snapshots), I didn't bother implementing SQLite → RAW_V2.

Yeah, that seems like the right call here.  Thanks.

PR #1272, opened 2024-04-24
Refactor downstairs writes

mkeeter:
https://github.com/oxidecomputer/crucible/pull/1206 reorganized the `Message::Write` variant so that the bulk data was all packed at the end, and could be sent in a single write (instead of incurring the cost of serialization).

The `Downstairs` then took that data and split it into a `Vec<crucible_protocol::Write>` objects, which each look like this:
```rust
pub struct Write {
    pub eid: u64,
    pub offset: Block,
    pub data: Bytes,

    pub block_context: BlockContext,
}
```

Note that this `Write` does not do an extra copy; the `data: Bytes` is borrowed from the received data (this is good).

However, it's still an awkward API: a single contiguous `Bytes` object arrives on the wire, then we split it into a bunch of smaller (block-sized) chunks, then we write those chunks to contiguous blocks with `pwritev`.

Put a different way, it's awkward to have `ExtentInner::write` take a `&[crucible_protocol::Write]`: the extent implementation has to jump through hoops to reassemble contiguous writes from the incoming slice of writes.  As a smaller papercut, it's also awkward that this struct contains an `eid`: what should happen if someone calls `ExtentInner::write` with an `eid` that doesn't match?

This PR reorganizes our writes into two new types in the `crucible-downstairs` crate:
```rust
/// Deserialized write operation, contiguous to a single extent
///
/// `data` should be a borrowed section of the received `Message::Write`, to
/// reduce memory copies.
#[derive(Debug, Clone, Eq, PartialEq)]
pub(crate) struct ExtentWrite {
    pub offset: Block, // starting position
    pub block_contexts: Vec<BlockContext>, // one per block
    pub data: bytes::Bytes, // num blocks * block size
}

/// `ExtentWrite` plus an extent tag
#[derive(Debug, Clone, Eq, PartialEq)]
pub(crate) struct Write {
    extent: u64,
    write: ExtentWrite,
}
```

The `ExtentWrite` is represents a set of contiguous writes, because it simply doesn't store one offset per block; it just has a starting offset.  Note that it does not contain an extent tag, so `ExtentInner::write` no longer needs to think about what happens if there's a mismatch.  The target extent lives in a new `crucible_downstairs::Write` object, where it's used for dispatch at the `Region` level (but not passed to the `Extent`).

The process of splitting the write into contiguous, separated-by-extent subwrites still has to happen, but it's now a conversion from `Message::Write → Vec<Write>` that happens when messages are received.

This means that the extent implementations can use `pwrite` instead of `pwritev`, and no longer need to consider whether the incoming writes are contiguous.

There should be no change to on-disk formats; this is purely a change to how we wrangle data from the wire into the extent data file.

Performance benchmarking to come, but this feels like the correct architecture to me.

### DIFF
--- a/downstairs/src/extent_inner_raw.rs
+++ b/downstairs/src/extent_inner_raw.rs
@@ -9,9 +9,9 @@ use crate::{
         pread_all, pwrite_all, OnDiskMeta, BLOCK_META_SIZE_BYTES,
     },
     integrity_hash, mkdir_for_file,
-    region::{BatchedPwritev, JobOrReconciliationId},
+    region::JobOrReconciliationId,
     Block, BlockContext, CrucibleError, ExtentReadRequest, ExtentReadResponse,
-    JobId, RegionDefinition,
+    ExtentWrite, JobId, RegionDefinition,
 };
 
 use itertools::Itertools;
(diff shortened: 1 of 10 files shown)


downstairs/src/extent_inner_raw.rs
```diff
@@ -149,38 +149,172 @@ impl ExtentInner for RawInner {
...
-                    &writes[start..i],
-                    only_write_unwritten,
-                    iov_max,
-                )?;
-                // Keep going, resetting the hashmap and start position
-                if i != writes.len() {
-                    seen.clear();
-                    seen.insert(writes[i].offset.value);
-                    start = i;
+        check_input(self.extent_size, write.offset, &write.data)?;
```
mkeeter, line 160: This diff is transplanting the old `write_without_overlaps` into `write`, because that's now guaranteed by the API.

downstairs/src/extent_inner_raw.rs
```diff
@@ -2450,66 +2387,4 @@ mod test {
         let mut meta_buf = [0u8; BLOCK_META_SIZE_BYTES as usize];
         bincode::serialize_into(meta_buf.as_mut_slice(), &Some(m)).unwrap();
     }
-
-    /// Test that multiple writes to the same location work
-    #[test]
-    fn test_multiple_writes_to_same_location_raw() -> Result<()> {
```
mkeeter, line 2456: This can no longer happen, so we can delete the test!

faithanalog (commented):
I've been hoping for this for quite awhile. My main concern was with `only_write_unwritten` but it looks like you've handled that as far as I can tell.

mkeeter:
Doing a performance test on Madrid, I don't see any concerns; it looks slightly faster.

| blocks | `struct-extent-write`        | `main`                       |
|--------|------------------------------|------------------------------|
| 1      | 13.7 MiB/sec ± 395.1 KiB/sec | 14.6 MiB/sec ± 63 KiB/sec    |
| 2      | 28.1 MiB/sec ± 1 MiB/sec     | 28.4 MiB/sec ± 190.1 KiB/sec |
| 4      | 54.7 MiB/sec ± 579.5 KiB/sec | 52.7 MiB/sec ± 568.5 KiB/sec |
| 8      | 97.9 MiB/sec ± 3.9 MiB/sec   | 94.1 MiB/sec ± 1.9 MiB/sec   |
| 16     | 167 MiB/sec ± 5.6 MiB/sec    | 166.6 MiB/sec ± 6.5 MiB/sec  |
| 32     | 270.1 MiB/sec ± 3.1 MiB/sec  | 260.6 MiB/sec ± 3.9 MiB/sec  |
| 64     | 359.1 MiB/sec ± 6.8 MiB/sec  | 347.8 MiB/sec ± 5.2 MiB/sec  |
| 128    | 562.1 MiB/sec ± 8.9 MiB/sec  | 540.8 MiB/sec ± 5.6 MiB/sec  |
| 192    | 681 MiB/sec ± 20.7 MiB/sec   | 659.3 MiB/sec ± 13.4 MiB/sec |
| 256    | 760.4 MiB/sec ± 14.8 MiB/sec | 740.7 MiB/sec ± 17.3 MiB/sec |
| 320    | 662.6 MiB/sec ± 18.8 MiB/sec | 659.2 MiB/sec ± 16.9 MiB/sec |
| 384    | 721.1 MiB/sec ± 14.2 MiB/sec | 691.5 MiB/sec ± 13.5 MiB/sec |
| 448    | 777.2 MiB/sec ± 18.4 MiB/sec | 733.4 MiB/sec ± 11.7 MiB/sec |
| 512    | 817.6 MiB/sec ± 14.4 MiB/sec | 768.7 MiB/sec ± 14.2 MiB/sec |
| 576    | 742.2 MiB/sec ± 12.4 MiB/sec | 701.5 MiB/sec ± 12.5 MiB/sec |
| 640    | 770.3 MiB/sec ± 13.2 MiB/sec | 707.3 MiB/sec ± 24.7 MiB/sec |
| 704    | 790.7 MiB/sec ± 16.9 MiB/sec | 728.9 MiB/sec ± 24.2 MiB/sec |
| 768    | 823.6 MiB/sec ± 18.1 MiB/sec | 738.8 MiB/sec ± 37 MiB/sec   |
| 832    | 786.9 MiB/sec ± 18.2 MiB/sec | 726.7 MiB/sec ± 17 MiB/sec   |
| 896    | 778.6 MiB/sec ± 21 MiB/sec   | 722.6 MiB/sec ± 25.4 MiB/sec |
| 960    | 832.7 MiB/sec ± 17.7 MiB/sec | 730 MiB/sec ± 23.6 MiB/sec   |
| 1024   | 787.5 MiB/sec ± 97 MiB/sec   | 758.1 MiB/sec ± 46.9 MiB/sec |

downstairs/src/lib.rs
```diff
@@ -1898,11 +2022,14 @@ impl Downstairs {
                     data,
                 }))
             }
-            IOop::WriteUnwritten { writes, .. } => {
+            IOop::WriteUnwritten { .. } => {
                 /*
                  * Any error from an IO should be intercepted here and passed
                  * back to the upstairs.
                  */
+                let IOop::WriteUnwritten { writes, .. } = job.work else {
```
jmpesp, line 2030: Why not change to `match job.work` above?

leftwo, line 2030: What's the purpose here of this check?  Are we not already matching on `job.work`?

downstairs/src/lib.rs
```diff
@@ -1922,10 +2049,14 @@ impl Downstairs {
                     result,
                 }))
             }
-            IOop::Write {
-                dependencies,
-                writes,
-            } => {
+            IOop::Write { .. } => {
+                let IOop::Write {
```
leftwo, line 2053: Same question here about the check on `job.work`.

downstairs/src/lib.rs
```diff
@@ -1922,10 +2049,14 @@ impl Downstairs {
...
+                    writes,
+                    dependencies,
+                } = job.work
+                else {
+                    unreachable!();
+                };
```
faithanalog, line 2059: what does this change (and the similar one above it) do? Is this something to make the rust compiler happy?

downstairs/src/region.rs
```diff
@@ -758,27 +758,40 @@ impl Region {
...
+                        integrity_hash(&[&block])
+                    };
 
-            if computed_hash != write.block_context.hash {
-                error!(self.log, "Failed write hash validation");
-                // TODO: print out the extent and block where this failed!!
-                crucible_bail!(HashMismatch);
+                if computed_hash != ctx.hash {
+                    error!(self.log, "Failed write hash validation");
+                    // TODO: print out the extent and block where this failed!!
```
leftwo, line 792: [I opened this issue ](https://github.com/oxidecomputer/crucible/issues/1307) for this.  In the event that we do get a hash validation error, all the data we can dump can only help us figure out what has happened.

faithanalog (approved):

downstairs/src/region.rs
```diff
@@ -799,36 +812,22 @@ impl Region {
...
+            let extent = self.get_opened_extent_mut(req.extent as usize);
             extent
-                .write(job_id, &writes[..], only_write_unwritten)
+                .write(job_id, req.write, only_write_unwritten)
                 .await?;
-        }
 
-        // Mark any extents we sent a write-command to as potentially dirty
-        self.dirty_extents.extend(batched_writes.keys());
+            // Mark any extents we sent a write-command to as potentially dirty
```
leftwo, line 828: Is this in the correct place?  (It was here before you got here) but I'm wonding
if the mark as dirty should happen before we do the actual write?

leftwo (approved):
I had a question, comment, or nit, but nothing (I think) that is serious.

downstairs/src/lib.rs
mkeeter: Good catch – that was previously not possible due to the types in `IOop::ReadRequest`, but that was fixed by #1283 .  Fixed in b70e4f1f118f

downstairs/src/lib.rs
mkeeter: This was intentional at the time: the `match` borrowed `job.work`, and then this statement transferred ownership in only particular cases.  However, it's not necessary; see https://github.com/oxidecomputer/crucible/pull/1272#discussion_r1600406065

downstairs/src/extent_inner_raw.rs
```diff
@@ -134,38 +134,172 @@ impl ExtentInner for RawInner {
...
 
+    /// Performs a single write within this extent
     fn write(
         &mut self,
         job_id: JobId,
-        writes: &[crucible_protocol::Write],
+        write: &ExtentWrite,
         only_write_unwritten: bool,
-        iov_max: usize,
+        _iov_max: usize,
```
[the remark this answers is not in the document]
mkeeter: It's sticking around because it will be used in `RawFileV2` 😉

downstairs/src/extent_inner_raw.rs
```diff
@@ -134,38 +134,172 @@ impl ExtentInner for RawInner {
...
-                    iov_max,
-                )?;
-                // Keep going, resetting the hashmap and start position
-                if i != writes.len() {
-                    seen.clear();
-                    seen.insert(writes[i].offset.value);
-                    start = i;
+        check_input(self.extent_size, write.offset, write.data.len())?;
+        /*
+         * In order to be crash consistent, perform the following steps in
```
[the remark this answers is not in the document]
mkeeter: It's mostly the same, except that we don't need to check for contiguous regions here (because it's guaranteed by the type coming in)

downstairs/src/extent_inner_sqlite.rs
```diff
@@ -2,14 +2,16 @@
 use crate::{
     cdt, crucible_bail,
     extent::{check_input, extent_path, DownstairsBlockContext, ExtentInner},
+    extent_inner_raw_common::pwrite_all,
```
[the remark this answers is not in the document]
mkeeter: 😆 that's fair; we've also got pretty extensive unit tests backing us up (and they're run against both extent backends)

downstairs/src/lib.rs
```diff
@@ -307,6 +307,130 @@ pub(crate) struct ExtentReadResponse {
     data: BytesMut,
 }
 
+/// Deserialized write operation, contiguous to a single extent
+///
+/// `data` should be a borrowed section of the received `Message::Write`, to
+/// reduce memory copies.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub(crate) struct ExtentWrite {
+    pub offset: Block,
```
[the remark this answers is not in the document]
mkeeter: Sure, added docstrings in 23019c92

downstairs/src/lib.rs
```diff
@@ -307,6 +307,130 @@ pub(crate) struct ExtentReadResponse {
...
+            return Err(CrucibleError::InvalidNumberOfBlocks(format!(
+                "got {} contexts but {} blocks",
+                blocks.len(),
+                buf.len() / block_size
+            )));
+        }
+        let mut out = Self(vec![]);
+        for b in blocks {
+            let append = out.can_append(b);
+            if append {
```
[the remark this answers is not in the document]
mkeeter: Good catch, done in 94fd44609a03

downstairs/src/lib.rs
mkeeter: Congrats to all three reviewers for noticing this 😂 

https://github.com/oxidecomputer/crucible/pull/1272#discussion_r1600406065 explains the issue

downstairs/src/region.rs
mkeeter: Interesting – I don't know whether a failed write should qualify an extent for sync-to-disk, but it's probably safer.  I moved this above the write attempt in b9b087e1f731

jmpesp (approved):

downstairs/src/region.rs
jmpesp: I don't think this is right, though the problem might have existed before this PR too - if all writes are skipped, then `set_dirty` isn't called. `dirty_extents`' comment says

> Should be true if the dirty flag in the extent's metadata is set.

which might drift in that case.

PR #1274, opened 2024-04-25
Add some DTrace scripts to the package.

leftwo:
Added a few dtrace scripts to the pantry and downstairs packages. 
Updated the upstairs_info.d DTrace script to handle multiple volumes under the same PID.  
More work is needed there, but this at least will print correct numbers.

### DIFF
--- a/package-manifest.toml
+++ b/package-manifest.toml
@@ -6,6 +6,7 @@ source.rust.release = true
 source.paths = [
   { from = "agent/smf", to = "/var/svc/manifest/site/crucible" },
   { from = "agent/downstairs_method_script.sh", to = "/opt/oxide/lib/svc/manifest/crucible/downstairs.sh" },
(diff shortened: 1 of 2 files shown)


package-manifest.toml
```diff
@@ -16,5 +17,10 @@ source.rust.binary_names = ["crucible-pantry"]
 source.rust.release = true
 source.paths = [
   { from = "pantry/smf/pantry.xml", to = "/var/svc/manifest/site/crucible/pantry.xml" },
+  { from = "tools/dtrace/upstairs_info.d", to = "/opt/oxide/dtrace/upstairs_info.d" },
```
jmpesp, line 20: I think `from` can be a directory, if we want to ship the whole thing

jmpesp (approved):
This is really cool! :ice_cube:

package-manifest.toml
leftwo: I'm going to continue to iterate on these, but yeah, just including them all might not be too bad of an idea.

PR #1275, opened 2024-04-25
replace needing no work should not be an error

jmpesp:
Previously, if a caller requested a target replacement, and supplied the same construction request for the original and replacement arguments, a `ReplaceRequestInvalid` error would be returned. This isn't exactly an error, it's a state where no work needs to be done.

Return `VcrMatches` as a `ReplaceResult` if this is the case, not an error. Note `compare_vcr_for_replacement` still returns all the same errors otherwise.


leftwo (approved):
Questions addressed in Chat.

PR #1277, opened 2024-04-25
Don't allow replacement before the upstairs is active

leftwo:
Return an error if a downstairs replacement arrives before the upstairs is active.

### DIFF
--- a/crucible-client-types/src/lib.rs
+++ b/crucible-client-types/src/lib.rs
@@ -104,6 +104,7 @@ pub enum ReplaceResult {
     CompletedAlready,
     Missing,
     VcrMatches,
+    NotActive,
 }
 
 impl Debug for ReplaceResult {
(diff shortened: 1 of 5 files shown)


leftwo:
This blocks replacement requests when talking to a running upstairs and that upstairs is not `Active`.

My concern here is making sure we have some path out of the situation where an upstairs has started but has not completed reconciliation, possibly never able to make progress (like one of the downstairs is gone).

In that case, we can't replace a downstairs (because we refuse it).
But, we also will never go active..

So, for this to work in the big picture, we have to be able to tell an instance or pantry to stop (or fail, or something) so
we can tear it down and try again.

leftwo:
We do need to support the ability to replace a downstairs before active.
We have to be careful though, as a replacement can't be considered done till an upstairs
has reconciled the three downstairs.

PR #1281, opened 2024-04-29
Allow replacement of a target in a read_only_parent

leftwo:
Update the VCR validation and replacement code to accept changes to the read_only_parent field.  

For this, I was able to re-use the same code that was checking sub_volumes, by pulling that common
part out to a sub-function which we could use to also check read only parents.  The meat of the
comparison tests is all the same, just moved down.  Since we now compare both sub_volume and
read_only_parent, we have to then compare the results with each other and be sure that only one
of the two changes.

New tests added for read only parent replacement.

Cleaned up a few copy paste errors in other test comments.


upstairs/src/volume.rs
```diff
@@ -792,9 +792,25 @@ impl BlockIO for Volume {
...
+                | ReplaceResult::StartedAlready
+                | ReplaceResult::CompletedAlready
+                | ReplaceResult::VcrMatches => {
+                    // Found a match in our read only parent
+                    info!(self.log, "ROP has something here");
+                    return Ok(result);
+                }
+
+                ReplaceResult::Missing => {
+                    // keep looking!
```
jmpesp, line 810: `810Gdd`

upstairs/src/volume.rs
```diff
@@ -1238,14 +1234,130 @@ impl Volume {
...
+                    "read_only_parent does not exist in replacement"
+                        .to_string(),
+                ))
+            }
+        };
+
+        // Determine the sub_volume difference if there is any.
+        let sub_volume_result = Self::compare_vcr_region_for_replacement(
+            &o_sub_volumes[0],
+            &n_sub_volumes[0],
```
jmpesp, line 1284: This part and probably a whole lot of other areas are going to need revisiting when we move to the future where Nexus is allocating multiple sub-volume entries, but that shouldn't block this PR.

upstairs/src/volume.rs
```diff
@@ -1238,14 +1234,130 @@ impl Volume {
...
+                // error, otherwise, we return what the sub_volume found.
+                match rop_result {
+                    Ok(Some((_, _))) => {
+                        // Both have changes, this is no good
+                        crucible_bail!(
+                            ReplaceRequestInvalid,
+                            "Both sub_volume and read_only_parent have changes"
+                        )
+                    }
+                    _ => Ok(Some((old, new))),
```
jmpesp, line 1310: should this match on `rop_result` being `Ok(None)` ?

upstairs/src/volume.rs
```diff
@@ -1238,14 +1234,130 @@ impl Volume {
...
+            Ok(None) => {
+                // The sub_volume VCR is the same as what we have, but with
+                // updated generation numbers, so this qualifies for the
+                // Ok(None) state.  Check what the ROP found.
+                match rop_result {
+                    Ok(Some((read_only_old, read_only_new))) => {
+                        // Just the ROP has changes, so let's return those.
+                        Ok(Some((read_only_old, read_only_new)))
+                    }
+                    _ => Ok(None),
```
jmpesp, line 1322: this too?

upstairs/src/volume.rs
leftwo: :w

upstairs/src/volume.rs
leftwo: Yeah,  I know of a few places as well where we will either explode or report failure if it finds more than one sub-volume.

upstairs/src/volume.rs
leftwo: Yes... I believe the code as is would ignore an error in `rop_result` if `sub_volume_result` is
`Some()` and instead of returning error it returns what the `SubVolume` found.  This is tricky
because, when comparing two VCRs,  we can have only one of these:
1) A valid ROP difference only, SubVolumes are the same.
2) The ROPs are identical and the new SubVolume is different.
3) The Old has a ROP and the new does not, and the SubVolume is different.

I'll ponder this..

integration_tests/src/lib.rs
```diff
@@ -5695,6 +5695,140 @@ mod test {
...
+            gen: 1,
+        }];
+
+        let new_vol: VolumeConstructionRequest =
+            VolumeConstructionRequest::Volume {
+                id: sv_volume_id,
+                block_size: BLOCK_SIZE as u64,
+                sub_volumes: new_sub_vol.clone(),
+                read_only_parent: Some(rop),
+            };
```
mkeeter, line 5780:
```suggestion
        let new_vol = VolumeConstructionRequest::Volume {
            id: sv_volume_id,
            block_size: BLOCK_SIZE as u64,
            sub_volumes: new_sub_vol.clone(),
            read_only_parent: Some(rop),
        };
```

integration_tests/src/lib.rs
```diff
@@ -5695,6 +5695,140 @@ mod test {
...
+
+        // Our "new" VCR must have a new downstairs in the opts, and have
+        // the generation number be larger than the original.
+        let replacement: VolumeConstructionRequest =
+            VolumeConstructionRequest::Volume {
+                id: sv_volume_id,
+                block_size: BLOCK_SIZE as u64,
+                sub_volumes: new_sub_vol.clone(),
+                read_only_parent: Some(new_rop),
+            };
```
mkeeter, line 5816:
```suggestion
        let replacement = VolumeConstructionRequest::Volume {
            id: sv_volume_id,
            block_size: BLOCK_SIZE as u64,
            sub_volumes: new_sub_vol.clone(),
            read_only_parent: Some(new_rop),
        };
```

upstairs/src/volume.rs
```diff
@@ -1238,14 +1234,130 @@ impl Volume {
...
+                }
+            } else {
+                // New read_only_parent is none, so the read_only_parents can't
+                // be considered for replacement.
+                Err(CrucibleError::ReplaceRequestInvalid(
+                    "read_only_parent does not exist in replacement"
+                        .to_string(),
+                ))
+            }
+        };
```
mkeeter, line 1279: Nit: I would find this more readable as a `match` statement, e.g.

```suggestion
        let rop_result = match (n_read_only_parent, o_read_only_parent) {
            (Some(..), None) => {
                // We support a VCR where a read_only_parent has gone away,
                // but we should never see addition of a read_only_parent
                // to a VCR where there was not one originally.
                crucible_bail!(
                    ReplaceRequestInvalid,
                    "Unexpected read_only_parent on new VCR"
                )
            }
            (Some(new_rop), Some(old_rop)) if old_rop == new_rop => {
                // If the read_only_parents are the same, then this is
                // technically an error (from the ROP replacement point of
                // view) as it means the targets and generation numbers
                // are the same as what we have.
                Err(CrucibleError::ReplaceRequestInvalid(
                    "read_only_parent generation numbers match".to_string(),
                ))
            }
            (Some(new_rop), Some(old_rop)) => {
                // So, the ROPs are different.  Now let's be sure
                // they are different in only a target.
                Self::compare_vcr_region_for_replacement(&old_rop, &new_rop)
            }
            (None, _) => {
                // New read_only_parent is none, so the read_only_parents can't
                // be considered for replacement.
                Err(CrucibleError::ReplaceRequestInvalid(
                    "read_only_parent does not exist in replacement"
                        .to_string(),
                ))
            }
        };
```

upstairs/src/volume.rs
```diff
@@ -1238,14 +1234,130 @@ impl Volume {
...
+                match rop_result {
+                    Ok(Some((ro, rn))) => {
+                        // Just the ROP has changes, return those.
+                        Ok(Some((ro, rn)))
+                    }
+                    Ok(None) => Ok(None),
+                    Err(_) => Err(e),
+                }
+            }
+        }
```
mkeeter, line 1338: I'd recommend matching against the tuple of `(sub_volume_result, rop_result)`, to make more clear what cases are being handled.  Here's my translation of the current code:
```suggestion
        // So, now time for the truth.
        // We use the same function to compare both the region in the
        // sub_volumes and the region in the read_only_parent.  Only one
        // can be different, so we walk the cases.
        match (sub_volume_result, rop_result) {
            (Ok(Some(_)), Ok(Some(_))) => {
                // Both have changes, this is no good
                crucible_bail!(
                    ReplaceRequestInvalid,
                    "Both sub_volume and read_only_parent have changes"
                )
            }
            (r @ Ok(Some(_)), _) => {
                // Return what the subvolume found
                r
            }
            // The sub_volume VCR is the same as what we have, but with
            // updated generation numbers, so this qualifies for the
            // Ok(None) state.  Check what the ROP found.
            (Ok(None), r @ Ok(Some(_))) => {
                // Just the ROP has changes, so let's return those.
                r
            }
            (Ok(None), _) => Ok(None),

            // The sub_volume replacement is not acceptable.  Check if
            // the read_only_parent has a valid change.  If not, then
            // return the error from the sub_volume.
            (Err(_), r @ Ok(_)) => r,
            (Err(e), _) => Err(e),
        }
```

upstairs/src/volume.rs
leftwo: I've refactored a bunch of this and it now should cover all cases as well as
(I hope) make better sense.

integration_tests/src/lib.rs
leftwo: Done

integration_tests/src/lib.rs
leftwo: Done

upstairs/src/volume.rs
leftwo: I like the idea, but I did a big refactor here and this is different now.  This logic
has moved inside the `compare_vcr_region_for_replacement()` function

upstairs/src/volume.rs
leftwo: This got a little more complex with the refactor.  There are now 4 possible
return states.  It's possible though that the tuple match is still more clear than
what I ended up with in the refactor.  Take a look and tell me what you think here.

upstairs/src/volume.rs
mkeeter: Okay, I'll take a look and the new code now

upstairs/src/volume.rs
mkeeter: Okay, I'll read the new code and start a new thread

integration_tests/src/lib.rs
```diff
@@ -5694,6 +5694,138 @@ mod test {
...
+                block_size: BLOCK_SIZE as u64,
+                sub_volumes: vec![VolumeConstructionRequest::Region {
+                    block_size: BLOCK_SIZE as u64,
+                    blocks_per_extent: tds.blocks_per_extent(),
+                    extent_count: tds.extent_count(),
+                    opts: opts.clone(),
+                    gen: 1,
+                }],
+                read_only_parent: None,
+            };
```
mkeeter, line 5721:
```suggestion
        let original = VolumeConstructionRequest::Volume {
            id: volume_id,
            block_size: BLOCK_SIZE as u64,
            sub_volumes: vec![VolumeConstructionRequest::Region {
                block_size: BLOCK_SIZE as u64,
                blocks_per_extent: tds.blocks_per_extent(),
                extent_count: tds.extent_count(),
                opts: opts.clone(),
                gen: 1,
            }],
            read_only_parent: None,
        };
```

upstairs/src/volume.rs
```diff
@@ -42,6 +42,18 @@ fn build_region_definition(
     Ok(region_def)
 }
 
+#[derive(Debug)]
```
mkeeter, line 45:
```suggestion
/// Result of comparing two [`VolumeConstructRequest`] values
#[derive(Debug)]
```

upstairs/src/volume.rs
```diff
@@ -1238,25 +1246,167 @@ impl Volume {
...
+                warn!(log, "read_only_parent error: {}", e);
+                return Err(e);
+            }
+            Ok(rr) => rr,
+        };
+
+        // Determine the sub_volume difference if there is any.
+        let sub_volume_result = match Self::compare_vcr_region_for_replacement(
+            Some(Box::new(o_sub_volumes[0].clone())),
+            Some(Box::new(n_sub_volumes[0].clone())),
```
mkeeter, line 1268: How about making `compare_vcr_region_for_replace` take an `Option<&VolumeConstructionRequest>`, to avoid these shenanigans?  Here's the complete diff:
```diff
diff --git a/upstairs/src/volume.rs b/upstairs/src/volume.rs
index 782dd6d0c2...841d0d90f7 100644
--- a/upstairs/src/volume.rs
+++ b/upstairs/src/volume.rs
@@ -1252,8 +1252,8 @@
         // update, only one of the sub_volume or read_only_parent targets
         // can change.
         let rop_result = match Self::compare_vcr_region_for_replacement(
-            o_read_only_parent,
-            n_read_only_parent,
+            o_read_only_parent.as_deref(),
+            n_read_only_parent.as_deref(),
         ) {
             Err(e) => {
                 warn!(log, "read_only_parent error: {}", e);
@@ -1264,8 +1264,8 @@

         // Determine the sub_volume difference if there is any.
         let sub_volume_result = match Self::compare_vcr_region_for_replacement(
-            Some(Box::new(o_sub_volumes[0].clone())),
-            Some(Box::new(n_sub_volumes[0].clone())),
+            Some(&o_sub_volumes[0]),
+            Some(&n_sub_volumes[0]),
         ) {
             Err(e) => {
                 warn!(log, "sub_volume has invalid differences: {}", e);
@@ -1364,8 +1364,8 @@
     // Any other difference is an error, and an error returned here means the
     // VCRs are incompatible in a way that prevents one from replacing another.
     fn compare_vcr_region_for_replacement(
-        opt_o_vol: Option<Box<VolumeConstructionRequest>>,
-        opt_n_vol: Option<Box<VolumeConstructionRequest>>,
+        opt_o_vol: Option<&VolumeConstructionRequest>,
+        opt_n_vol: Option<&VolumeConstructionRequest>,
     ) -> Result<VCRDelta, CrucibleError> {
         if opt_n_vol.is_some() {
             if opt_o_vol.is_none() {
```

upstairs/src/volume.rs
```diff
@@ -1238,25 +1246,167 @@ impl Volume {
...
+    // Any other difference is an error, and an error returned here means the
+    // VCRs are incompatible in a way that prevents one from replacing another.
+    fn compare_vcr_region_for_replacement(
+        opt_o_vol: Option<Box<VolumeConstructionRequest>>,
+        opt_n_vol: Option<Box<VolumeConstructionRequest>>,
+    ) -> Result<VCRDelta, CrucibleError> {
+        if opt_n_vol.is_some() {
+            if opt_o_vol.is_none() {
+                // We support a VCR where a read_only_parent has gone away,
+                // but we should never see addition of a read_only_parent
```
mkeeter, line 1373: Is this comment still meaningful, considering we also use `compare_vcr_region_for_replacement` to check subregions?

upstairs/src/volume.rs
```diff
@@ -1238,25 +1246,167 @@ impl Volume {
...
+            n_read_only_parent,
+        ) {
+            Err(e) => {
+                warn!(log, "read_only_parent error: {}", e);
+                return Err(e);
+            }
+            Ok(rr) => rr,
+        };
+
+        // Determine the sub_volume difference if there is any.
```
mkeeter, line 1265: Add an assertion that `o_sub_volumes.len() == 1` and `n_sub_volumes.len() == 1`?

upstairs/src/volume.rs
```diff
@@ -1238,25 +1246,167 @@ impl Volume {
...
+                // We support a VCR where a read_only_parent has gone away,
+                // but we should never see addition of a read_only_parent
+                // to a VCR where there was not one originally.
+                crucible_bail!(
+                    ReplaceRequestInvalid,
+                    "VCR added where there should not be one"
+                )
+            }
+        } else {
+            // New VCR is none. When comparing read_only_parents this is
```
mkeeter, line 1381: Same question here, the comment refers to "comparing read_only_parents" but this is also used for subretions

upstairs/src/volume.rs
```diff
@@ -1238,25 +1246,167 @@ impl Volume {
...
+            // okay.  We indicate that with this return status and let the
+            // caller decide what action to take
+            return Ok(VCRDelta::NewMissing);
+        }
+
+        // To get here, both VCRs are populated.
+        let o_vol = opt_o_vol.unwrap();
+        let n_vol = opt_n_vol.unwrap();
```
mkeeter, line 1389: I still prefer a tuple match here, which is also shorter:
```suggestion
        let (o_vol, n_vol) = match (opt_o_vol, opt_n_vol) {
            (.., None) => {
                // New VCR is none. When comparing read_only_parents this is
                // okay.  We indicate that with this return status and let the
                // caller decide what action to take
                return Ok(VCRDelta::NewMissing);
            }
            (None, Some(..)) => 
                // We support a VCR where a read_only_parent has gone away,
                // but we should never see addition of a read_only_parent
                // to a VCR where there was not one originally.
                crucible_bail!(
                    ReplaceRequestInvalid,
                    "VCR added where there should not be one"
                )
            }
            (Some(o_vol), Some(n_vol)) => (o_vol, n_vol),
        };
```

upstairs/src/volume.rs
```diff
@@ -1238,25 +1246,167 @@ impl Volume {
...
+            sub_volume_result,
+            rop_result
+        );
+
+        // We use the same function to compare both the region in the
+        // sub_volumes and the region in the read_only_parent.  Only one
+        // can have a different target, so we walk the cases.  As some
+        // deltas are okay for a read only parent that are not for
+        // a sub_volume, we need to look at all the possible permutations.
+        match sub_volume_result {
```
mkeeter, line 1289: I find tuple matches to be more clear than nested matches, e.g. `match (sub_volume_result, rop_result)`

integration_tests/src/lib.rs
leftwo: I went through this file and removed all the unneeded `: VolumeConstructionRequest`

leftwo: I also went through `/upstairs/src/volume.rs` and changed all of those.  It added a bunch of noise to the
tests, but there should be no functional change to any of them.

upstairs/src/volume.rs
leftwo: Done

upstairs/src/volume.rs
leftwo: It is valid in the case it describes, but could be better for the `sub_volume` case, I'll reword it.

upstairs/src/volume.rs
leftwo: We do check that the sub_volume lengths are 1 above. (line ~1244).

upstairs/src/volume.rs
leftwo: I've updated the comment to take into account both options.

upstairs/src/volume.rs
leftwo: Done in latest commit

upstairs/src/volume.rs
leftwo: I made this section the tuple match, let me know if that works for you.

PR #1284, opened 2024-04-30
DTrace probes for IO on/off the network

leftwo:
Put in DTrace probes for when a message is put on the wire to a downstairs, and a corresponding 
probe when a message is pulled off the wire from a downstairs.

Updated the README, and moved a few D scripts around.

Left the existing probes that were previously marking networking but had drifted off to measuring 
something else where they were, but updated the comments to reflect their new role.

Added a new D script, `perf-ds-net.d`, for the new probes.  Moved the old `perfdw.d` to be
`perf-ds-client.d`.

### DIFF
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -2665,6 +2665,7 @@ impl ClientIoTask {
             self.response_tx.clone(),
             fr,
             self.log.clone(),
+            self.client_id,
         )));
 
         let mut ping_interval = deadline_secs(PING_INTERVAL_SECS);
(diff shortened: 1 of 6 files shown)


upstairs/src/client.rs
```diff
@@ -2833,6 +2837,44 @@ where
     }
 }
 
+fn update_net_start_probes(m: &Message, cid: ClientId) {
+    match m {
+        Message::ReadRequest { ref job_id, .. } => {
```
mkeeter, line 2842: Nit: all 8x of these `ref` are unnecessary, it compiles fine without them 😄

mkeeter (approved):
Thanks for throwing this together!

PR #1286, opened 2024-05-01
Use MessageWriter downstairs

mkeeter:
While staring at flamegraphs, I noticed something surprising:

Screenshot 2024-05-01 at 3 00 31 PM

The whole point of #1206 was to reorganize messages so that bulk data could be sent down the wire in-place, without an intermediate copy.  However, it turns out that I forgot to actually use the new `MessageWriter` in the Crucible Downstairs!

±15 LOC later, the flamegraph is no longer problematic:

Screenshot 2024-05-01 at 3 02 55 PM

(although I'm giving that `memset` call some side-eye)

It's hard to tell if there's a performance impact, but this still seems worthwhile – if only to remove another big chunk of the flamegraph.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -19,20 +19,20 @@ use crucible_common::{
     verbose_timeout, Block, CrucibleError, RegionDefinition, MAX_BLOCK_SIZE,
 };
 use crucible_protocol::{
-    BlockContext, CrucibleDecoder, CrucibleEncoder, JobId, Message,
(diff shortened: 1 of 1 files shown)


faithanalog:
at this point MessageWriter is just doing the same wire-format as the FramedWrite but faster, so this isn't a protocol change, right?

mkeeter:
Yup, that's correct – it's sending the same bytes on the wire, but manually splitting up the header versus data (rather than serializing a full `Message`).

faithanalog (approved):

jmpesp (approved):

PR #1267, opened 2024-04-19
Remove SQLite backend

mkeeter:
The time is near!

(closes #1265)

PR #1287, opened 2024-05-01
Use uninitialized memory when doing reads

mkeeter:
While fixing #1286, I noticed that 8% of on-CPU time being spend doing `memset` in `read_into`:

Screenshot 2024-05-01 at 4 59 46 PM

This is unnecessary work: the buffer has sufficient capacity, and we're about to overwrite that region of the buffer.

In this PR, we switch to using `libc::pread` into the uninitialized portion of the buffer, then manually set the length with `set_len`.  Sure enough, this removes the `memset`:

Screenshot 2024-05-01 at 5 06 31 PM

Note that the new code is identical between `extent_inner_sqlite.rs` and `extent_inner_raw.rs`, because they have the same data format; since we're planning to remove SQLite in #1267 , I don't feel the need to refactor into helper functions.

There's also a drive-by fix revealed by the new assertion: we need to reattach the remaining `buf` to `out.data` (by calling `unsplit(..)`); otherwise, it will force a reallocation.  This will only trigger for reads which span multiple extents, so it's probably not a big deal in the wild, but let's do the correct thing instead of the wrong thing.

### DIFF
--- a/downstairs/src/extent_inner_raw.rs
+++ b/downstairs/src/extent_inner_raw.rs
@@ -223,7 +223,7 @@ impl ExtentInner for RawInner {
             // memory in the buffer.  This is fine; we sized the buffer
             // appropriately in advance (and will panic here if we messed up).
             let expected_bytes = n_contiguous_blocks * block_size as usize;
-            assert!(buf.capacity() >= expected_bytes);
+            assert!(buf.spare_capacity_mut().len() >= expected_bytes);
 
             let first_resp = &out.blocks[resp_run_start];
             check_input(self.extent_size, first_resp.offset, &buf)?;
--- a/downstairs/src/extent_inner_raw.rs
+++ b/downstairs/src/extent_inner_raw.rs
@@ -21,7 +21,7 @@ use slog::{error, Logger};
 use std::collections::HashSet;
 use std::fs::{File, OpenOptions};
 use std::io::{BufReader, Read};
-use std::os::fd::AsFd;
+use std::os::fd::{AsFd, AsRawFd};
 use std::path::Path;
 
 /// Equivalent to `DownstairsBlockContext`, but without one's own block number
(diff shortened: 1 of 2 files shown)


downstairs/src/extent_inner_sqlite.rs
```diff
@@ -342,12 +342,37 @@ impl SqliteMoreInner {
                 (job_id.0, self.extent_number, n_contiguous_blocks as u64)
             });
 
-            nix::sys::uio::pread(
```
mkeeter, line 345: For some reason, the SQLite backend wasn't checking the returned number of bytes here.

downstairs/src/extent_inner_raw.rs
```diff
@@ -233,15 +233,21 @@ impl ExtentInner for RawInner {
...
-                &mut buf,
-                first_req.offset.value as i64 * block_size as i64,
-            )
-            .map_err(|e| {
+            // SAFETY: the buffer has sufficient capacity, and this is a valid
+            // file descriptor.
+            let r = unsafe {
+                libc::pread(
+                    self.file.as_raw_fd(),
+                    buf.spare_capacity_mut().as_mut_ptr().cast(),
```
hawkw, line 241: extremely nitpicky, take it or leave it: personally, i try to avoid the use of un-turbofished `ptr::cast()`, as it obscures the type being cast to; i would generally either pronounce this as
```suggestion
                    buf.spare_capacity_mut().as_mut_ptr().cast::<libc::c_void>(),
```
or
```suggestion
                    buf.spare_capacity_mut().as_mut_ptr() as *mut libc::c_void,
```

but, feel free to ignore me!

downstairs/src/extent_inner_sqlite.rs
```diff
@@ -342,12 +342,37 @@ impl SqliteMoreInner {
...
-                &mut buf,
-                first_req.offset.value as i64 * block_size as i64,
-            )
-            .map_err(|e| CrucibleError::IoError(e.to_string()))?;
+            // SAFETY: the buffer has sufficient capacity, and this is a valid
+            // file descriptor.
+            let r = unsafe {
+                libc::pread(
+                    self.file.as_raw_fd(),
+                    buf.spare_capacity_mut().as_mut_ptr().cast(),
```
hawkw, line 350: similarly, IMHO, `ptr::cast()` without a type annotation obscures the type being cast to...

hawkw (approved):

downstairs/src/extent_inner_raw.rs
mkeeter: Sure, done in 028153cafc22c275282c946a6d0c131fcd41be76

PR #1295, opened 2024-05-13
Speed up `crutest fill`

mkeeter:
The compiler faceplants at optimizing our current `fill_vec` implementation:

Screenshot 2024-05-13 at 6 03 07 PM

Writing directly into the uninitialized buffer makes filling about 30% faster.  Testing on a 128 GiB region, filling goes from 4:05 → 2:43; verification remains the same at about 2:30.

Screenshot 2024-05-13 at 6 04 18 PM

### DIFF
--- a/crutest/src/main.rs
+++ b/crutest/src/main.rs
@@ -1342,16 +1342,22 @@ fn fill_vec(
      * the write buffer with the correct seed value.
      */
     let mut vec = BytesMut::with_capacity(blocks * bs as usize);
-    for block_offset in block_index..(block_index + blocks) {
-        /*
-         * The start of each block contains that blocks index mod 255
-         */
-        vec.extend(&[(block_offset % 255) as u8]);
-        /*
-         * Fill the rest of the buffer with the new write count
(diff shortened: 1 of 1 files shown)


crutest/src/main.rs
```diff
@@ -1342,16 +1342,22 @@ fn fill_vec(
...
-        /*
-         * The start of each block contains that blocks index mod 255
-         */
-        vec.extend(&[(block_offset % 255) as u8]);
-        /*
-         * Fill the rest of the buffer with the new write count
-         */
+    for (block_offset, chunk) in (block_index..(block_index + blocks))
+        .zip(vec.spare_capacity_mut().chunks_mut(bs as usize))
+    {
```
leftwo, line 1347: We should keep the comment here that the start of the block is the block's index
mod 255.

leftwo (approved):

crutest/src/main.rs
mkeeter: good catch, done

PR #1294, opened 2024-05-13
Improve deferred write encryption

mkeeter:
- Uses a per-volume pool, instead of a global pool (#1285)
- Make encryption infallible; because the only failure conditions are for sizes that we can't hit (and this makes the next step easier)
- Don't defer encryption for small writes, using the same logic as reads

PR #1308, opened 2024-05-14
Reduce Rayon pool size for Crucible Downstairs

mkeeter:
The Downstairs equivalent to #1294, but even simpler.

### DIFF
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -15,6 +15,15 @@ use crucible_common::*;
 use crucible_protocol::SnapshotDetails;
 use repair_client::Client;
 
+/// Number of worker threads in the Rayon thread pool
+///
+/// This is only used when synching extents to disk; we spawn separate Rayon
+/// tasks to call `fdsync` on each extent in parallel.  The value is picked
+/// somewhat arbitrarily as a number that doesn't seem to affect performance,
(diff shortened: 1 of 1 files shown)


leftwo (commented):
The changes look all good.
Do we have any perf numbers for this to see if it has any impact?
With all the other changes flowing in at the moment, it might make
sense to just merge them all and then run another pass and see
how we are and fine tune as necessayr.

mkeeter:
Doing a sweep with `crutest rand-write`, it looks _very slightly_ worse than `main`.

I suspect that saving 120 threads a' spinning is still probably it, especially for heavily-loaded systems.

Screenshot 2024-05-15 at 1 25 39 PM

faithanalog (approved):
all my homies hate contention. we don't need rayon at all downstairs once we get the FIOFFS-equivalent working for flush right? or do we rayon the reads and writes too?

leftwo (approved):

mkeeter:
> all my homies hate contention. we don't need rayon at all downstairs once we get the FIOFFS-equivalent working for flush right? or do we rayon the reads and writes too?

Correct, although it _may_ stick around for the non-illumos systems?

(now that I say that, I'm questioning whether it's necessary, since our performance/complexity tradeoff curve is different on non-Gimlet hardware...)

PR #1309, opened 2024-05-14
Added more DTrace scripts.

leftwo:
Additional scripts, single_up_info.d and sled_upstairs_info.d to give you a 
specific script for a specific PID, or for an entire system.

A new script, all_downstairs.d, to display IO stats for all downstairs running on a system.

Updated the README

`single_up_info.d` requires a PID and adds a `SESSION` column to tell apart different
upstairs inside a single process.
`sled_upstairs_info.d` has both a `PID` and a `SESSION` column and will print out
stats for the whole system.

`all_downstairs.d` looks like this:
```
EVT22200005 # dtrace -s /alan/dtrace/all_downstairs.d
  PID   F>   F<   W>   W<    R>    R<    WS   WIP    WD
13790   10    9 1911 1835     0     0  1943  1867  1867
25574   10   10 2204 2082     0     0  2237  2115  2114
25442   10   10 2204 2089     0     0  2236  2122  2121
  PID   F>   F<   W>   W<    R>    R<    WS   WIP    WD
17147    2    2    0    0   389   389   391   391   391
25492    2    2    0    0   389   389   391   391   391
25627    2    2    0    0   389   389   391   391   391
25442   10    9 2283 2177     0     0  2315  2207  2208
25574   10    9 2283 2184     0     0  2314  2214  2215
13790   10   10 2054 2030     0     0  2085  2061  2061
  PID   F>   F<   W>   W<    R>    R<    WS   WIP    WD
17147    2    2    2    2     0     0     4     4     4
25492    2    2    2    2     0     0     4     4     4
25627    2    2    2    2     0     0     4     4     4
13790   10   10 1961 1985     0     0  1994  2018  2018
25442   10   10 2042 2185     0     0  2074  2218  2217
25574   10   10 2045 2185     0     0  2077  2218  2217
```

### DIFF
--- a/tools/dtrace/README.md
+++ b/tools/dtrace/README.md
@@ -1,5 +1,46 @@
 # Oxide DTrace Crucible scripts
 
+## all_downstairs.d
+A DTrace script to show IOs coming and going on all downstairs as well as the
+work task geting new work, performing the work and completing the work. Stats
+are printed at a 4 second interval.
+
+The columns show counts in the last 4 seconds of:
+F>  Flush coming in from the upstairs
+F<  Flush completed message being sent back to the upstairs.
+W>  Write coming in from the upstairs
+W<  Write completed message being sent back to the upstairs.
+R>  Read coming in from the upstairs
+R<  Read completed message being sent back to the upstairs.
+WS  An IO has been submitted to the work task in the downstairs
+WIP An IO is taken off the work queue by the downstairs work task.
+WD  An IO is completed by the downstairs work task.
+
+If a downstairs has not done any IOs it will either print no line, or
+print a line of zeros.
+
+```
+EVT22200005 # dtrace -s /alan/dtrace/all_downstairs.d
+  PID   F>   F<   W>   W<    R>    R<    WS   WIP    WD
+13790   10    9 1911 1835     0     0  1943  1867  1867
+25574   10   10 2204 2082     0     0  2237  2115  2114
+25442   10   10 2204 2089     0     0  2236  2122  2121
+  PID   F>   F<   W>   W<    R>    R<    WS   WIP    WD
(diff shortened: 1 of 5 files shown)


tools/dtrace/README.md
```diff
@@ -1,5 +1,46 @@
...
+W>  Write coming in from the upstairs
+W<  Write completed message being sent back to the upstairs.
+R>  Read coming in from the upstairs
+R<  Read completed message being sent back to the upstairs.
+WS  An IO has been submitted to the work task in the downstairs
+WIP An IO is taken off the work queue by the downstairs work task.
+WD  An IO is completed by the downstairs work task.
+
+If a downstairs has not done any IOs it will either print no line, or
+print a line of zeros.
```
faithanalog, line 20: How would it be that it prints no line? I don't understand from reading the dtrace code how that would happen instead of printing a line of zeroes.

also I think it would be ideal if it always printed a line for each time period if that's possible instead of dropping lines

leftwo: If none of the probes have fired at all, then dtrace won't know about them and will have nothing to print,
but that only happens when the script first starts.  I had a different version of the script that would just
run for 5 seconds and then print a line and exit, and the comment was more applicable there.  I did want
to leave the comment just in case someone ran the script for 10 seconds, hit control-C, they might not
get any output if all the downstairs were idle.

Once any of the downstairs probes in the script has fired, we populate the variable in the script and it
 will print zeros (or new values) for that PID going forward.  It's just we need one IO to prime the pump.

faithanalog: > but that only happens when the script first starts ...  It's just we need one IO to prime the pump.

ah perfect, seems fine to me

faithanalog (approved):

mkeeter (approved):

PR #1310, opened 2024-05-15
Unified `ExtentId` type

mkeeter:
Across the codebase, we refer to extent IDs as either `u32`, `u64`, or `usize`.

This PR switches to an `ExtentId(pub u32)` type, and uses it everywhere.  It is a **breaking change** to the message protocol, because extent IDs were previously encoded as `u64`'s in messages.

For consistency, it also changes `extent_count: u64` to `extent_count: u32` in various places.

This is Episode 5 of a long-running series; previously:

- https://github.com/oxidecomputer/crucible/pull/919
- https://github.com/oxidecomputer/crucible/pull/925
- https://github.com/oxidecomputer/crucible/pull/928
- https://github.com/oxidecomputer/crucible/pull/1076


upstairs/src/downstairs.rs
```diff
@@ -3011,15 +3016,15 @@ impl Downstairs {
                 }
                 IOop::ExtentFlushClose { extent, .. } => {
                     let job_type = "FClose".to_string();
-                    (job_type, *extent)
+                    (job_type, extent.0 as usize) // XXX this isn't blocks!
```
jmpesp, line 3019: Hmm, make it zero instead?

jmpesp (approved):
10/10, would watch episode 5 again

upstairs/src/downstairs.rs
mkeeter: Yup, done.

We noted in chat that `ioop_summary` is also returning the ambiguous blocks / extent field, but it's printed as "BL/E" so I left it alone for now.

downstairs/src/dump.rs
```diff
@@ -114,7 +114,7 @@ pub async fn dump_region(
      *       compare every time.
      */
     if let Some(ce) = cmp_extent {
-        if ce >= total_extents {
+        if ce.0 >= total_extents {
```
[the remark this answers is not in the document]
mkeeter: I thought about it, but ended up leaning towards **not** having automatic conversions.  Quoth [the docs](https://doc.rust-lang.org/std/ops/trait.Deref.html#when-to-implement-deref-or-derefmut):

> In general, deref traits **should** be implemented if:
>
> 1. a value of the type transparently behaves like a value of the target type;
> 2. the implementation of the deref function is cheap; and
> 3. users of the type will not be surprised by any deref coercion behaviour.

I don't think we want `ExtentId` to always behave like a `u32`, and the automatic conversion would be surprising (to me at least).

upstairs/src/downstairs.rs
leftwo: Yeah, that was me being lazy and re-using a column instead of making a new one.
All the repair operations overload that blocks column and use it for reporting extent id.

faithanalog (approved):
i like it

upstairs/src/downstairs.rs
leftwo: Rather than just print 0 in this column, why not change the column to be 'BL/E' like `ioop_summary` so we
can still print and see the information here?

mkeeter: Okay, done in 43c97ee42704 (using `BKS/EXT` as the column header, since we have 7 characters to play with)

PR #1313, opened 2024-05-16
Downstairs task cleanup

mkeeter:
This is standalone preparation for a single-task refactoring of the Crucible downstairs.

The goal of this PR is to reduce places where we take an `Arc<Mutex<Downstairs>>`.

There are four things happening in the PR, split into four commits:

- Merge `pf_task` and `dw_task`, removing the job channel doorbell
- Consolidate flags into a new `DownstairsFlags` struct (because the `Downstairs` was getting unwieldy)
- Add a Downstairs task runner, which listens on a queue and hands Downstairs-specific actions
  - Eventually, this will be the single-task event handler!
  - For now, we're just using it for the repair API, removing their `Arc<Mutex<Downstairs>>` argument
- Move `clone_region` into a method on the `Downstairs`, removing its `Arc<Mutex<Downstairs>>` argument


downstairs/src/lib.rs
```diff
@@ -1467,9 +1439,9 @@ where
      *        │                 │errors, pings     │responses
      *        │                 │                  │
      *        │                 │                  │
-     *   ┌────▼────┐ message   ┌┴──────┐  job     ┌┴────────┐
-     *   │resp_loop├──────────►│pf_task├─────────►│ dw_task │
-     *   └─────────┘ channel   └──┬────┘ channel  └▲────────┘
+     *   ┌────▼────┐ message   ┌┴──────────────────┴────────┐
+     *   │resp_loop├──────────►│        pf_task             │
+     *   └─────────┘ channel   └──┬─────────────────────────┘
```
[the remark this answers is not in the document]
mkeeter: Good catch, done in 68d4afcde80a

mkeeter:
> I wonder if this will have any impact on performance.

Update: it's bad!

read_compare

write_compare

This is testing 1f7683d384abd91bab6c3eee90b18d55dac7e2c4 (this branch) against current main (a3b920bd), using `crutest rand-read/write` on the London minicluster (128 GiB disk, 4K blocks, 16384 blocks per extent, encrypted).

I'm puzzled as to why this is worse; I had though that all the expensive (IO) work was _already_ done while holding the Downstairs lock, so combining tasks shouldn't have a noticeable impact.  Clearly, I'm missing something in my mental model.

downstairs/src/lib.rs
```diff
@@ -1727,23 +1686,39 @@ impl DownstairsBuilder<'_> {
             region.def().extent_count(),
         );
 
+        let (request_tx, request_rx) = mpsc::channel(256); // TODO size?
```
jmpesp, line 1689: Probably need to revisit this, unless this channel goes away in a future commit?

jmpesp (approved):
Looks good

downstairs/src/lib.rs
mkeeter: Good catch; I switch to an `unbounded_channel` later on in this stack of commits (because it's eventually limited by Upstairs backpressure)

PR #1314, opened 2024-05-22
Update crutest replace test, and mismatch printing.

leftwo:
Updated the replace test to better reflect replacing a downstairs when IO is underway.  We spawn a task to do the replacement and wait for it to finish while the main task just sends IO in a loop.

Updated the CI tests for live repair to make a little larger region so we can catch IO in progress while doing live repair. Updated the test to also collect dtrace stats.

Updated the crutest block mismatch printing code to detect a mismatch range and just print a summary at the end of the range instead of printing a line for every byte that does not match what is expected.
Here is an example mismatch where I have written one more IO to 5 blocks than the verify file expects:
(recreating an extra write, though a missing write would look similar)
```
Mismatch     Block::0 bo:1-511 Volume offset:0-0  Expected:12 Got:13
Mismatch     Block::1 bo:1-511 Volume offset:512-512  Expected:19 Got:20
Mismatch     Block::2 bo:1-511 Volume offset:1024-1024  Expected:27 Got:28
Mismatch     Block::3 bo:1-511 Volume offset:1536-1536  Expected:30 Got:31
Mismatch     Block::4 bo:1-511 Volume offset:2048-2048  Expected:31 Got:32
Error in block range 0 -> 8
```

Here is an example mismatch when I've corrupted block 2 and 3 with the
contents from block 6 and 7 (recreating a write to the wrong location)
```
Mismatch Block Index Block:2 Offset:1024 Expected:2 Got:6
Mismatch     Block::2 bo:1-511 Volume offset:1024-1024  Expected:30 Got:26
Mismatch Block Index Block:3 Offset:1536 Expected:3 Got:7
Mismatch     Block::3 bo:1-511 Volume offset:1536-1536  Expected:35 Got:20
Error in block range 0 -> 8
```

### DIFF
--- a/crutest/src/main.rs
+++ b/crutest/src/main.rs
@@ -25,6 +25,7 @@ use signal_hook::consts::signal::*;
 use signal_hook_tokio::Signals;
 use tokio::sync::mpsc;
 use tokio::time::{Duration, Instant};
+use tokio_util::sync::CancellationToken;
 use uuid::Uuid;
 
 mod cli;
@@ -1087,10 +1088,7 @@ async fn main() -> Result<()> {
             // Build the list of targets we use during the replace test.
             // The first three are provided in the crucible opts, and the
             // final one (the replacement) is a test specific arg.
-            let mut targets = Vec::with_capacity(4);
-            for t in opt.target.iter() {
-                targets.push(*t);
-            }
+            let mut targets = opt.target.clone();
             targets.push(replacement);
             replace_workload(
                 &guest,
@@ -1493,51 +1491,41 @@ fn validate_vec<V: AsRef<[u8]>>(
 
                 // Check to see if we have an error already
                 block_err = match block_err {
-                    None => Some(BlockErr {
-                        block: block_offset,
-                        starting_offset: i,
-                        ending_offset: i,
-                        starting_volume_offset: byte_offset,
-                        ending_volume_offset: byte_offset,
-                        expected_value: seed,
-                        found_value: data[data_offset + i],
-                        status: res.clone(),
-                        msg,
-                    }),
-                    Some(mut be) => {
-                        // We have an existing error, does this extend our
-                        // range or start a new one?
+                    // The error is contiguous with our current error
+                    Some(mut be)
                         if be.status == res
                             && be.found_value == data[data_offset + i]
-                            && be.ending_offset + 1 == i
-                        {
-                            be.ending_offset = i;
-                            be.ending_volume_offset = byte_offset;
-                            Some(be)
-                        } else {
-                            // Print the current range
+                            && be.ending_offset + 1 == i =>
+                    {
+                        be.ending_offset = i;
+                        be.ending_volume_offset = byte_offset;
+                        Some(be)
+                    }
+                    _ => {
+                        // Print the previous range, if it exists
+                        if let Some(be) = block_err {
                             println!("{}", be);
-                            // Start a new range.
-                            Some(BlockErr {
-                                block: block_offset,
-                                starting_offset: i,
-                                ending_offset: i,
-                                starting_volume_offset: byte_offset,
-                                ending_volume_offset: byte_offset,
-                                expected_value: seed,
-                                found_value: data[data_offset + i],
-                                status: res.clone(),
-                                msg,
-                            })
                         }
+
+                        // Start a new range.
+                        Some(BlockErr {
+                            block: block_offset,
+                            starting_offset: i,
+                            ending_offset: i,
+                            starting_volume_offset: byte_offset,
+                            ending_volume_offset: byte_offset,
+                            expected_value: seed,
+                            found_value: data[data_offset + i],
+                            status: res.clone(),
+                            msg,
+                        })
                     }
                 };
             } else {
                 // This check was valid. If we have seen previous errors,
                 // print out the summary report now.
-                if let Some(be) = block_err {
+                if let Some(be) = block_err.take() {
                     println!("{}", be);
-                    block_err = None;
                 }
             }
         }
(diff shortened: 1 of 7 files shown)


crutest/src/main.rs
```diff
@@ -1103,14 +1084,19 @@ async fn main() -> Result<()> {
...
-            // This should already be setup for us.
-            let full_target = full_target.unwrap();
+            // Build the list of targets we use during the replace test.
+            // The first three are provided in the crucible opts, and the
+            // final one (the replacement) is a test specific arg.
+            let mut targets = Vec::with_capacity(4);
+            for t in opt.target.iter() {
+                targets.push(*t);
+            }
+            targets.push(replacement);
```
mkeeter, line 1094:
```suggestion
            let mut targets = opt.target.clone();
            targets.push(replacement);
```

crutest/src/main.rs
```diff
@@ -1470,17 +1491,62 @@ fn validate_vec<V: AsRef<[u8]>>(
...
+                                starting_volume_offset: byte_offset,
+                                ending_volume_offset: byte_offset,
+                                expected_value: seed,
+                                found_value: data[data_offset + i],
+                                status: res.clone(),
+                                msg,
+                            })
+                        }
+                    }
+                };
```
mkeeter, line 1534: Nit: we can reduce duplication here by combining the `None` and non-matching cases
```suggestion
                block_err = match block_err {
                    // The error is contiguous with our current error
                    Some(mut be)
                        if be.status == res
                            && be.found_value == data[data_offset + i]
                            && be.ending_offset + 1 == i =>
                    {
                        be.ending_offset = i;
                        be.ending_volume_offset = byte_offset;
                        Some(be)
                    }
                    _ => {
                        // Print the current range, if it exists
                        if let Some(be) = block_err {
                            println!("{}", be);
                        }
                        // Start a new range.
                        Some(BlockErr {
                            block: block_offset,
                            starting_offset: i,
                            ending_offset: i,
                            starting_volume_offset: byte_offset,
                            ending_volume_offset: byte_offset,
                            expected_value: seed,
                            found_value: data[data_offset + i],
                            status: res.clone(),
                            msg,
                        })
                    }
                };
```

crutest/src/main.rs
```diff
@@ -1470,17 +1491,62 @@ fn validate_vec<V: AsRef<[u8]>>(
...
+            } else {
+                // This check was valid. If we have seen previous errors,
+                // print out the summary report now.
+                if let Some(be) = block_err {
+                    println!("{}", be);
+                    block_err = None;
```
mkeeter, line 1540: Nit: use `Option::take` here to idiomatically remove the value (setting the variable to `None` automatically)
```suggestion
                if let Some(be) = block_err.take() {
                    println!("{}", be);
```

crutest/src/main.rs
```diff
@@ -1924,77 +1990,118 @@ async fn replace_workload(
...
-            tokio::time::sleep(tokio::time::Duration::from_secs(4)).await;
+            // See (if provided) we have reached a requested number of loops
+            if let Some(stop_at) = stop_at {
+                if c >= stop_at {
+                    println!("[{c}] Replace task ends as count was reached");
+                    break;
+                }
+            }
+
+            // No stopping yet, lets do another loop.
```
mkeeter, line 2071:
```suggestion
            // No stopping yet, let's do another loop.
```

crutest/src/main.rs
```diff
@@ -1924,77 +1990,118 @@ async fn replace_workload(
...
+    // Make a copy of the stop at counter if one was provided so the
+    // IO task and the replace task don't have to share wtq
+    let stop_at = match wtq {
+        WhenToQuit::Count { count } => Some(*count),
+        _ => None,
+    };
 
-        // Wait for all IO to settle down before we continue
+    // Start up a task to do the replacement workload
+    let (work_stop_tx, mut work_stop_rx) = mpsc::channel(1);
```
mkeeter, line 2009: It looks like this is used as a oneshot; could we use `tokio::sync::oneshot::channel()` instead?

crutest/src/main.rs
```diff
@@ -1924,77 +1990,118 @@ async fn replace_workload(
...
+    let (work_count_tx, mut work_count_rx) = mpsc::channel(1);
+    let guest_c = guest.clone();
+    let handle = tokio::spawn(async move {
+        let mut c: usize = 1;
```
mkeeter, line 2013: Nit: `c` is only used as a counter variable, so the `loop {` could instead be `for c in 1.. {`

crutest/src/main.rs
mkeeter: Or even a `tokio_util::sync::CancellationToken`, which is designed to cancel things!

crutest/src/main.rs
```diff
@@ -2005,9 +2112,31 @@ async fn replace_workload(
         }
     }
 
+    // The replace task may have already exited.
+    let _ = work_stop_tx.send(true).await;
+    if let Err(e) = handle.await.unwrap() {
+        bail!("Failed in IO workload: {:?}", e);
```
mkeeter, line 2118: I'm confused by the error message here – isn't `handle` the replace workload (and the IO workload is done in the loop right above, in this same task)?

jmpesp (approved):

mkeeter (approved):

crutest/src/main.rs
leftwo: Done

crutest/src/main.rs
leftwo: I got the CancellationToken to work (and added that to my toolbox for later)

PR #1317, opened 2024-05-27
Reuse a reqwest client when creating Nexus clients

### LINKED ISSUE #1316: Notifying nexus of live repair progress updates eats way too much CPU because of creating reqwest clients
image

image

This rack is very distressed. Some disks have gone into live-repair. That distress is being made even worse because **8% of the rack's CPU is being spent notifying nexus of live-repair status**, and most of that is being spent creating Reqwest clients. this is the same class of problem as https://github.com/oxidecomputer/omicron/issues/5717 which was fixed by https://github.com/oxidecomputer/omicron/pull/5750, but we are having it over here too.

Fixing this requires caching the **Reqwest client**, but it **does not require caching the nexus client**. This is an important distinction because caching the nexus client would cause problems when nexus moves around. But caching the *reqwest* client is fine, because it's just the thing that does HTTP requests, it doesn't care where they're going.

In short, we want to be calling `nexus_client::Client::new_with_client()` instead of `nexus_client::Client::new()` here https://github.com/oxidecomputer/crucible/blob/da333e91484652e3d4a4a73b94a11e5409820e04/upstairs/src/lib.rs#L1816-L1819

A Reqwest client is internally Arc'd, so we just need one laying around that we can clone every time we make a nexus client

@iliana tagging you to make sure I'm explaining this right

jmpesp:
Creating Reqwest clients is expensive, and creating them every time Nexus is notified is therefore an awful idea. Reuse a single Reqwest client instead.

Fixes #1316.

### DIFF
--- a/upstairs/src/downstairs.rs
+++ b/upstairs/src/downstairs.rs
@@ -131,6 +131,9 @@ pub(crate) struct Downstairs {
     ///
     /// This must be handled after every event
     ackable_work: BTreeSet<JobId>,
+
+    /// A reqwest client, to be reused when creating Nexus clients
+    reqwest_client: reqwest::Client,
 }
 
 /// Helper struct to contain a count of backpressure bytes
@@ -338,6 +341,11 @@ impl Downstairs {
             log: log.new(o!("" => "downstairs".to_string())),
             ackable_work: BTreeSet::new(),
             repair: None,
+            reqwest_client: reqwest::ClientBuilder::new()
+                .connect_timeout(std::time::Duration::from_secs(15))
+                .timeout(std::time::Duration::from_secs(15))
+                .build()
+                .unwrap(),
         }
     }
 
@@ -3951,9 +3959,10 @@ impl Downstairs {
         // Spawn a task so we don't block the main loop talking to
         // Nexus.
         let target_addrs = self.get_target_addrs();
+        let client = self.reqwest_client.clone();
         tokio::spawn(async move {
             let Some(nexus_client) =
-                get_nexus_client(&log, &target_addrs).await
+                get_nexus_client(&log, client, &target_addrs).await
             else {
                 // Exit if no Nexus client returned from DNS - our notification
                 // is best effort.
@@ -4031,9 +4040,10 @@ impl Downstairs {
         // Spawn a task so we don't block the main loop talking to
         // Nexus.
         let target_addrs = self.get_target_addrs();
+        let client = self.reqwest_client.clone();
         tokio::spawn(async move {
             let Some(nexus_client) =
-                get_nexus_client(&log, &target_addrs).await
+                get_nexus_client(&log, client, &target_addrs).await
             else {
                 // Exit if no Nexus client returned from DNS - our notification
                 // is best effort.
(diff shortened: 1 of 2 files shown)


faithanalog:
this looks right to me. I want to see if I can get a before/after flamegraph under more controlled conditions than the original issue to conclusively prove that it's doing the right thing, but from reading the code I'm pretty sure it is.

upstairs/src/downstairs.rs
```diff
@@ -131,6 +131,9 @@ pub(crate) struct Downstairs {
     ///
     /// This must be handled after every event
     ackable_work: BTreeSet<JobId>,
+
+    /// A reqwest client, to be reused when creating Nexus clients
+    reqwest_client: reqwest::Client,
```
smklein, line 136: This connects you to a *single* nexus, right? Wouldn't you want to be connected to a set of Nexuses, and use "whichever one is available"?

I'd be happy to talk through integration details, but this seems like a great fit for https://github.com/oxidecomputer/qorb

jmpesp: I don't think so, I think this just creates the reqwest client, which is then cloned and used in `get_nexus_client` which will select a Nexus based on DNS at the time the notification needs to go out.

smklein: (also, I'm not trying to say "skip this PR" -- I still think this is a worthwhile measurable improvement -- just that this use-case, and dipping the toe into caching land, seems like it has overlap with the connection pooling library more generally)

jmpesp: I do think qorb is a better fit, to be sure - for the reasons you posted, but especially since `retry_until_known_result` is used here, which will indefinitely try to talk with the Nexus that was chosen in the beginning of the spawned task. If it's gone, that's a bummer.

I'd like to talk integration details (maybe not for this PR so we can get that immediate win). Is there a basic `get_me_a_reqwest_client` function in qorb?

smklein: I just updated the README in https://github.com/oxidecomputer/qorb - to make a connection pool, you'll need to pick:

- a resolver (How do I find backends)
- a connector (How do I create/maintain connections to the backends I've found)

You can almost certainly use the default [DNS resolver](https://github.com/oxidecomputer/qorb/blob/master/src/resolvers/dns.rs), but you'll probably need to make a custom connector. Here's an example of the "TCP connector": https://github.com/oxidecomputer/qorb/blob/master/src/connectors/tcp.rs -- but without knowing the server we're talking to, I can't generically implement an "are you still alive" check. 

You can implement a connector with no health checks by just implementing the `Connector::connect` method and nothing else, which is basically:

- Input: SocketAddr, plus whatever state you have cached in your connector
- Output: A new connection (presumably, a Nexus reqwest client)

smklein: Also, just updated the docs with a little example for the pool creation itself: https://github.com/oxidecomputer/qorb/blob/f2b2b15d21ae967c234b5297b803d09f2719e038/src/pool.rs#L371-L392

faithanalog:
Tested this on my bench sled.

Before:

flamegraph-lr-before

---

After:

flamegraph-lr-after

Good! the big reqwest overhead is gone. Of the remaining CPU time, the bulk is still notifying nexus and DNS and such, but considering none of the stuff in the second flamegraph even showed up at all in the first flamegraph, probably a very slim amount of actual CPU usage.

faithanalog (approved):

PR #1318, opened 2024-05-29
Move negotiation into `Downstairs` as well

mkeeter:
This PR moves two operations into the `struct Downstairs`, without changing their behavior:

- What happens when the Upstairs disconnects (`Downstairs::on_disconnect`)
- How to handle a new negotiation message (`Downstairs::on_negotiation_step`)

State which was previously local to the function is moved into a `struct NegotiationData`, which is passed to `Downstairs::on_negotiation_step` as a mutable reference.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -911,6 +911,34 @@ pub struct UpstairsConnection {
     gen: u64,
 }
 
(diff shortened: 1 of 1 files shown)


mkeeter:
Closing in favor of #1320, so I can stack on top of it

PR #1324, opened 2024-05-29
Reuse a reqwest client when creating repair client

jmpesp:
Similar to #1317, create a reqwest client once, and clone it when creating progenitor clients.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -1704,6 +1704,11 @@ impl DownstairsBuilder<'_> {
             log,
             request_tx,
             request_rx: Some(request_rx),
+            reqwest_client: reqwest::ClientBuilder::new()
+                .connect_timeout(std::time::Duration::from_secs(15))
+                .timeout(std::time::Duration::from_secs(15))
+                .build()
+                .unwrap(),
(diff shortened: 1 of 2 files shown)


mkeeter (approved):

leftwo:
Just to be sure, this will re-use the part of a client independent of where it's connecting to, right?  Like I don't
want my current repair to connect to the previous downstairs I repaired from, as that may not be the same as
the current downstairs I need to repair from.

jmpesp:
> Just to be sure, this will re-use the part of a client independent of where it's connecting to, right? Like I don't want my current repair to connect to the previous downstairs I repaired from, as that may not be the same as the current downstairs I need to repair from.

Right - the repair (progenitor) client is what is passed the address of what downstairs to connect to, which is at a higher level than the reqwest client it's reusing.

leftwo:
I fixed the problem with the live repair test (that my PR never seemed to hit), so if you merge main into this
it should run clean.  Even if not, the failures reported are due to log files changing after the test completed
so we can safely ignore the failure.

jmpesp:
Taking this advice, merged in main.

PR #1328, opened 2024-06-03
Make the Downstairs own per-connection state

mkeeter:
(stacked on top of #1327)

This PR moves ownership of the `ConnectionState` from a local variable in `proc_task` into the `struct Downstairs`.

It is preparation for consolidating Downstairs logic into a single task, instead of the current task-per-connection architecture (in which multiple tasks share an `Arc<Mutex<Downstairs>>`).

In the `Downstairs`, per-connection state is indexed by a new `ConnectionId` type, which is a monotonically increasing counter that increments on every new connection.  We're using a new type because upstairs / session UUIDs are sometimes reused if a new connection is made, and we specifically _don't_ want to reuse `ConnectionState` between different connections to an upstairs with a maybe-matching UUID.  In the previous code, this was ensured by using local variables in `proc_task` (which was spawned on a per-connection basis); now, we ensure it by using `ConnectionId` (which is also unique on a per-connection basis).

This reorganization had a few other side effects:

- Downstairs replacement is extracted from `proc_task` into `Downstairs::on_new_connection_replacing` (with minimal logical changes)
- Message handling is extracted from `proc_task` into `Downstairs::on_message_for` (with minimal logical changes)
- `proc_frame` now takes a `&mut Downstairs` (instead of a `Arc<Mutex<Downstairs>>`), because it is called from `on_message_for` (which already has the lock)

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -880,6 +880,7 @@ fn is_message_valid(
 
 async fn proc_stream(
     ads: &Arc<Mutex<Downstairs>>,
+    id: ConnectionId,
     stream: WrappedStream,
 ) -> Result<()> {
     match stream {
@@ -889,7 +890,7 @@ async fn proc_stream(
             let fr = FramedRead::new(read, CrucibleDecoder::new());
             let fw = MessageWriter::new(write);
 
-            proc(ads, fr, fw).await
+            proc(ads, id, fr, fw).await
         }
         WrappedStream::Https(stream) => {
             let (read, write) = tokio::io::split(stream);
@@ -897,7 +898,7 @@ async fn proc_stream(
             let fr = FramedRead::new(read, CrucibleDecoder::new());
             let fw = MessageWriter::new(write);
 
-            proc(ads, fr, fw).await
+            proc(ads, id, fr, fw).await
         }
     }
 }
@@ -910,6 +911,14 @@ pub struct UpstairsConnection {
     gen: u64,
 }
 
+/// Unique ID identifying a single connection
+///
+/// This value is assigned and incremented every time `TcpListener::accept()`
+/// receives a new connection.  Note that it is **not** the upstairs or session
+/// UUID, which may collide under certain circumstances.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+struct ConnectionId(u64);
+
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 enum NegotiationState {
     Start,
@@ -919,6 +928,7 @@ enum NegotiationState {
     Ready,
 }
 
+#[derive(Debug)]
 struct ConnectionState {
     /// State machine progress
     negotiated: NegotiationState,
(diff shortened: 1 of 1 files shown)


downstairs/src/lib.rs
```diff
@@ -3237,6 +3162,111 @@ impl Downstairs {
         }
         Ok(())
     }
+
+    /// Binds the given `ConnectionId` to a new connection state
+    ///
+    /// # Panics
```
jmpesp, line 3168: when?

downstairs/src/lib.rs
```diff
@@ -3237,6 +3162,111 @@ impl Downstairs {
...
+    fn new_connection(
+        &mut self,
+        id: ConnectionId,
+        reply_channel_tx: mpsc::UnboundedSender<Message>,
+    ) -> oneshot::Receiver<UpstairsConnection> {
+        // Prepare for the autonegotiation / IO loop loop
```
jmpesp, line 3174: yo dawg

jmpesp (approved):

downstairs/src/lib.rs
leftwo, line 3168: Panics when there is no repair_address or if a previous connection state
exists at the same ID?

downstairs/src/lib.rs
```diff
@@ -3732,6 +3762,7 @@ pub async fn start_downstairs(
          * multiple Upstairs connecting but only one active one.
          */
         info!(log, "downstairs listening on {}", listen_on);
+        let mut id = ConnectionId(0);
```
leftwo, line 3765: We already have a connection counter in ds.dss.add_connection()
Could we re-use that counter here?  Or would that complicate things
because it live in the stat code?

leftwo (approved):
Is it worth adding the `ConnectionId` to the logging output?  I know having the session ID is useful when trying to debug multiple connection issues with activation and just
in general tracking down what has happened, but I'm not sure if the addition of
`ConnectionId` will help us here.  Maybe just having it log what ID it is at when it
creates a new connection, like in `new_connection()`?

faithanalog (approved):

downstairs/src/lib.rs
mkeeter: fixed in 22f4e4538bcd

downstairs/src/lib.rs
mkeeter: loop loop

downstairs/src/lib.rs
mkeeter: I don't feel too strongly about it, but am leaning towards keeping it separate.  For example, I could imagine adding a "reset stats" function at some point in the future, which would then break `ConnectionId` uniqueness if we using that as the connection counter.

downstairs/src/lib.rs
mkeeter: yup, done in 22f4e4538bcd8719fe7b23ab113c02c3439f39e1

mkeeter:
@leftwo I'm going to add the `ConnectionId` to the log in #1329, since I agree it could be useful

downstairs/src/lib.rs
leftwo: Sure, that's fine.

PR #1329, opened 2024-06-03
Move cancellation into Downstairs, using a token to kill IO tasks

mkeeter:
(staged on top of #1328)

The IO tasks are now cancelled by a `tokio_util::sync::CancellationToken`, instead of a oneshot channel.

The token is wrapped in a `tokio_util::sync::DropGuard` and held in the `ConnectionState`, meaning the IO tasks are cancelled automatically when we drop the `ConnectionState` – no more accidentally forgetting to stop them!

Downstairs connection replacement is now handled in the `Downstairs` itself, rather than in the IO task.  Previously, the downstairs replacement process was as follows:

- `Downstairs::promote_to_active` notices that replacement is necessary
- Downstairs signals to the old IO task on the `oneshot`
- IO task receives this signal in its loop
- IO task locks the `Downstairs` and calls `Downstairs::on_new_connection_replacing`, then exits

Now, it looks like the following:

- `Downstairs::promote_to_active` notices that replacement is necessary
- `Downstairs` calls `on_new_connection_replacing` directly
  - The previous `ConnectionState` is removed and dropped
  - This automatically cancels the IO task

This puts tricky replacement logic in just one place, and makes the IO task simpler: from its perspective, it just gets cancelled.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -1061,7 +1061,9 @@ async fn proc_task(
     let (log, cancel_io) = {
         let mut ds = ads.lock().await;
         (
-            ds.log.new(o!("task" => "proc_task".to_string())),
+            ds.log.new(
+                o!("task" => "proc_task".to_string(), "id" => id.0.to_string()),
+            ),
             ds.new_connection(id, reply_channel_tx),
         )
     };
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -939,13 +939,9 @@ struct ConnectionState {
     /// Upstairs connection, populated at `HelloItsMe`
     upstairs_connection: Option<UpstairsConnection>,
 
-    /// Channel used to kill this Downstairs' IO tasks
-    ///
-    /// The value is moved out of the option when
-    /// `Downstairs::promote_to_active` is called, i.e. between
-    /// `NegotiationState::ConnectedToUpstairs` and
-    /// `NegotiationState::PromotedToActive`
-    another_upstairs_active_tx: Option<oneshot::Sender<UpstairsConnection>>,
+    /// Token used to cancel the IO tasks
+    #[allow(unused)]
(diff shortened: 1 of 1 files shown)


downstairs/src/lib.rs
```diff
@@ -1062,8 +1058,13 @@ async fn proc_task(
     reply_channel_tx: mpsc::UnboundedSender<Message>,
 ) -> Result<()> {
     // Populate our state in the Downstairs, keyed by our `id`
-    let mut another_upstairs_active_rx =
-        ads.lock().await.new_connection(id, reply_channel_tx);
+    let (log, cancel_io) = {
+        let mut ds = ads.lock().await;
+        (
+            ds.log.new(o!("task" => "proc_task".to_string())),
```
leftwo, line 1064: From a previous review comment, maybe here we add `ConnectionId` to the log line?

downstairs/src/lib.rs
```diff
@@ -1087,35 +1088,15 @@ async fn proc_task(
...
-                    Ok(new_upstairs_connection) => {
-                        let mut ds = ads.lock().await;
-                        ds.on_new_connection_replacing(
-                            id, new_upstairs_connection);
-                        return Ok(());
-                    }
-                }
+            _ = cancel_io.cancelled() => {
+                info!(log, "proc_loop cancelled; returning now");
+                return Ok(());
```
leftwo, line 1093: This was returning Error before, but I don't think the difference between error
and Ok resulted in any difference (other than the message).

leftwo (approved):

downstairs/src/lib.rs
mkeeter: Sure, done in 6dbde31b7be5

downstairs/src/lib.rs
mkeeter: I agree, we basically just want to exit the IO task

PR #1331, opened 2024-06-03
Fix outdated comments

mkeeter:
(stacked on top of #1330)

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -1065,28 +1065,8 @@ async fn proc_task(
     let handle = ads.lock().await.handle();
     let cancel_io = handle.new_connection(id, reply_channel_tx).await?;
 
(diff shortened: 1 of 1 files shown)


downstairs/src/lib.rs
```diff
@@ -2836,6 +2816,13 @@ impl Downstairs {
         Ok(())
     }
 
+    /// See the comment in the proc() function on the upstairs side that
```
leftwo, line 2819: This is good, but we don't have `proc()` in the upstairs any longer :)
I think this should be the method: `continue_negotiation()`

leftwo (approved):

downstairs/src/lib.rs
mkeeter: Fixed!

PR #1332, opened 2024-06-03
Move disconnect logic to the Downstairs

mkeeter:
(stacked on top of #1331)

This PR moves the responsibility for "disconnecting when an error occurs" into the `struct Downstairs`.

Previously, `Downstairs::on_message_for` returned a `Result`.  If the result was an error (or `false`), the IO task would exit, but the Downstairs wouldn't necessarily notice until someone tried to communicate with that IO task.

This PR makes `Downstairs::on_message_for` a fire-and-forget operation from the perspective of the IO task, returning nothing.  If the message _does_ result in an error, the `Downstairs` itself will note the error and drop that connection.  When the connection is dropped, the cancellation token held in the `ConnectionState` will automatically stop the IO tasks.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -1074,7 +1074,12 @@ async fn proc_task(
 
             new_read = msg_channel_rx.recv() => {
                 let mut ds = ads.lock().await;
+                let empty = new_read.is_none();
                 ds.on_message_for(id, new_read).await;
+                if empty {
+                    info!(log, "proc_loop read `None`; returning now");
+                    break Ok(());
+                }
             }
         }
     }
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -1074,10 +1074,7 @@ async fn proc_task(
 
             new_read = msg_channel_rx.recv() => {
                 let mut ds = ads.lock().await;
-                if !ds.on_message_for(id, new_read).await? {
-                    info!(log, "on_message_for returned False; returning now");
-                    return Ok(());
-                }
+                ds.on_message_for(id, new_read).await;
             }
         }
     }
@@ -2781,10 +2778,7 @@ impl Downstairs {
     }
 
     /// Function called when the given upstairs disconnects
-    fn on_disconnected(
-        &mut self,
-        upstairs_connection: UpstairsConnection,
-    ) -> Result<()> {
+    fn on_disconnected(&mut self, upstairs_connection: UpstairsConnection) {
         // If our upstairs never completed activation,
         // or some other upstairs activated, we won't
         // be able to report how many jobs.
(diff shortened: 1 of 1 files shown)


downstairs/src/lib.rs
```diff
@@ -3139,28 +3143,25 @@ impl Downstairs {
...
+        if let Some(upstairs_connection) = state.upstairs_connection {
+            self.on_disconnected(upstairs_connection);
+        } else {
+            info!(self.log, "unknown upstairs ({id:?}) removed");
+        }
+    }
+
     /// Handles a single message (or empty channel condition)
     ///
     /// Returns `true` if we should keep going, `false` (or an error) otherwise
```
leftwo, line 3158: This comment is no longer true.

downstairs/src/lib.rs
```diff
@@ -1074,9 +1074,11 @@ async fn proc_task(
 
             new_read = msg_channel_rx.recv() => {
                 let mut ds = ads.lock().await;
-                if !ds.on_message_for(id, new_read).await? {
-                    info!(log, "on_message_for returned False; returning now");
-                    return Ok(());
+                let empty = new_read.is_none();
+                ds.on_message_for(id, new_read).await;
```
leftwo, line 1078: I find this a little confusing as we are deciding if we want to quit based
on the message being `None`, but we still send the message to `on_message_for()`
anyway, which also behaves differently if the message is `None`.  It kind of splits
the logic into two places.  Is this all going to be refactored more later?

leftwo (commented):
Apoligies, but I'm still a little fuzzy on what happens when `on_message_for()` 
decides that we need to disconnect from the upstairs.  I see that it does
some steps to remove the connection, but how does the rest of the downstairs
get the message that it is time for it to disconnect and tasks to end?

downstairs/src/lib.rs
mkeeter: Yup, fixed in 826433b8812e

mkeeter:
> Apoligies, but I'm still a little fuzzy on what happens when on_message_for()
decides that we need to disconnect from the upstairs. I see that it does
some steps to remove the connection, but how does the rest of the downstairs
get the message that it is time for it to disconnect and tasks to end?

When we drop the `ConnectionState` in the `Downstairs`, this drops two interesting things:

- A `DropGuard` which cancels the `CancellationToken` held in `proc_task`; this cancels `proc_task`, which in turn aborts `recv_task`
- A `mpsc::Sender<Message>` which feeds the `reply_task`.  When this is dropped, `reply_task` starts reading `None` values and exits

When either `proc_task`, `reply_task`, or `recv_task` ends, `proc` any upstream tasks then exits itself

(this is _also_ simplified further in #1336, along with a longer description of cancellation orders)

jmpesp (approved):

PR #1333, opened 2024-06-03
Consolidate `on_disconnected` and `remove_connection`

mkeeter:
(stacked on top of #1332)

This is a purely mechanical change that combines logic that was unnecessarily factored out.


leftwo (approved):
This is almost too small :)

jmpesp (approved):

PR #1336, opened 2024-06-04
Remove proc_task / proc, sending tagged message to Downstairs

mkeeter:
(stacked on top of #1333)

This PR removes `proc_task` and the persistent `proc`.  Previously, `proc_task` was responsible for calling `Downstairs::on_message_for` when messages arrived from `recv_task`, and `proc` was responsible for monitoring the suite of IO tasks.

Now, `proc` spawns `recv_task` and `reply_task`, then returns.  `recv_task` sends messages to the `Downstairs` directly, through the `MessageRequest` channel.  Coordinated task shutdown is based on the `CancelGuard` and when message channels run dry; see the block comment in `proc` for details.


downstairs/src/lib.rs
```diff
@@ -989,101 +989,61 @@ where
...
+    //
+    // ## Intentional disconnection by the `Downstairs`
+    // - The `Downstairs` drops the `ConnectionState
+    // - `reply_task` stops as above when its `Receiver` reads `None`
+    // - The `ConnectionState` owns a `DropGuard` and `recv_task` polls the
+    //   corresponding `CancellationToken`; when dropped, `recv_task` exits.
+    //
+    // ## Error in `reply_task`
+    // - `reply_task` will exit upon error
+    // - When the `Downstairs` tries to write to it, the write will fail
```
mkeeter, line 1024: Should we be more proactive here, e.g. stopping `recv_task` if `reply_task` exits?

leftwo: We should stop trying to do any work once we know the connection is closed.
If reply_task ends, will that signal to `Downstairs::run` that it should stop as well?
That's not really an answer to your question directly, but maybe if recv task ends,
it won't be trying to put more work on the queue.

downstairs/src/lib.rs
```diff
@@ -2749,7 +2738,7 @@ impl Downstairs {
 
     async fn run(
```
leftwo, line 2739: We should add some comments here to signify this is now the "main loop" for
doing work on the downstairs.

leftwo (commented):
If we run with lossy set, will the `Downstairs::run` task still make
forward progress?  I'm just thinking of this now and probably should have
considered it before.  I don't think we have any tests that just run with
lossy on one downstairs and verify that the lossy downstairs does not
get stuck.

downstairs/src/lib.rs
mkeeter: I made `recv_task` responsible for monitoring `reply_task` as well in aa6167a69b57, so it should cleanly shut down under all circumstances.

downstairs/src/lib.rs
```diff
@@ -989,101 +989,61 @@ where
     //  │ recv_task  │       │   reply_task   │
     //  └─────┬──────┘       └────────▲───────┘
     //        │                       │
-    //        │message          reply │responses,
-    //        │channel         channel│invalid frame
+    //        │downstairs       reply │responses,
```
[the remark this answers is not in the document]
mkeeter: I did a pass through and updated comments + logic for the new "spawn-only" `proc` behavior, see 2c93629b2210

downstairs/src/lib.rs
mkeeter: Done in 2c93629b2210

mkeeter:
@leftwo 

> If we run with lossy set, will the `Downstairs::run` task still make
> forward progress?  I'm just thinking of this now and probably should have
> considered it before.  I don't think we have any tests that just run with
> lossy on one downstairs and verify that the lossy downstairs does not
> get stuck.

I don't think this behavior has changed – `lossy` still means "occasionally move a job to the end of the queue in `do_work_for`", and we still require all ready jobs in the queue to be finished before `do_work_for` returns.

leftwo (approved):

downstairs/src/lib.rs
```diff
@@ -940,6 +941,11 @@ struct ConnectionState {
     /// Upstairs connection, populated at `HelloItsMe`
     upstairs_connection: Option<UpstairsConnection>,
 
+    // It's important for `cancel` to be above `reply_channel_tx`, to ensure
+    // that `recv_task` knows it has been cancelled by the time `reply_task`
+    // has stopped.  This lets us distinguish between "`reply_task` stopped due
+    // to an error" and "`reply_task` stopped because `reply_channel_tx` was
+    // dropped by the `Downstairs`".
```
jmpesp, line 948: This feels important enough to `impl Drop for ConnectionState` and do it ourselves?

downstairs/src/lib.rs
```diff
@@ -1132,29 +1098,65 @@ where
...
-                        break Ok(());
+                        info!(log, "upstairs disconnected");
+                        send_disconnect();
+                        break;
                     }
                     Some(Err(e)) => {
-                        // XXX "unexpected end of file" can occur if upstairs
-                        // terminates, we don't yet have a HangUp message
-                        bail!("Error reading from Upstairs: {}", e);
+                        warn!(log, "upstairs read an error: {e:?}");
```
jmpesp, line 1147: this warn line could be more specific: `error reading from framedreader` or something

jmpesp (approved):
:rocket: except for the comments!

downstairs/src/lib.rs
mkeeter: I realized that we still need to be robust against ordering shenanigans, so I'm just going to remove the comment entirely, since it's not load-bearing anymore.

The specific ordering issue that _isn't_ prevented by careful `Drop` ordering is

- `recv_task` exits with an error
- Downstairs tries to write to it, notices that it's gone, removes that connection (which cancels the `reply_task` token)
- `reply_task` notices that `recv_task` exited, tells the Downstairs to remove the connection (before noticing that its token has been cancelled)

downstairs/src/lib.rs
mkeeter: Yup, done in fa3fca1f4757

PR #1339, opened 2024-06-05
Make `ConnectionState` an enum type

mkeeter:
The connection state has three different sub-states:

- Connection is open, but we haven't received `HelloItsMe` yet
- Negotiating
- Active and running

This was previously implicit in the `ConnectionState` type:

- `upstairs_connection: Option<UpstairsConnection>` was populated by `HelloItsMe`
- Negotiation is complete when `negotiated == NegotiationState::Ready`

This PR converts `struct ConnectionState` into `enum ConnectionState`, which separates the three possible substates.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -3256,7 +3256,11 @@ impl Downstairs {
 
     /// Handles a single message, either negotiation or doing IO
     async fn on_message_for(&mut self, id: ConnectionId, m: Message) {
-        if matches!(self.connection_state[&id], ConnectionState::Running(..)) {
+        let Some(state) = self.connection_state.get_mut(&id) else {
+            warn!(self.log, "got message for disconnected id {id:?}; ignoring");
+            return;
+        };
+        if matches!(state, ConnectionState::Running(..)) {
             match self.proc_frame(m, id).await {
                 // If we added work, then do it!
                 Ok(Some(new_ds_id)) => {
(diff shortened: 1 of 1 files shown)


downstairs/src/lib.rs
```diff
@@ -2189,8 +2242,14 @@ impl Downstairs {
         m: Message,
         id: ConnectionId,
     ) -> Result<Option<JobId>> {
```
jmpesp, line 2244: I think we need

```rust
        let Some(state) = self.connection_state.get_mut(&id) else {
            warn!(self.log, "got message for disconnected id {id:?}; ignoring");
            return;
        };
```
        
here too?

leftwo (approved):
Does this enum type smooth the way for future work?

leftwo:
> Does this enum type smooth the way for future work?

Okay, reading the next PR, I see where this is going.  Ignore the question

downstairs/src/lib.rs
mkeeter: I don't think so – `proc_frame` is only called in `on_message_for`, which does that check itself:

```rust
    async fn on_message_for(&mut self, id: ConnectionId, m: Message) {
        let Some(state) = self.connection_state.get_mut(&id) else {
            warn!(self.log, "got message for disconnected id {id:?}; ignoring");
            return;
        };
        if matches!(state, ConnectionState::Running(..)) {
            match self.proc_frame(m, id).await {
```

I added the panic condition to the `proc_frame` docstring in aeba76b4e25c

jmpesp (approved):

PR #1348, opened 2024-06-11
Move `proc_frame` into `ActiveConnection`

mkeeter:
(stacked on top of #1341)

Github's rendering absolutely butchers the diff.  `jj show --git` does a much better job; [here's the output](https://gist.github.com/mkeeter/d2e3fda882fe7701d95273e2ab75a006).

The changes are as follows:

- Add a `log` field to `ActiveConnection`
- Add `ActiveConnection::new` constructor
- Added `ActiveConnection::reply` as a convenience function
- Moved `Downstairs::proc_frame` into `ActiveConnection::proc_frame`
- Call `ActiveConnection::proc_frame` in `Downstairs::on_message_for`
- Fix up some unit tests

PR #1350, opened 2024-06-12
Move `do_work_for` and `do_work` into `ActiveConnection`

mkeeter:
(staged on top of #1348)

Continuing the pattern of moving logic from the `Downstairs` into the `ActiveConnection`:

- `Downstairs::do_work_for` is moved to `ActiveConnection::do_ready_work`; the former remains as a thin test-only wrapper
- `Downstairs::do_work` is moved to `ActiveConnection::do_work`
- `Work::in_progress` now returns a `bool` instead of an `Option<JobId>`, because we already know the `JobId` when we call the function
- Unit tests no longer call `in_progress` / `do_work` / `complete_work` separately, because those functions are all integrated into `ActiveConnection::do_ready_work`
    - To get the reply message during unit tests (previously returned by `do_work`), `add_fake_connection` now returns a `mpsc::Receiver`.  This `Receiver` is the output channel for the connection and `recv(..)` returns the message in unit tests.
- `Downstairs::complete_work_stat` is moved to `DsStateOuter::on_complete`, since it's 50% stat-updates by volume (the other 50% is DTrace probes)

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -1489,8 +1489,6 @@ impl ActiveConnection {
                     warn!(self.log, "returning error on read!");
                     Err(CrucibleError::GenericError("test error".to_string()))
                 } else {
-                    // This clone shouldn't be too expensive, since it's only
-                    // 32 bytes per extent (and should usually only be 1 extent)
                     region.region_read(&requests, job_id).await
                 };
                 debug!(
@@ -3236,9 +3234,9 @@ impl Work {
     /// If the dependencies are not met, move the state to `DepWait` and return
     /// `false`.
     ///
-    /// If this job is not new, then just return `false`.
-    ///
     /// If the job already is `InProgress`, return `true` (to be idempotent).
+    ///
+    /// If this job is `Done`, then just return `false`.
     fn in_progress(&mut self, ds_id: JobId) -> bool {
         let Some(job) = self.active.get_mut(&ds_id) else {
             panic!("called in_progress for invalid job");
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -1361,6 +1361,409 @@ impl ActiveConnection {
         };
         Ok(r)
     }
+
+    /// Processes all ready jobs
+    async fn do_ready_work(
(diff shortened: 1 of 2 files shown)


mkeeter:
I've found that looking at the commit locally with `git show --color-moved --patience` works pretty well (better than the Github diff)

downstairs/src/lib.rs
```diff
@@ -1361,846 +1361,861 @@ impl ActiveConnection {
...
+            } => {
+                /*
+                 * Any error from an IO should be intercepted here and passed
+                 * back to the upstairs.
+                 */
+                let response = if flags.read_errors && random() && random() {
+                    warn!(self.log, "returning error on read!");
+                    Err(CrucibleError::GenericError("test error".to_string()))
+                } else {
+                    // This clone shouldn't be too expensive, since it's only
```
leftwo, line 1492: I wonder what clone this comment referred to, as I don't see any clones here.

downstairs/src/lib.rs
```diff
@@ -3283,22 +3230,16 @@ impl Work {
...
-         */
+    /// If the requested job is still new, and the dependencies are all met,
+    /// move the state of the job to `InProgress` and return `true`.
+    ///
+    /// If the dependencies are not met, move the state to `DepWait` and return
+    /// `false`.
+    ///
+    /// If this job is not new, then just return `false`.
+    ///
+    /// If the job already is `InProgress`, return `true` (to be idempotent).
```
leftwo, line 3241: We should move this comment above the "not new, return false" comment, so there is no
confusion on what the flow through here is.

And, while we are at it, we should change the "not new" to be `Done`, as that is the only
other choice besides `InProgress`, `New`, or `DepWait`.

leftwo (approved):

downstairs/src/lib.rs
mkeeter: Outdated, good catch!

downstairs/src/lib.rs
mkeeter: Good idea, fixed

PR #1352, opened 2024-06-13
Remove the SQLite backend

mkeeter:
Duplicate of #1267, because I somehow deleted the branch by accident?


leftwo (approved):

downstairs/src/region.rs
```diff
@@ -639,17 +577,7 @@ impl Region {
             "eid:{} Found repair files: {:?}", eid, repair_files
         );
 
-        // Depending on if this is a clone or not, we have a different
-        // set of files we expect to find.
```
jmpesp, line 643: What about different backends in the future? I guess we would have to imagine a new multi-file backend to argue for some version of this code to stick around, so maybe this is ok.

jmpesp (approved):

downstairs/src/region.rs
mkeeter: Yeah, we can always bring multi-file support back, but I don't have any plans for multiple-file-per-extent backends...

PR #1355, opened 2024-06-14
Remove `Workstate::Done` (unused)

mkeeter:
(staged on top of #1354)

We never actually assigned anything to `WorkState::Done`, so we can remove it.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -3329,12 +3329,6 @@ impl Work {
                 // return idempotently.
                 true
             }
-            WorkState::Done => {
-                /*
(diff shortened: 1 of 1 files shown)


jmpesp (approved):

downstairs/src/lib.rs
```diff
@@ -3333,12 +3333,6 @@ impl Work {
                 // return idempotently.
                 true
             }
-            WorkState::Done => {
```
leftwo, line 3336: We can remove the comment up on 3235 as it's not possible for that situation
to exist.

```
If this job is not new, then just return `false`.
```

leftwo (approved):

PR #1356, opened 2024-06-14
Make many `Region` functions synchronous

mkeeter:
(staged on top of #1355)

For performance, we want to make these functions synchronous and use `block_in_place`.  This PR does the first half, which is purely mechanical removing of `async` and `.await`.

(Previously, we were considering per-extent IO workers, but I haven't gotten that to perform as well)


jmpesp (approved):

leftwo (approved):
I am curious to see how this performs

PR #1357, opened 2024-06-14
Using `block_in_place` for IO operations

mkeeter:
(staged on top of #1356)

`tokio::task::block_in_place` may cause a lot of problems, but it also helps a lot of IOs get done on time, so, it;s impossible to say if its bad or not

### DIFF
--- a/downstairs/src/region.rs
+++ b/downstairs/src/region.rs
@@ -23,6 +23,11 @@ use repair_client::Client;
 /// (i.e. 128 for the Gimlet server sleds).
 const WORKER_POOL_SIZE: usize = 8;
 
+/// Size at which reads and writes should be done in the Tokio blocking pool
+///
+/// This is chosen somewhat arbitrarily.
+const MIN_BLOCKING_SIZE: usize = 64 * 1024; // 64 KiB
+
 use super::*;
 use crate::extent::{
     copy_dir, extent_dir, extent_file_name, move_replacement_extent,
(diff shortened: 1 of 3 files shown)


leftwo:
> (staged on top of #1356)
> 
> `tokio::task::block_in_place` may cause a lot of problems, but it also helps a lot of IOs get done on time, so, it;s impossible to say if its bad or not

I think the performance tests will tell us if it is bad or not :)

downstairs/src/region.rs
```diff
@@ -874,7 +885,13 @@ impl Region {
         for req in req.iter() {
             let extent = self.get_opened_extent_mut(req.extent);
             let req = response.request(req.offset, req.count.get());
-            let out = extent.read(job_id, req)?;
+
+            // Run sufficiently large reads on in the blocking pool
```
leftwo, line 889: Nit: Remove the `on` here.

leftwo (approved):
Would DTrace stats on which path IOs are taking (run blocking or not) be of any value
here?

downstairs/src/region.rs
mkeeter: Fixed!

mkeeter:
> Would DTrace stats on which path IOs are taking (run blocking or not) be of any value
here?

I think we could check using existing probes (since we measure job size), so it doesn't seem critical (and they're easy to add later if we change our minds).

downstairs/src/region.rs
```diff
@@ -23,6 +23,11 @@ use repair_client::Client;
 /// (i.e. 128 for the Gimlet server sleds).
 const WORKER_POOL_SIZE: usize = 8;
 
+/// Size at which reads and writes should be done in the Tokio blocking pool
```
jmpesp, line 26: "size above which" I think?

jmpesp (approved):

downstairs/src/region.rs
mkeeter: Yup, fixed

PR #1361, opened 2024-06-18
Remove `DownstairsWork` from map when handling it

mkeeter:
Previously, we borrowed the `DownstairsWork` from the map when doing IO work.

In this PR, we switch to a strongly-owned model: the `DownstairsWork` is removed from the map when it's ready to be handled.  Removing the job from the map also eliminates a `clone()` in `get_ready_job`, which could be expensive for large writes (where the `Vec<BlockContext>` is non-trivial).  To prevent clones from sneaking in, we remove `#[derive(Clone)]` from `IOop` and `DownstairsWork`.

This is a building block for processing ready IOs immediately as they arrive, instead of putting them in the map at all.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -55,8 +55,12 @@ pub use dump::dump_region;
 pub use dynamometer::*;
 pub use stats::{DsCountStat, DsStatOuter};
 
+/// Single IO operation
+///
+/// `Clone` is deliberately **not** derived on this type, because cloning the
+/// internal `Vec<..>` objects can be expensive.
 #[allow(clippy::derive_partial_eq_without_eq)]
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, PartialEq)]
 enum IOop {
     Write {
         dependencies: Vec<JobId>, // Jobs that must finish before this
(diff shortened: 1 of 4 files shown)


downstairs/src/lib.rs
```diff
@@ -3220,139 +3221,125 @@ impl Work {
...
-                    // is called with a FlushAck so this early out cannot be
-                    // removed unless that is changed too.
-                    if dep <= &self.last_flush {
-                        continue;
-                    }
+        for dep in dep_list.iter() {
+            // The Downstairs currently assumes that all jobs previous
+            // to the last flush have completed, hence this early out.
+            //
+            // Currently `work.completed` is cleared out when
```
leftwo, line 3253: Not sure when `Downstairs::complete_work` went away, but it's not there now.
Non test jobs updates `work.completed` in `do_ready_work()`, so we should replace
`Downstairs::complete_work` with that.

leftwo (approved):

downstairs/src/lib.rs
```diff
@@ -1496,16 +1498,15 @@ impl ActiveConnection {
     /// an error field in the returned `Message`.
     async fn do_work(
         &mut self,
-        job_id: JobId,
+        job: &DownstairsWork,
```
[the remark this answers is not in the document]
mkeeter: Yup, fixed in 9587ae9b547f811f862236b5793174fa87e2d594

downstairs/src/lib.rs
mkeeter: Also fixed in 9587ae9b547f811f862236b5793174fa87e2d594

jmpesp (approved):

PR #1365, opened 2024-06-21
Move 'do work for one job' into a helper function

mkeeter:
(staged on top of #1361)

Renames `do_work` → `do_work_inner`, and make `do_work` the canonical function to handle a `DownstairsWork`.  This is a building block for processing work immediately when it arrives, instead of having to put it into the `active` map then immediately pull it back out.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -1420,9 +1420,15 @@ impl ActiveConnection {
     /// Do work for the given IO job
     ///
     /// This function will either complete the work (adding it to the completed
-    /// list), fail to complete the work (re-adding it to the active list and
-    /// returning the `JobId`), or fail to communicate when replying to the
-    /// Upstairs (which returns an error).
+    /// list and returning `None`), fail to complete the work (re-adding it to
+    /// the active list, sending an `ErrorPort`, and returning the `JobId`), or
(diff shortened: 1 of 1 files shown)


leftwo (approved):

downstairs/src/lib.rs
```diff
@@ -1410,56 +1408,83 @@ impl ActiveConnection {
...
-                // If the job errored, do not consider it completed.
-                // Retry it.
-                new_work.push_back(new_id);
-                self.work.active.insert(dw.ds_id, dw);
+    /// Do work for the given IO job
+    ///
+    /// This function will either complete the work (adding it to the completed
+    /// list), fail to complete the work (re-adding it to the active list and
+    /// returning the `JobId`), or fail to communicate when replying to the
+    /// Upstairs (which returns an error).
```
jmpesp, line 1425: There should be an explicit mention of when this returns None, as well as a mention of when the entire task should bail due to an error during live repair.

jmpesp (approved):

downstairs/src/lib.rs
mkeeter: Yup, done in 872351db691c0f1debf1264adb77e1060741068b

PR #1369, opened 2024-06-24
update omicron deps; use re-exported dropshot types in oximeter-producer configuration

ahl:
Trying to get my arms around syncing nexus and its satellites with https://github.com/oxidecomputer/dropshot/pull/1028

### DIFF
--- a/openapi/crucible-agent.json
+++ b/openapi/crucible-agent.json
@@ -439,6 +439,7 @@
           },
           "source": {
             "nullable": true,
+            "default": null,
             "type": "string"
           },
           "state": {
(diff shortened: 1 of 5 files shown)


openapi/crucible-agent.json
```diff
@@ -439,6 +439,7 @@
           },
           "source": {
             "nullable": true,
+            "default": null,
```
bnaecker, line 442: I'm guessing this is `schemars` unbreaking itself after that optional argument bug was introduced, but just checking.

bnaecker (approved):
👍

leftwo (approved):

PR #1370, opened 2024-06-25
Remove `WorkState`, because it's implicit

mkeeter:
- Any job that's in the `Work::active` map is either in `WorkState::New` or `WorkState::DepWait`; these two variants are always handled the same.
- Any job that's _not_ in the `active` map is `WorkState::InProgress`.
- Because the work state is precisely determined by whether the job is in the map or not, we don't need to store it separately.-

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -684,7 +684,7 @@ pub fn show_work(ds: &mut Downstairs) {
     for upstairs_connection in active_upstairs_connections {
         let work = ds.work(upstairs_connection);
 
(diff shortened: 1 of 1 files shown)


jmpesp (approved):

leftwo (approved):
Basically, a job comes off the wire and we either do the job, or we put it on
the "active" list because it has dependencies that are not met.

We might even consider renaming the `active` map to be `dep_wait` map, as
those are the only type of job that ends up on that list.

mkeeter:
Sure, renamed in 06e5d4b51218

PR #1372, opened 2024-06-25
Clean up dependency checking, fixing space leak

mkeeter:
(staged on top of #1371)

We only ever use the **count** of outstanding dependencies, so we don't need to actually accumulate the specific jobs.  This PR changes `deps_outstanding: Vec<JobId>` into `num_jobs_outstanding: usize`.  It also fixes a space leak in the `outstanding_deps` map; previously, jobs were added (if they were ever skipped due to missing deps) and never removed.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -3312,22 +3312,21 @@ impl Work {
     /// Updates `self.outstanding_deps` and prints a warning message the first
     /// time a job with unmet dependencies is checked.
     fn check_ready(&mut self, ds_id: JobId, work: &mut IOop) -> bool {
-        // Before we can make this in_progress, we have to check the dep
-        // list if there is one and make sure all dependencies are
-        // completed.
+        // Before we can do work for this job, we have to check the dep list if
+        // there is one and make sure all dependencies are completed.
         //
-        // The Downstairs currently assumes that all jobs previous
-        // to the last flush have completed, hence this early out.
+        // The Downstairs currently assumes that all jobs previous to the last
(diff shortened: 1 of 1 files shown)


downstairs/src/lib.rs
```diff
@@ -3308,45 +3312,35 @@ impl Work {
...
-            // unless that is changed too.
-            if dep <= &self.last_flush {
-                continue;
-            }
-
-            if !self.completed.contains(dep) {
-                deps_outstanding.push(*dep);
-            }
-        }
+        // Before we can make this in_progress, we have to check the dep
```
leftwo, line 3315: With the recent changes, we don't make things "in_progress" any more, so the
language here is a little strange.  Perhaps this could read:
`Before we can perform an IO, we have to check ...`

leftwo (approved):
This PR is good, but I have comments here that should have come on a previous
PR :)

Because we don't have an "active" list any longer the comment on line 1509 could
change to reflect the new name.

downstairs/src/lib.rs
```diff
@@ -3391,6 +3385,7 @@ impl Work {
         };
 
         if self.check_ready(ds_id, &mut work) {
+            let _ = self.outstanding_deps.remove(&ds_id);
```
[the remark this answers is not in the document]
mkeeter: Yup, updated in beea9bc1

downstairs/src/lib.rs
mkeeter: Also updated in beea9bc1

Issue #1377, opened 2024-06-29
Snapshot downstairs still wanting to talk to SQLite


askfongjojo:
The issue was seen on rack2 after updating to `659130e0939fe27c4cabd38fad56734f89734407`:
```
root@oxz_switch0:~# pilot host exec -c 'svcs -xZ' 0-31
 6  BRM23230018        ok: 
 7  BRM27230045        ok: 
 8  BRM44220011        ok: svc:/oxide/crucible/downstairs:snapshot-fda723de-5177-4592-9963-f8f10fbd054e-e523f7a6-01db-4df5-bdab-a70e273be322 (Oxide Crucible Downstairs)
  Zone: oxz_crucible_0b41c560-3b20-42f4-82ad-92f5bb575d6b
 State: offline since Sat Jun 29 00:20:35 2024
Reason: Start method is running.
   See: http://illumos.org/msg/SMF-8000-C4
   See: /pool/ext/b93f880e-c55b-4d6c-9a16-939d84b628fc/crypt/zone/oxz_crucible_0b41c560-3b20-42f4-82ad-92f5bb575d6b/root/var/svc/log/oxide-crucible-downstairs:snapshot-fda723de-5177-4592-9963-f8f10fbd054e-e523f7a6-01db-4df5-bdab-a70e273be322.log
Impact: This service is not running.
 9  BRM44220005        ok: 
10  BRM42220009        ok: svc:/oxide/crucible/downstairs:snapshot-2d8b7eb9-32c9-48fb-88e7-197ac86dca8c-ea9ef2e0-8e90-451d-921a-97f1b625dd52 (Oxide Crucible Downstairs)
  Zone: oxz_crucible_cf3b2d54-5e36-4c93-b44f-8bf36ac98071
 State: offline since Sat Jun 29 00:20:38 2024
Reason: Start method is running.
   See: http://illumos.org/msg/SMF-8000-C4
   See: /pool/ext/d0584f4a-20ba-436d-a75b-7709e80deb79/crypt/zone/oxz_crucible_cf3b2d54-5e36-4c93-b44f-8bf36ac98071/root/var/svc/log/oxide-crucible-downstairs:snapshot-2d8b7eb9-32c9-48fb-88e7-197ac86dca8c-ea9ef2e0-8e90-451d-921a-97f1b625dd52.log
Impact: This service is not running.
11  BRM42220006        ok: 
12  BRM42220057        ok: 
[...]

PR #1378, opened 2024-07-01
Restore SQLite backend

mkeeter:
See #1377 for context

This PR effectively undoes #1352 and #1360.  It's _mostly_ a straight-forward revert, but enough things have changed that there are some subtleties.

In particular, #1363 switches to a single context per block.  This means that the SQLite extent implementation now has to decide which of multiple possible contexts are valid.  It can do so using `on_disk_hash`; see `SqliteMoreInner::get_block_contexts` for that implementation.  Some unit tests had to be updated, but `extent_inner_sqlite.rs` is otherwise unchanged.

I also made region / extent creation APIs more consistent:
- Removed `options: RegionOptions` from `Region::open`, because they were unused (basically all callers were passing in `Default::default()`)
- Added a `Backend` argument to `Region::extend(..)`, because it's the only thing which actually creates new extents on disk
- Distinguish between `Region::open_extents` (which opens existing files) and `Region::create_extents` (which takes a backend and creates new extents); previously, both behaviors were handled by `Region::open_extents` depending on a flag
- For simplicity, region migration _always happens_ in `Region::open` (for read-write regions).  To avoid migration in integration tests, I added `DownstairsBuilder::from_region`, which reuses an already-open region (skipping the reopening step which would otherwise trigger migration).


downstairs/src/extent_inner_sqlite.rs
```diff
@@ -0,0 +1,1741 @@
...
+        // buffer a certain amount of writes, only actually writing that
+        // buffer when either a flush is issued or the buffer exceeds some
+        // set size (based on our memory constraints). This would have
+        // benefits on any workload that frequently writes to the same block
+        // between flushes, would have benefits for small contiguous writes
+        // issued over multiple write commands by letting us batch them into
+        // a larger write, and (speculation) may benefit non-contiguous writes
+        // by cutting down the number of sqlite transactions. But, it
+        // introduces complexity. The time spent implementing that would
+        // probably better be spent switching to aio or something like that.
```
faithanalog, line 515: hah, i see my past self was on the same train of thought as me now. though i think i'm more in favor of doing that kind of thing upstairs closer to the guest these days

faithanalog (approved):
I think the region and migration code looks right to me. Matches my memory of what was going, and seems like it's integrated with the current refactoring correctly (with the changes to ExtentWrite, etc.)

I'm less familiar with the integration test stuff, so I can't say I understand that the tests are testing what they should be.

leftwo (approved):
The code changes (un-changes?) look good.
I'm going to continue to try and get a test to run on the bench gimlet, but that
work can proceed in parallel.

PR #1379, opened 2024-07-08
Add `BlockOffset` and `BlockIndex` types

mkeeter:
The `Block` type is very overloaded:

- Sometimes, its value represents an offset relative to the start of an extent
- Other times, its value represents an absolute block index
- In some cases, it represents an extent size!
- `Block` contains both a `value` and a `shift`; `shift` usually means _log2(block size)_
  - `shift` is rarely used and only sometimes checked
  - It's basically only relevant when `Block` represents an extent size
  - However, it's always present, so we have to provide it to `Block::new` even when irrelevant

This PR adds two new types to clarify the meaning of `Block`:

- `BlockIndex` is an absolute block index
- `BlockOffset` is an offset relative to the start of an extent

Both of these types are wrappers around a single `u64` and **do not** store block size.  This is good, because it simplifies the conceptual space: since the `Block` argument includes the block size, we have to handle the case where the block size is **wrong**.  If we simply don't store the block size, then we eliminate all of that logic.

The rest of this PR is using `BlockIndex` and `BlockOffset` types wherever possible, and removing superfluous functions on the original `Block` type.

⚠️ This is a change to wire format, because we're using `BlockOffset` (one `u64`) instead of `Block` (a `u64` + `u32`) in various messages.

This does **not** change the `BlockIO` API, although I'm planning to do that in a subsequent PR.

See also https://github.com/oxidecomputer/crucible/issues/923 , although I don't change `RegionDefinition` in this PR (because the existing format has been serialized to disk).

### DIFF
--- a/common/src/impacted_blocks.rs
+++ b/common/src/impacted_blocks.rs
@@ -258,23 +258,20 @@ impl ImpactedBlocks {
 pub fn extent_from_offset(
     ddef: &RegionDefinition,
     offset: BlockIndex,
-    num_blocks: Block,
+    num_blocks: u64,
 ) -> ImpactedBlocks {
     let extent_size = ddef.extent_size().value;
-    let shift = ddef.block_size().trailing_zeros();
 
     // If we get invalid data at this point, something has gone terribly
     // wrong further up the stack. So here's the assumptions we're making:
-    // First, our blocks are actually the same size as this region's blocks
-    assert_eq!(num_blocks.shift, shift);
 
-    // Second, the offset is actually within the region
+    // First, the offset is actually within the region
     assert!(offset.0 < ddef.extent_count() as u64 * extent_size);
 
-    // Third, the last block is also within the region. These are widened to
+    // Second, the last block is also within the region. These are widened to
     // u128 in order to catch the case where first_block + n_blocks == u64::MAX.
     assert!(
-        offset.0 as u128 + num_blocks.value as u128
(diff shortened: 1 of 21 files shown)


leftwo (approved):
Just a nit

common/src/impacted_blocks.rs
```diff
@@ -240,7 +232,7 @@ impl ImpactedBlocks {
     }
 }
 
-/// Given an offset and number of blocks, compute a list of individually
+/// Given an global offset and number of blocks, compute a list of individually
```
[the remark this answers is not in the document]
mkeeter: Fixed!

common/src/region.rs
```diff
@@ -282,20 +252,35 @@ impl RegionDefinition {
...
+        offset: BlockIndex,
         length: usize,
+        ddef: &RegionDefinition,
     ) -> Result<(), CrucibleError> {
-        if offset.shift != self.extent_size.shift {
-            return Err(CrucibleError::BlockSizeMismatch);
-        }
-
-        let final_offset = offset.byte_value() + length as u64;
+        let final_offset = offset.0 * ddef.block_size + length as u64;
```
jmpesp, line 259: Why not use `self.block_size` here? I don't grok why there's a `RegionDefinition` argument now.

mkeeter: lol no idea what I was doing, fixed

jmpesp (approved):

PR #1380, opened 2024-07-08
Support deactivation when a downstairs is offline

leftwo:
This adds support for deactivation of the upstairs when one or more downstairs is offline.

### DIFF
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -1082,7 +1082,14 @@ impl DownstairsClient {
                     | DsState::Replay
                     | DsState::LiveRepair
                     | DsState::LiveRepairReady
+                    | DsState::Offline
                     | DsState::Reconcile => {} // Okay
+                    DsState::Faulted => {
+                        if matches!(up_state, UpstairsState::Active) {
+                            // Can't transition like this when active
+                            panic_invalid();
+                        }
+                    }
                     _ => {
                         panic_invalid();
                     }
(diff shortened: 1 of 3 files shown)


upstairs/src/downstairs.rs
```diff
@@ -812,7 +812,7 @@ impl Downstairs {
          * TODO: Handle deactivation when a downstairs is offline.
```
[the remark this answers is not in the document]
leftwo: Yes, and actually,I wonder if this check for Offline should even happen, as we don't 
do anything different here if there are Offline downstairs other than print a message.

I think I'll take the comment out of this method and move it where it makes better sense.

mkeeter (approved):

upstairs/src/client.rs
```diff
@@ -1082,7 +1082,14 @@ impl DownstairsClient {
                     | DsState::Replay
                     | DsState::LiveRepair
                     | DsState::LiveRepairReady
+                    | DsState::Offline
```
mkeeter, line 1085: Is this transition actually a thing?  It looks like we added an `Offline` → `Faulted` transition, not directly `Offline` → `Deactivated`

leftwo: Yeah, if we are offline and there are no jobs in the work queue, we will go from `Offline` to `Deactivated`

jmpesp (approved):

upstairs/src/client.rs
mkeeter: Got it, I see that transition happens before the new `Offline` → `Faulted` transition

Issue #1376, opened 2024-06-26
abort_reconciliation() does not support being called more than once.


leftwo:
Method `abort_reconciliation()` is called when we are giving up on a reconciliation.
As part of that, it asserts `self.reconcile.is_some()` and then proceeds to clear it.
However, both `reinitialize()` and `on_reconciliation_ack()` both call that method in a loop:
```
        if self.clients.iter().any(|c| c.state() == DsState::Reconcile) {
            self.abort_reconciliation(up_state);
        }
```

Whomever gets there 2nd will fail the assert.

Issue #1375, opened 2024-06-26
replace should work for downstairs before activation if UUIDs change.


leftwo:
If we have not activated, but are connected to a downstairs, we should support replacing it and
not expecting the UUID to match.
```
Jun 26 18:13:23.454 INFO connecting to 0.0.0.0:8840, : io task, client: 1, : downstairs, session_id: 33e5cf3f-781f-4fbc-9d2c-97f81f296b9b
Jun 26 18:13:23.454 INFO ds_connection connected from Ok(127.0.0.1:59309), : io task, client: 1, : downstairs, session_id: 33e5cf3f-781f-4fbc-9d2c-97f81f296b9b
Jun 26 18:13:23.454 INFO ds_transition from Replaced to WaitActive, client: 1, : down
stairs, session_id: 33e5cf3f-781f-4fbc-9d2c-97f81f296b9b
Jun 26 18:13:23.455 INFO [1] Transition from Replaced to WaitActive, client: 1, : downstairs, session_id: 33e5cf3f-781f-4fbc-9d2c-97f81f296b9b
Jun 26 18:13:23.455 INFO thread 'downstairs client at Some(0.0.0.0:8840) has region UUID 12345678-0000-0000-0000-000000008840tokio-runtime-worker' panicked at , upstairs/
src/client.rs:1921:29:
New client:1 uuid:12345678-0000-0000-0000-000000008840 does not match existing 12345678-0000-0000-0000-000000008820
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace        
client: 1, : downstairs, session_id: 33e5cf3f-781f-4fbc-9d2c-97f81f296b9b
Abort trap: 6 
```

PR #1381, opened 2024-07-09
Replace a downstairs before we are active.

leftwo:
Allow a downstairs to be replaced if the request comes in before the upstairs is active.  
Fix for: https://github.com/oxidecomputer/crucible/issues/871
(Also fixed: https://github.com/oxidecomputer/crucible/issues/1375)
(Also fixed: https://github.com/oxidecomputer/crucible/issues/1376)

This actually was a pretty small change to the upstairs code itself as the general idea of
replacement before active fits within the overall "make all downstairs consistent before starting".

If the activate has not been requested from the guest, then we can do the replacement in
place (as we have not started negotiation yet).  If reconciliation is in progress, it will be aborted and 
then restarted.   The code already checked for any state change in a downstairs and aborted the
reconciliation.  All we had to change here was allowing the reconciliation to restart after a replacement.

Created two new crutest tests that will test replacement before active for both before reconciliation has
started and after reconciliation has started.  Added a clone of the logger for crutest itself to use.

Updated the crutest cli to allow us to send an activation without waiting for the result.  This then allows 
us to trigger a replacement from the cli.

Added a new test, test_replace_special.sh that will spin up downstairs and then call crutest and test 
the replacement of a downstairs while reconciliation is underway.

Added two tests to `test_up.sh`. One that will test replacement of a downstairs before activation and before reconciliation. A second test will test replacement before activation but during reconciliation.

Added the test_replace_special.sh to the nightly test.

### DIFF
--- a/crutest/src/cli.rs
+++ b/crutest/src/cli.rs
@@ -94,6 +94,15 @@ enum CliCommand {
         #[clap(long, short, default_value = "1", action)]
         gen: u64,
     },
+    /// Send an activation message and don't wait for an answer.
+    ///
+    /// This will spawn a background task that will wait on the result of
+    /// the request and display a message with the result when it comes.
+    ActivateRequest {
+        /// Specify this generation number to use when requesting activation.
+        #[clap(long, short, default_value = "1", action)]
+        gen: u64,
+    },
     /// Commit the current write_log data to the minimum expected counts.
     Commit,
     /// Deactivate the upstairs
@@ -467,6 +476,9 @@ async fn cmd_to_msg(
         CliCommand::Activate { gen } => {
             fw.send(CliMessage::Activate(gen)).await?;
         }
+        CliCommand::ActivateRequest { gen } => {
+            fw.send(CliMessage::ActivateRequest(gen)).await?;
+        }
         CliCommand::Commit => {
             fw.send(CliMessage::Commit).await?;
         }
@@ -742,7 +754,11 @@ async fn process_cli_command(
     verify_output: Option<PathBuf>,
 ) -> Result<()> {
     match cmd {
-        CliMessage::Activate(gen) => {
+        CliMessage::Activate(gen) => match guest.activate_with_gen(gen).await {
+            Ok(_) => fw.send(CliMessage::DoneOk).await,
+            Err(e) => fw.send(CliMessage::Error(e)).await,
+        },
+        CliMessage::ActivateRequest(gen) => {
             let gc = guest.clone();
             let _handle = tokio::spawn(async move {
                 match gc.activate_with_gen(gen).await {
@@ -754,6 +770,9 @@ async fn process_cli_command(
                     }
                 }
             });
+            // We return OK here as we have sent off the request.  It's up to
+            // the caller to now check is-active to determine if the
+            // activation has completed.
             fw.send(CliMessage::DoneOk).await
         }
         CliMessage::Deactivate => match guest.deactivate().await {
--- a/crutest/src/cli.rs
+++ b/crutest/src/cli.rs
@@ -742,10 +742,20 @@ async fn process_cli_command(
     verify_output: Option<PathBuf>,
 ) -> Result<()> {
     match cmd {
-        CliMessage::Activate(gen) => match guest.activate_with_gen(gen).await {
-            Ok(_) => fw.send(CliMessage::DoneOk).await,
-            Err(e) => fw.send(CliMessage::Error(e)).await,
-        },
+        CliMessage::Activate(gen) => {
+            let gc = guest.clone();
+            let _handle = tokio::spawn(async move {
+                match gc.activate_with_gen(gen).await {
+                    Ok(_) => {
+                        println!("Activate Successful");
+                    }
+                    Err(e) => {
+                        println!("Activate failed: {:?}", e);
+                    }
+                }
+            });
+            fw.send(CliMessage::DoneOk).await
+        }
         CliMessage::Deactivate => match guest.deactivate().await {
             Ok(_) => fw.send(CliMessage::DoneOk).await,
             Err(e) => fw.send(CliMessage::Error(e)).await,
@@ -1076,14 +1086,14 @@ pub async fn start_cli_server(
                         },
                         Some(cmd) => {
                             process_cli_command(
-                                guest,
-                                &mut fw,
-                                cmd,
-                                &mut ri,
-                                &mut wc_filled,
-                                verify_input.clone(),
-                                verify_output.clone()
-                            ).await?;
+                                    guest,
+                                    &mut fw,
+                                    cmd,
+                                    &mut ri,
+                                    &mut wc_filled,
+                                    verify_input.clone(),
+                                    verify_output.clone()
+                                ).await?;
                         }
                     }
                 }
--- a/crutest/src/cli.rs
+++ b/crutest/src/cli.rs
@@ -1105,14 +1105,14 @@ pub async fn start_cli_server(
                         },
                         Some(cmd) => {
                             process_cli_command(
-                                    guest,
-                                    &mut fw,
-                                    cmd,
-                                    &mut ri,
-                                    &mut wc_filled,
-                                    verify_input.clone(),
-                                    verify_output.clone()
-                                ).await?;
+                                guest,
+                                &mut fw,
+                                cmd,
+                                &mut ri,
+                                &mut wc_filled,
+                                verify_input.clone(),
+                                verify_output.clone()
+                            ).await?;
                         }
                     }
                 }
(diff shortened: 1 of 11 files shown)


upstairs/src/client.rs
```diff
@@ -571,9 +571,6 @@ impl DownstairsClient {
                 );
                 true
             }
-            DsState::Offline => {
-                panic!("can't deactivate while a downstairs is offline")
-            }
             s => {
```
leftwo, line 574: We no longer need this "protection" panic that was here before we supported replacing while Offline.

upstairs/src/downstairs.rs
```diff
@@ -819,12 +819,12 @@ impl Downstairs {
             return true;
         }
         // If this client is offline, then move it to faulted after skipping
-        // all the outstanding jobs.  This will shut the door on this client till
-        // the deactivation has finished on all downstairs and the upstairs has
-        // completed deactivation.  After skipping jobs and setting faulted,
-        // we return false here and let the faulting framework take care of
-        // clearing out the skipped jobs.  This then allows the requested
-        // deactivation to finish.
+        // all the outstanding jobs.  This will shut the door on this client
```
leftwo, line 822: Fixing this one line that was too long resulted in changing the entire comment.
No actual difference here, just moving words around.

crutest/src/cli.rs
```diff
@@ -742,10 +742,20 @@ async fn process_cli_command(
...
+                match gc.activate_with_gen(gen).await {
+                    Ok(_) => {
+                        println!("Activate Successful");
+                    }
+                    Err(e) => {
+                        println!("Activate failed: {:?}", e);
+                    }
+                }
+            });
+            fw.send(CliMessage::DoneOk).await
```
mkeeter, line 757: Can you add a comment explaining why it's fine to return `DoneOk` even though the activation may not be complete?  Or should we leave `CliMessage::Activate` unchanged and add `CliMessage::StartActivate` to start activation but not wait for it?

jmpesp: I'm also for having a separate command to spawn an activation task

crutest/src/main.rs
```diff
@@ -1968,6 +2046,353 @@ async fn replay_workload(
...
+                            bail!("Requested volume verify failed: {:?}", e)
+                        }
+                    }
+                    _ => {} // Ignore everything else
+                }
+            }
+        }
+    }
+
+    info!(log, "Test replace_special has completed");
```
jmpesp, line 2240: replace_while_reconcile ?

tools/test_up.sh
```diff
@@ -149,7 +148,8 @@ res=0
 gen=1
 # Keep deactivate test last, and add 15 to the generation number so
 # tests that follow will be able to activate.
-test_list="span big dep balloon deactivate"
+#test_list="span big dep balloon deactivate"
+test_list="span"
```
jmpesp, line 152: was this intentional?

crutest/src/cli.rs
leftwo: I'll both comment and make a new command :)

leftwo: Done in 4f9da70

Here is how the new CLI looks:
```
>> help
Usage: <COMMAND>

Commands:
  activate          Send an activation message and wait for an answer
  activate-request  Send an activation message and don't wait for an answer
  commit            Commit the current write_log data to the minimum
                        expected counts
  deactivate        Deactivate the upstairs
  dsc               DSC
  expected          Report the expected read count for an offset
  export            Export the current write count to the verify out file
  fill              Run the fill then verify test
  flush             Flush
  generic           Run Generic workload
  info              Request region information
  is-active         Report if the Upstairs is ready for guest IO
  perf              Run the client perf test
  quit              Quit the CLI
  read              Read from a given block offset
  replace
  rr                Issue a random read
  rw                Issue a random write
  show              Show the work queues
  wait              Change the wait state between true and false
  verify            Read and verify the whole volume
  write             Write to a given block offset
  write-unwritten   WriteUnwritten to a given block offset. Note that the
                        cli will decide if a block is written to based on the
                        cli's internal write log.  This log may not match if the
                        block actually has or has not been written to
  uuid              Get the upstairs UUID
  help              Print this message or the help of the given
                        subcommand(s)

Options:
  -h, --help  Print help

>> activate -g 2
Ok
>> info
Got info: 4096 100 122880000
>> is-active
Active is: true
>> deactivate
Ok
>> dsc stop -c 0
Got res: Ok(())
>> activate-request -g 3
Ok
>> is-active
Active is: false
>> dsc start -c 0
Got res: Ok(())
>> is-active
Active is: true
>> 
```

crutest/src/main.rs
leftwo: Fixed in 4f9da70

tools/test_up.sh
leftwo: It was not.  Fixed in 4f9da70

jmpesp (approved):

crutest/src/cli.rs
```diff
@@ -1076,14 +1105,14 @@ pub async fn start_cli_server(
...
                             process_cli_command(
-                                guest,
-                                &mut fw,
-                                cmd,
-                                &mut ri,
-                                &mut wc_filled,
-                                verify_input.clone(),
-                                verify_output.clone()
-                            ).await?;
+                                    guest,
```
mkeeter, line 1108: Is this `rustfmt`'s preference?

crutest/src/main.rs
```diff
@@ -1968,6 +2046,353 @@ async fn replay_workload(
...
+                    log,
+                    "[{c}] Adjusting wait time smaller to catch reconcile",
+                );
+                continue;
+            } else {
+                // A second was not long enough to catch the reconcile, so
+                // the downstairs region size is too small for this test.
+                panic!(
+                    "Downstairs reconciliation too fast for this test, \
+                        consider making a larger downstairs region"
```
mkeeter, line 2147:
```suggestion
                    consider making a larger downstairs region"
```

crutest/src/main.rs
```diff
@@ -1968,6 +2046,353 @@ async fn replay_workload(
...
 
+// Test that a downstairs can be replaced while the initial reconciliation
+// is underway.
+//
+// This test makes use of the dsc client to stop and start downstairs that
+// will allow us to create a mismatch between downstairs which will then
+// trigger a reconcile.
+async fn replace_while_reconcile(
+    guest: &Arc<Guest>,
+    wtq: &mut WhenToQuit,
```
mkeeter, line 2057: Could this be passed by value?  It looks like the caller doesn't use it afterwards.

crutest/src/main.rs
```diff
@@ -1968,6 +2046,353 @@ async fn replay_workload(
...
+    Ok(())
+}
+
+// Test that a downstairs can be replaced before activation when the original
+// downstairs is offline.
+// Make use of dsc to stop our downstairs before activation.
+// Do a fill on each loop so every extent will need to be repaired.
+async fn replace_before_active(
+    guest: &Arc<Guest>,
+    wtq: &mut WhenToQuit,
```
mkeeter, line 2250: Same question here

tools/test_live_repair.sh
```diff
@@ -91,7 +91,7 @@ args+=( -t "127.0.0.1:8830" )
 gen=1
 # Initial seed for verify file
 if ! "$crucible_test" fill "${args[@]}" -q -g "$gen"\
-          --verify-out alan --retry-activate >> "$test_log" 2>&1 ; then
+          --verify-out alan >> "$test_log" 2>&1 ; then
```
mkeeter, line 94: Is this change intentional?

crutest/src/cli.rs
leftwo: Looks like no.
I originally changed this section (where is was all shifted right), but then reverted it.
Seems like `cargo fmt` just left it where it was.

I removed the extra space and `cargo fmt` did not change it back.
I'll push an update with this reverted.

tools/test_live_repair.sh
leftwo: It is intentional yeah, though it's a drive by change and unrelated to this PR.
That option is not required for the live-repair test and was just copied
over from the previous test.  I noticed it when I copied `test_live_repair.sh` to make
`test_repair_special.sh`.  :)

There are probably a pile of things like this, and I debated if I should change
this here, or make a whole new PR with a bunch of clean up.  I then forgot
about it and it remained changed.

upstairs/src/downstairs.rs
```diff
@@ -856,7 +856,7 @@ impl Downstairs {
             if state == &IOState::New || state == &IOState::InProgress {
                 info!(
                     self.log,
-                    "[{}] deactivate job {} not {:?} flush, NO",
+                    "[{}] cannot deactivate, job {} in state {:?}",
```
mkeeter, line 859: thank you for cleaning this up

mkeeter (approved):

tools/test_live_repair.sh
mkeeter: Okay, I'm fine keeping it in this PR

crutest/src/main.rs
leftwo: Yes, both can be passed by value and I did that in 85fc1e20

PR #1387, opened 2024-07-18
Make read and write messages contiguous

mkeeter:
The `BlockIO` trait only allows us to access contiguous sets of blocks for reading and writing.

However, our internal `Message` type converts these contiguous spans into random-access block lists.  This adds complexity: we have to support the possibility for disjoint operations everywhere, even though they should never actually happen.

This PR changes the `Message` types to accept a starting index + block count (for reads), or a starting index + block and context data (for writes).  In addition, the `ReadResponse` is simplified to not include per-block indexes, because they're implied by the start index + position in the list.

Various tests for disjoint reads and writes are removed.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -14,9 +14,9 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use crucible_common::{
-    build_logger, impacted_blocks::extent_from_offset, integrity_hash,
-    mkdir_for_file, verbose_timeout, Block, BlockIndex, BlockOffset,
-    CrucibleError, ExtentId, RegionDefinition, MAX_BLOCK_SIZE,
+    build_logger, integrity_hash, mkdir_for_file, verbose_timeout, Block,
+    BlockIndex, BlockOffset, CrucibleError, ExtentId, RegionDefinition,
+    MAX_BLOCK_SIZE,
 };
 use crucible_protocol::{
     BlockContext, CrucibleDecoder, JobId, Message, MessageWriter,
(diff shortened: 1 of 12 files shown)


upstairs/src/downstairs.rs
```diff
@@ -1891,15 +1894,24 @@ impl Downstairs {
         ));
         debug!(self.log, "IO Write {} has deps {:?}", ds_id, dependencies);
 
+        // TODO: can anyone actually give us an empty write?
```
jmpesp, line 1897: I don't think there's anything in the spec that prevents this? Even if there was something, I'd still argue we have to support it anyway.

downstairs/src/lib.rs
```diff
@@ -1648,31 +1647,7 @@ impl ActiveConnection {
                 // offset); we inject that extra information in this terrible
                 // match statement.
```
jmpesp, line 1648: is this comment now out of date? :)

jmpesp (approved):

protocol/src/lib.rs
```diff
@@ -926,7 +900,8 @@ impl CrucibleEncoder {
                     session_id: Uuid::new_v4(),
                     job_id: JobId(1),
                     dependencies: vec![JobId(1)],
-                    blocks: vec![block_meta; mid],
+                    start: BlockIndex(0),
+                    contexts: vec![ctx; n],
```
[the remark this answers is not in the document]
mkeeter: good catch, fixed in 175f66ebefdf

downstairs/src/lib.rs
mkeeter: Fixed in c1d472b8361f; we can also replace `r.blocks.clone()` with `r.blocks` 😎

faithanalog (approved):

PR #1388, opened 2024-07-18
Update Omicron and Dropshot dependency

bnaecker:
Picks up new `log_headers` field in Dropshot configuration type


bnaecker:
I'm doing some metrics updates that touch Omicron, Crucible, and Propolis, so time for the quarterly slog through the update trenches :)

mkeeter (approved):

PR #1393, opened 2024-08-15
Add worker tasks to the agent.

leftwo:
Added more tasks to the crucible-agent.

We now spawn up to 5 tasks to handle a region creation operation.  This includes a clone operation where we create an empty region, then fill it by copying from another running downstairs.

I attempted to do this in a way that preserved much of how the worker tasks operated, instead of refactoring it.  
I moved what was in the main `worker()` loop for region create into a spawned task, it should be the same set of operations that existed in the previous version.

The existing doorbell mechanism used in the first_in_states()` loop is also used to indicate that a worker task has finished.  `first_in_states()` will also now check for jobs that have completed and remove them from the work queue.

To keep SMF happy, and to not clobber ourselves, I put a lock around the SMF update.  This seemed simpler then trying to break apart `apply_smf()` into the various things it does and then try to only call the part I need when I need it. 

Other actions besides region creation will still be processed by the agent's current worker task. 

Updated agent-antagonist test to print a little less in some places and a little more in other.
Because agent-antagonist will go much faster than Nexus (and has fewer restrictions) I added a loop in the clone path that will try to connect to the source endpoint and verify the remote downstairs is responding before trying to clone from it.  I don't believe this kind of check is needed in Omicron.

agent-antagonist now has a new test that just tests downstairs cloning operations in a loop.

I added a "work queue" API endpoint to the agent.  I'll come back when this lands in Omicron and add that to `omdb`.

### DIFF
--- a/agent-antagonist/src/main.rs
+++ b/agent-antagonist/src/main.rs
@@ -7,7 +7,7 @@ use rand::random;
 use signal_hook::consts::signal::*;
 use signal_hook_tokio::Signals;
 use slog::Logger;
-use slog::{info, warn};
+use slog::{debug, info, warn};
 use std::net::SocketAddr;
 use std::process::Command;
 use std::sync::{
@@ -27,8 +27,11 @@ use crucible_agent_client::{
     about = "Stress tester for the crucible agent"
 )]
 enum Args {
-    /// Run a number of antagonist loop tasks that will do the following:
+    /// Run a number of antagonist loop tasks.
+    /// There are two options for what the test will do.
+    /// All tests operate on a 1GB region.
     ///
+    /// The default test loop will do this loop:
     /// - Create a 1GB region
     ///
     /// - Randomly:
@@ -40,6 +43,14 @@ enum Args {
     ///
     /// - Delete the region
     ///
+    /// The clone stress test will do this:
+    /// - Create a 1GB region
+    /// - Create a snapshot of that region.
+    /// - Loop on:
+    ///   - Clone the snapshot to a new region.
+    ///   - Delete the region.
+    ///
+    ///
     /// Additionally, one task is spawned that will:
     ///
     /// - Get a list of regions
@@ -50,6 +61,10 @@ enum Args {
         /// crucible zone
         #[clap(short, long)]
         agent: Option<SocketAddr>,
+
+        /// Stress test the downstairs clone of a snapshot operation.
+        #[clap(long, action)]
+        clone_stress: bool,
 
         /// Dataset for the crucible agent - leave blank to autodetect if in the
         /// crucible zone
@@ -69,7 +84,7 @@ fn command(log: &Logger, bin: &'static str, args: &[&str]) -> Result<String> {
     info!(log, "{} {:?} took {:?}", bin, args, elapsed);
 
     if !cmd.status.success() {
-        bail!("zfs list failed!");
+        bail!("command: {} {:?} failed {:?}", bin, args, cmd);
     }
 
     Ok(String::from_utf8(cmd.stdout)?.trim_end().to_string())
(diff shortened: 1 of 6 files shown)


agent-antagonist/src/main.rs
```diff
@@ -69,7 +84,7 @@ fn command(log: &Logger, bin: &'static str, args: &[&str]) -> Result<String> {
     info!(log, "{} {:?} took {:?}", bin, args, elapsed);
 
     if !cmd.status.success() {
-        bail!("zfs list failed!");
+        bail!("command: {} {:?} failed {:?}", bin, args, cmd);
```
jmpesp, line 87: where's the dunce emoji

agent/src/datafile.rs
```diff
@@ -783,20 +884,46 @@ impl DataFile {
...
-                for (rid, r) in &inner.running_snapshots {
+                for (rid, r) in &outer.inner.running_snapshots {
                     for (name, rs) in r {
                         if &rs.state == s {
-                            return Resource::RunningSnapshot(
-                                rid.clone(),
-                                name.clone(),
-                                rs.clone(),
-                            );
+                            if outer.work_queue.contains_key(rid) {
```
jmpesp, line 918: With this branch, nothing currently does work on running snapshots right?

agent/src/main.rs
```diff
@@ -400,15 +404,17 @@ where
      */
     let expected_downstairs_instances = regions
         .iter()
-        .filter(|r| r.state == State::Created)
+        .filter(|r| r.state == State::Created || r.state == State::Requested)
```
jmpesp, line 407: I don't think this is right - if I remember correctly this would start a downstairs process for a region that either hasn't had its dataset created yet, or could still be cloning.

agent/src/main.rs
```diff
@@ -400,15 +404,17 @@ where
...
         .map(|r| format!("{}-{}", downstairs_prefix, r.id.0))
         .collect::<HashSet<_>>();
 
     let expected_snapshot_instances: Vec<String> = running_snapshots
         .iter()
         .flat_map(|(_, n)| {
             n.iter()
-                .filter(|(_, rs)| rs.state == State::Created)
+                .filter(|(_, rs)| {
+                    rs.state == State::Created || rs.state == State::Requested
```
jmpesp, line 416: Similarly, I don't think this is right, as it would start a downstairs process for a snapshot's `.zfs/snapshot/...` directory when that might not exist yet

agent/src/main.rs
```diff
@@ -799,6 +805,10 @@ where
                     }
                     crucible_smf::CommitResult::OutOfDate => {
                         error!(log, "concurrent modification?!");
+                        panic!(
```
jmpesp, line 808: why panic here and not in the other check for concurrent modification?

agent/src/main.rs
```diff
@@ -1633,107 +1745,45 @@ fn worker(
...
-                        if let Err(e) = res {
-                            error!(
-                                log,
-                                "region {:?} create failed: {:?}", r.id.0, e
+                    State::Requested => {
+                        // first_in_states has given us a new region to create
+                        // that does not already have a job on work queue, so
+                        // now we will create that job and spawn a task for
+                        // it to do the work in.
+                        let log0 = log.new(o!("component" => "worktask"));
```
jmpesp, line 1753: this log context should also have the region id as a tag / whatever slog calls it :)

jmpesp (changes requested):
Some changes needed:

- adding work to the queue has to be idempotent, as the agent is called from a saga context
- what happens when there is work on the queue and someone tombstones the region? does that tombstone serialize behind the region creation job?

agent-antagonist/src/main.rs
jmpesp: to be clear I meant myself! :o

leftwo:
> adding work to the queue has to be idempotent, as the agent is called from a saga context

Do you mean calling the agent to take an action has to be idempotent?  If we call the agent several times it should only do the work once?
 
> what happens when there is work on the queue and someone tombstones the region? does that tombstone serialize behind the region creation job?

It should, yes (though I'll verify).

Issue #1425, opened 2024-08-20
crucible upstairs fails to replace a downstairs if the new downstairs is unavailable


leftwo:
Found this while stress testing downstairs replacement.

When a downstairs is replaced we transition from:
`Active` -> `Replacing` -> `Replaced`.

This causes the client task to restart itself.

When the client task comes back online, it then tries to connect to the
new downstairs.  If it does, then it starts negotiation and because the
current state is `Replaced`, the negotiation steps will allow the new 
downstairs to proceed to a point where we set it to `LiveRepairReady`
and eventually do the repair.

However, if we are in state `Replaced` and the client task restarts and
it is not able to connect to the new downstairs, it will eventually timeout
and then mark the downstairs as `Failed`

This setting for `Failed` means the upstairs now forgets that this downstairs
was `Replaced`.  When the downstairs does come online, and we try to
negotiate again, we find a downstairs with a different UUID than we expect
and (because we forgot this was replaced) we panic.

leftwo:
Fixed in 1426

PR #1426, opened 2024-08-21
Remember that a replaced downstairs is replaced

leftwo:
Leave a downstairs in state Replaced after a replacement request until we connect to the new downstairs and either validate it is correct or determine it is bad.

This fixes a bug where a replaced downstairs would go to failed, then cause problems when we eventually connect to that downstairs.

Fix for: https://github.com/oxidecomputer/crucible/issues/1425


leftwo:
I thought I had already merged this..

PR #1427, opened 2024-08-21
Use new syncfs syscall

mkeeter:
Opening this as a draft right now, to check whether it works on Buildomat's new helios image.

### DIFF
--- a/downstairs/src/extent_inner_sqlite.rs
+++ b/downstairs/src/extent_inner_sqlite.rs
@@ -36,13 +36,32 @@ impl ExtentInner for SqliteInner {
         self.0.lock().unwrap().dirty()
     }
 
-    fn flush(
+    fn pre_flush(
         &mut self,
         new_flush: u64,
         new_gen: u64,
         job_id: JobOrReconciliationId,
     ) -> Result<(), CrucibleError> {
-        self.0.lock().unwrap().flush(new_flush, new_gen, job_id)
+        self.0.lock().unwrap().pre_flush(new_flush, new_gen, job_id)
+    }
+
+    fn flush_inner(
+        &mut self,
+        job_id: JobOrReconciliationId,
+    ) -> Result<(), CrucibleError> {
+        self.0.lock().unwrap().flush_inner(job_id)
+    }
+
+    fn post_flush(
+        &mut self,
+        new_flush: u64,
+        new_gen: u64,
+        job_id: JobOrReconciliationId,
+    ) -> Result<(), CrucibleError> {
+        self.0
+            .lock()
+            .unwrap()
+            .post_flush(new_flush, new_gen, job_id)
     }
 
     fn read(
@@ -194,10 +213,10 @@ impl SqliteMoreInner {
         Ok(self.dirty.get())
     }
 
-    fn flush(
+    fn pre_flush(
         &mut self,
-        new_flush: u64,
-        new_gen: u64,
+        _new_flush: u64,
+        _new_gen: u64,
         job_id: JobOrReconciliationId,
     ) -> Result<(), CrucibleError> {
         // Used for profiling
@@ -207,12 +226,19 @@ impl SqliteMoreInner {
             (job_id.get(), self.extent_number.0, n_dirty_blocks)
         });
 
+        Ok(())
+    }
+
+    fn flush_inner(
+        &mut self,
+        job_id: JobOrReconciliationId,
+    ) -> Result<(), CrucibleError> {
         /*
          * We must first fsync to get any outstanding data written to disk.
          * This must be done before we update the flush number.
          */
         cdt::extent__flush__file__start!(|| {
-            (job_id.get(), self.extent_number.0, n_dirty_blocks)
+            (job_id.get(), self.extent_number.0)
         });
         if let Err(e) = self.file.sync_all() {
             /*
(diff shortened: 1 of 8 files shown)


downstairs/src/region.rs
```diff
@@ -1053,6 +1014,136 @@ impl Region {
...
+        }
+
+        // Execute the syscall
+        let dir = self.dir.clone();
+
+        // Do the flush in a worker thread if possible, to avoid blocking the
+        // main tokio thread pool.
+        run_blocking(move || -> Result<(), CrucibleError> {
+            let file = File::open(&dir)?;
+            let rc = unsafe { syncfs(file.as_raw_fd()) };
```
faithanalog, line 1127: Think there'd be any benefit to keeping this fd around persistently instead of re-opening every flush?

I don't think I saw this at all on the flamegraphs so I think the answer is no

downstairs/src/region.rs
```diff
@@ -114,6 +114,7 @@ pub struct Region {
     log: Logger,
 
     /// Thread pool for doing long-running CPU work outside the Tokio runtime
+    #[allow(unused)]
     pool: rayon::ThreadPool,
```
faithanalog, line 118: maybe this could be `#[cfg(not(feature = "omicron-build"))]` instead of `allow(unused)`?

faithanalog:
built a tuf repo with this + propolis main + omicron main. threw the tuf repo on madrid. my instance has been "starting" for a suspiciously long time now, so maybe current propolis main + current omicron main is a spicy combo or something. but before that point, uploading a boot iso worked just fine so i think that means the crucible side of things is happy

downstairs/src/region.rs
mkeeter: Removing it entirely is awkward, because then I have to sprinkle `#[cfg(not(feature = "omicron-build"))]` in a bunch of other places.  How about making the `allow(unused)` conditional instead?
```rust
#[cfg_attr(feature = "omicron-build", allow(unused))]
```

faithanalog (approved):

leftwo:
> built a tuf repo with this + propolis main + omicron main. threw the tuf repo on madrid. my instance has been "starting" for a suspiciously long time now, so maybe current propolis main + current omicron main or current propolis main + crucible main isn't a good combo.
> 
> but before that point, uploading a boot iso worked just fine so i think that means the crucible + omicron + illumos side of things is happy (because it had to get through uploading blocks)

Did you ever track down where this was getting stuck?  Do we think it's something to do with the changes here?

downstairs/src/extent_inner_sqlite.rs
```diff
@@ -36,13 +36,32 @@ impl ExtentInner for SqliteInner {
...
         &mut self,
         new_flush: u64,
         new_gen: u64,
         job_id: JobOrReconciliationId,
     ) -> Result<(), CrucibleError> {
-        self.0.lock().unwrap().flush(new_flush, new_gen, job_id)
+        self.0.lock().unwrap().pre_flush(new_flush, new_gen, job_id)
+    }
+
+    fn flush_inner(
```
leftwo, line 48: If things go well, we should never be running this code, as we hope all SQLite extents are read only.

downstairs/src/region.rs
```diff
@@ -1053,6 +1014,136 @@ impl Region {
         Ok(())
     }
 
+    #[cfg(not(feature = "omicron-build"))]
+    #[allow(clippy::unused_async)]
```
leftwo, line 1018: What uses async here? Or, what does not use async that we are allowing?

downstairs/src/region.rs
```diff
@@ -1053,6 +1014,136 @@ impl Region {
...
+            } else {
+                Ok(())
+            }
+        })?;
+
+        // After the bits have been committed to durable storage, execute any
+        // post flush routine that needs to happen
+        for eid in flushed_extents {
+            let extent = self.get_opened_extent_mut(eid);
+            extent.post_flush(flush_number, gen_number, job_id)?;
```
leftwo, line 1141: If we do all the flushes in one pass, then do the `post_flush` in a second loop, our DTrace counters for `cdt::extent__flush__done` will not fire till the `post_flush` loop.  I don't know if that really matters, but we should at least understand it if we are looking at per-extent flush times, as I believe the first extent we flush will include in its final `extent_flush_done` value the time for all the other extents to flush as well.

We still have the per extent `file_flush_start` and `file_flush_done` counters, which probably are the details we care about, but if we are trying to profile the "post flush work per extent" I don't think we will be gathering that info correctly.

leftwo (approved):
Looks goood, some DTrace thoughs.

I'm also curious to know why @faithanalog reported an issue on Madrid and if we know why those instances were not starting, but barring a problem there, I think this is good to go.

downstairs/src/region.rs
mkeeter: At some point, we may have per-extent IO workers, so we deliberately left some functions as async.

faithanalog:
> > built a tuf repo with this + propolis main + omicron main. threw the tuf repo on madrid. my instance has been "starting" for a suspiciously long time now, so maybe current propolis main + current omicron main or current propolis main + crucible main isn't a good combo.
> > but before that point, uploading a boot iso worked just fine so i think that means the crucible + omicron + illumos side of things is happy (because it had to get through uploading blocks)
> 
> Did you ever track down where this was getting stuck? Do we think it's something to do with the changes here?

Haven't. I have not had a chance to do another racklette run. I was thinking to try this branch rebased on a crucible commit that's actually been landed into omicron along with a matching propolis to see if that works fine

PR #1429, opened 2024-08-23
Update oximeter dependency

bnaecker:
- Also updates other omicron deps that come along for the ride
- Updates the pinned toolchain, which is required for the version of `parse-display` that is used here.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.76.0"
+channel = "1.80.0"
 profile = "default"


Cargo.toml
```diff
@@ -113,13 +113,13 @@ uuid = { version = "1", features = [ "serde", "v4" ] }
 
 # git
 dropshot = { git = "https://github.com/oxidecomputer/dropshot", branch = "main", features = [ "usdt-probes" ] }
-omicron-common = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
-nexus-client = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
-internal-dns = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
-omicron-uuid-kinds = { git = "https://github.com/oxidecomputer/omicron", branch = "main" }
+omicron-common = { git = "https://github.com/oxidecomputer/omicron", branch = "new-virtual-disk-timeseries-definition" }
```
bnaecker, line 116: Note to self: Point these back to main once https://github.com/oxidecomputer/omicron/pull/6420 merges

leftwo (approved):
This is fine, but you need to fix the `cargo hakari` issue.

Cargo.toml
bnaecker: Done in a19a366e3afb6a49f000e022b86e2f44beac6b1a

Issue #1258, opened 2024-04-12
Consider merging `DsState::{Faulted, Offline}`


mkeeter:
The documentation for `enum DsState` includes this lovely diagram:
```
 *                       │
 *                       ▼
 *                       │
 *                  ┌────┴──────┐
 *   ┌───────┐      │           ╞═════◄══════════════════╗
 *   │  Bad  │      │    New    ╞═════◄════════════════╗ ║
 *   │Version├──◄───┤           ├─────◄──────┐         ║ ║
 *   └───────┘      └────┬───┬──┘            │         ║ ║
 *                       ▼   └───►───┐       │         ║ ║
 *                  ┌────┴──────┐    │       │         ║ ║
 *                  │   Wait    │    │       │         ║ ║
 *                  │  Active   ├─►┐ │       │         ║ ║
 *                  └────┬──────┘  │ │  ┌────┴───────┐ ║ ║
 *   ┌───────┐      ┌────┴──────┐  │ └──┤            │ ║ ║
 *   │  Bad  │      │   Wait    │  └────┤Disconnected│ ║ ║
 *   │Region ├──◄───┤  Quorum   ├──►────┤            │ ║ ║
 *   └───────┘      └────┬──────┘       └────┬───────┘ ║ ║
 *               ........▼..........         │         ║ ║
 *  ┌─────────┐  :  ┌────┴──────┐  :         ▲         ║ ║
 *  │ Failed  │  :  │ Reconcile │  :         │       ╔═╝ ║
 *  │Reconcile├─◄───┤           ├──►─────────┘       ║   ║
 *  └─────────┘  :  └────┬──────┘  :                 ║   ║
 *  Not Active   :       │         :                 ▲   ▲  Not Active
 *  .............. . . . │. . . . ...................║...║............
 *  Active               ▼                           ║   ║  Active
 *                  ┌────┴──────┐         ┌──────────╨┐  ║
 *              ┌─►─┤  Active   ├─────►───┤Deactivated│  ║
 *              │   │           │  ┌──────┤           ├─◄──────┐
 *              │   └─┬───┬───┬─┘  │      └───────────┘  ║     │
 *              │     ▼   ▼   ▲    ▲                     ║     │
 *              │     │   │   │    │                     ║     │
 *              │     │   │   │    │                     ║     │
 *              │     │   │   ▲  ┌─┘                     ║     │
 *              │     │   │ ┌─┴──┴──┐                    ║     │
[...]

PR #1431, opened 2024-08-26
Make backpressure ignore `Offline` downstairs

mkeeter:
One of my complaints in #1258 is that the `Offline` → `Faulted` transition depends on reaching a certain number of jobs in the queue, but accumulating those jobs is subject to backpressure.

This has a few awkward consequences:
- The time it takes to transition from `Offline` → `Faulted` depends on backpressure tuning, and we have to pick backpressure parameters that make this take a reasonable amount of time (tested in `check_offline_timeout`)
- The offline Downstairs slows down the other Downstairs based on the number of jobs in its queue, meaning the system as a whole gets slower and slower until that Downstairs is marked as faulted

This PR tweaks the backpressure implementation so that only `Active` Downstairs are counted.  To do so, we move the `write_bytes_outstanding` to a per-client field.  The backpressure implementation is also moved to a separate module, since we'll be adding more to it in the future.

The `check_offline_timeout` test is no longer relevant and has been removed.

### DIFF
--- a/upstairs/src/backpressure.rs
+++ b/upstairs/src/backpressure.rs
@@ -1,0 +1,45 @@
+// Copyright 2023 Oxide Computer Company
+
+use crate::{ClientId, DownstairsIO, IOop};
+
+/// Helper struct to contain a count of backpressure bytes
+#[derive(Debug)]
+pub struct BackpressureBytes(u64);
+
+impl BackpressureBytes {
+    pub fn new() -> Self {
+        BackpressureBytes(0)
+    }
+
+    pub fn get(&self) -> u64 {
+        self.0
+    }
+
+    /// Ensures that the given `DownstairsIO` is counted for backpressure
+    ///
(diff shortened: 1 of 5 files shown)


leftwo (approved):
We don't count backpressure bytes when a downstairs is in `LiveRepair` state, but I think that's okay as the LR condition will force all downstairs to sync up around the repair commands, so the other `Active` downstairs will handle any backpressure.

If a downstairs was offline, then comes back, will the outstanding bytes for that downstairs jump to match all the IO it now must replay?

mkeeter:
> If a downstairs was offline, then comes back, will the outstanding bytes for that downstairs jump to match all the IO it now must replay?

Yes (or at least that's the design intent).

This means that the "backpressure goes to infinity" value must be greater than the "offline → faulted" transition point; otherwise, a Downstairs could come back and immediately trigger Infinite Backpressure upon replay.

PR #1432, opened 2024-08-26
Add `crutest bufferbloat` subcommand

mkeeter:
This subcommand does the same write pattern as `crutest rand-write`, but (1) without all of the benchmarking and (2) with a single read at the end, to see how long it takes.

### DIFF
--- a/crutest/src/main.rs
+++ b/crutest/src/main.rs
@@ -106,6 +106,11 @@ enum Workload {
         #[clap(flatten)]
         cfg: RandReadWriteWorkload,
     },
+    /// Send many random reads, then report how long a single write takes
+    Bufferbloat {
+        #[clap(flatten)]
+        cfg: BufferbloatWorkload,
+    },
     Repair,
     /// Test the downstairs replay path.
     /// Stop a downstairs, then run some IO, then start that downstairs back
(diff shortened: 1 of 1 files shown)


crutest/src/main.rs
```diff
@@ -106,6 +106,11 @@ enum Workload {
         #[clap(flatten)]
         cfg: RandReadWriteWorkload,
     },
+    /// Send many random reads, then report how long a single write takes
```
leftwo, line 109: It's the other way, right?  Do a bunch of writes then a single read?

crutest/src/main.rs
```diff
@@ -1053,6 +1107,17 @@ async fn main() -> Result<()> {
                 return Ok(());
             }
         }
+        Workload::Bufferbloat { cfg } => {
```
leftwo, line 1110: The workloads were in alphabetical order.  Maybe put this
one up between `Workload::Biggest` and `Workload::Burst`?

crutest/src/main.rs
```diff
@@ -3120,6 +3185,119 @@ async fn rand_read_write_workload(
...
+    // Join all workers
+    for h in workers {
+        h.await??;
+    }
+
+    println!("\nawaiting final read...");
+    let read_time = read.await?;
+    println!("read took {read_time:?}");
+
+    guest.deactivate().await?;
```
leftwo, line 3296: I would not deactivate here, as there is stuff that the outer loop might do after
the match on `opt.workload`.  It's likely to not be meaningful for this test, but in
general none of the workloads deactivate when they are done and this allows us
to call them from the crutest CLI.

Sometimes I'll also poke at the still running crutest with DTrace or the control 
interface to look at left over job queues or stuff like that.

leftwo (approved):
I like it, but some suggestions :)

crutest/src/main.rs
mkeeter: Fixed in 7465eb040360

crutest/src/main.rs
mkeeter: I added this deactivation because otherwise – if we call `crutest -q` to exit after the test is done – we have to go through reconciliation when `crutest` is next called.

There are two fixes that come to mind:
- `main` could deactivate the guest if `-q` is provided
- We could call `guest.flush(None).await?` here, instead of `guest.deactivate()`

I'm inclined towards the latter, because it's a smaller change.

leftwo: Yeah, adding a flush and then waiting for it to finish should avoid having to reconcile
on the next startup.

There is also a `--stable` option to crutest you can use instead of `-q` that will
wait for all jobs to finish on all downstairs before it exits.  Sometimes I do want to
"leave things dirty" and `-q` will do that for me.

But, this is your test, so you can add a flush on the end if that is what you want :)

PR #1437, opened 2024-08-29
Make build / test consistent in buildomat

mkeeter:
- `build-release.sh` was already building with `--all-features,` but it should also run tests with `--release --all-features`
- `build.sh` should both build and run tests with `--all-features`

Using `--all-features` means we'll get `omicron-build` once #1427 is merged and will start testing `syncfs`

### DIFF
--- a/.github/buildomat/jobs/build-release.sh
+++ b/.github/buildomat/jobs/build-release.sh
@@ -72,7 +72,7 @@ banner rbuild
 ptime -m cargo build --verbose --release --all-features
 
 banner rtest
-ptime -m cargo test --verbose -- --nocapture > /tmp/cargo-test-out.log 2>&1
+ptime -m cargo test --verbose --release --all-features -- --nocapture > /tmp/cargo-test-out.log 2>&1
 
 banner output
 mkdir -p /work/rbins
--- a/.github/buildomat/jobs/build.sh
+++ b/.github/buildomat/jobs/build.sh
@@ -28,7 +28,7 @@ pfexec coreadm -i /tmp/core.%f.%p \
  -e global-setid
 
 banner build
-ptime -m cargo build --verbose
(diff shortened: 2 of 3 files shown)


leftwo (approved):

mkeeter:
Whoops, this doesn't actually work – `cargo test ... --all-features` enables Nexus notifications, which means the Downstairs tries to spawn a Tokio task in a non-Tokio test.

```
---- downstairs::test::reconcile_repair_workflow_1 stdout ----
thread 'downstairs::test::reconcile_repair_workflow_1' panicked at upstairs/src/downstairs.rs:4283:9:
there is no reactor running, must be called from the context of a Tokio 1.x runtime
```

I'm going to go through add switch those tests to `#[tokio::test]` so that they work.

leftwo:
> Whoops, this doesn't actually work – `cargo test ... --all-features` enables Nexus notifications, which means the Downstairs tries to spawn a Tokio task in a non-Tokio test.
> 
> ```
> ---- downstairs::test::reconcile_repair_workflow_1 stdout ----
> thread 'downstairs::test::reconcile_repair_workflow_1' panicked at upstairs/src/downstairs.rs:4283:9:
> there is no reactor running, must be called from the context of a Tokio 1.x runtime
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> I'm going to go through add switch those tests to `#[tokio::test]` so that they work.

I was just looking at that exact section of code, I should have seen that too :)

mkeeter:
Upon further consideration, this is a fool's errand – there are features which explicitly shouldn't be enabled for tests!
```toml
# importantly, don't use features = ["zfs_snapshot"] here, this will cause
# cleanup issues!
crucible-downstairs = { workspace = true, features = ["integration-tests"] }
```

PR #1438, opened 2024-08-29
[prep deps] omicron branch

zeeshanlakhani:
Prep to update lock to account for changes in omicron:

- [x] https://github.com/oxidecomputer/omicron/pull/6414
- [x] and used in https://github.com/oxidecomputer/propolis/pull/747

I will revert the branch names and update Cargo.lock, once https://github.com/oxidecomputer/omicron/pull/6414 lands.


leftwo (approved):
Adjust the branches as necessary

PR #1440, opened 2024-08-30
Adjust backpressure tuning

mkeeter:
Testing with `crutest bufferbloat` reveals that we can get multi-second delays if we send a single read after filling up the write queues.

This PR turns up backpressure gain to make steady-state queue lengths less egregiously long.

I've done some casual testing, and now see turnaround times in the 10-100s of milliseconds.  @faithanalog it would be great to get an `iodriver` run on this PR as well.

### DIFF
--- a/upstairs/src/backpressure.rs
+++ b/upstairs/src/backpressure.rs
@@ -93,39 +93,63 @@ impl Default for BackpressureConfig {
 pub struct BackpressureChannelConfig {
     /// When should backpressure start
     pub start: u64,
-    /// Value at which backpressure goes to infinity
+    /// Value at which backpressure is saturated
     pub max: u64,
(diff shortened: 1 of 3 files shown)


mkeeter:
Here's `main` running a 30-seconds `crutest rand-write` with 4 KiB blocks.  Notice that it takes an additional 30 seconds to actually finish the work; if someone sent a `read` at that point, it would be delayed by up to 30 seconds.

Screenshot 2024-08-30 at 11 47 42 AM

Here's this branch, which finishes the jobs at a reasonable speed:

Screenshot 2024-08-30 at 11 48 05 AM

mkeeter:
Closing in favor of a larger stack of small changes

PR #1442, opened 2024-09-03
Fine-tuning backpressure clamping, and API cleanups

mkeeter:
Right now, backpressure goes to infinity as you approach the `max` value.  However, it's possible to exceed `max` if you (1) start below `max` and (2) send a single job that pushes over the limit (e.g. a large write).  If this happens, then we special-case a 1 hour backpressure delay.

All of this is kinda awkward.

This PR cleans it up:
- There's an explicit `delay_max` stored in the backpressure configuration, which is both a clamp and the delay used if our counter exceeds `max_value`
- Backpressure now has an explicit `Saturated` value, which is used if we exceed `max_value`
- If backpressure is saturated, then we return an error to the guest immediately (instead of delaying the IO for 1 hour)

### DIFF
--- a/upstairs/src/backpressure.rs
+++ b/upstairs/src/backpressure.rs
@@ -75,7 +75,7 @@ impl Default for BackpressureConfig {
     fn default() -> BackpressureConfig {
         // `max_value` values below must be higher than `IO_OUTSTANDING_MAX_*`;
         // otherwise, replaying jobs to a previously-Offline Downstairs could
-        // immediately kick us into the saturated regine, which would be
+        // immediately kick us into the saturated regime, which would be
         // unfortunate.
         BackpressureConfig {
             // Byte-based backpressure
@@ -123,6 +123,9 @@ pub enum BackpressureAmount {
 impl std::cmp::Ord for BackpressureAmount {
     fn cmp(&self, other: &Self) -> std::cmp::Ordering {
         match (self, other) {
+            (BackpressureAmount::Saturated, BackpressureAmount::Saturated) => {
+                std::cmp::Ordering::Equal
+            }
             (BackpressureAmount::Saturated, _) => std::cmp::Ordering::Greater,
             (_, BackpressureAmount::Saturated) => std::cmp::Ordering::Less,
             (
@@ -254,7 +257,7 @@ mod test {
         assert!(
             timeout >= Duration::from_secs(30),
             "max byte-based backpressure delay is too low;
-            expected > 30 secs, got {}",
+            expected >= 30 secs, got {}",
             humantime::format_duration(timeout)
         );
 
@@ -270,8 +273,30 @@ mod test {
         assert!(
             timeout >= Duration::from_secs(30),
             "max job-based backpressure delay is too low;
-            expected > 30 secs, got {}",
+            expected >= 30 secs, got {}",
             humantime::format_duration(timeout)
         );
+    }
+
+    #[test]
+    fn check_saturated_backpressure() {
+        let cfg = BackpressureConfig::default();
+        assert!(matches!(
+            cfg.get_backpressure(IO_OUTSTANDING_MAX_BYTES * 2 + 1, 0),
+            BackpressureAmount::Saturated
+        ));
+        assert!(matches!(
+            cfg.get_backpressure(IO_OUTSTANDING_MAX_BYTES * 2, 0),
+            BackpressureAmount::Saturated
+        ));
+
+        assert!(matches!(
+            cfg.get_backpressure(0, IO_OUTSTANDING_MAX_JOBS as u64 * 2 + 1),
+            BackpressureAmount::Saturated
+        ));
+        assert!(matches!(
+            cfg.get_backpressure(0, IO_OUTSTANDING_MAX_JOBS as u64 * 2),
+            BackpressureAmount::Saturated
+        ));
     }
 }
--- a/upstairs/src/backpressure.rs
+++ b/upstairs/src/backpressure.rs
@@ -204,7 +204,7 @@ impl BackpressureChannelConfig {
         };
 
         BackpressureAmount::Duration(
-            self.delay_scale.mul_f64(v).min(self.delay_max),
+            self.delay_scale.mul_f64(v.powi(2)).min(self.delay_max),
         )
     }
 }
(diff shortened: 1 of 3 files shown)


upstairs/src/backpressure.rs
```diff
@@ -88,40 +100,120 @@ impl Default for BackpressureConfig {
...
+pub enum BackpressureAmount {
+    Duration(Duration),
+    Saturated,
+}
+
+impl std::cmp::Ord for BackpressureAmount {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        match (self, other) {
+            (BackpressureAmount::Saturated, _) => std::cmp::Ordering::Greater,
+            (_, BackpressureAmount::Saturated) => std::cmp::Ordering::Less,
```
faithanalog, line 127: I think there should be a third case before these two that returns Equal if both are saturated. otherwise our Ord/PartialOrd impl disagrees with our derived Eq/PartialEq

> The methods of this trait must be consistent with each other and with those of [PartialEq](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html). The following conditions must hold:
> 1. a == b if and only if partial_cmp(a, b) == Some(Equal).

(from https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html)

upstairs/src/backpressure.rs
```diff
@@ -67,19 +73,25 @@ pub struct BackpressureConfig {
 
 impl Default for BackpressureConfig {
     fn default() -> BackpressureConfig {
+        // `max_value` values below must be higher than `IO_OUTSTANDING_MAX_*`;
+        // otherwise, replaying jobs to a previously-Offline Downstairs could
+        // immediately kick us into the saturated regine, which would be
```
faithanalog, line 78: what is "regine"

upstairs/src/backpressure.rs
```diff
@@ -130,36 +222,55 @@ impl BackpressureConfig {
...
             humantime::format_duration(timeout)
         );
         assert!(
-            timeout > Duration::from_secs(60 * 60),
+            timeout >= Duration::from_secs(30),
             "max job-based backpressure delay is too low;
-            expected > 1 hr, got {}",
+            expected > 30 secs, got {}",
             humantime::format_duration(timeout)
         );
```
faithanalog, line 275: for completion i think we should have an explicit test below here that we do in fact get Saturated at the max value

faithanalog (changes requested):
requested some changes but the main logic of the PR makes sense to me

upstairs/src/backpressure.rs
mkeeter: Done in dc3130f1bcea

upstairs/src/backpressure.rs
mkeeter: Good catch, done in dc3130f1bcea

upstairs/src/backpressure.rs
mkeeter: Either "region" or "regime", take your pick (switched to "regime" in dc3130f1bcea)

upstairs/src/backpressure.rs
```diff
@@ -88,40 +100,123 @@ impl Default for BackpressureConfig {
...
+        // Return a special value if we're saturated
+        if value >= self.max_value {
+            return BackpressureAmount::Saturated;
         }
 
-        // These ratios start at 0 (at *_start) and hit 1 when backpressure
-        // should be infinite.
-        let frac = value.saturating_sub(self.start) as f64
-            / (self.max - self.start) as f64;
+        // This ratio start at 0 (at start_value) and hits 1 when backpressure
```
leftwo, line 195: Nit, `This ratio starts at ..`

upstairs/src/guest.rs
```diff
@@ -413,13 +413,25 @@ impl Guest {
...
-            drop(_guard);
+    /// Sleeps for a backpressure-dependent amount, holding the lock
+    ///
+    /// If backpressure is saturated, logs and returns an error.
+    async fn backpressure_sleep(&self) -> Result<(), CrucibleError> {
+        let bp = self.backpressure.load();
+        match bp {
+            BackpressureAmount::Saturated => {
+                let err = "write queue is saturated";
+                error!(self.log, "{err}");
```
leftwo, line 424: I wonder a few things here.

1) If we are saturated, what impact will we see from logging a bunch of messages.  Like, will it further impact our ability to process IOs?  Or, are things so bad at this point that it does not matter that we are spending time logging things?

2) I do like the idea of knowing this has happened, and if we don't log a message then bumping some internal counter I think would be good as it could help us explain why some VM got a bunch of errors yesterday but now things look okay.

3) And, finally, how difficult is it to trigger this condition from a VM?

leftwo (commented):
I'm concerned about returning errors rather than retrying forever (albit with longer and longer delays), so I thought I would at least document my concerns to allow for more discussion around them.

At some point, we don't have the memory to keep taking IOs in from the guest, so I guess we can't just keep queueing forever.  That being said, what would it take for a guest to generate the traffic to get us saturated?  I suppose it all depends on how fast we can process incoming IOs and what happens if the overall system is so busy that crucible is not getting the resources that it needs?

upstairs/src/backpressure.rs
mkeeter: Fixed!

upstairs/src/guest.rs
mkeeter: 1. In the saturated case, perhaps we should still sleep for `delay_max` here before returning the error code?  That would limit us to one error every 30 seconds, which seems perfectly reasonable.  We could also log an error then delay some long amount of time (e.g. 1 hour), which would match our previous behavior but seems fraught; I worry that delaying a guest by 1 hour would be more confusing than just returning an error.

2. Agreed, we should definitely record it somewhere if not in an error message

3. From a VM, you should only see this error condition if writes take > 30 seconds to execute in Crucible.  The VM can't really affect this (assuming it's already spamming IOs as quickly as possible); the saturated error indicates that even at maximum backpressure, IOs are not being answered fast enough.

leftwo (approved):
I think we have a pile of backpressure work in flight, so we will be further updates in this area and I think this can go in while we iterate.

faithanalog (approved):

PR #1443, opened 2024-09-03
Use RAII handles for backpressure

mkeeter:
staged on top of #1442 

Right now, we're manually tracking backpressure bytes by incrementing / decrementing a counter any time a job becomes active / inactive.  This can be error-prone!

This PR makes tracking backpressure values automatic, through the power of RAII:

- We add a new `struct BackpressureCounters`, which contains the three flavors of counters (write bytes, total bytes, job count)
- `BackpressureCounters::increment` increments counters and hands you a `BackpressureGuard`
- When dropped, the `BackpressureGuard` decrements the counters automatically

### DIFF
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -1,6 +1,6 @@
 // Copyright 2023 Oxide Computer Company
 use crate::{
-    backpressure::BackpressureBytes, cdt, integrity_hash,
+    backpressure::BackpressureCounters, cdt, integrity_hash,
     live_repair::ExtentInfo, upstairs::UpstairsConfig, upstairs::UpstairsState,
     ClientIOStateCount, ClientId, CrucibleDecoder, CrucibleError, DownstairsIO,
     DsState, EncryptionContext, IOState, IOop, JobId, Message, RawReadResponse,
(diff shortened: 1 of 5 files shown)


upstairs/src/client.rs
```diff
@@ -380,16 +375,14 @@ impl DownstairsClient {
 
         // Update our bytes-in-flight counter
         if was_running && !is_running {
-            self.bytes_outstanding = self
-                .bytes_outstanding
-                .checked_sub(job.work.job_bytes())
-                .unwrap();
-            self.write_bytes_outstanding.decrement(job, self.client_id);
+            job.backpressure_guard.take(&self.client_id);
```
faithanalog, line 378: I think I am understanding this correct: this takes the counter and drops it to do a decrement. And we do that here so that this job stops contributing to backpressure since it's not actually running. We expect this to be how most jobs decrement the backpressure counters, but if somehow they don't go through this path they will get cleaned up whenever that job finally gets dropped

faithanalog (approved):

upstairs/src/client.rs
mkeeter: Yup, that's right.  I realized that my comment got lost; added it back in 7e11a7e3ccf5

upstairs/src/client.rs
```diff
@@ -380,16 +375,17 @@ impl DownstairsClient {
...
+            // Because the job is no longer running, it shouldn't count for
+            // backpressure.  Remove the backpressure guard for this client,
+            // which decrements backpressure counters on drop.
+            job.backpressure_guard.take(&self.client_id);
         } else if is_running && !was_running {
             // This should only happen if a job is replayed, but that still
             // counts!
-            self.bytes_outstanding += job.work.job_bytes();
-            self.write_bytes_outstanding.increment(job, self.client_id);
+            job.backpressure_guard.insert(
```
leftwo, line 385: I realize this is not a change in when we adjust backpressure, but how are we not counting replay IOs twice here?  Once for the original IO and a 2nd time on the replay?  I think I've missed a decrement step somewhere.

leftwo (approved):
A question, but assuming that question is answered I'm good to go here. :)

upstairs/src/client.rs
mkeeter: The decrement happens automatically when we replace the `BackpressureGuard` in the `ClientMap`, because the old value is dropped (and decrements on drop).

PR #1444, opened 2024-09-03
Increment write backpressure before deferred encryption

mkeeter:
(Staged on top of #1443 )

This PR tightens the feedback loop for backpressure, by incrementing our counters _before_ encryption is complete.  It uses the new RAII `BackpressureGuard` type, so we don't need to worry about weird accounting failures.

Before:
```
                ┌────────────┐ incr             
             ┌──┤backpressure│◄───┐             
             │  └────────────┘    │             
┌─────┐  ┌───▼──┐    ┌──────────┐ │    ┌───────┐
│Guest├─►│submit├───►│encryption├─┼───►│clients│
└───▲─┘  └──────┘    └──────────┘ │    └───────┘
    │                             │             
    └─────────────────────────────┘             
              ack                               
```

After:
```
                ┌────────────┐                  
             ┌──┤backpressure│                  
             │  └─▲──────────┘                  
             │    │incr                         
┌─────┐  ┌───▼──┐ │  ┌──────────┐      ┌───────┐
│Guest├─►│submit├─┴─►│encryption├─┬───►│clients│
└───▲─┘  └──────┘    └──────────┘ │    └───────┘
    │                             │             
    └─────────────────────────────┘             
              ack                               
```

PR #1445, opened 2024-09-03
Do fast-ack immediately after a write is submitted

mkeeter:
Staged on top of #1444

This PR moves the fast-ack reply to happen before encryption, to avoid an accidental source of backpressure.

Before:
```
                ┌────────────┐                  
             ┌──┤backpressure│                  
             │  └─▲──────────┘                  
             │    │incr                         
┌─────┐  ┌───▼──┐ │  ┌──────────┐      ┌───────┐
│Guest├─►│submit├─┴─►│encryption├─┬───►│clients│
└───▲─┘  └──────┘    └──────────┘ │    └───────┘
    │                             │             
    └─────────────────────────────┘             
              ack                               
```

After:
```
                ┌────────────┐                  
             ┌──┤backpressure│                  
             │  └─▲──────────┘                  
             │    │incr                         
┌─────┐  ┌───▼──┐ │  ┌──────────┐      ┌───────┐
│Guest├─►│submit├─┼─►│encryption├─────►│clients│
└───▲─┘  └──────┘ │  └──────────┘      └───────┘
    │             │                             
    └─────────────┘                             
              ack                               
```

This requires a bunch of changes to unit tests in the `Downstairs`, because of changes in `submit_write`:

- Previously, writes were added to `ackable_work`
- Now, writes must be acked _beforehand_ by the `Upstairs`, and are created with `acked: true`

Opening as a draft for now, because we should decide whether we want to keep the difference in behavior between `Write` (which does fast-ack) and `WriteUnwritten` (which does not).

### DIFF
--- a/upstairs/src/downstairs.rs
+++ b/upstairs/src/downstairs.rs
@@ -1852,22 +1852,28 @@ impl Downstairs {
             extent_id: ExtentId(0),
             block: BlockOffset(0),
         });
-        let awrite = if is_write_unwritten {
-            IOop::WriteUnwritten {
-                dependencies,
-                start_eid: start.extent_id,
-                start_offset: start.block,
-                data: write.data.freeze(),
-                blocks: write.blocks,
-            }
+        let (awrite, acked) = if is_write_unwritten {
+            (
+                IOop::WriteUnwritten {
+                    dependencies,
+                    start_eid: start.extent_id,
+                    start_offset: start.block,
+                    data: write.data.freeze(),
+                    blocks: write.blocks,
+                },
+                false,
+            )
         } else {
-            IOop::Write {
-                dependencies,
-                start_eid: start.extent_id,
-                start_offset: start.block,
-                data: write.data.freeze(),
-                blocks: write.blocks,
-            }
+            (
+                IOop::Write {
+                    dependencies,
+                    start_eid: start.extent_id,
+                    start_offset: start.block,
+                    data: write.data.freeze(),
+                    blocks: write.blocks,
+                },
+                true,
+            )
         };
 
         let io = DownstairsIO {
@@ -1875,7 +1881,7 @@ impl Downstairs {
             guest_id: gw_id,
             work: awrite,
             state: ClientData::new(IOState::New),
-            acked: false,
+            acked,
             replay: false,
             data: None,
             read_validations: Vec::new(),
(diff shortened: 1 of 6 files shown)


mkeeter:
Two failed CI runs, which I can't reproduce:
https://github.com/oxidecomputer/crucible/pull/1445/checks?check_run_id=29621238596
https://github.com/oxidecomputer/crucible/pull/1445/checks?check_run_id=29621239477

I'm going to rerun them and see what happens!

leftwo (approved):

upstairs/src/downstairs.rs
```diff
@@ -1848,30 +1848,36 @@ impl Downstairs {
...
         });
-        let awrite = if is_write_unwritten {
-            IOop::WriteUnwritten {
-                dependencies,
-                start_eid: start.extent_id,
-                start_offset: start.block,
-                data: write.data.freeze(),
-                blocks: write.blocks,
-            }
+        let (awrite, acked) = if is_write_unwritten {
```
jmpesp, line 1851: > Opening as a draft for now, because we should decide whether we want to keep the difference in behavior between Write (which does fast-ack) and WriteUnwritten (which does not).

I don't remember why we decided this, and without knowing that I say we remove this difference and have both fast ack. WriteUnwritten may even do no work at all, meaning whatever overhead we can get rid of would be a win.

mkeeter: Okay, done in 75c8c5e4dab28a02297dab9246185fb8e53eab43

upstairs/src/downstairs.rs
```diff
@@ -2530,22 +2535,18 @@ impl Downstairs {
...
         }
+        assert_eq!(is_write, io.acked);
 
-        if skipped == 3 || is_write {
-            let job = self.ds_active.get_mut(&ds_id).unwrap();
-            assert!(!job.acked);
-            self.ackable_work.insert(ds_id);
-        }
+        // Puts the IO onto the downstairs work queue.
+        self.ds_active.insert(ds_id, io);
```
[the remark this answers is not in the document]
mkeeter: Correct, no one can interfere as long as we're holding the `&mut self`.

upstairs/src/downstairs.rs
```diff
@@ -4521,8 +4517,11 @@ pub(crate) mod test {
                 None,
             );
         }
-
-        ds.ack(ds_id);
+        // Writes are fast-acked when first submitted
+        if !matches!(ds.ds_active.get(&ds_id).unwrap().work, IOop::Write { .. })
```
jmpesp, line 4521: Even though this is just for tests, should this check for WriteUnwritten here too?

upstairs/src/downstairs.rs
```diff
@@ -2530,22 +2529,19 @@ impl Downstairs {
             }
         }
 
-        // Make sure this job is counted for backpressure
-        let is_write = matches!(io.work, IOop::Write { .. });
-
-        // Puts the IO onto the downstairs work queue.
+        let is_write =
+            matches!(io.work, IOop::Write { .. } | IOop::WriteUnwritten { .. });
```
jmpesp, line 2533: Should places that match against either IOop instead call `work.is_write` or something equivalent?

jmpesp (approved):

upstairs/src/downstairs.rs
mkeeter: Yeah, done in 28f79c258115

upstairs/src/downstairs.rs
mkeeter: Added `IOop::is_write` in 28f79c258115, and used it in a few places.

PR #1452, opened 2024-09-06
Make crutest use BlockIO trait instead of a Guest

leftwo:
This enables future work where we plan to use a `Volume` in crutest instead of a `Guest`.

Update crutest to use `<T: BlockIO + Send + Sync + 'static>` instead of `Guest`.  
This involved adding a few more methods to the BlockIO trait, moving a few guest specific methods over to BlockIO, and adding support for these new methods in places where they did not have them.

This should result in no differences at run time as we are not doing anything different than before. 

I wanted this change first to reduce some of the boiler plate noise in future PRs.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -76,7 +76,7 @@ pub enum CrucibleError {
     #[error("Offset past end of extent")]
     OffsetInvalid,
 
-    #[error("Upstairs is already active")]
+    #[error("Upstairs activation is in progress")]
     UpstairsActivateInProgress,
 
     #[error("Upstairs is deactivating")]
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -77,7 +77,7 @@ pub enum CrucibleError {
     OffsetInvalid,
 
     #[error("Upstairs is already active")]
-    UpstairsAlreadyActive,
+    UpstairsActivateInProgress,
 
     #[error("Upstairs is deactivating")]
     UpstairsDeactivating,
@@ -132,6 +132,9 @@ pub enum CrucibleError {
 
     #[error("Generation number is too low: {0}")]
     GenerationNumberTooLow(String),
+
+    #[error("Active with different generation number")]
+    GenerationNumberInvalid,
 
     #[error("No longer active")]
     NoLongerActive,
@@ -389,6 +392,7 @@ impl From<CrucibleError> for dropshot::HttpError {
             | CrucibleError::Disconnect
             | CrucibleError::EncryptionError(_)
             | CrucibleError::GenerationNumberTooLow(_)
+            | CrucibleError::GenerationNumberInvalid
             | CrucibleError::GenericError(_)
             | CrucibleError::HashMismatch
             | CrucibleError::InvalidExtent
@@ -403,7 +407,7 @@ impl From<CrucibleError> for dropshot::HttpError {
             | CrucibleError::RwLockError(_)
             | CrucibleError::SendError(_)
             | CrucibleError::SubvolumeSizeMismatch
-            | CrucibleError::UpstairsAlreadyActive
+            | CrucibleError::UpstairsActivateInProgress
             | CrucibleError::UpstairsDeactivating
             | CrucibleError::UuidMismatch
             | CrucibleError::MissingContextSlot(..)
--- a/crutest/src/cli.rs
+++ b/crutest/src/cli.rs
@@ -223,8 +223,8 @@ enum CliCommand {
  * After verify, we truncate the data to 10 fields and return that so
  * the cli server can send it back to the client for display.
  */
-async fn cli_read(
-    guest: &Arc<Guest>,
+async fn cli_read<T: BlockIO + Send + Sync + 'static>(
+    guest: &Arc<T>,
     ri: &mut RegionInfo,
     block_index: usize,
     size: usize,
@@ -264,8 +264,8 @@ async fn cli_read(
 /*
  * A wrapper around write that just picks a random offset.
  */
-async fn rand_write(
-    guest: &Arc<Guest>,
+async fn rand_write<T: BlockIO + Send + Sync + 'static>(
+    guest: &Arc<T>,
     ri: &mut RegionInfo,
 ) -> Result<(), CrucibleError> {
     /*
@@ -290,8 +290,8 @@ async fn rand_write(
  * Data is generated based on the value in the internal write counter.
  * Update the internal write counter so we have something to compare to.
  */
-async fn cli_write(
-    guest: &Arc<Guest>,
+async fn cli_write<T: BlockIO + Send + Sync + 'static>(
+    guest: &Arc<T>,
     ri: &mut RegionInfo,
     block_index: usize,
     size: usize,
@@ -334,8 +334,8 @@ async fn cli_write(
  * If we do believe the block is written to, then we don't update our
  * internal counter and we don't expect our write to change the contents.
  */
-async fn cli_write_unwritten(
-    guest: &Arc<Guest>,
+async fn cli_write_unwritten<T: BlockIO + Send + Sync + 'static>(
+    guest: &Arc<T>,
     ri: &mut RegionInfo,
     block_index: usize,
 ) -> Result<(), CrucibleError> {
@@ -744,8 +744,8 @@ pub async fn start_cli_client(attach: SocketAddr) -> Result<()> {
 /**
  * Process a CLI command from the client, we are the server side.
  */
-async fn process_cli_command(
-    guest: &Arc<Guest>,
+async fn process_cli_command<T: BlockIO + Send + Sync + 'static>(
+    guest: &Arc<T>,
     fw: &mut FramedWrite<tokio::net::tcp::OwnedWriteHalf, CliEncoder>,
     cmd: protocol::CliMessage,
     ri: &mut RegionInfo,
@@ -1040,14 +1040,14 @@ async fn process_cli_command(
 /*
  * Server for a crucible client CLI.
  * This opens a network port and listens for commands from the cli_client.
- * When it receives one, it translates it into the crucible Guest command
+ * When it receives one, it translates it into the crucible BlockIO command
  * and passes it on to the Upstairs.
  * State is kept here.
  * No checking is done.
  * Wait here if you want.
  */
-pub async fn start_cli_server(
-    guest: &Arc<Guest>,
+pub async fn start_cli_server<T: BlockIO + Send + Sync + 'static>(
+    guest: &Arc<T>,
     address: IpAddr,
     port: u16,
     verify_input: Option<PathBuf>,
--- a/crutest/src/cli.rs
+++ b/crutest/src/cli.rs
@@ -223,8 +223,8 @@ enum CliCommand {
  * After verify, we truncate the data to 10 fields and return that so
  * the cli server can send it back to the client for display.
  */
-async fn cli_read<T: BlockIO + Send + Sync + 'static>(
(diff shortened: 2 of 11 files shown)


crutest/src/main.rs
```diff
@@ -804,6 +806,7 @@ async fn main() -> Result<()> {
         }
         _ => slog::Level::Info,
     };
+
```
mkeeter, line 809: nit: remove newline unless it's intentional, to minimize the diff

crutest/src/main.rs
```diff
@@ -844,7 +847,7 @@ async fn main() -> Result<()> {
         pr = None;
     }
 
-    let _join_handle = up_main(crucible_opts, opt.gen, None, io, pr)?;
+    let _join_handle = up_main(crucible_opts.clone(), opt.gen, None, io, pr)?;
```
mkeeter, line 850: This seems to build fine without the added `.clone()`

mkeeter:
Totally optional: It looks like we only need to pass in a `&Arc<T>` with `Send + Sync + 'static` constraints if we'll be using `tokio::spawn` in the `crutest` function, so we could take this opportunity to simplify a few `crutest` functions.

```diff
diff --git a/crutest/src/main.rs b/crutest/src/main.rs
index c32262df5e...ca5cd07518 100644
--- a/crutest/src/main.rs
+++ b/crutest/src/main.rs
@@ -931,7 +931,7 @@
     match opt.workload {
         Workload::Balloon => {
             println!("Run balloon test");
-            balloon_workload(&guest, &mut region_info).await?;
+            balloon_workload(guest.as_ref(), &mut region_info).await?;
         }
         Workload::Big => {
             println!("Run big test");
@@ -978,7 +978,7 @@
         Workload::Dirty => {
             println!("Run dirty test");
             let count = opt.count.unwrap_or(10);
-            dirty_workload(&guest, &mut region_info, count).await?;
+            dirty_workload(guest.as_ref(), &mut region_info, count).await?;
 
             /*
              * Saving state here when we have not waited for a flush
@@ -1680,8 +1680,8 @@
  * I named it balloon because each loop on a block "balloons" from the
  * minimum IO size to the largest possible IO size.
  */
-async fn balloon_workload<T: BlockIO + Send + Sync + 'static>(
-    guest: &Arc<T>,
+async fn balloon_workload<T: BlockIO>(
+    guest: &T,
     ri: &mut RegionInfo,
 ) -> Result<()> {
     for block_index in 0..ri.total_blocks {
@@ -1826,8 +1826,8 @@
  * Do a single random write to every extent, results in every extent being
  * touched without having to write to every block.
  */
-async fn fill_sparse_workload<T: BlockIO + Send + Sync + 'static>(
-    guest: &Arc<T>,
+async fn fill_sparse_workload<T: BlockIO>(
+    guest: &T,
     ri: &mut RegionInfo,
 ) -> Result<()> {
     let mut rng = rand_chacha::ChaCha8Rng::from_entropy();
@@ -2127,7 +2127,7 @@
     info!(log, "Begin replacement while reconciliation test");
     loop {
         info!(log, "[{c}] Touch every extent part 1");
-        fill_sparse_workload(guest, ri).await?;
+        fill_sparse_workload(guest.as_ref(), ri).await?;
 
         info!(log, "[{c}] Stop a downstairs");
         // Stop a downstairs, wait for dsc to confirm it is stopped.
@@ -2141,7 +2141,7 @@
             tokio::time::sleep(tokio::time::Duration::from_secs(4)).await;
         }
         info!(log, "[{c}] Touch every extent part 2");
-        fill_sparse_workload(guest, ri).await?;
+        fill_sparse_workload(guest.as_ref(), ri).await?;
 
         info!(log, "[{c}] Deactivate");
         guest.deactivate().await.unwrap();
@@ -2318,7 +2318,7 @@
     let mut new_ds = 3;
     for c in 1.. {
         info!(log, "[{c}] Touch every extent");
-        fill_sparse_workload(guest, ri).await?;
+        fill_sparse_workload(guest.as_ref(), ri).await?;
 
         guest.deactivate().await.unwrap();
         loop {
@@ -2466,7 +2466,7 @@
     assert!(targets.len() == 4);
 
     if fill {
-        fill_sparse_workload(guest, ri).await?;
+        fill_sparse_workload(guest.as_ref(), ri).await?;
     }
     // Make a copy of the stop at counter if one was provided so the
     // IO task and the replace task don't have to share wtq
@@ -2613,8 +2613,8 @@
  * We are trying to leave extents "dirty" so we want to exit before the
  * automatic flush can come through and sync our data.
  */
-async fn dirty_workload<T: BlockIO + Send + Sync + 'static>(
-    guest: &Arc<T>,
+async fn dirty_workload<T: BlockIO>(
+    guest: &T,
     ri: &mut RegionInfo,
     count: usize,
 ) -> Result<()> {
```

crutest/src/cli.rs
```diff
@@ -223,8 +223,8 @@ enum CliCommand {
  * After verify, we truncate the data to 10 fields and return that so
  * the cli server can send it back to the client for display.
  */
-async fn cli_read(
-    guest: &Arc<Guest>,
+async fn cli_read<T: BlockIO + Send + Sync + 'static>(
```
mkeeter, line 226: See my comment in the main PR conversation; this only actually requires `BlockIO` and can take a `&T` instead of an `&Arc<T>`

upstairs/src/lib.rs
```diff
@@ -215,6 +219,17 @@ pub trait BlockIO: Sync {
 
         self.activate().await
     }
+    /// Conditional activate if not active.
+    async fn conditional_activate_with_gen(
```
mkeeter, line 223: It looks like this function is only needed by `Volume::activate_with_gen`, so I'd suggest removing it from the trait and just doing this logic in the body of that function (`volume.rs:615,626`).

upstairs/src/volume.rs
```diff
@@ -610,6 +610,72 @@ impl BlockIO for Volume {
 
         Ok(())
     }
+    async fn activate_with_gen(&self, gen: u64) -> Result<(), CrucibleError> {
+        for sub_volume in &self.sub_volumes {
+            sub_volume.conditional_activate_with_gen(gen).await?;
```
mkeeter, line 615: The correctness of this isn't obvious: why is it okay to consider yourself activated with a particular gen if a subvolume is activated with a _different_ gen?

upstairs/src/lib.rs
mkeeter: Same with `conditional_activate`, now that I look at it!

upstairs/src/volume.rs
```diff
@@ -610,6 +610,72 @@ impl BlockIO for Volume {
...
+            all_wq.ds_count += rop_wq.ds_count;
+            all_wq.active_count += rop_wq.active_count;
+        }
+
+        Ok(all_wq)
+    }
+
+    // Return a vec of these?
+    // Return a struct with a vec for SV and Some/None for ROP?
+    async fn query_extent_size(&self) -> Result<Block, CrucibleError> {
```
jmpesp, line 662: What's this used for? I wonder why the caller is ask and if that could be changed.

jmpesp (changes requested):

upstairs/src/volume.rs
leftwo: We use it in crutest to construct specific kinds of tests where we cross the extent boundary with IOs.  It could be done other ways, like we pass the test itself information externally.  Having the test figure it out without having to supply anything is an easy way to do it.

leftwo: The `sparse_fill()` is another good example.  We want to do an IO to every extent so they all end up dirty, but dont' want to have to do IO to every block in the extent.

crutest/src/main.rs
leftwo: Ah, that was leftover from when I did need it for code that I've since removed.
I'm suprosed clippy did not find that.

upstairs/src/volume.rs
leftwo: So, `activate_with_gen()` has been updated to:
1) Allow activation with a higher number than the upstairs was started with, but only if the upstairs is not active yet.
2) If we are active, and the generation number we are active with is the same as what we requested, then return OK,
3) Any other combination should cause `activate_with_gen()` to fail.

I'd like to eventually remove `activate_with_gen()`, but not here in this PR till as that will involve some more serious test restructuring, and I'm not ready for that yet.

upstairs/src/lib.rs
leftwo: I've removed all the `conditional_activate()` and now the behavior is:
If we are active with the same gen as requested, then return OK()
If we are not active, then activate with the provide gen.
Else, return error.

crutest/src/cli.rs
leftwo: Done

crutest/src/main.rs
leftwo: Fixed

crutest/src/main.rs
leftwo: Fixed.

common/src/lib.rs
```diff
@@ -77,7 +77,7 @@ pub enum CrucibleError {
     OffsetInvalid,
 
     #[error("Upstairs is already active")]
```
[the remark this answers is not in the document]
leftwo: Done in 0a8efdd, thanks

integration_tests/src/lib.rs
```diff
@@ -3646,6 +3644,94 @@ mod test {
...
+
+        guest.activate().await?;
+        assert!(guest.activate_with_gen(3).await.is_err());
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn integration_test_guest_reactivate_same() -> Result<()> {
+        // Verify we fail activate_with_gen() if we are already active
```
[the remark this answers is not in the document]
leftwo: done in 0a8efdd

leftwo:
I'll  fix the merge conflicts my merging main into the branch.
I did not want to do this till everyone had a chance to look at the changes so far.

upstairs/src/lib.rs
```diff
@@ -104,11 +104,15 @@ pub const IO_OUTSTANDING_MAX_JOBS: usize = 10000;
 #[async_trait]
 pub trait BlockIO: Sync {
     async fn activate(&self) -> Result<(), CrucibleError>;
+    async fn activate_with_gen(&self, gen: u64) -> Result<(), CrucibleError>;
```
jmpesp, line 107: Eventually we want to get rid of `activate_with_gen` - for now, should we mark this as test-only, or hide it behind a flag? I'm also fine not doing this, so consider this a non-blocking suggestion :)

upstairs/src/pseudo_file.rs
```diff
@@ -221,6 +210,16 @@ impl<T: BlockIO> CruciblePseudoFile<T> {
 
         Ok(())
     }
+    pub async fn activate_with_gen(
+        &self,
+        gen: u64,
+    ) -> Result<(), CrucibleError> {
+        self.block_io.activate_with_gen(gen).await
```
jmpesp, line 217: CruciblePseudoFile needs to do some queries to set some of its fields after activation, those need to be duplicated here.

upstairs/src/lib.rs
leftwo: I want to focus on getting rid of it!

leftwo:
from chat:
```
Alan (he/him)
You have any more thoughts on : https://github.com/oxidecomputer/crucible/pull/1452
I think I've addressed all comments.
James (he/him)
naw ship it
```

jmpesp (approved):

PR #1453, opened 2024-09-06
Cache `ddef` in Downstairs instead of passing it in

mkeeter:
This is a small cleanup to make the Downstairs more standalone.

### DIFF
--- a/upstairs/src/downstairs.rs
+++ b/upstairs/src/downstairs.rs
@@ -124,6 +124,9 @@ pub(crate) struct Downstairs {
     /// This must be handled after every event
     ackable_work: BTreeSet<JobId>,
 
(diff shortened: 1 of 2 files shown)


leftwo (commented):
Is the cached version the same one that we would pass in otherwise?

This won't conflict with any of the downstairs replacement stuff, right?

mkeeter:
@leftwo 

> Is the cached version the same one that we would pass in otherwise?

Correct, we write it to the `Downstairs` at the same time that it's stored in the `Upstairs`

> This won't conflict with any of the downstairs replacement stuff, right?

Yeah, the `RegionDefinition` is very much not allowed to change

leftwo (approved):

PR #1456, opened 2024-09-16
Volumes in crutest plus supporting region info

leftwo:
**Part of the Guest to Volume work**

This is another step towards testing from the `Volume` layer.  It's a sprinking of changes here and there.  The most interesting part is probalby in crutest where we create a Volume instaead of creating a Guest.  Most of the rest of the changes are pretty minor.

Made crutest always use a `Volume` instead of a `Guest`.  If we are just provided targets (the current method) then we make a `Volume` with a single sub-volume.

Added support to crutest to read a VCR file as input.  This allows us to build a multi level sub-volume.

Moved the `config_path()` function to crucible-common so others can make use of it.  This function takes a region directory path and appends the name of the expected region description file to the end of it.

Moved `RegionExtentInfo` to crucible-common so others can make use of it.

Updated dsc to record and report region information about the regions (downstairs) it is running.  Updated dsc to report back port numbers for clients.  These changes should allow crutest to not need to know targets/ports for the downstairs and can instead just ask dsc.

Made dsc's type for block_size the same as everyone else.

Fixed a typo in tools/test_up.sh and the long broken tools/test_dsc.sh

**Still to do:**

Update the Volume layer to return a better format for the crutest internal `RegionInfo` structure.  Right now it is assumed we only have a single sub-volume and all the tests make assumptions based on that. We need to update the Volume layer to return more information so the tests can make better decisions.

Update all the tests to just ask DSC for region info instead of having to pass that information on the command line.

There are 245 places where we use "guest" that I want to change to "block_io", but I'll do that in a follow up PR which will be just that name change.


crucible-client-types/src/lib.rs
```diff
@@ -139,3 +139,12 @@ pub enum ReplacementRequestCheck {
     /// original.
     ReplacementMatchesOriginal,
 }
+
+/// Extent information about a region.
+#[derive(Serialize, Deserialize, Clone, Debug, JsonSchema)]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub struct RegionExtentInfo {
```
mkeeter, line 146: Nit: docstrings would be nice here (especially for `block_size`, which is presumably in bytes?)

crutest/src/main.rs
```diff
@@ -846,8 +844,79 @@ async fn main() -> Result<()> {
...
+                Ok(vcr) => vcr,
+                Err(e) => {
+                    bail!("Error {:?} reading VCR from {:?}", e, vcr_file)
+                }
+            };
+            let volume =
+                Volume::construct(vcr, pr, guest_logger).await.unwrap();
+            Arc::new(volume)
+        } else {
+            // We were not provided a VCR, so, we have to make one by using
```
mkeeter, line 865: Vague architectural question: do we need to get this from the repair port, or could we get it by building a `Guest` instead (then use that `Guest` to build a `Volume`)?

dsc/src/control.rs
```diff
@@ -154,6 +156,37 @@ async fn dsc_get_pid(
...
+    if cid_bad(&api_context.dsci, cid).await {
+        return Err(HttpError::for_bad_request(
+            Some(String::from("BadInput")),
+            format!("Invalid client id: {}", cid),
+        ));
+    }
+    let ds_port = api_context.dsci.get_ds_port(cid).await.map_err(|e| {
+        HttpError::for_bad_request(
+            None,
+            format!("failed to state for downstairs {}: {:#}", 0, e),
```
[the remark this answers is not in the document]
mkeeter: Also, the error messages indicates that this is always downstairs 0 – is that correct?

dsc/src/control.rs
```diff
@@ -373,6 +406,29 @@ async fn dsc_enable_restart_all(
...
+async fn dsc_get_region_info(
+    rqctx: RequestContext<Arc<DownstairsControl>>,
+) -> Result<HttpResponseOk<RegionExtentInfo>, HttpError> {
+    let api_context = rqctx.context();
+
+    let region_info =
+        api_context.dsci.get_region_info().await.map_err(|e| {
+            HttpError::for_bad_request(
+                None,
+                format!("failed to state for downstairs {}: {:#}", 0, e),
```
mkeeter, line 425:
```suggestion
                format!("failed to get region info for downstairs {}: {:#}", 0, e),
```

For bonus points, this could also be fixed in `dsc_get_pid` ("failed to get PID") and `dsc_get_ds_state` ("failed to get state" instead of "failed to state")

dsc/src/main.rs
```diff
@@ -484,13 +488,22 @@ impl DscInfo {
...
-        block_size: u32,
+        block_size: u64,
         quiet: bool,
         encrypted: bool,
     ) -> Result<f32> {
         // Create the path for this region by combining the region
         // directory and the port this downstairs will use.
         let mut rs = self.rs.lock().await;
+
+        // If we don't have region info yet, set it now.
```
mkeeter, line 499: Should we check against existing `region_info` and bail if they don't match?

dsc/src/main.rs
```diff
@@ -615,6 +653,10 @@ impl DscInfo {
             port += rs.port_step;
         }
 
+        // Update our region information with what we found in the config file.
```
mkeeter, line 656: Same question here: should we `bail!` if they don't match?

dsc/src/main.rs
```diff
@@ -1399,6 +1458,7 @@ fn main() -> Result<()> {
             if cleanup {
                 crate::cleanup(output_dir.clone(), region_dir.clone())?;
             }
+
```
mkeeter, line 1461: Nit: bonus whitespace added in this file

dsc/src/main.rs
```diff
@@ -1814,12 +1880,17 @@ mod test {
                 *port,
             )
             .unwrap();
-            fs::create_dir_all(ds_region_dir).unwrap();
+            fs::create_dir_all(ds_region_dir.clone()).unwrap();
```
mkeeter, line 1883: Does this work?
```suggestion
            fs::create_dir_all(&ds_region_dir).unwrap();
```

dsc/src/control.rs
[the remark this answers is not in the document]
leftwo: Yeah, I noticed that and fixed it in a coming PR as well as found other places where the same copy/pasta error was.  I might as well fix them all here.

dsc/src/main.rs
leftwo: I considered that, but decided not to, as we could be testing that a mismatch is caught by the upstairs.  I could make a check and warning here though.

leftwo: Also, currently DSC does not allow you to create regions that differ from each other, so to get something like this to exist, you would have to run dsc multiple times, then try to stitch all the parts together after the fact to get a single DSC to serve the results.

crucible-client-types/src/lib.rs
leftwo: Will do.  This was just moved but I can improve the comment(s) on it.

crutest/src/main.rs
leftwo: Maybe.. we could build it backwards like that.
There exists a `Volume`:

```
    // Create a simple Volume from a single BlockIO                             
    pub async fn from_block_io(                                                 
        block_io: Arc<dyn BlockIO + Sync + Send>,                               
        log: Logger,
    ) -> Result<Volume, CrucibleError> {                                       
```

But, I would have to play around with it to be sure it does what we need.

We could also possibly just make a `Guest` here instead of a `Volume`, though I was not sure if I could have the
```
let guest = {
```
Return either a `Volume` or a `Guest`.
Let me see if the `from_block_io` works like I expect.

dsc/src/control.rs
leftwo: And, there is no downstairs whom we are specifically asking either.  I'll fix them all

dsc/src/main.rs
leftwo: It complains because the `let cp = config_path` line below also wants to use `ds_region_dir`

crutest/src/main.rs
leftwo: Ah, so, the problem is that the guest layer can't return a valid `get_block_size()` until the
upstairs is activated.  This means the `from_block_io()` volume creation attempt fails.
Even if `get_block_size()` worked, we also need `total_size()` to work as well.

So, for a Volume, we have to know this information ahead of time.  We could require it be passed in, or
use the method in the PR to scrape it from a downstairs.

dsc/src/control.rs
leftwo: Fixed in 8acd4d6

dsc/src/main.rs
leftwo: Added warnings in 8acd4d6

dsc/src/main.rs
leftwo: Warnings in 8acd4d6

dsc/src/main.rs
leftwo: Removed in 8acd4d6

dsc/src/control.rs
```diff
@@ -422,7 +422,7 @@ async fn dsc_get_region_info(
         api_context.dsci.get_region_info().await.map_err(|e| {
             HttpError::for_bad_request(
                 None,
-                format!("failed to state for downstairs {}: {:#}", 0, e),
+                format!("failed to region info {:#}", e),
```
mkeeter, line 425: "failed to get region info", perhaps?

crutest/src/main.rs
mkeeter: Okay, seems fine in that case

dsc/src/control.rs
leftwo: almost fixed in 30e467f, no actual diff for 8acd4d6, but then really fixed in f5cf0d1

dsc/src/control.rs
leftwo: sigh.. yeah.  fixed

PR #1458, opened 2024-09-18
Name change in crutest

leftwo:
Follow on to https://github.com/oxidecomputer/crucible/pull/1456
This is just a blanket name change from `guest` to `block_io`.


leftwo:
Now that I've done this, I wonder if I should just have named it "volume" everywhere.
crutest no longer uses anything but `Volume`.  And, given the changes coming for support of larger disks, I think that Volume is going to be the way forward.  At any rate, let me know what you all think.

mkeeter:
You should edit the target branch of this PR to `alan/crutest-volume-2`, so that it only shows the relative changes.  Once #1456 merges, this PR will automatically be changed to target `main`

mkeeter (approved):

PR #1466, opened 2024-09-20
Combine `enqueue` and `enqueue_repair`

mkeeter:
The only difference between `enqueue` and `enqueue_repair` is that the latter _always_ sends jobs, instead of checking against live-repair state.

We can instead move that logic into `send_io_live_repair`, always sending live-repair IO jobs (based on their `IOop` variants)


leftwo (approved):
So, I think this is good, but I have a question/concern.

A tricky part of `LIveRepair` is that we build IOs and send them off to the downstairs
based on the state of things at the time the IO is generated.  Specifically I'm thinking
about the `last_repair_extent`.  This value needs to be determined at the time the IO is created.  

I believe this constraint is still honored, as we are creating, enque-ing, and sending all at the same time, so the `last_repair_extent` that we collect in enqueue is the one we use when we put the IO on the wire.

I wish there was a way to make this requirement.. better enforced in the code.

jmpesp (approved):

faithanalog (approved):

PR #1467, opened 2024-09-20
Remove `IOState::New` (6/n)

mkeeter:
staged on top of #1466

At long last, we're no longer using `IOState::New`; jobs are immediately sent to the IO worker tasks when they are created, so we can remove this variant.

Reconciliation jobs – which _do_ use `IOState::New` but never used `IOState::Error` – now get their own `enum ReconcileIOState`, limited to their subset of states.

### DIFF
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -413,10 +413,11 @@ impl DownstairsClient {
         job
     }
 
-    /// Ensures that the given job is in the job queue
+    /// Sets the given job's state to [`IOState::InProgress`]
+    ///
+    /// # Panics
+    /// If the job's state is [`IOState::Done`] but the job has not been acked
     pub(crate) fn replay_job(&mut self, job: &mut DownstairsIO) {
-        // If the job is InProgress, then we can just go back to New and no
-        // extra work is required.
         // If it's Done, then by definition it has been acked; test that here
         // to double-check.
         if IOState::Done == job.state[self.client_id] && !job.acked {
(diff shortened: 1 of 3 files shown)


upstairs/src/client.rs
```diff
@@ -414,7 +413,7 @@ impl DownstairsClient {
         job
     }
 
-    /// Ensures that the given job is in the job queue and in `IOState::New`
+    /// Ensures that the given job is in the job queue
     pub(crate) fn replay_job(&mut self, job: &mut DownstairsIO) {
         // If the job is InProgress, then we can just go back to New and no
```
leftwo, line 418: The comment here about state `New` is no longer correct.

upstairs/src/client.rs
```diff
@@ -439,7 +433,7 @@ impl DownstairsClient {
     /// If the job is not new or in-progress
```
leftwo, line 433: Nit, no more `new` state

leftwo (approved):

upstairs/src/client.rs
mkeeter: fixed!

upstairs/src/client.rs
mkeeter: Fixed!

upstairs/src/downstairs.rs
```diff
@@ -2531,7 +2531,7 @@ impl Downstairs {
         self.ds_active.for_each(|ds_id, job| {
             let state = &job.state[client_id];
 
-            if matches!(state, IOState::InProgress | IOState::New) {
```
jmpesp, line 2534: nit: don't need `matches!` here anymore

faithanalog (approved):

PR #1470, opened 2024-09-24
Test read from 3x faulted Downstairs

mkeeter:
This should return an error, because the IO is skipped on all 3x Downstairs.  Previously, it did not!


jmpesp (approved):
Alan says ship it!

leftwo (approved):
James says whatever I tell him to say

Issue #1459, opened 2024-09-18
crutest cli does not let you change from the default dsc server


leftwo:
The crutest cli has support for connection to a dsc server and running commands on it.
However, you don't get a choice for where to connect, it always select the default.

Issue #1457, opened 2024-09-17
Update crutest+dsc to auto-generate volumes


leftwo:
Crutest will soon support working at the volume layer, with support for reading a VCR from a file.

Once this work lands, we need to grow it to have crutest be able to ask dsc for either a VCR, or enough information so it can create a VCR based on what dsc tells it.  It seems like having crutest build the VCR by asking dsc for the region information it needs could result in a framework that can create the scenarios we have now.

For example, If we have crutest itself generate the VCR, that could allow crutest to have multiple dsc "region sets" where each could have it's own region information that need not be the same as other region sets.

This flexibility could also allow us run a number of different tests of crutest in a loop where we change the dsc (and hence VCR) on each loop and crutest will just figure it out each time, possibly reducing the size of bash scripts we have now.

leftwo:
If dsc returns the number of "region sets" it is responsible for, this could enable a single dsc to serve a number of (identical) region sets.

This could be combined with also supporting multiple dsc servers and crutest could be smart enough to build a VCR based on both of these.

leftwo:
With #1472, this work is done

Issue #1455, opened 2024-09-13
Some BlockOps need more logic at the Volume layer


leftwo:
Some BlockOp enums might need to be updated to provide additional context.

Specifically:
````
    QueryExtentSize {                                                           
        done: BlockRes<Block>,                                                  
    },                                                                          
    QueryWorkQueue {                                                            
        done: BlockRes<WQCounts>,                                               
    },                                                                          
    ShowWork {                                                                  
        done: BlockRes<WQCounts>,                                               
    }
```

These block ops make sense when a single upstairs exists behind the object that supports the trait.
For example, `Guest` only ever has three downstairs and one work queue.

However, with a `Volume`, you could have multiple sub-volumes and a read only parent.  So what is returned in this case can no longer be represented by a single object, or if it is, it has lost some information that the caller might want.
In the case of `QueryExtentSize`, a single value here for a `Volume` is only correct if everyone happens to share the same extent size, or if there is only a single sub volume.

leftwo:
1508 updated how `query_extent_size` needed to change, that work is done.

Left to do is if we want job level support for Volumes, or just lump it all together like we do now.

Issue #1454, opened 2024-09-11
crutest's `RegionInfo` needs to support Volume level.


leftwo:
The current way crutest works for many tests is to build a `RegionInfo` struct which looks as follows:

```
pub struct RegionInfo {                                                             
    block_size: u64,                                                                
    extent_size: Block,                                                             
    total_size: u64,                                                                
    total_blocks: usize,                                                            
    write_log: WriteLog,                                                            
    max_block_io: usize,                                                            
}
```

This allows it to figure out where the extents end and from that how to create tests that catch edge conditions for us without having to first inspect the region then manually create a specific test for it.

In the volume centric world, where we can have multiple `SubVolumes`, this single structure is no longer contains the information we need, and can't actually represent what we want.

leftwo:
Perhaps a new struct, replace `RegionInfo` with `VolumeInfo`

For a Volume, We could store
`<Vec<RegionInfo>>` for each sub_volume
`<Option<RegionInfo>>` for read_only_parent

With a geust, it would just be:
A vec, with just one `RegionInfo`
`None` for read_only_parent

Something like this that mimics the way we do volumes:
```
struct VolumeInfo {
    // Block size for the volume,
    block_size: u64,
    // Total blocks in the volume
    total_blocks: u64,
    // SubVolumes that make up this volume.
    sub_volume: Vec<SubVolumeInfo>
    read_only_parent: Option<SubVolumeInfo>
    // write log for the volume
    write_log: WriteLog,
} 
```

With
```
struct SubVolumeInfo{
    // Number blocks per extent
    extent_size: Block,
    // Total extents in this sub volume
    extent_count: u64,
    // Total size in bytes for this sub volume
    total_size: u64,
    // Range this sub volume covers.
    range: Range,
}
```

I'm not even sure if we need to (at first) support read only parents in crutest, but we can if we want.

leftwo:
Crutest now supports volume layer actions better.

PR #1472, opened 2024-09-25
More crutest and dsc updates to support Volume layer activities.

leftwo:
This adds support for crutest to use a provided dsc endpoint to construct a Volume object.

These changes should be a fix for both issues:
https://github.com/oxidecomputer/crucible/issues/1451 and https://github.com/oxidecomputer/crucible/issues/1457

Moved the existing volume creation steps to a new function, and added another option on how we can create a Volume.  The two previous ways of creating a volume are not changed (though I changed a log message and added some warnings).  The new code is in taking the dsc provided endpoint and using that to construct a volume.

Additional dsc changes were made to help provide Volume info.
Renamed things in dsc to better reflect what information they hold. Specifically, update a bunch of region set comments, as dsc just controls crucible-downstairs processes, and does not know which ones are part of what region set.

**New dsc commands**
* get_ds_uuid: Returns the UUID for the given client ID
* all_running: Returns true if all downstairs that dsc knows about are currently in Running state.
* get_region_count: Returns the total number of regions that dsc knows about.

**New dsc behavior**
* dsc will now wait on all downstairs starting before taking any commands. The ability for dsc to answer a request can be used by a test to confirm that all downstairs had started.
* Add the ability to supply a dsc endpoint to crutest-cli (fix https://github.com/oxidecomputer/crucible/issues/1459)

**Other changes**
`tools/test_replay.sh` transitioned to using the new --dsc option, as that test already required a dsc endpoint and was using a hard coded default value for it.

`tools/test_restart_repair.sh` was updated to wait for dsc to report that all downstairs are online after a restart.  This avoids a race where we told dsc to start, and then start crutest, but the downstairs are not yet online.

All the tests that use dsc will eventually transition to using it to construct a Volume, but I'm pushing that work to another PR.

**There are more changes coming, specifically**
* Updating replace-before-acive and replace-reconcile to use the new dsc option correctly instead of a default.
* Updating tools/test_* to not require targets for crutest and instead use dsc option.
* More updates in crutest to update the current RegionInfo struct to be aware of multiple sub-volumes.
https://github.com/oxidecomputer/crucible/issues/1454
* Possibly some updates to the BlockIO trait.
https://github.com/oxidecomputer/crucible/issues/1455

**Other work this enables**
* Making tests that use multiple sub-volumes.
* Layering the current set of tests in a way so we can run the same tests with a single sub-volume and with multiple sub-volumes

### DIFF
--- a/crutest/src/cli.rs
+++ b/crutest/src/cli.rs
@@ -445,18 +445,13 @@ async fn handle_dsc(
                 println!("Got res: {:?}", res);
             }
         }
+    } else if let DscCommand::Connect { server } = dsc_cmd {
+        let url = format!("http://{}", server).to_string();
+        println!("Connecting to {:?}", url);
+        let rs = Client::new(&url);
+        *dsc_client = Some(rs);
     } else {
-        match dsc_cmd {
-            DscCommand::Connect { server } => {
-                let url = format!("http://{}", server).to_string();
-                println!("Connecting to {:?}", url);
-                let rs = Client::new(&url);
-                *dsc_client = Some(rs);
-            }
-            _ => {
-                println!("dsc: Need to be connected first");
-            }
-        }
+        println!("dsc: Need to be connected first");
     }
 }
 /*
--- a/crutest/src/cli.rs
+++ b/crutest/src/cli.rs
@@ -25,8 +25,9 @@ pub struct CliAction {
 #[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(Debug, Parser, PartialEq)]
 pub enum DscCommand {
-    /// Connect to the default DSC server (http://127.0.0.1:9998)
-    Connect,
+    /// IP:Port for a dsc server
+    ///  #[clap(long, global = true, default_value = "127.0.0.1:9998", action)]
+    Connect { server: SocketAddr },
     /// Disable random stopping of downstairs
     DisableRandomStop,
     /// Disable auto restart on the given downstairs client ID
@@ -380,7 +381,7 @@ async fn handle_dsc(
 ) {
     if let Some(dsc_client) = dsc_client {
         match dsc_cmd {
-            DscCommand::Connect => {
+            DscCommand::Connect { .. } => {
                 println!("Already connected");
             }
             DscCommand::DisableRandomStop => {
@@ -444,13 +445,18 @@ async fn handle_dsc(
                 println!("Got res: {:?}", res);
             }
         }
-    } else if dsc_cmd == DscCommand::Connect {
-        let url = "http://127.0.0.1:9998".to_string();
-        println!("Connect to {:?}", url);
-        let rs = Client::new(&url);
-        *dsc_client = Some(rs);
     } else {
-        println!("dsc: Need to be connected first");
+        match dsc_cmd {
+            DscCommand::Connect { server } => {
+                let url = format!("http://{}", server).to_string();
+                println!("Connecting to {:?}", url);
+                let rs = Client::new(&url);
+                *dsc_client = Some(rs);
+            }
+            _ => {
+                println!("dsc: Need to be connected first");
+            }
+        }
     }
 }
 /*
--- a/crutest/src/main.rs
+++ b/crutest/src/main.rs
@@ -760,7 +760,7 @@ async fn make_a_volume(
     let mut crucible_opts = CrucibleOpts {
         id: up_uuid,
         target: opt.target.clone(),
-        lossy: opt.lossy,
+        lossy: false,
         flush_timeout: opt.flush_timeout,
         key: opt.key.clone(),
         cert_pem: opt.cert_pem.clone(),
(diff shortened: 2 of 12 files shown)


crutest/src/cli.rs
```diff
@@ -444,13 +445,18 @@ async fn handle_dsc(
...
             }
         }
-    } else if dsc_cmd == DscCommand::Connect {
-        let url = "http://127.0.0.1:9998".to_string();
-        println!("Connect to {:?}", url);
-        let rs = Client::new(&url);
-        *dsc_client = Some(rs);
     } else {
-        println!("dsc: Need to be connected first");
+        match dsc_cmd {
```
mkeeter, line 449: could you write this as
```rust
} else if let DscCommand::Connect { server } = dsc_cmd {
```
instead of the nested `match`?

crutest/src/main.rs
```diff
@@ -139,7 +137,7 @@ enum Workload {
     ReplaceBeforeActive {
         /// URL location of the running dsc server
         #[clap(long, default_value = "http://127.0.0.1:9998", action)]
-        dsc: String,
+        dsc_str: String,
```
mkeeter, line 140: Could we use the new global `dsc` argument here (and in `ReplaceReconcile`)?

crutest/src/main.rs
```diff
@@ -744,9 +742,190 @@ async fn handle_signals(
...
+        };
+        info!(test_log, "Using VCR: {:?}", vcr);
+
+        if opt.gen != 0 {
+            warn!(test_log, "gen option is ignored when VCR is provided");
+        }
+        if !opt.target.is_empty() {
+            warn!(test_log, "targets are ignored when VCR is provided");
+        }
+        let volume = Volume::construct(vcr, pr, block_io_logger).await.unwrap();
```
mkeeter, line 797: Should we warn or error if `dsc` is provided here?

Personally, I would vote for errors (including for the lines that are warnings above), since invalid CLI arguments are probably a sign that the user is doing something wrong (and no one reads the logs).

crutest/src/main.rs
```diff
@@ -744,9 +742,190 @@ async fn handle_signals(
...
+        let extent_info = RegionExtentInfo {
+            block_size: ri.block_size,
+            blocks_per_extent: ri.blocks_per_extent,
+            extent_count: ri.extent_count,
+        };
+
+        let res = dsc_client.dsc_get_region_count().await.unwrap();
+        let regions = res.into_inner();
+
+        let sv_count = regions / 3;
```
mkeeter, line 824: Error if `sv_count` is 0?

crutest/src/main.rs
```diff
@@ -744,9 +742,190 @@ async fn handle_signals(
...
+        // be detected later in the process and handled by the upstairs.
+        let mut extent_info_result = None;
+        for target in &crucible_opts.target {
+            let port = target.port() + crucible_common::REPAIR_PORT_OFFSET;
+            println!("look at: http://{}:{} ", target.ip(), port);
+            let repair_url = format!("http://{}:{}", target.ip(), port);
+            let repair_client = repair_client::new(&repair_url);
+            match repair_client.get_region_info().await {
+                Ok(ri) => {
+                    println!("RI is: {:?}", ri);
```
mkeeter, line 889: Same here

crutest/src/main.rs
```diff
@@ -744,9 +742,190 @@ async fn handle_signals(
...
+                    extent_info_result = Some(RegionExtentInfo {
+                        block_size: ri.block_size(),
+                        blocks_per_extent: ri.extent_size().value,
+                        extent_count: ri.extent_count(),
+                    });
+                    break;
+                }
+                Err(e) => {
+                    println!(
```
mkeeter, line 898: Same here (`warn!` perhaps?)

mkeeter (approved):
A few small comments, but generally LGTM

crutest/src/main.rs
leftwo: We can and will!  Coming  in my next PR.
It also requires some changes to the tests themselves, and I felt this PR had enough in it.

crutest/src/main.rs
leftwo: I think we are not quite ready to make these hard errors yet, though you are right that no one reads the logs, unless there is a failure.  This could lead to someone thinking they are testing one thing, when in fact they are testing another, which is not good.

My next PR (yes, I'm using that again) has more changes that will get us closer to not
needing both targets and a VCR file.  The 

There are some cases where we use the targets provided for building test specific args.  The `ReplaceBeforeActive`test you pointed to above is one of these cases.

I've made [an issue](https://github.com/oxidecomputer/crucible/issues/1473) to come back and harden up the errors here.

For `dsc` specifically, we could, in theory, create a volume from a VCR file, then use DSC to connect and control the downstairs themselves.  I'm trying to figure out how we will support having sub-volumes with different sizes and how to ingest that into crutest, and I was able to do that by layering a few dsc create commands in a way that did it.  
Another idea is to support multiple `dsc` endopints.  I'm still not sure which way is better.

crutest/src/cli.rs
leftwo: Yes, that's much better.  Done in 11fb248

crutest/src/main.rs
leftwo: Done in 11fb248

crutest/src/main.rs
```diff
@@ -744,9 +742,190 @@ async fn handle_signals(
...
+
+        // For each sub-volume, we need to know:
+        // block_size, blocks_per_extent, and extent_size.  We can get any
+        // of the target downstairs to give us this info, if they are
+        // running.  We don't care which one responds.  Any mismatch will
+        // be detected later in the process and handled by the upstairs.
+        let mut extent_info_result = None;
+        for target in &crucible_opts.target {
+            let port = target.port() + crucible_common::REPAIR_PORT_OFFSET;
+            println!("look at: http://{}:{} ", target.ip(), port);
```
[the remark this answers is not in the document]
leftwo: I forgot I had a logger here (or possibly when I wrote this before I moved it)
Fixed in 11fb248

crutest/src/main.rs
leftwo: Fixed in 11fb248

crutest/src/main.rs
leftwo: Yeah, I consider myself warned in 11fb248

leftwo:
If anyone else want's to comment (or has comments in progress) let me know, otherwise I'll merge this
when the CI tests pass.

PR #1480, opened 2024-09-27
Update more tests to use dsc

leftwo:
This is plucking out non `BlockIO` parts from: https://github.com/oxidecomputer/crucible/pull/1474

Give crutest the list of targets it will be operating on so some tests that need that list of targets can still get it.

Updated more tests to use dsc to determine targets instead of having to provide them on the command line.

Updated most of test_up.sh to use dsc for targets instead of supplying them on the command line.
There is still some work required to get hammer tests to use dsc, if we decide to do that.

### DIFF
--- a/crutest/src/main.rs
+++ b/crutest/src/main.rs
@@ -135,20 +135,12 @@ enum Workload {
     },
     /// Test that we can replace a downstairs when the upstairs is not active.
     ReplaceBeforeActive {
-        /// URL location of the running dsc server
-        #[clap(long, default_value = "http://127.0.0.1:9998", action)]
-        dsc_str: String,
-
         /// The address:port of a running downstairs for replacement
         #[clap(long, action)]
         replacement: SocketAddr,
     },
     /// Test replacement of a downstairs while doing the initial reconciliation.
     ReplaceReconcile {
-        /// URL location of the running dsc server
-        #[clap(long, default_value = "http://127.0.0.1:9998", action)]
-        dsc_str: String,
-
         /// The address:port of a running downstairs for replacement
         #[clap(long, action)]
         replacement: SocketAddr,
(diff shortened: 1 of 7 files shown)


crutest/src/main.rs
```diff
@@ -1344,11 +1344,16 @@ async fn main() -> Result<()> {
...
             .await?;
         }
-        Workload::ReplaceBeforeActive {
-            dsc_str,
-            replacement,
-        } => {
-            let dsc_client = Client::new(&dsc_str);
+        Workload::ReplaceBeforeActive { replacement } => {
+            let dsc_client = match opt.dsc {
+                Some(dsc_addr) => {
```
mkeeter, line 1349: This is a small behavior change: previously, there was a default value of `http://127.0.0.1:9998`; now, we return an error.  Dunno if the change is intentional or not!

crutest/src/main.rs
```diff
@@ -1375,11 +1379,16 @@ async fn main() -> Result<()> {
             )
             .await?;
         }
-        Workload::ReplaceReconcile {
-            dsc_str,
-            replacement,
-        } => {
-            let dsc_client = Client::new(&dsc_str);
+        Workload::ReplaceReconcile { replacement } => {
+            let dsc_client = match opt.dsc {
```
mkeeter, line 1383: Same here (there was previously a default value for `dsc_str`)

mkeeter (approved):

crutest/src/main.rs
leftwo: Yeah, before it assumed that any `dsc` it finds at the default would be correct, but that
may not be what the user wanted.  More so now that there is a global `dsc` option which
should be what these tests use.  Hopefully fewer surprises as this makes the dsc server
target required.

PR #1482, opened 2024-09-30
Remove `GuestWorkId` (2/2)

mkeeter:
(staged on top of #1481)

`GuestWorkId` is always 1:1 with `JobId`, so let's get rid of it!

It's not exposed to the guest or used by anything outside of Crucible internals.

The only exception is DTrace probes, which now use the `JobId` everywhere (instead of a mix of `JobId` and `GuestWorkId`).  The `gw__*__start` probes now fire immediately after a `JobId` is assigned, rather than when the `GuestWorkId` was assigned.


jmpesp:
Sorry if this is missing something, I haven't reviewed the entirety of the PR - do we need to keep the gw probes?

mkeeter:
> Sorry if this is missing something, I haven't reviewed the entirety of the PR - do we need to keep the gw probes?

Yeah, I think they're still valuable as the "earliest / latest possible" probe point for IOs.

leftwo:
> > Sorry if this is missing something, I haven't reviewed the entirety of the PR - do we need to keep the gw probes?
> 
> Yeah, I think they're still valuable as the "earliest / latest possible" probe point for IOs.

I hope/thought we still had probe points for:
1) IO lands in crucible.
2) IO is acked back to propolis
3) IO is completed on all downstairs.
4) All traces of that IO are done in crucible and we drop it.

I think that's still true.

faithanalog:
> > > Sorry if this is missing something, I haven't reviewed the entirety of the PR - do we need to keep the gw probes?
> > 
> > 
> > Yeah, I think they're still valuable as the "earliest / latest possible" probe point for IOs.
> 
> I hope/thought we still had probe points for:
> 
>     1. IO lands in crucible.
> 
>     2. IO is acked back to propolis
> 
>     3. IO is completed on all downstairs.
> 
>     4. All traces of that IO are done in crucible and we drop it.
> 
> 
> I think that's still true.

yeah I believe the gw probes are for "2. IO is acked back to propolis". And this PR leaves them in.

upstairs/src/guest.rs
```diff
@@ -34,11 +34,6 @@ use uuid::Uuid;
  */
```
faithanalog, line 34: Is this comment out of date now that `ds_id` is gone here?

mkeeter:
@leftwo 

> I think that's still true.

I think so too!  I updated some of the text in the probe descriptions, so it should be accurate.

upstairs/src/upstairs.rs
```diff
@@ -1488,28 +1483,20 @@ impl Upstairs {
          * Grab this ID after extent_from_offset: in case of Err we don't
          * want to create a gap in the IDs.
          */
-        let (gw_id, _) = self.guest.guest_work.submit_job(
-            |gw_id| {
-                if write.is_write_unwritten {
-                    cdt::gw__write__unwritten__start!(|| (gw_id.0));
-                } else {
-                    cdt::gw__write__start!(|| (gw_id.0));
-                }
```
faithanalog, line 1497: do we lose anything with these `gw__write__start`s going away? I know there's an `up_to_ds` probe write below that's telling us the same thing, im not sure if they mean two semantically different things or what

mkeeter: It's still around, but moved to the beginning of `Downstairs::create_and_enqueue_write_eob` (right after we get the `JobId`)

upstairs/src/guest.rs
mkeeter: It's kinda weird; edited to read `JobId` instead of `ds_id` (because we still have a `HashMap<JobId, GtoS>`

faithanalog (approved):

leftwo (approved):

PR #1484, opened 2024-09-30
More multiple region support.

leftwo:
Even more updates to support Volume layer activities.

Added a new test to CI that will run "test_up.sh encrypted" with two region sets (6 downstairs).

Crutest changes:
Added some comments to crutest tests that were missing them.
Crutest will now determine the total number of downstairs it expects to be using and compare that value when looking for an expected number of downstairs in an active state.
Updated replay, replace, replace-before-active, and replace-while-reconcile tests to support having more than one region set.
Fixed a bug in the replace-before-active test where previous tests could leave things broken in a way the test could not repair.

Removed the no longer valid NEW columns from the DTrace `single_up_info.d` script.

test_up.sh changes:
Added support for running all the tests with a user supplied number of region sets instead of just defaulting to one (three downstairs).
Some other misc cleanup of test_up.sh.

### DIFF
--- a/crutest/src/main.rs
+++ b/crutest/src/main.rs
@@ -2472,7 +2472,7 @@ async fn replace_while_reconcile(
                 wc.active_count
             );
             if wc.up_count + wc.ds_count == 0 && wc.active_count == ds_total {
-                info!(log, "[{c}] Replay: All jobs finished, all DS active.");
+                info!(log, "[{c}] All jobs finished, all DS active.");
                 break;
             }
             tokio::time::sleep(tokio::time::Duration::from_secs(4)).await;
@@ -2528,6 +2528,12 @@ async fn replace_before_active(
     assert!(targets.len() % 3 == 1);
 
     info!(log, "Begin replacement before activation test");
+    // We need to start from a known state and be sure that all three of the
+    // current downstairs are consistent with each other. To guarantee this
+    // we write to every block, then flush, then read.  This way we know
+    // that the initial downstairs are all synced up on the same flush and
+    // generation numbers.
+    fill_workload(volume, ri, true).await?;
     let ds_total = targets.len() - 1;
     let mut old_ds = 0;
     let mut new_ds = targets.len() - 1;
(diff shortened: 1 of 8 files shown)


leftwo:
I believe I've found all the problems CI was tripping over.. finally!

leftwo:
Fixing https://github.com/oxidecomputer/crucible/issues/1491 was part of the changes done here as well.

crutest/src/main.rs
```diff
@@ -2431,15 +2471,15 @@ async fn replace_while_reconcile(
...
+            if wc.up_count + wc.ds_count == 0 && wc.active_count == ds_total {
+                info!(log, "[{c}] All jobs finished, all DS active.");
                 break;
             }
             tokio::time::sleep(tokio::time::Duration::from_secs(4)).await;
         }
 
-        old_ds = (old_ds + 1) % 4;
-        new_ds = (new_ds + 1) % 4;
+        old_ds = (old_ds + 1) % (ds_total as u32 + 1);
```
mkeeter, line 2481: Nit: what's with the `as u32` cast here (and below)?

tools/test_up.sh
```diff
@@ -212,8 +237,8 @@ echo cp -r "${testdir}/${port}" "${testdir}/previous"
 cp -r "${testdir}/${port}" "${testdir}/previous"
 
 new_args+=( --verify-in "${testdir}/verify_file" )
-echo "$ct" repair -g "$gen" -q "${new_args[@]}"
-if ! "$ct" repair -g "$gen" -q "${new_args[@]}"; then
+echo "$ct" fill -g "$gen" -q "${new_args[@]}"
+if ! "$ct" fill -g "$gen" -q "${new_args[@]}"; then
```
mkeeter, line 241: Is this change intentional?

leftwo: Yes, this was one of the test bugs.
Sometimes the repair test would not actually change the downstairs that we dump below.
The `fill` touches every extent, so we know there should be a difference when we dump.

crutest/src/main.rs
leftwo: I'm just doing what the compiler tells me, man!

`old_ds` which is sent to dsc_client functions wants it to be u32.
However, when we use it as an index for `targets[`,  both `old_ds` and `new_ds` need to be usize.

If you think it's better, I could make it a `usize`, and cast to` u32` when I send it to dsc?

mkeeter (approved):

Issue #1368, opened 2024-06-22
README should point to RFDs + podcast episode


bcantrill:

tim-seoss:
Since I'd just excavated these myself...  LMK if I'm missing any RFDs etc.

PR #1486, opened 2024-10-02
Expand summary and add documentation references to the README.

tim-seoss:
Fixes upstream bug #1368

If there are other RFDs or other public docs which I'm missing, please let me know!


leftwo (approved):
Thanks!

PR #1489, opened 2024-10-03
Move complete job tracker to a helper object

mkeeter:
This reduces visible implementation details, in case we want to change this object in the future (e.g. using a `BTreeSet` or a sorted `Vec`).

### DIFF
--- a/downstairs/src/complete_jobs.rs
+++ b/downstairs/src/complete_jobs.rs
@@ -1,0 +1,55 @@
+use crate::JobId;
+
+/// Stores a flush operation and a set of complete jobs
+///
+/// The flush operation is not included in the list of complete jobs, but acts
+/// as a lower bound for dependencies; any job older than the flush is assumed
+/// to have completed.
+#[derive(Debug)]
+pub struct CompletedJobs {
+    last_flush: Option<JobId>,
+    completed: Vec<JobId>,
+}
+
+impl CompletedJobs {
+    pub fn new(last_flush: Option<JobId>) -> Self {
+        Self {
+            last_flush,
+            completed: vec![],
+        }
+    }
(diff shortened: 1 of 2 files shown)


downstairs/src/lib.rs
```diff
@@ -3316,9 +3312,7 @@ impl Work {
         let num_deps_outstanding = work
```
jmpesp, line 3312: non-blocking: what about `num_deps_outstanding` moving into the helper too?

downstairs/src/complete_jobs.rs
```diff
@@ -0,0 +1,55 @@
...
+/// as a lower bound for dependencies; any job older than the flush is assumed
+/// to have completed.
+#[derive(Debug)]
+pub struct CompletedJobs {
+    last_flush: Option<JobId>,
+    completed: Vec<JobId>,
+}
+
+impl CompletedJobs {
+    pub fn new(last_flush: Option<JobId>) -> Self {
```
jmpesp, line 15: is the last_flush argument always going to be `None`? we could drop the argument if so

jmpesp (approved):
one suggestion, and one non-blocking suggestion!

downstairs/src/complete_jobs.rs
mkeeter: Not quite, it's `Some(..)` when we've received `Message::LastFlush` during negotiation!

faithanalog (approved):

downstairs/src/lib.rs
mkeeter: I will consider it; not sure if I want `CompletedJobs` to know about the `IOop`, though...

PR #1490, opened 2024-10-03
Simplify matches

mkeeter:
(Staged on top of #1489)

Drive-by cleanup to simplify match statements:

- Use `..` instead of `foobar: _` to ignore fields
- Consolidate `match` statement with identical branches

(I don't think we're ever deliberately destructuring objects to check their shape, but let me know if that's an incorrect assumption!)

### DIFF
--- a/protocol/src/lib.rs
+++ b/protocol/src/lib.rs
@@ -592,53 +592,55 @@ impl Message {
     /// Return true if this message contains an Error result
     pub fn err(&self) -> Option<&CrucibleError> {
         match self {
-            Message::HereIAm { .. } => None,
-            Message::YesItsMe { .. } => None,
-            Message::VersionMismatch { .. } => None,
-            Message::ReadOnlyMismatch { .. } => None,
-            Message::EncryptedMismatch { .. } => None,
-            Message::PromoteToActive { .. } => None,
-            Message::YouAreNowActive { .. } => None,
-            Message::YouAreNoLongerActive { .. } => None,
-            Message::UuidMismatch { .. } => None,
-            Message::Ruok { .. } => None,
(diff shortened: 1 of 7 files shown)


protocol/src/lib.rs
```diff
@@ -592,53 +592,55 @@ impl Message {
...
+            | Message::ExtentLiveNoOp { .. }
+            | Message::Flush { .. }
+            | Message::ReadRequest { .. }
+            | Message::WriteUnwritten { .. }
+            | Message::Unknown(..) => None,
+
+            Message::ExtentError { error, .. }
+            | Message::ErrorReport { error, .. } => Some(error),
 
             Message::ExtentLiveCloseAck { result, .. } => result.as_ref().err(),
```
leftwo, line 631: I don't know if this makes it any simpler, but I also don't really care one way or the other.  I do think it's all part of your master plan to touch every line in Crucible :)

However, you should be able to put `Message::ExtentLiveCloseAck` with the rest just below, right?

leftwo (approved):

jmpesp (approved):
same question as Alan, otherwise :rocket:

protocol/src/lib.rs
mkeeter: > I don't know if this makes it any simpler, but I also don't really care one way or the other. I do think it's all part of your master plan to touch every line in Crucible :)

🤫 

> However, you should be able to put `Message::ExtentLiveCloseAck` with the rest just below, right?

They're actually different types `Result<(), CrucibleError>` versus `Result<(u64, u64, bool), CrucibleError>`, so you can't match on them simultaneously (even though we're only using the error).

faithanalog (approved):

PR #1492, opened 2024-10-04
Add `VolumeBuilder` type

mkeeter:
The `Volume` type does too many things:

- It is a pseudo-builder, with mutating functions like `add_subvolume` and `add_read_only_parent`
- It is cloneable, but cloning it could be expensive (because it contains `Vec`)
- It implements the `BlockIO` trait (and can hence be nested)

The combination of mutability and cloneability opens the possibility for confusion:

```rust
let mut foo = Volume::create(..);
let bar = foo.clone();
foo.add_subvolume(..);
```

Here, `bar` will share the subvolumes of `foo` before the clone, but will not contain the new subvolume.

In addition, we mix `Volume` and `Arc<Volume>` widely in the codebase.  Even within `Volume`, we have an `Option<Arc<SubVolume>>`, which in turn contains `Arc<dyn BlockIO>` – two layers of indirection!

This PR cleans everything up!

- `Volume` is retconned into a wrapper type: it wraps a single `Arc<VolumeInner>`
- `Volume` implements `Clone` (which is cheap, because it's just a reference increment)
- `VolumeInner` does not implement `Clone`, because it could be expensive
- We remove (almost) all instances of `Arc<Volume>` from the codebase.
- The inner `Arc` from `SubVolume` is removed
- `Arc<AtomicXYZ>` in `Volume` become `AtomicXYZ` in `VolumeInner`, since the sharing happens at the `Volume` level
- All of the builder-like functionality moves into a `VolumeBuilder` object
- As an `Arc<..>` wrapper, the `Volume` is immutable once constructed


crutest/src/main.rs
```diff
@@ -829,7 +828,8 @@ async fn make_a_volume(
         }
 
         // We start by creating the overall volume.
```
leftwo, line 830: Nit: We start by creating the volume builder.

leftwo (approved):
I think I found the answer to my own question about being able to add a sub-volume to an existing volume, and if you can then I'm all good here.

crutest/src/main.rs
mkeeter: Fixed!

leftwo:
> > Over in propolis, if we decide that we want to grow a volume by adding sub volumes to the existing volume, could we still do that?
> 
> I'm curious what answer you have found for yourself, because I think it's not always possible (after this PR).
> 
> You can call `Volume::deconstruct`, but that will fail if someone else has cloned the `Volume` (i.e. there are >1 objects using the `Arc<VolumeInner>`). The easiest way do accomplish the same goal would be to build a new `Volume`, which wraps your previous volume along with a new subvolume at the end.

Ah, yeah, if a clone blocks it, I missed that part.

Ideally, the same volume would still exist (and any IOs outstanding would continue to flow).  We would just be tacking something on the end of the volume and increasing the LBA range the volume has.  I'm not even sure if that would work now as we never tried it, but it was the idea back when the Volume layer was created.

leftwo:
From chat, I wanted to record the discussion:

`Matt Keeter` thinking about the immutable Volume PR, how intent are you on adding new subvolumes while a Volume is alive and possibly running?

`Alan (he/him)` It's currently a dream

`Matt Keeter` I'm thinking about things like CrucibleBackend::create in Propolis, which spawns a scrubber task on the volume – if that scrubber is running but we want the underlying Volume to be mutable, it seems potentially messy
since the scrubber will be operating on the original volume, should modifications to scrub_point etc be shared with the extended volume?
(I don't expect an answer now, but just an example of things that become harder if we want mutable volumes)

`Alan (he/him)` Yeah, these are the questions I want to consider.

`Matt Keeter` my inclination is to merge the immutable volumes PR, because we can always go back to mutable volumes – but it forces us to do so intentionally
versus the current semi-mutable state of volumes, which has the potential for confusion

`Alan (he/him)` Yeah, I'm thinking about what to do with the immutable volumes PR too.

`Matt Keeter` semi-related, I was threatening James with this earlier:
```
enum VolumeTree {
    Concatenate(Box<VolumeTree>, Box<VolumeTree>),
    Interleave(Box<VolumeTree>, Box<VolumeTree>),
    Leaf(Volume),
}
```
😛

`Alan (he/him)` But, I don't want us to close the door on mutable volumes, if that is our path to growing a volume live.

`Matt Keeter`: eh, none of this is serialized to files / databases or defined as stable network messages, so I'm less worried about closing doors – they can be reopened without too much difficulty once we figure out what we want / need

`Alan (he/him)` Yeah, that part is true.
I worry that we would build something else that took advantage of the immutable state of a volume, then had to undo/redo whatever that was in the future.
But, I guess it's us that would be building it.

`Matt Keeter` yeah, and even right now, we're (kinda) relying on the volumes being immutable, e.g. by cloning them into worker tasks in Propolis
if someone goes and extends the original volume, all of the IO worker tasks in Propolis aren't going to see that!
because only the Arc<..> members (scrub point, etc) are shared; the Vec<SubVolume> is not
I could see switching to an Arc<RwLock<VolumeInner>> if we decide that mutation is necessary, which would be a pretty easy change

`Alan (he/him)` And, to update that, we would not have to deconstruct and re-create the volume object in Propolis?

`Matt Keeter` correct – we'd have to get a reader lock to do IO (and we can have many reader locks simultaneously, so worker threads can cooperate)
and then if we wanted to extend the subvolume list, we'd get a writer lock and all of the IO tasks would have to wait for that to be released
(the more I think about it, the more this feels like the correct next step if we decide to reintroduce mutability)

mkeeter:
Shorter version of the above:

- We're **already** assuming that `Volume` objects are immutable; e.g. we clone them into Propolis worker tasks.  If we then modify the original volume, those changes wouldn't be reflected in the workers!
- If we do want to mutate volume objects, we could use an `Arc<RwLock<VolumeInner>>`, which would let us safely use + modify volumes between multiple tasks / threads

jmpesp (approved):
These changes make working with the objects more type safe, so I'm for it.

leftwo:
From chat:

```
I'm going to stage the Propolis changes before merging, to make sure there are no surprises there
yeah looks fine

@@ -182,8 +183,8 @@ impl CrucibleBackend {
             block_size,
             size as usize,
         ));
-        let mut volume = Volume::new(block_size, log);
-        volume
+        let mut builder = VolumeBuilder::new(block_size, log);
+        builder
             .add_subvolume(mem_disk)
             .await
             .map_err(|e| std::io::Error::from(e))?;
@@ -191,7 +192,7 @@ impl CrucibleBackend {
         Ok(Arc::new(CrucibleBackend {
             state: Arc::new(WorkerState {
                 attachment: block::BackendAttachment::new(),
-                volume,
+                volume: builder.into(),
                 info: block::DeviceInfo {
                     block_size: block_size as u32,
                     total_size: size / block_size,
```

PR #1493, opened 2024-10-04
Simplify `CompleteJobs`

mkeeter:
We never actually use the "jobs before `last_flush` count as completed" property of `CompleteJobs`.

From discussions in chat, this may have been a legacy from when we sent literally every previous job in the dependency list.
Now that flushes act as a full dependency barrier, it shouldn't be possible to depend on anything before a flush.

This PR removes special-casing of `last_flush`; instead it becomes first among equals in `completed: Vec<JobId>`.  Most of the changes are to edit test cases that violate the "flushes are a full barrier" property.

(This is preparation for a new `Barrier` operation, which is part of RFD 518)

### DIFF
--- a/downstairs/src/complete_jobs.rs
+++ b/downstairs/src/complete_jobs.rs
@@ -7,15 +7,13 @@ use crate::JobId;
 /// to have completed.
 #[derive(Debug)]
 pub struct CompletedJobs {
-    last_flush: Option<JobId>,
     completed: Vec<JobId>,
 }
 
 impl CompletedJobs {
     pub fn new(last_flush: Option<JobId>) -> Self {
         Self {
-            last_flush,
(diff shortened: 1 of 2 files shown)


jmpesp (approved):

downstairs/src/complete_jobs.rs
```diff
@@ -7,15 +7,13 @@ use crate::JobId;
 /// to have completed.
```
faithanalog, line 7: nit: this comment is wrong now

faithanalog (approved):

downstairs/src/complete_jobs.rs
mkeeter: thanks, fixed!

PR #1494, opened 2024-10-07
Add `IOop::Barrier`

mkeeter:
(staged on top of #1493)

The `Barrier` op does no work, but serves as a full dependency barrier (like a flush) so that the Downstairs can reset its list of completed jobs.  Note that it does _not_ update `last_flush` or retire jobs in the Upstairs, because it doesn't actually persist data to disk.

This is a building block for removing automatic flushes (see RFD 518).

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -82,6 +82,9 @@ enum IOop {
         snapshot_details: Option<SnapshotDetails>,
         extent_limit: Option<ExtentId>,
     },
+    Barrier {
+        dependencies: Vec<JobId>, // Jobs that must finish before this
+    },
     /*
      * These operations are for repairing a bad downstairs
      */
@@ -114,6 +117,7 @@ impl IOop {
         match &self {
             IOop::Write { dependencies, .. }
             | IOop::Flush { dependencies, .. }
+            | IOop::Barrier { dependencies, .. }
             | IOop::Read { dependencies, .. }
             | IOop::WriteUnwritten { dependencies, .. }
             | IOop::ExtentClose { dependencies, .. }
@@ -687,6 +691,9 @@ pub fn show_work(ds: &mut Downstairs) {
                     IOop::Read { dependencies, .. } => ("Read", dependencies),
                     IOop::Write { dependencies, .. } => ("Write", dependencies),
                     IOop::Flush { dependencies, .. } => ("Flush", dependencies),
+                    IOop::Barrier { dependencies, .. } => {
+                        ("Barrier", dependencies)
+                    }
                     IOop::WriteUnwritten { dependencies, .. } => {
                         ("WriteU", dependencies)
                     }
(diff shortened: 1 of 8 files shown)


faithanalog:
It looks like this just adds the IOop, and doesn't change replay behavior currently, right? Seems fine to me

faithanalog (approved):

faithanalog (commented):
oh i need to fill this text field to make octocat happy

mkeeter:
> It looks like this just adds the IOop, but the operation won't be submitted until a future PR, and doesn't change replay behavior currently, right? Seems fine to me, as long as that CI failure was a flake and not something real

Yup, this is just adding the `IOop`, but doesn't use it (yet)

downstairs/src/lib.rs
```diff
@@ -3674,7 +3719,10 @@ mod test {
             .iter()
             .all(|dep| work.completed.is_complete(*dep)));
 
-        if matches!(job, IOop::Flush { .. }) {
+        // Flushes and barriers both guarantee that no future jobs will depend
+        // on jobs that preceded them, so we reset the completed jobs list to
+        // only include their job id.
+        if matches!(job, IOop::Flush { .. } | IOop::Barrier { .. }) {
```
leftwo, line 3725: This means the lowest job in `work.completed` will always be either a flush or a barrier?

upstairs/src/client.rs
```diff
@@ -1508,6 +1512,16 @@ impl DownstairsClient {
                     }
                     self.last_flush = ds_id;
                 }
+                IOop::Barrier { .. } => {
+                    assert!(read_data.blocks.is_empty());
+                    assert!(read_data.data.is_empty());
+                    assert!(extent_info.is_none());
+
+                    if jobs_completed_ok == 2 {
```
leftwo, line 1520: For the barrier, do we want `jobs_completed_ok == 3`?

And, we are not acking this to anyone, but do you plan to use ackable later to indicate that future action can move forward?

downstairs/src/lib.rs
mkeeter: Yup, exactly

upstairs/src/client.rs
mkeeter: I'm ambivalent!  Like you said, we're not actually acking this to anyone; I also don't plan to use presence in `ackable_work` as an indication for anything here (unlike how it's used in live-repair).

This means the only thing that happens when the job is acked is that it moves from `GuestWork::active` to `GuestWork::completed`.  Setting the ack to happen at `jobs_completed_ok == 3` seems reasonable to me.

leftwo: If we are not using `ackable` for anything, then I would vote for not setting it at all.
Otherwise I'll come back here and see it set and wonder what was consuming that and be confused when I found nothing.

I do care about the dtrace probe though, that one we might actually want to know when this operation has finised on all three downstairs. I don't think I would ever care to know that the barrier had finished on 2/3 downstairs.

leftwo (approved):

upstairs/src/client.rs
mkeeter: We do need to use `ackable`, because the job is stored in `GuestWork` so we need to "ack" it to remove it from that map (even though nothing else happens in the ack).

I changed the threshold to 3 in ed52bf5

leftwo:
Ignore the failure in  `test-up-2region-encrypted`.  I'm working on a fix and I have another PR out to disable it till I get it fixed.

PR #1496, opened 2024-10-07
Add region set options to more tests.

leftwo:
Give test_live_repair, test_replace_special, and test_replay an option to indicate how many region sets should be created for the test.  This allows us to use the same script to test single and multiple sub-volumes.

Updated a few of the tests to no longer run a smaller test in a loop, but instead just run a longer test.  This makes better sense when testing many sub-volumes as a smaller test might not cover everything. It has the side effect of not printing regular status messages, but if we find that is important we can come up with another way of displaying test progress.

Future PRs will use these options to increase test coverage on multiple sub-volumes.

### DIFF
--- a/tools/test_live_repair.sh
+++ b/tools/test_live_repair.sh
@@ -18,6 +18,7 @@ trap ctrl_c INT
 function ctrl_c() {
     echo "Stopping at your request"
     ${dsc} cmd shutdown
+    exit 1
 }
 
 REGION_ROOT=${REGION_ROOT:-/var/tmp/test_live_repair}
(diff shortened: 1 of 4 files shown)


tools/test_live_repair.sh
```diff
@@ -91,63 +108,44 @@ if ! ps -p $dsc_pid > /dev/null; then
 fi
 
 gen=1
-# Initial seed for verify file
+# Seed the initial volume
+echo "$(date) Begin pretest initial fill" | tee -a "$test_log"
 if ! "$crucible_test" fill --dsc 127.0.0.1:9998 -q -g "$gen"\
-          --verify-out "$verify_log" >> "$test_log" 2>&1 ; then
-    echo Failed on initial verify seed, check "$test_log"
+          --skip-verify >> "$test_log" 2>&1 ; then
```
mkeeter, line 114: Is this just to go faster?

mkeeter (approved):

tools/test_live_repair.sh
leftwo: The addition of `--skip-verify`, yeah, that's just to go a little faster.
We don't bother with the `--verify-out` any longer because we are not stopping and starting crutest several times, so we no longer need to read in the verify info from the previous loop.

The stopping/starting part of the test here was causing problems when we have multiple region sets.  The region set that we ended the previous test loop on might not match up with the region set that we start with by default.

Issue #1358, opened 2024-06-17
revisit flush_timeout


faithanalog:
## Background:

Upstairs periodically sends flush commands to Downstairs, even if the guest is not asking for disk flushes. Originally this was every 5 seconds. I *believe*, though am not sure, that it was added to prevent jobs piling up in the upstairs queue.

We adjusted this to 0.5 seconds back when we were still using the sqlite: https://github.com/oxidecomputer/crucible/blob/8757b3fb55a1763382ad7111144bc60ec72af23d/upstairs/src/lib.rs#L9675

We adjusted to 0.5 because, at the time, the cost of an extent flush required a lot of work clearing old metadata contexts out of the sqlite database. That work scaled up directly with the number of writes that had hit an extent since the last flush, and was causing some pretty terrible latency bubbles that we wanted to avoid. Bryan did some testing, found 0.5 gave the best results ( https://github.com/oxidecomputer/crucible/issues/757 ). Note that this was also pre fast-write-ack.

And it remains 0.5 to this day:

https://github.com/oxidecomputer/crucible/blob/7d6c7e1e71d0b389999be06515db855bf273989e/upstairs/src/upstairs.rs#L396

This is configurable in the VCR, but Nexus is written to always pass None and accept our default.

## Why we might want to change it

Well for one thing, we are not on the sqlite backend anymore, and our new one has different flush performance characteristics. But also, we may be sending more fsyncs to zfs than the guest actually cares about. That has a cost to it.

## Questions

- Does adjusting the flush timeout affect performance or CPU usage?
- If it is also still serving the role of preventing job pileup, is it still the right mechanism for that?

leftwo:
The flush timeout does still play a role in preventing job pileup.
For replay to work, we keep a list of all jobs since the last ack'd flush, and if a
downstairs goes away and comes back, we replay all the work since that last confirmed flush.
The frequent flush allows us to discard any work and not have to keep that work in memory.

So, we would need to determine a new way of allowing replay to work correctly if we want
to increase the flush timeout.

PR #1505, opened 2024-10-14
Only send flushes when Downstairs is idle; send Barrier otherwise

mkeeter:
This PR removes automatic flushes, per [RFD 518](https://rfd.shared.oxide.computer/rfd/518).  Instead, the new `Barrier` operation is sent.  If the system is idle for a particular amount of time, we send a final `Flush` to put everything into a known state.

When the Upstairs retires jobs after a barrier operation, the system as a whole becomes ineligible for replay.  This state determines whether the new Downstairs reconnects through `Offline` (which does replay) or `Faulted` (which does live-repair instead).

Removing automatic flushes is a noticeable performance improvement:

Screenshot 2024-10-11 at 11 06 26 AM

(tested on the London mini-cluster, with Upstairs and 3x Downstairs on different sleds)

```
region info:
  block size:      4096 bytes
  blocks / extent: 16384
  extent size:     64 MiB
  extent count:    2048
  total blocks:    33554432
  total size:      128 GiB
  encryption:      yes
```

### DIFF
--- a/upstairs/src/downstairs.rs
+++ b/upstairs/src/downstairs.rs
@@ -88,6 +88,19 @@ pub(crate) struct Downstairs {
     /// (including automatic flushes).
     next_flush: u64,
 
+    /// Indicates whether we are eligible for replay
+    ///
+    /// We are only eligible for replay if all jobs since the last flush are
+    /// buffered (i.e. none have been retired by a `Barrier` operation).
+    can_replay: bool,
+
+    /// How many `Flush` or `Barrier` operations are pending?
+    ///
+    /// We only want to send a `Barrier` if there isn't already one pending, so
+    /// we track it here (incrementing in `submit_flush` / `submit_barrier` and
+    /// decrementing in `retire_check`).
+    pending_barrier: usize,
+
     /// Ringbuf of recently acked job IDs.
     acked_ids: AllocRingBuffer<JobId>,
 
@@ -291,6 +304,8 @@ impl Downstairs {
             },
             cfg,
             next_flush: 0,
+            can_replay: true,
+            pending_barrier: 0,
             ds_active: ActiveJobs::new(),
             gw_active: HashSet::new(),
             acked_ids: AllocRingBuffer::new(2048),
@@ -523,7 +538,7 @@ impl Downstairs {
 
         // Restart the IO task for that specific client, transitioning to a new
         // state.
-        self.clients[client_id].reinitialize(up_state);
+        self.clients[client_id].reinitialize(up_state, self.can_replay);
 
         for i in ClientId::iter() {
             // Clear per-client delay, because we're starting a new session
@@ -1887,6 +1902,7 @@ impl Downstairs {
             extent_limit: extent_under_repair,
         };
 
+        self.pending_barrier += 1;
         self.enqueue(
             next_id,
             flush,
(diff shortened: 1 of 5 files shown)


mkeeter:
Rebased to stage on top of #1507, because we want to only send `Barrier` operations if there's enough bytes / jobs buffered (which requires keeping track!)

upstairs/src/dummy_downstairs_tests.rs
```diff
@@ -1473,6 +1528,17 @@ async fn test_byte_fault_condition() {
...
+
+        // Once one of the Downstairs has cached more that a certain amount of
+        // bytes, it will automatically send a Barrier operation.
+        if i == barrier_point {
+            let m = harness.ds1().recv().await.unwrap();
+            if !matches!(m, Message::Barrier { .. }) {
+                panic!("expected Barrier, got message {m:?}");
+            }
+            harness.ds2.ack_barrier().await;
+            harness.ds3.ack_barrier().await;
```
leftwo, line 1540: Do we have any way to check that replay is no longer available to this downstairs?  I'm not sure if the test has access at this layer to it.

leftwo:
With this, we can probably close https://github.com/oxidecomputer/crucible/issues/1358 as fixed

upstairs/src/dummy_downstairs_tests.rs
mkeeter: The end of this unit test is checking that live-repair works, so I think we're implicitly testing that we don't do replay.

leftwo: Ah, I was thinking more that the sending of a barrier operation has also flipped the `can_replay` in the `Downstairs` struct.  I'm not sure if we can do that though here (or in a test elsewhere), as `can_replay` might not be exposed like that.

mkeeter: Yeah, it's not easy to check from the outside (when we only have the `Guest` handle).  If you really want it, we could probably add it to the `downstairs_state` helper function.

upstairs/src/downstairs.rs
```diff
@@ -2474,6 +2538,13 @@ impl Downstairs {
                 "downstairs failed, too many outstanding bytes {byte_count}"
             );
             Some(ClientStopReason::TooManyOutstandingBytes)
+        } else if !self.can_replay {
+            // XXX can this actually happen?
+            warn!(
+                self.log,
+                "downstairs became ineligible for replay while offline"
```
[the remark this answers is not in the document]
mkeeter: I don't think it's possible, but the state machines are complex enough that I'm not 100% certain; faulting the downstairs is a correct behavior, so I'd rather do that versus panicking.  I did try running unit tests with a panic here, and it didn't fire, but that could just mean that it's an unlikely corner case.

> It also seems like the statement downstairs became ineligible for replay while offline is not the unusual part, is it? If a downstairs is Offline and we continue to get IO, then it will become ineligible for replay and that's kind of the expected behavior, though maybe not in this method?

The `can_replay` flag is set to `false` when we remove _completed_ jobs that would be needed for a replay ("completed" = done / skipped / error on all three downstairs).  That shouldn't happen with one downstairs Offline, because jobs are left `InProgress` for the offline downstairs (for exactly this reason).

(We can also become ineligible for replay when too many `InProgress` jobs pile up, which is checked right here)

upstairs/src/dummy_downstairs_tests.rs
leftwo: I don't think it's worth adding it to the `downstairs_state` helper function. 
I don't see anywhere else that seems like a good place either.  Nothing in the upstairs tests cover this case.

leftwo (approved):

faithanalog (approved):
We should probably tighten up `IO_CACHED_MAX_BYTES` soon. Right now we can theoretically buffer up to 1GiB of jobs for replay, and that feels like quite a lot to me. We usually won't- we will only get to that point if the guest is doing large amounts of IO, continuously so we never go idle, and the guest never sends a flush. That's unlikely/rare during normal filesystem operations (but will be easier to hit when writing to the raw block device like iodriver does). Because of that, we shouldn't expect more than a few VMs to hit this under normal operation. But in interest in not overcommitting resources, I think that bound should be lower.

jmpesp (approved):

PR #1506, opened 2024-10-15
Remove IOP/BW limits (for now)

mkeeter:
(staged on top of #1502)

Right now, IOP / bandwidth limiting will buffer incoming jobs in the `Guest`, then trickle them out to the `Upstairs`.  The current implementation doesn't effectively implement backpressure (see [RFD 445 § 4.1](https://rfd.shared.oxide.computer/rfd/445#_iop_and_bandwidth_limits_do_not_cause_backpressure)), meaning guests could eat an unexpected amount of RAM by submitting many jobs to an IOP / bandwidth-limited Upstairs.

Fortunately, we never actually turn on IOP / bandwidth limiting in production!

I'm currently working on IO limits to control maximum queue lengths in a backpressure-friendly way.  Once that's complete, it will be easy to add IOP / bandwidth limiting within that same framework.  In the meantime, it would be helpful if the current implementation was removed from the codebase.

(I'm also planning to revive equivalent unit tests once the reimplementation is complete, but want them out of the way for now)


measure_iops/src/main.rs
```diff
@@ -91,15 +91,15 @@ async fn main() -> Result<()> {
...
     };
 
-    let (guest, mut io) = Guest::new(None);
+    let (guest, io) = Guest::new(None);
     let guest = Arc::new(guest);
 
-    if let Some(iop_limit) = opt.iop_limit {
-        io.set_iop_limit(16 * 1024 * 1024, iop_limit);
+    if let Some(_iop_limit) = opt.iop_limit {
+        bail!("IOP limit is not implemented"); // XXX fix this once implemented
```
jmpesp, line 98: should be `CrucibleError::Unsupported` instead

jmpesp (approved):
:wave:

measure_iops/src/main.rs
mkeeter: Done in 25ec12c50f5d, although I'm ambivalent (the function is returning `anyhow::Result` already, and it's a CLI tool rather than an internal library)

PR #1507, opened 2024-10-15
Track both jobs and bytes in each IO state

mkeeter:
- Refactors `ClientIOStateCount` to be generic and implement `std::ops::Index/IndexMut`
- Add a `io_state_byte_count: ClientIOStateCount<u64>` to track number of bytes associated with each IO state
- Rename `io_state_count` to `io_state_job_count`
- Make `DownstairsIO::io_size()` just call into `IOop::job_bytes()`, which does the same thing


upstairs/src/client.rs
```diff
@@ -2252,7 +2279,8 @@ impl DownstairsClient {
     }
 
     pub(crate) fn total_live_work(&self) -> usize {
-        (self.io_state_count.new + self.io_state_count.in_progress) as usize
+        (self.io_state_job_count.new + self.io_state_job_count.in_progress)
```
leftwo, line 2282: I don't know how this snuck through, but the `io_state_job_count.new` has no `IOState` equivalent any longer.  Since your in the neighborhood can you just remove the new field from `ClientIOStateCount`?

upstairs/src/lib.rs
```diff
@@ -1423,36 +1410,30 @@ impl fmt::Display for IOState {
...
-#[derive(Debug, Copy, Clone, Serialize, Deserialize)]
-pub struct ClientIOStateCount {
-    pub new: u32,
-    pub in_progress: u32,
-    pub done: u32,
-    pub skipped: u32,
-    pub error: u32,
+#[derive(Debug, Default, Copy, Clone, Serialize, Deserialize)]
+pub struct ClientIOStateCount<T = u32> {
+    pub new: T,
```
leftwo, line 1415: Yeah, the new here is now vestigial and we can remove it.

leftwo (approved):

upstairs/src/client.rs
mkeeter: Good catch, removed in 663ae0d00acb

This will change the DTrace probe format, but it looks like no one is using `ok.ds_io_count.new`

Issue #1477, opened 2024-09-27
Crutest span test should cover IO spanning sub_volumes


leftwo:
The crutest span test will verify IOs that span two extents work as expected.
This test should be updated to generate IOs that span sub-volumes.

leftwo:
This is now fixed

PR #1508, opened 2024-10-15
Still more updates to support Volume layer activities.

leftwo:
This is a re-do, re-work of what was in: Make crutest know extent info for each sub volume. #1474

Two new structs to describe a Volume:
```
pub struct VolumeInfo {
    pub block_size: u64,
    pub volumes: Vec<SubVolumeInfo>,
}
pub struct SubVolumeInfo {
    pub blocks_per_extent: u64,
    pub extent_count: u32,
}
```

Added a new method to the `Volume` type that will gather extent info from each sub-volume and return a `VolumeInfo` struct that describes the whole volume.

In the `BlockIO` trait, replaced `query_extent_size` with `query_extent_info`.  Objects that support this call will return the `RegionExtentInfo` type filled out.  Things that do not will just return `None`.

Updated crutest's `RegionInfo` struct to have a `VolumeInfo` field.

Updated the fill_sparse and span crutest tests to perform their work on all sub-volumes.

print_region_description looks like this:
```
----------------------------------------------
random read with 4 KiB chunks (1 block)
region info:
  block size:                   4096 bytes
  sub_volume 0 blocks / extent: 100
  sub_volume 0 extent count:    15
  sub_volume 0 extent size:     400 KiB
  sub_volume 1 blocks / extent: 100
  sub_volume 1 extent count:    15
  sub_volume 1 extent size:     400 KiB
  total blocks:                 3000
  total size:                   11.7 MiB
  encryption:                   no
----------------------------------------------
```

Added fill-sparse test to crutest cli

Coming in a follow up PR I would like to rename `RegionInfo` in crutest to be `DiskInfo`.  This would just be name changes, no functional changes.  I decided to break that out to reduce review overhead.

### DIFF
--- a/crutest/src/cli.rs
+++ b/crutest/src/cli.rs
@@ -283,7 +283,7 @@ async fn rand_write(
      * IO size.
      */
     let size = 1;
-    let block_max = ri.total_blocks - size + 1;
+    let block_max = ri.volume_info.total_blocks() - size + 1;
     let block_index = rng.gen_range(0..block_max);
 
     cli_write(volume, ri, block_index, size).await
@@ -311,8 +311,9 @@ async fn cli_write(
      * If so, then don't update any write counts and just make
      * the correct size buffer with all zeros.
      */
-    let data = if block_index + size > ri.total_blocks {
-        println!("Skip write log for invalid size {}", ri.total_blocks);
+    let total_blocks = ri.volume_info.total_blocks();
+    let data = if block_index + size > total_blocks {
+        println!("Skip write log for invalid size {}", total_blocks);
         let mut out = BytesMut::new();
         out.resize(size * ri.volume_info.block_size as usize, 0);
         out
@@ -581,8 +582,8 @@ async fn cmd_to_msg(
         Some(CliMessage::MyUuid(uuid)) => {
             println!("uuid: {}", uuid);
         }
-        Some(CliMessage::Info(vi, bs, bl)) => {
-            println!("Got info: {:?} {} {}", vi, bs, bl);
+        Some(CliMessage::Info(vi)) => {
+            println!("Got info: {:?}", vi);
         }
         Some(CliMessage::DoneOk) => {
             println!("Ok");
@@ -759,7 +760,7 @@ async fn process_cli_command(
     volume: &Volume,
     fw: &mut FramedWrite<tokio::net::tcp::OwnedWriteHalf, CliEncoder>,
     cmd: protocol::CliMessage,
-    ri: &mut RegionInfo,
+    ri_option: &mut Option<RegionInfo>,
     wc_filled: &mut bool,
     verify_input: Option<PathBuf>,
     verify_output: Option<PathBuf>,
@@ -793,14 +794,21 @@ async fn process_cli_command(
             Err(e) => fw.send(CliMessage::Error(e)).await,
         },
         CliMessage::Commit => {
-            if ri.write_log.is_empty() {
+            if let Some(ri) = ri_option {
+                if ri.write_log.is_empty() {
+                    fw.send(CliMessage::Error(CrucibleError::GenericError(
+                        "Info not initialized".to_string(),
+                    )))
+                    .await
+                } else {
+                    ri.write_log.commit();
+                    fw.send(CliMessage::DoneOk).await
+                }
+            } else {
                 fw.send(CliMessage::Error(CrucibleError::GenericError(
                     "Info not initialized".to_string(),
                 )))
                 .await
-            } else {
-                ri.write_log.commit();
-                fw.send(CliMessage::DoneOk).await
             }
         }
         CliMessage::Expected(offset) => {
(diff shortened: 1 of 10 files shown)


crutest/src/cli.rs
```diff
@@ -1071,9 +1100,12 @@ pub async fn start_cli_server(
      * If write_log len is zero, then the RegionInfo has
      * not been filled.
      */
+    let volume_info = VolumeInfo {
```
mkeeter, line 1103: Should we be pulling this info from the `volume` argument?

crutest/src/cli.rs
```diff
@@ -1071,9 +1100,12 @@ pub async fn start_cli_server(
...
+        block_size: 512,
+        volumes: Vec::new(),
+    };
     let mut ri: RegionInfo = RegionInfo {
-        block_size: 0,
-        extent_size: Block::new_512(0),
+        volume_info,
```
mkeeter, line 1108: Should we remove `total_size` and `total_blocks` here, because they can both be calculated from `volume_info`?

crutest/src/main.rs
```diff
@@ -2045,24 +2047,32 @@ async fn fill_sparse_workload(
...
+                ri.volume_info.block_size,
+            );
 
-        println!("[{extent}/{extents}] Write to block {}", block_index);
-        volume.write(offset, data).await?;
+            println!(
+                "[{index}][{extent}/{extents}] Write to block {}",
+                block_index
+            );
+            volume.write(offset, data).await?;
```
mkeeter, line 2074: This doesn't look right to me, because `offset` here is a relative offset within the subvolume (but ignores the offset of that subvolume within the whole volume).

I think we need to accumulate a global offset (cumulative sum of `sv.extent_count * sv.blocks_per_extent`) and add it to `offset` here.

crutest/src/main.rs
```diff
@@ -3089,10 +3111,13 @@ async fn perf_workload(
...
     let mut read_buffers: Vec<Buffer> = (0..io_depth)
-        .map(|_| Buffer::new(blocks_per_io, ri.block_size as usize))
+        .map(|_| Buffer::new(blocks_per_io, ri.volume_info.block_size as usize))
         .collect();
 
-    let es = ri.extent_size.value;
+    let mut es = 0;
+    for sv in ri.volume_info.volumes.iter() {
+        es += sv.blocks_per_extent;
+    }
```
mkeeter, line 3120: This doesn't make sense to me; the sum of blocks-per-extent across all subvolumes isn't a meaningful value.

In a subvolume-flavored world, there is no single extent size (`es`), so we should just remove it.  Extent count `ec` remains meaningful, but should be a function on `VolumeInfo`, e.g.
```rust
impl VolumeInfo {
    /// Returns the total number of extent files in this volume
    fn extent_count(&self) -> usize {
        self.subvolumes.iter().map(|s| s.extent_count).sum().unwrap_or(0)
    }
}
```

crutest/src/cli.rs
mkeeter: Upon further reading, it would minimize the diff to also add a `block_size()` helper function, since lots of the diff in this PR is `s/block_size/volume_info.block_size`.

```rust
fn block_size(&self) -> u64 {
    self.volume_info.block_size
}
```

crutest/src/protocol.rs
```diff
@@ -34,9 +34,11 @@ pub enum CliMessage {
     Export,
     // Run the fill test.
     Fill(bool),
+    // Run the sparse fill test.
+    FillSparse,
     Flush,
     Generic(usize, bool),
-    Info(u64, u64, u64),
+    Info(VolumeInfo, u64, usize),
```
mkeeter, line 41: We should just pass `VolumeInfo`, because total size and total blocks can be calculated from it.

upstairs/src/guest.rs
```diff
@@ -342,9 +343,13 @@ impl Guest {
         rx.wait().await
     }
 
-    pub async fn query_extent_size(&self) -> Result<Block, CrucibleError> {
-        self.send_and_wait(|done| BlockOp::QueryExtentSize { done })
-            .await
+    pub async fn query_extent_info(
```
mkeeter, line 346: Why is this implemented twice (once in `impl Guest` and once in `impl BlockIO for Guest`)?

upstairs/src/upstairs.rs
```diff
@@ -1049,11 +1050,16 @@ impl Upstairs {
...
                 // Yes, test only
                 match self.ddef.get_def() {
                     Some(rd) => {
-                        done.send_ok(rd.extent_size());
+                        let ei = RegionExtentInfo {
+                            block_size: rd.block_size(),
+                            blocks_per_extent: rd.extent_size().value,
+                            extent_count: rd.extent_count(),
+                        };
+                        done.send_ok(ei);
```
mkeeter, line 1062: Edit the warning messages below to say "extent info not available" (instead of "extent size")?

upstairs/src/volume.rs
```diff
@@ -611,6 +611,36 @@ impl Volume {
...
+    pub async fn volume_extent_info(
+        &self,
+    ) -> Result<VolumeInfo, CrucibleError> {
+        // A volume has multiple levels of extent info, not just one.
+        let mut volumes = Vec::new();
+
+        for sub_volume in &self.sub_volumes {
+            match sub_volume.query_extent_info().await? {
+                Some(ei) => {
+                    assert_eq!(self.block_size, ei.block_size);
```
mkeeter, line 624: Add a comment here as to why this can never happen (is it checked on Volume construction?)

upstairs/src/volume.rs
```diff
@@ -611,6 +611,36 @@ impl Volume {
...
+                    let svi = SubVolumeInfo {
+                        blocks_per_extent: ei.blocks_per_extent,
+                        extent_count: ei.extent_count,
+                    };
+                    volumes.push(svi);
+                }
+                None => {
+                    continue;
```
mkeeter, line 632: I would lean towards "mixing extent and non-extent flavored subvolumes is an error", instead of ignoring it.

(In other words, return an error if we get a `None` for a subvolume and `!volumes.is_empty()`)

crutest/src/cli.rs
leftwo: It was that way because I could not figure out how to make in an `Option` and pass it back and forth between `process_cli_command` and `start_cli_server`.   Let me see if I can figure it out now

crutest/src/cli.rs
leftwo: I liked the idea of computing it once, then just passing it around, but don't really care if we compute it every time we need it.  I don't think its anywhere inside the performance path, and if I find a place I'll handle it there.

crutest/src/main.rs
leftwo: Yeah, for this I'm going to remove the whole crutest `perf` test completely, so this was just to make it compile.
It's all going away.

crutest/src/cli.rs
leftwo: I added an <Option> and I pass that around instead.

crutest/src/main.rs
leftwo: Yes, this was wrong and not doing what I wanted.
I've updated it with an additional global offset and I can see from the test output that it is touching all extents

upstairs/src/upstairs.rs
leftwo: Done

upstairs/src/guest.rs
leftwo: I don't know why. Maybe an artifact of existing before BlockIO trait?
`activate_with_gen()` and `query_work_queue()` are also doubled up.  I'll remove the duplicates as
I don't see any differences that we need.

leftwo:
Should also fix: 
https://github.com/oxidecomputer/crucible/issues/1477

Partial fix for this https://github.com/oxidecomputer/crucible/issues/1455, though we may want to re-think how jobs are returned and displayed at the volume level.

We can probably call this one fixed, as we are full volume now: https://github.com/oxidecomputer/crucible/issues/1454

crutest/src/cli.rs
leftwo: I've updated the struct and added new methods on the type.

crutest/src/protocol.rs
leftwo: Done.

upstairs/src/volume.rs
leftwo: I've updated this to return an error if the subvolumes don't match.

crutest/src/cli.rs
```diff
@@ -799,76 +817,130 @@ async fn process_cli_command(
...
-                    Err(e) => {
-                        println!("Failed writing to {vo:?} with {e}");
-                        fw.send(CliMessage::Error(CrucibleError::GenericError(
-                            "Failed writing to file".to_string(),
-                        )))
-                        .await
+            if let Some(ri) = ri_option {
+                if ri.write_log.is_empty() {
+                    fw.send(CliMessage::Error(CrucibleError::GenericError(
+                        "Info not initialized".to_string(),
```
mkeeter, line 843: Is this the correct error message?  It's confusing to me, because this message is used for both `write_log.is_empty()` and `ri_option.is_none()` cases.  Above, we use `"Internal write count buffer empty"` for the former.

(same question applies below in a bunch of places)

crutest/src/main.rs
```diff
@@ -3993,41 +4048,100 @@ async fn demo_workload(
...
+    let last_sub_volume = ri.volume_info.volumes.len() - 1;
+    let block_size = ri.volume_info.block_size;
+    for (index, sv) in ri.volume_info.volumes.iter().enumerate() {
+        // Pick the last block in the first extent
+        let extent_size = sv.blocks_per_extent as usize;
+        let block_index = extent_block_start + extent_size - 1;
+
+        // Update the counter for the blocks we are about to write.
+        ri.write_log.update_wc(block_index);
+        ri.write_log.update_wc(block_index + 1);
```
mkeeter, line 4065: Why does this not fail on the last subvolume?  I'd expect it to write out of bounds.

mkeeter (approved):

crutest/src/cli.rs
leftwo: It's not the correct message any longer, but even better, it's also not possible.
If we are at the place where we check for a write log, then it must exist, as we now have an `<Option<RegionInfo>>` instead of an "empty" one.  So, if the RegionInfo exists, then we have a write log and we can avoid this whole check.  I'll make that change in all the places where we check for it.

PR #1509, opened 2024-10-15
Fix new `rustc` and `clippy` warnings

mkeeter:
`cargo check --all-targets` and `cargo clippy --all-targets` are now both clean, after the bump to `rustc 1.81.0`

In addition, we can use the new `#[expect(unused)]` for a `DropGuard` that is held purely for its on-drop side effects.


leftwo (approved):
I'll delete my branch with the same changes

PR #1515, opened 2024-10-23
Remove delay-based backpressure in favor of explicit queue limits

mkeeter:
Right now, we have a backpressure delay that's a function of `(job count, bytes in flight)`.

The actual queue length is set implicitly by the shape of that function (e.g. its gain) and the actual downstairs IO time.  If a downstairs slows down due to load, then the queue length has to grow to compensate!

(for more on this, come to my OxCon talk this afternoon!)

Having an implicitly-defined queue length is tricky, because it's not well controlled.  Queue length also affects latency in subtle ways, so directly controlling the queue length would be helpful.

This PR removes the delay-based backpressure implementation in favor of a simple pair of semaphores: there are a certain number of job and byte permits available, and the `Guest` has to acquire them before sending a job to the `Upstairs`.

In other words, writes will be accepted as fast as possible _until_ we run out of permits; we will then shift to a one-in, one-out operation mode where jobs have to be completed by the downstairs before a new job is submitted.  The maximum queue length (either in jobs or bytes) is well known and set by global constants.

Architecturally, we replace the `BackpressureGuard` with a new `IOLimitGuard`, which is claimed by the `Guest` and stored in relevant `BlockOp` variants (then moved into the `DownstairsIO`).  It still uses RAII to automatically release permits when dropped, and we still manually drop it early (once a job is complete on a particular downstairs).

The `IOLimitGuard` will also be used (eventually) for _configured_ IOP and bandwidth limiting, which was removed in #1506; we can use a strategy similar to [this example from the Tokio docs](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#rate-limiting-using-a-token-bucket).


upstairs/src/lib.rs
```diff
@@ -86,17 +83,20 @@ mod downstairs;
...
+/// If we exceed this value, the upstairs will give up and mark the offline
+/// downstairs as faulted.
+const IO_OUTSTANDING_MAX_BYTES: u64 = 50 * 1024 * 1024; // 50 MiB
 
 /// Max number of outstanding IOs between the upstairs and an offline downstairs
 ///
 /// If we exceed this value, the upstairs will give up and mark that offline
 /// downstairs as faulted.
-pub const IO_OUTSTANDING_MAX_JOBS: usize = 10000;
+pub const IO_OUTSTANDING_MAX_JOBS: usize = 1000;
```
leftwo, line 99: I'm curious to see how 1000 jobs does with replay tests.
We might need to change a few things around in the tests to be sure they stick under this limit, otherwise we are testing live repair.

leftwo (commented):
Some other random thoughts

Maybe not now, but if setting tokens was part of the BlockIO trait, then we would be setup to pass them down and controll them from the Volume layer? 

If an IO blocks in read/write/flush because there are not enough tokens, the rest of the upstairs still makes progress right?  Like, we don't need the same task that is stuck in a read to pull something else off the queue elsewhere to free up resources or anything like that?  Or, is that what the _try stuff is all about?

upstairs/src/downstairs.rs
```diff
@@ -2107,27 +2106,26 @@ impl Downstairs {
...
-                        let g = client.backpressure_counters.increment(&io);
-                        bp_guard.insert(cid, g);
-                    }
-                    if matches!(r, EnqueueResult::Send) {
-                        self.send(ds_id, io.clone(), cid);
-                    }
+                EnqueueResult::Send => self.send(ds_id, io.clone(), cid),
+                EnqueueResult::Hold => (),
+                EnqueueResult::Skip => {
+                    let _ = io_limits.take(&cid);
```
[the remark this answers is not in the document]
mkeeter: That's correct; I added a comment in f9abea59

mkeeter:
> Maybe not now, but if setting tokens was part of the BlockIO trait, then we would be setup to pass them down and controll them from the Volume layer?

That's something that we'll have to figure out – previously, IO limits were attached to the `GuestIOHandle`, but in a world of multi-`Guest` volumes, that may need to change.  We'll have to see whether we can attach limits to the Volume layer, or whether we have to add new APIs to `BlockIO`.

(`Guest` is also increasingly misnamed, but that's neither here nor there)

> If an IO blocks in read/write/flush because there are not enough tokens, the rest of the upstairs still makes progress right? Like, we don't need the same task that is stuck in a read to pull something else off the queue elsewhere to free up resources or anything like that? Or, is that what the _try stuff is all about?

Good question!  The intention is for the `Guest` task to claim IO permits (blocking until they're available), then pass an `IOLimitGuard` into the `Upstairs` task.  It's fine for the guest task to block on obtaining permits, as long as the upstairs task continues running, because it's the thing which actually completes IO.  I tweaked the API slightly in f9abea59 so that `async fn claim(..0` is _only_ on the `struct IOLimitView` (available to the `Guest`), and not present on the `struct IOLimits` (available in the `Upstairs`).  This should make it hard to deadlock oneself!

upstairs/src/guest.rs
```diff
@@ -382,11 +356,20 @@ impl BlockIO for Guest {
             assert_eq!(chunk.len() as u64 % bs, 0);
 
             let offset_change = chunk.len() as u64 / bs;
+            let io_guard =
+                self.io_limits.claim(chunk.len() as u32).await.map_err(
```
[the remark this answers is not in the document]
mkeeter: Yup!  The only possible error here occurs if the semaphore is closed; we never explicitly close the semaphore, but it may happen on drop?

upstairs/src/downstairs.rs
```diff
@@ -2666,10 +2664,10 @@ impl Downstairs {
                 // **not** when they are retired.  We'll do a sanity check here
                 // and print a warning if that's not the case.
                 for c in ClientId::iter() {
-                    if job.backpressure_guard.contains(&c) {
+                    if job.io_limits.contains(&c) {
                         warn!(
                             self.log,
-                            "job {ds_id} had pending backpressure bytes \
```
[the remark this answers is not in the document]
mkeeter: I did a few fixes in 267a5b71af26.  We still have per-client backpressure to keep queues roughly in sync, so the word isn't totally gone (and even the semaphore is a form of backpressure writ large).

upstairs/src/lib.rs
mkeeter: Good catch, this was left in by accident.

I didn't see any performance differences, though, so I'm tempted to stick with the smaller numbers...

leftwo: I'm thinking more about when we can still replay vs. LiveRepair.  If this gives us a little more time before we give up on a downstairs, that should result in less work when it comes back.

mkeeter: I think we're in the clear: in the [most recent buildomat logs for `test_replay`](https://buildomat.eng.oxide.computer/wg/0/artefact/01JBCZMVYBQTGG2HBRSYCVXT0Q/FG5ETDGnFlNC9JUCKj8scbyHSSbqVZ6JUvlECF0ZFdbOLvaU/01JBCZNGW1ZQ9EP560WE4FBSXJ/01JBD13WSNDTA8MN8SB302NZ0X/test_replay.log), there are indications that it's going through replay (and not live-repair)

```json
{"msg":"[2] client re-new 302 jobs since flush 3114","v":0,"name":"crucible","level":30,"time":"2024-10-29T20:50:36.963940864Z","hostname":"w-01JBD0H1N4K3YEM72S3XWCQD40","pid":1198,"":"downstairs","":"downstairs","session_id":"18a6fd2d-1f19-4552-b9d6-5ab6b53fbb51"}
{"msg":"[2] Replayed 302 jobs since flush: 3114","v":0,"name":"crucible","level":30,"time":"2024-10-29T20:50:36.964171465Z","hostname":"w-01JBD0H1N4K3YEM72S3XWCQD40","pid":1198,"":"downstairs","":"downstairs","session_id":"18a6fd2d-1f19-4552-b9d6-5ab6b53fbb51"}
```

tools/dtrace/single_up_info.d
```diff
@@ -62,7 +62,6 @@ crucible_upstairs*:::up-status
      * Job ID delta and backpressure
      */
     json(copyinstr(arg1), "ok.next_job_id"),
-    json(copyinstr(arg1), "ok.up_backpressure"),
```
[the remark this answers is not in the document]
mkeeter: It's pretty easy to add in Crucible, but we probably don't have room for all 6x values (bytes and jobs available for each client) in the DTrace UIs...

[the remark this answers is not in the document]
leftwo: Oh, actually displaying them will be an exercise left to the DTracer.  I just wanted to know they exist.

I'll spend way way to much time crafting another script to show them :)

leftwo (approved):
If you don't want to add the DTrace collection points now, expect a PR from me soon that does have them :)

faithanalog (approved):
the only part of this i dont think i caught is where the disable_backpressure flag actually comes into effect. the rest of it I understand (and the way these async semaphores work is quite neat to learn about)

mkeeter:
@faithanalog 
> the only part of this i dont think i caught is where the disable_backpressure flag actually comes into effect. the rest of it I understand (and the way these async semaphores work is quite neat to learn about)

Good question – the `disable_backpressure` flag in `GuestIOHandle` doesn't do anything itself anymore, but it's checked in `up_main` to disable per-client backpressure (which still exists!)

upstairs/src/io_limits.rs
```diff
@@ -0,0 +1,158 @@
...
+    ///
+    /// Returns `Err((ClientId, e))` if any of the claims fail
+    pub fn try_claim(
+        &self,
+        bytes: u32,
+    ) -> Result<IOLimitGuard, (ClientId, TryAcquireError)> {
+        let mut out = ClientData::from_fn(|_| None);
+        for i in ClientId::iter() {
+            out[i] = Some(self.0[i].try_claim(bytes).map_err(|e| (i, e))?);
+        }
```
jmpesp, line 86: If any claim succeeds here, they won't be released if a subsequent `try_claim` fails.

upstairs/src/io_limits.rs
```diff
@@ -0,0 +1,158 @@
...
+    /// should not be called from the same task which is processing jobs (since
+    /// that could create a deadlock).
+    pub async fn claim(
+        &self,
+        bytes: u32,
+    ) -> Result<IOLimitGuard, AcquireError> {
+        let mut out = ClientData::from_fn(|_| None);
+        let lim = &self.0;
+        for i in ClientId::iter() {
+            out[i] = Some(lim.0[i].claim(bytes).await?);
```
jmpesp, line 113: Same here.

upstairs/src/io_limits.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    /// Tries to claim a certain number of bytes (and one job)
+    pub fn try_claim(
+        &self,
+        bytes: u32,
+    ) -> Result<ClientIOLimitGuard, TryAcquireError> {
+        let io_bytes = self.io_bytes.clone().try_acquire_many_owned(bytes)?;
+        let jobs = self.jobs.clone().try_acquire_owned()?;
+        Ok(ClientIOLimitGuard { io_bytes, jobs })
+    }
```
jmpesp, line 49: Coming back to these two, same suggestion: if there's an error acquiring permits for the jobs semaphore, the acquired permits for the io_bytes semaphore should be released

upstairs/src/io_limits.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+/// Handle owning some amount of per-client IO
+///
+/// The IO permits are released when this handle is dropped
+#[derive(Debug)]
+pub struct ClientIOLimitGuard {
+    #[expect(unused)]
```
jmpesp, line 126: TIL!

upstairs/src/lib.rs
```diff
@@ -86,17 +83,20 @@ mod downstairs;
...
+use io_limits::IOLimitGuard;
+
 /// Max number of write bytes between the upstairs and an offline downstairs
 ///
-/// If we exceed this value, the upstairs will give
-/// up and mark the offline downstairs as faulted.
-const IO_OUTSTANDING_MAX_BYTES: u64 = 1024 * 1024 * 1024; // 1 GiB
+/// If we exceed this value, the upstairs will give up and mark the offline
+/// downstairs as faulted.
+const IO_OUTSTANDING_MAX_BYTES: u64 = 50 * 1024 * 1024; // 50 MiB
```
jmpesp, line 93: Where did this number come from?

jmpesp (approved):
Ship it, pending some comments about handling chained semaphore acquires!

upstairs/src/lib.rs
mkeeter: Mostly vibes!  This was previously the point at which backpressure delays started.

upstairs/src/io_limits.rs
jmpesp: You most kindly pointed out in the DMs that dropping `out` will drop any permits claimed, so ignore these!

upstairs/src/lib.rs
jmpesp: To be clear, I don't think this is a bad number! Just wondering, esp when you were looking at the graphs showing the buffering cliff that real devices had.

mkeeter: One data point: with 10x `crutest` instances doing 4K random writes on a single Gimlet (so 30x Downstairs), we end up buffering about 8 seconds of writes:
Screenshot 2024-11-12 at 10 31 47 AM

(there's 60 seconds of IO, then you can watch the number of active jobs drain)

This is a significant improvement from the ~50 seconds that we see on `main`:
Screenshot 2024-11-12 at 10 31 57 AM

It may still be too much, but it's easy to tune these values later.

Issue #1513, opened 2024-10-18
Pantry panic in a4x2: Job completed while not InProgress: Done


jgallagher:
I was trying to import a disk in a4x2 via this Oxide CLI invocation:

```
oxide --timeout 120 disk import \
    --path debian-12-genericcloud-amd64.raw \
    --disk bookworm-boot \
    --project prototypes \
    --description "Debian bookworm disk" \
    --snapshot bookworm-snapshot \
    --image bookworm \
    --image-description "Debian bookworm image" \
    --image-os Debian \
    --image-version 12
```

After not quite 2 minutes it failed with

```
sending chunk to thread failed: channel closed
```

which seems similar to the test flake oxidecomputer/omicron#6771, but in this case it looks like the pantry panicked:

```
thread 'tokio-runtime-worker' panicked at upstairs/src/client.rs:1280:13:
[1] Job 1252 completed while not InProgress: Done DownstairsIO { ... trim a couple megabytes ... }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[ Oct 18 18:21:28 Stopping because all processes in service exited. ]
```

The full pantry log: [pantry.log](https://github.com/user-attachments/files/17439584/pantry.log)

mkeeter:
This looks like a race condition in replay!

When a downstairs is `Offline`, we still send jobs to it:

https://github.com/oxidecomputer/crucible/blob/bfcc86f79dfa14b619735d4b81a005d262ad1e59/upstairs/src/client.rs#L968-L974

We actually send IO jobs to the client task, which sends them out!  However, they're ignored by the Downstairs while it's in negotiation.

Now, imagine sending a write _exactly when_ the downstairs finishes negotiation:

- Downstairs sends the final negotiation message (`Message::ExtentVersions`) and decides it's done with negotiation!
- A write is sent to the ostensibly-offline Downstairs
- The Downstairs thinks it's online, so it executes the write and sends a reply
- The Upstairs receives the final negotiation message
  - It replays all IO jobs, including the one it just sent to the Downstairs!
  - The Downstairs replies to that one as well
- The upstairs receives both replies
  - It panics because the job is completed twice

--------------------------------------------------------------------------------

Evidence for this theory is in the `JobId` numbers:

```
[...]

PR #1519, opened 2024-10-24
Add test and fix for replay race condition

mkeeter:
Fix for #1513 

When submitting IO to a Dowstairs in `DsState::Offline`, we need to mark that IO as `IOState::InProgress` (and not `IOState::Skipped`), because it's eligible for replay.

However, we don't want to actually send that IO down the wire, because there's a race condition if the Downstairs has just finish negotiation and would actually reply to it (which would be bad).

This PR distinguishes between the three cases in `DownstairsClient::enqueue` with a new `enum EnqueueResult`.  Bikeshedding is welcome on the variant names.

It's hard to unit-test the race condition specifically, so I'm testing the more general rule that an Offline downstairs shouldn't receive IO messages.

### DIFF
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -905,7 +905,8 @@ impl DownstairsClient {
 
     /// Checks whether the given job should be sent
     ///
-    /// Returns `true` if it should be sent and `false` otherwise
+    /// Returns an [`EnqueueResult`] indicating how the caller should handle the
+    /// packet.
     ///
     /// If the job should be skipped, then it is added to `self.skipped_jobs`.
     /// `self.io_state_job_count` is updated with the incoming job state.
@@ -940,8 +941,8 @@ impl DownstairsClient {
 
     /// Checks whether the given job should be sent or skipped
     ///
-    /// Returns `true` if the job should be sent and `false` if it should be
-    /// skipped.
+    /// Returns an [`EnqueueResult`] indicating how the caller should handle the
+    /// packet.
     #[must_use]
     fn should_send(
         &self,
@@ -977,7 +978,7 @@ impl DownstairsClient {
             // cleared out by a subsequent flush (so we'll be able to bring that
             // client back into compliance by replaying jobs).
             DsState::Active => EnqueueResult::Send,
-            DsState::Offline => EnqueueResult::Drop,
+            DsState::Offline => EnqueueResult::Store,
 
             DsState::New
             | DsState::BadVersion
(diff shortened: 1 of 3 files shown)


upstairs/src/client.rs
```diff
@@ -946,13 +947,13 @@ impl DownstairsClient {
         &self,
```
leftwo, line 947: Update comment on 943 also

upstairs/src/client.rs
```diff
@@ -2334,6 +2340,31 @@ enum NegotiationState {
...
+pub(crate) enum EnqueueResult {
+    /// The given job should be marked as in progress and sent
+    Send,
+
+    /// The given job should be marked as in progress, but not sent
+    ///
+    /// This is used when the Downstairs is Offline; we want to mark the job as
+    /// in-progress so that it's eligible for replay, but the job should not
+    /// actually go out on the wire.
+    Drop,
```
leftwo, line 2353: What about the name `Hold` instead of `Drop`?

mkeeter: Much better, changed!

upstairs/src/client.rs
mkeeter: Done!

upstairs/src/client.rs
```diff
@@ -911,7 +915,7 @@ impl DownstairsClient {
         ds_id: JobId,
```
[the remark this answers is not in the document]
mkeeter: Done!

upstairs/src/dummy_downstairs_tests.rs
```diff
@@ -2646,3 +2646,57 @@ async fn test_write_replay() {
...
+    // Check that a write doesn't make it to DS1
+    let write_handle = harness.spawn(|guest| async move {
+        let mut data = BytesMut::new();
+        data.resize(512, 1u8);
+        guest.write(BlockIndex(0), data).await.unwrap();
+    });
+    harness.ds2.ack_write().await;
+    harness.ds3.ack_write().await;
+    assert_eq!(harness.ds1().try_recv(), Err(TryRecvError::Empty));
+    write_handle.await.unwrap();
```
jmpesp, line 2701: Maybe extend this to bring ds1 online and expect the write is send?

jmpesp (approved):
:rocket: with one test suggestion

upstairs/src/dummy_downstairs_tests.rs
mkeeter: Good idea, done

faithanalog (approved):

leftwo (approved):
I ran some 600 replays using this branch and no problems.

PR #1520, opened 2024-10-24
Switch to using `Duration` for times

mkeeter:
In Crucible, we often see things like
```rust
const TIMEOUT_SECS: f32 = 10.0;
```

`Duration::from_secs` is now a `const fn`, so we can store actual `Duration` values for timeout intervals (instead of `f32`).

This PR also replaces the pattern of `sleep_until(deadline_secs(..))` with a plain `sleep`, e.g.
```diff
-tokio::time::sleep_until(deadline_secs(100.0));
+tokio::time::sleep(Duration::from_secs(100));
```

Finally, it makes time-related names more consistent: `*_interval` for the interval between periodic actions, and `*_deadline` for a time at which we perform an action.

This should be a pure cleanup PR, without any logic changes.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -2057,6 +2057,7 @@ async fn recv_task<RT>(
 ) where
     RT: tokio::io::AsyncRead + std::marker::Unpin + std::marker::Send,
 {
+    // Disconnect from the upstairs after 45 sec, logging a warning every 15s
     const TIMEOUT: VerboseTimeout = VerboseTimeout {
         tick: Duration::from_secs(15),
         count: 3,
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -12,8 +12,8 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use crucible_common::{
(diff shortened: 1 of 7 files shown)


faithanalog (approved):

downstairs/src/lib.rs
```diff
@@ -2057,11 +2057,10 @@ async fn recv_task<RT>(
 ) where
     RT: tokio::io::AsyncRead + std::marker::Unpin + std::marker::Send,
 {
-    // How long we wait before logging a message that we have not heard from
-    // the upstairs.
-    const TIMEOUT_SECS: f32 = 15.0;
-    // How many timeouts will tolerate before we disconnect from the upstairs.
-    const TIMEOUT_LIMIT: usize = 3;
+    const TIMEOUT: VerboseTimeout = VerboseTimeout {
```
leftwo, line 2060: I think it's worth keeping the comments here for what these values represent.

Though (reading further down), maybe better to put at `CLIENT_TIMEOUT`

leftwo (approved):

downstairs/src/lib.rs
mkeeter: Added comments in 6cc9128

PR #1524, opened 2024-10-30
Add unit test for VersionMismatch

mkeeter:
While looking at negotiation, I noticed that `VersionMismatch` is load-bearing: the Downstairs will send it if it receives a `HereIAm` with an incorrect version.  As such, we should be unit-testing that its representation doesn't change (same as `HereIAm` and `YesItsMe`).  I also made the `YesItsMe` test more exhaustive, testing the full encoding of both IPv4 and IPv6 addresses.

### DIFF
--- a/protocol/src/lib.rs
+++ b/protocol/src/lib.rs
@@ -1328,10 +1328,47 @@ mod tests {
         };
         let encoded = bincode::serialize(&m).unwrap();
         assert_eq!(
-            &encoded[0..8],
+            encoded,
             [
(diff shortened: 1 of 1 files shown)


leftwo (approved):
I think this is good, and since the `VersionMismatch` also includes what version the downstairs is expecting, that might give us a chance to handle it in the upstairs if we do ever want to change, though I can't quite come up with a scenario where it makes sense.

faithanalog (approved):

PR #1549, opened 2024-11-01
Consolidate `DownstairsClient::reinitialize`

mkeeter:
Previously, `Downstairs::reinitialize` would do the following:

- Call `DownstairsClient::on_missing`, which updates the client state
- Skip jobs (if necessary)
- Call `DownstairsClient::reinitialize`, which also updates client state

These used to be separate tasks, but they're now in the same place, so it's silly to have them both.

This PR consolidates all of the logic into `DownstairsClient::reinitialize`.

### DIFF
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -571,29 +571,35 @@ impl DownstairsClient {
 
         let current = &self.state;
         let new_state = match current {
-            DsState::Active | DsState::Offline => DsState::Offline,
+            DsState::Active => Some(DsState::Offline),
+            DsState::LiveRepair | DsState::LiveRepairReady => {
+                Some(DsState::Faulted)
+            }
 
-            DsState::Faulted
-            | DsState::LiveRepair
-            | DsState::LiveRepairReady => DsState::Faulted,
-
-            DsState::New
-            | DsState::Deactivated
+            DsState::Deactivated
             | DsState::Reconcile
             | DsState::Disconnected
             | DsState::WaitQuorum
             | DsState::WaitActive
-            | DsState::Disabled => DsState::New,
+            | DsState::Disabled => Some(DsState::New),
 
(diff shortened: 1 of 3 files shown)


upstairs/src/downstairs.rs
```diff
@@ -523,27 +515,19 @@ impl Downstairs {
         // Specifically, we want to skip jobs if the only path back online for
         // that client goes through live-repair; if that client can come back
         // through replay, then the jobs must remain live.
-        let new_state = self.clients[client_id].state();
-        if matches!(prev_state, DsState::LiveRepair | DsState::Active)
```
mkeeter, line 527: It was impossible to go from `Active` → `Faulted`; in the previous implementation of `on_missing`, `Active` always became `Offline`.  Therefore, this branch only fired if the incoming state is `LiveRepair`.

upstairs/src/client.rs
```diff
@@ -1858,7 +1849,6 @@ impl DownstairsClient {
                             String::new()
                         },
                     );
-                    self.checked_state_transition(up_state, DsState::New);
```
mkeeter, line 1861: This is a drive-by change: the calls to `restart_connection` below always update the state, so doing it here is superfluous.

upstairs/src/upstairs.rs
```diff
@@ -2008,17 +2008,8 @@ impl Upstairs {
         self.downstairs
             .notify_nexus_of_client_task_stopped(client_id, reason);
 
-        // If the upstairs is already active (or trying to go active), then the
-        // downstairs should automatically call PromoteToActive when it reaches
-        // the relevant state.
-        let auto_promote = match self.state {
```
mkeeter, line 2014: `auto_promote` is a deterministic function of `self.state`, so we don't need to compute it here (because we're passing `&self.state` into the function below).

jmpesp (approved):

PR #1551, opened 2024-11-05
Target replacement support for VCRs with multiple sub_volumes

leftwo:
Updated the volume layer code to support replacement of a target in a VCR that has multiple sub_volumes in it.  
Before this we only supported a single sub_volume.

Most of the changes here are for tests.  I've added a loop around the VCR replacement tests to create multiple sub volumes and iterate through the test for different VCR configurations.

In addition, it is now required that the generation number change for all sub-volumes when any target is changed.

Some tests that were duplicates have been removed.

### DIFF
--- a/integration_tests/src/lib.rs
+++ b/integration_tests/src/lib.rs
@@ -5695,13 +5695,13 @@ mod test {
             blocks_per_extent: sv_tds.blocks_per_extent(),
             extent_count: sv_tds.extent_count(),
             opts: sv_opts.clone(),
-            gen: 1,
+            gen: 2,
         }];
 
         let new_vol = VolumeConstructionRequest::Volume {
             id: sv_volume_id,
             block_size: BLOCK_SIZE as u64,
-            sub_volumes: new_sub_vol.clone(),
+            sub_volumes: new_sub_vol,
             read_only_parent: Some(rop),
         };
 
@@ -5714,6 +5714,14 @@ mod test {
         // Make one new downstairs
         let new_downstairs = tds.new_downstairs().await.unwrap();
         info!(log, "A New downstairs: {:?}", new_downstairs.address());
+
+        let more_sub_vol = vec![VolumeConstructionRequest::Region {
+            block_size: BLOCK_SIZE as u64,
+            blocks_per_extent: sv_tds.blocks_per_extent(),
+            extent_count: sv_tds.extent_count(),
+            opts: sv_opts.clone(),
+            gen: 3,
+        }];
 
         let mut new_opts = tds.opts().clone();
         new_opts.target[0] = new_downstairs.address();
@@ -5732,7 +5740,7 @@ mod test {
         let replacement = VolumeConstructionRequest::Volume {
             id: sv_volume_id,
             block_size: BLOCK_SIZE as u64,
-            sub_volumes: new_sub_vol.clone(),
+            sub_volumes: more_sub_vol,
             read_only_parent: Some(new_rop),
         };
 
--- a/upstairs/src/volume.rs
+++ b/upstairs/src/volume.rs
@@ -39,7 +39,7 @@ fn build_region_definition(
 }
 
 // Result of comparing two [`VolumeConstructionRequest::Region`] values
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 enum VCRDelta {
     /// The Regions are identical.
     Same,
@@ -50,7 +50,7 @@ enum VCRDelta {
 }
 
 // Result of comparing two [`VolumeConstructionRequest::Volume`] values
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 enum CompareResult {
     Volume {
         sub_compares: Vec<CompareResult>,
@@ -1398,7 +1398,6 @@ impl Volume {
 
                     parts.push_back(read_only_parent_compare);
                 }
-
                 CompareResult::Region { delta } => {
                     match delta {
                         VCRDelta::Same => {
@@ -1437,9 +1436,6 @@ impl Volume {
                 // Walk the cases. As some deltas are okay for a read only
                 // parent that are not for a sub_volume, we need to look at all
                 // the possible permutations Nexus supports.
-                //
-                // TODO also modify this when multiple subvolumes supported!
-
                 if targets.is_empty() {
                     // If here, we passed by the `all_same` bail above, meaning
                     // there's a generation delta somewhere.
@@ -1461,6 +1457,26 @@ impl Volume {
                 let VCRDelta::Target { old, new } = targets[0] else {
                     panic!("should only have pushed Target");
                 };
+
+                // We know we have one target that is different.  Verify that
+                // all the rest of the sub_volumes have a generation number
+                // difference.
+                for sc in sub_compares {
+                    match sc {
+                        CompareResult::Region { delta } => match delta {
+                            VCRDelta::Same => {
+                                crucible_bail!(
+                                    ReplaceRequestInvalid,
+                                    "SubVolumes don't have generation bump"
+                                );
+                            }
+                            VCRDelta::Generation | VCRDelta::Target { .. } => {}
+                        },
+                        _ => {
+                            panic!("Unsupported multi level CompareResult");
+                        }
+                    }
+                }
 
                 Ok(Some((*old, *new)))
             }
@@ -1487,6 +1503,12 @@ impl Volume {
         }
     }
 
+    // Given two VCRs, compare them to each other, including each sub_volume
+    // and the read only parent.  Return the collection of results in the
+    // CompareResult struct.  Any mismatch results in an error returned
+    // right away and no further processing.  The caller is responsible for
+    // parsing the CompareResult and deciding what to do if there are
+    // multiple changes.
     fn compare_vcr_for_replacement(
         log: &Logger,
         o_vol: &VolumeConstructionRequest,
@@ -1536,52 +1558,50 @@ impl Volume {
                     )
                 }
 
-                // Presently, we only support one sub_volume for replacement.
-                // If support for multiple sub_volumes is added, then this
-                // following section will need to be updated to loop over the
-                // sub_volume Vec and find the specific one with a difference,
-                // while verifying that all other sub_volumes are no different.
-                if n_sub_volumes.len() != 1 {
-                    crucible_bail!(
-                        ReplaceRequestInvalid,
-                        "Only a single sub_volume is supported"
-                    )
+                let mut sv_results = Vec::new();
+                // Walk the list of SubVolumes, get the compare result from
+                // each one and add it to list for the final result.
+                // If we get an error, then that is returned right away and
+                // we don't bother processing any further.
+                for (o_sv, new_sv) in
+                    o_sub_volumes.iter().zip(n_sub_volumes.iter())
+                {
+                    let sv_res = Self::compare_vcr_for_replacement(
+                        log, o_sv, new_sv, read_only,
+                    )?;
+                    sv_results.push(sv_res);
                 }
+                let read_only_parent_compare =
+                    Box::new(match (o_read_only_parent, n_read_only_parent) {
+                        (.., None) => {
+                            // The New VCR is none. When comparing
+                            // read_only_parents this is okay.
+                            CompareResult::NewMissing
+                        }
 
-                Ok(CompareResult::Volume {
-                    sub_compares: vec![Self::compare_vcr_for_replacement(
-                        log,
-                        &o_sub_volumes[0],
-                        &n_sub_volumes[0],
-                        read_only,
-                    )?],
+                        (None, Some(..)) => {
+                            // It's never valid when comparing VCRs to have
+                            // the original VCR be missing and the new VCR
+                            // is present.
+                            crucible_bail!(
+                                ReplaceRequestInvalid,
+                                "VCR added where there should not be one"
+                            );
+                        }
 
-                    read_only_parent_compare: Box::new(
-                        match (o_read_only_parent, n_read_only_parent) {
-                            (.., None) => {
-                                // The New VCR is none. When comparing
-                                // read_only_parents this is okay.
-                                CompareResult::NewMissing
-                            }
+                        (Some(o_vol), Some(n_vol)) => {
+                            Self::compare_vcr_for_replacement(
+                                log, o_vol, n_vol, true,
+                            )?
+                        }
+                    });
 
-                            (None, Some(..)) => {
-                                // It's never valid when comparing VCRs to have
-                                // the original VCR be missing and the new VCR
-                                // is present.
-                                crucible_bail!(
-                                    ReplaceRequestInvalid,
-                                    "VCR added where there should not be one"
-                                );
-                            }
+                let sv_res = CompareResult::Volume {
+                    sub_compares: sv_results,
+                    read_only_parent_compare,
+                };
 
-                            (Some(o_vol), Some(n_vol)) => {
-                                Self::compare_vcr_for_replacement(
-                                    log, o_vol, n_vol, true,
-                                )?
-                            }
-                        },
-                    ),
-                })
+                Ok(sv_res.clone())
             }
 
             (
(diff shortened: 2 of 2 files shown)


integration_tests/src/lib.rs
```diff
@@ -5695,13 +5695,13 @@ mod test {
             blocks_per_extent: sv_tds.blocks_per_extent(),
             extent_count: sv_tds.extent_count(),
             opts: sv_opts.clone(),
-            gen: 1,
+            gen: 2,
```
mkeeter, line 5698: Why does this `gen` need to change?  I would expect to only need a new `gen` for the replacement (below)

upstairs/src/volume.rs
```diff
@@ -1536,52 +1558,50 @@ impl Volume {
...
+                {
+                    let sv_res = Self::compare_vcr_for_replacement(
+                        log, o_sv, new_sv, read_only,
+                    )?;
+                    sv_results.push(sv_res);
                 }
+                let read_only_parent_compare =
+                    Box::new(match (o_read_only_parent, n_read_only_parent) {
+                        (.., None) => {
+                            // The New VCR is none. When comparing
```
mkeeter, line 1577: "The New VCR is none" is confusing.  I think it means "The new VCR has no read-only parent"?  Also, saying _why_ this is okay would be helpful (I think we need to support the case because it occurs after scrubbing is complete?)

upstairs/src/volume.rs
```diff
@@ -1536,52 +1558,50 @@ impl Volume {
...
-                    )
+                let mut sv_results = Vec::new();
+                // Walk the list of SubVolumes, get the compare result from
+                // each one and add it to list for the final result.
+                // If we get an error, then that is returned right away and
+                // we don't bother processing any further.
+                for (o_sv, new_sv) in
+                    o_sub_volumes.iter().zip(n_sub_volumes.iter())
+                {
+                    let sv_res = Self::compare_vcr_for_replacement(
```
mkeeter, line 1569: Any worries about unbounded recursion blowing up the stack here?

upstairs/src/volume.rs
```diff
@@ -1536,52 +1558,50 @@ impl Volume {
...
-                            }
-                        },
-                    ),
-                })
+                let sv_res = CompareResult::Volume {
+                    sub_compares: sv_results,
+                    read_only_parent_compare,
+                };
+
+                Ok(sv_res.clone())
```
mkeeter, line 1604: It looks like this is the only use of `CompareResult::clone`, and it's unnecessary.  Everything seems to compile fine if I remove it (assuaging my Bad Vibes from above):
```diff
diff --git a/upstairs/src/volume.rs b/upstairs/src/volume.rs
index f61f3f5303..43520a3155 100644
--- a/upstairs/src/volume.rs
+++ b/upstairs/src/volume.rs
@@ -39,7 +39,7 @@
 }

 // Result of comparing two [`VolumeConstructionRequest::Region`] values
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 enum VCRDelta {
     /// The Regions are identical.
     Same,
@@ -50,7 +50,7 @@
 }

 // Result of comparing two [`VolumeConstructionRequest::Volume`] values
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 enum CompareResult {
     Volume {
         sub_compares: Vec<CompareResult>,
@@ -1601,7 +1601,7 @@
                     read_only_parent_compare,
                 };

-                Ok(sv_res.clone())
+                Ok(sv_res)
             }

             (
```

upstairs/src/volume.rs
```diff
@@ -3490,55 +3510,116 @@ mod test {
...
                 block_size,
                 blocks_per_extent,
                 extent_count,
                 opts: opts.clone(),
                 gen: 2,
-            }],
-            read_only_parent: None,
+            };
+            sub_volumes.push(sv);
+        }
```
mkeeter, line 3579: Nit: this could be slightly shorter + immutable
```rust
        let sub_volumes = (0..sv_count)
            .map(|_| VolumeConstructionRequest::Region {
                block_size,
                blocks_per_extent,
                extent_count,
                opts: opts.clone(),
                gen: 2,
            })
            .collect();
```

upstairs/src/volume.rs
```diff
@@ -3490,55 +3510,116 @@ mod test {
...
+                opts: replacement_opts.clone(),
+                gen: replacement_gen,
+            };
+            replacement_sub_volumes.push(sv);
+        }
+        let replacement = VolumeConstructionRequest::Volume {
+            id: vol_id,
+            block_size,
+            sub_volumes: replacement_sub_volumes,
+            read_only_parent: rop.clone(),
```
mkeeter, line 3615: Nit: similarly, this could be
```rust
        let replacement_sub_volumes = (0..sv_count)
            .map(|i| {
                let mut replacement_opts = opts.clone();
                let replacement_gen = 3;

                if i == sv_changed {
                    replacement_opts.target[cid] = new_target;
                }

                VolumeConstructionRequest::Region {
                    block_size,
                    blocks_per_extent,
                    extent_count,
                    opts: replacement_opts.clone(),
                    gen: replacement_gen,
                }
            })
            .collect();
```

upstairs/src/volume.rs
```diff
@@ -3490,55 +3510,116 @@ mod test {
...
+    // sv_changed: The index (zero based) of the sub volume where we
+    // want the target to be different.
+    // cid: The client ID where the change will happen, which is the
+    // same as the index in the crucible opts target array.
+    // add_rop: True to create a read_only_parent, false for None
+    fn test_volume_replace_inner(
+        sv_count: usize,
+        sv_changed: usize,
+        cid: usize,
+        add_rop: bool,
```
mkeeter, line 3542: Nit: `add_rop` is confusing, because we don't change the ROP state.  Maybe `with_rop`?

mkeeter: Also, _should_ we test removing the read-only parent?  We could use a three-variant enum in these tests, e.g.

```rust
enum ReadOnlyParentMode {
    Absent,
    Present,
    Remove,
}
```

upstairs/src/volume.rs
```diff
@@ -3549,76 +3630,116 @@ mod test {
...
+            Some(rop)
+        } else {
+            None
+        };
 
-        let original = VolumeConstructionRequest::Volume {
-            id: vol_id,
-            block_size,
-            sub_volumes: vec![VolumeConstructionRequest::Region {
+        let mut sub_volumes = Vec::new();
```
mkeeter, line 3681: Same suggestion about building this immutably with `(0..sv_count).map(...)`

upstairs/src/volume.rs
```diff
@@ -3688,28 +3809,55 @@ mod test {
...
+    // Test with 1, 2, and 3 sub_volumes.
+    // Test both with and without a read_only_parent.
+    fn volume_replace_self() {
+        for sv in 1..4 {
+            test_volume_replace_self_inner(sv, false);
+            test_volume_replace_self_inner(sv, true);
+        }
+    }
+
+    fn test_volume_replace_self_inner(sv_count: usize, add_rop: bool) {
```
mkeeter, line 3824: Same comment here about `add_rop`

upstairs/src/volume.rs
```diff
@@ -3549,76 +3630,116 @@ mod test {
...
+                test_volume_replace_no_gen_inner(sv, sv_changed, false);
+                test_volume_replace_no_gen_inner(sv, sv_changed, true);
+            }
+        }
+    }
+
+    fn test_volume_replace_no_gen_inner(
+        sv_count: usize,
+        sv_changed: usize,
+        add_rop: bool,
```
mkeeter, line 3653: Same comment here about `add_rop`

upstairs/src/volume.rs
```diff
@@ -3824,44 +4042,79 @@ mod test {
...
 
-        let mut opts = generic_crucible_opts(vol_id);
-        let original = VolumeConstructionRequest::Volume {
-            id: vol_id,
-            block_size,
-            sub_volumes: vec![VolumeConstructionRequest::Region {
+        let opts = generic_crucible_opts(vol_id);
+
+        let mut sub_volumes = Vec::new();
+        for _ in 0..sv_count {
```
mkeeter, line 4067: You know, we should probably just have a `build_subvolume_vcr` helper function instead of copy-pasting this into every single function...

integration_tests/src/lib.rs
leftwo: This one does not.  I initially changed the wrong generation spot.  I've reverted this change.

upstairs/src/volume.rs
```diff
@@ -4501,101 +5030,21 @@ mod test {
...
-            blocks_per_extent,
-            extent_count,
-            o_opts,
-            n_opts
-        )
-        .is_err());
-    }
-
-    #[test]
-    fn volume_replace_with_rop() {
```
mkeeter, line 4516: Was this test folded into somewhere else?

upstairs/src/volume.rs
leftwo: Yep, I've made that change, thanks.

upstairs/src/volume.rs
leftwo: Since the loops are doing all the work here, we could even do:
```
ReadOnlyParentMode {
    Both,
    OnlyOriginal,
    OnlyReplacement,
    Neither,
}

The `OnlyReplacement` should fail 100% of the time.

File: upstairs/src/volume.rs (hunk 12 above)
Comment (leftwo) [reply upstairs/src/volume.rs:4516]: I think it may have been a duplicate before this PR, but for sure we cover it in the `volume_replace_basic()` tests

File: upstairs/src/volume.rs (hunk 3 above)
Comment (leftwo) [reply upstairs/src/volume.rs:1569]: Yes.  But, I'm not sure what to do about it.
We currently don't have any limits in Nexus, so you could craft VCRs forever until something broke.  I don't know what the limit is currently and what will break first.

File: upstairs/src/volume.rs (hunk 2 above)
Comment (leftwo) [reply upstairs/src/volume.rs]: Updated the comment in ab9e0223

File: upstairs/src/volume.rs (hunk 5 above)
Comment (leftwo) [reply upstairs/src/volume.rs]: ab9e0223 has this immutable update, plus I made a specific function and called it wherever I could.

File: upstairs/src/volume.rs (hunk 8 above)
Comment (leftwo) [reply upstairs/src/volume.rs]: Done in ab9e0223

File: upstairs/src/volume.rs (hunk 11 above)
Comment (leftwo) [reply upstairs/src/volume.rs]: Done in ab9e0223

File: upstairs/src/volume.rs (hunk 10 above)
Comment (leftwo) [reply upstairs/src/volume.rs]: With 2032eea3, no more add_rop.

File: upstairs/src/volume.rs (hunk 9 above)
Comment (leftwo) [reply upstairs/src/volume.rs]: With 2032eea3, no more add_rop.

File: upstairs/src/volume.rs (hunk 7 above)
Comment (leftwo) [reply upstairs/src/volume.rs]: I did this in 2032eea3

File: upstairs/src/volume.rs (hunk 6 above)
Comment (leftwo) [reply upstairs/src/volume.rs]: In 1949ae7d I did this as well as made a few into a common function.

**Review by mkeeter — APPROVED** [review]
LGTM

The only follow-up suggestion (optional) is to consolidate "test on every `ReadOnlyParentMode`", e.g.

```rust
impl ReadOnlyParentMode
    fn test_all<F: FnMut(Self)>(mut f: F) {
        f(ReadOnlyParentMode::OnlyOriginal);
        f(ReadOnlyParentMode::OnlyReplacement);
        f(ReadOnlyParentMode::Neither);
        f(ReadOnlyParentMode::Both);
    }
}
```

...which you could then use like

```rust
        for sv in 1..4 {
            for sv_changed in 0..sv {
                for cid in 0..3 {
                    ReadOnlyParentMode::test_all(|mode|
                        test_volume_replace_inner(
                            sv,
                            sv_changed,
                            cid,
                            mode,
                        )
                    );
                }
            }
        }
```

upstairs/src/volume.rs
```diff
@@ -1462,6 +1458,26 @@ impl Volume {
...
+                            VCRDelta::Same => {
+                                crucible_bail!(
+                                    ReplaceRequestInvalid,
+                                    "SubVolumes don't have generation bump"
+                                );
+                            }
+                            VCRDelta::Generation | VCRDelta::Target { .. } => {}
+                        },
+                        _ => {
+                            panic!("Unsupported multi level CompareResult");
```
jmpesp, line 1476: I think this should return `ReplaceRequestInvalid` instead of panicking

upstairs/src/volume.rs
```diff
@@ -3485,60 +3506,221 @@ mod test {
...
+        extent_count: u32,
+        opts: CrucibleOpts,
+        gen: u64,
+    ) -> Vec<VolumeConstructionRequest> {
+        (0..sv_count)
+            .map(|_| VolumeConstructionRequest::Region {
+                block_size,
+                blocks_per_extent,
+                extent_count,
+                opts: opts.clone(),
```
jmpesp, line 3586: The targets should be randomized here so that all sub volumes are not the same, using something like https://docs.rs/ipgen/1.0.2/ipgen/fn.ip.html

upstairs/src/volume.rs
```diff
@@ -3485,60 +3506,221 @@ mod test {
...
+
+                if i == sv_changed {
+                    replacement_opts.target[cid] = new_target;
+                }
+
+                VolumeConstructionRequest::Region {
+                    block_size,
+                    blocks_per_extent,
+                    extent_count,
+                    opts: replacement_opts.clone(),
```
jmpesp, line 3632: This too will have multiple subvolume targets that are the same

upstairs/src/volume.rs
```diff
@@ -3485,60 +3506,221 @@ mod test {
...
+        sv_changed: usize,
+        cid: usize,
+        rop_mode: ReadOnlyParentMode,
+    ) -> Result<()> {
+        assert!(sv_count > sv_changed);
+        assert!(cid < 3);
         // A valid replacement VCR is provided with a larger generation
         // number and only one target being different.
+        let vol_id = Uuid::new_v4();
+        let opts = generic_crucible_opts(vol_id);
```
jmpesp, line 3660: maybe `generic_crucible_opts` should also randomize the addresses?

upstairs/src/volume.rs
leftwo: I'm holding it wrong, somehow...

I get back:
`Err(InvalidIpNetwork("[BUG] failed to parse the generated IP address `a00:67` as IPv4"))`

And, the source says:
```
    // This error should never happen but I'm not a fan of panicking in libraries
    .map_err(|_| Error::InvalidIpNetwork(format!("[BUG] failed to parse the generated IP address `{}` as IPv4", ipv6_addr.trim_start_matches(':'))))
```

leftwo: Ah, it is a bug in the latest release 1.0.2. 
Fixed in main, but they have not rolled a new release out that contains the fix.

upstairs/src/volume.rs
leftwo: Fixed in 2e2ea36

upstairs/src/volume.rs
leftwo: I found another library that could generate random I.P. addresses for me and used that.

jmpesp (approved):

leftwo:
This is ready to go back, but we are holding it for the moment because this may require some changes on the Omicron side.
The changes would be that any change to a virtual disk target (a replacement) would require bumping the generation number on all Sub_Volumes that make up that virtual disk, including replacement of a read_only_parent target.   We need to determine if Omicron already does this, or if it needs updates, or if we need to relax this constraint for read_only_parent targets.

Also, this work should come behind https://github.com/oxidecomputer/omicron/issues/6172

leftwo:
Here is my test checklist

- [ ] Test ROP removal after scrubber runs (only attached, no pantry scrubber yet).
Boot an instance created from a snapshot or image.
Let the scrubber finish
Stop, then Start the instance.
Verify propolis logs don’t show ROP
- [ ] Replacement of a sub-volume region where there is no ROP
Create an instance with a data disk, replace a region on that data disk.
For the pantry, replace a region on an unattached disk.
- [ ] Replacement of a sub-volume region where there is a ROP.
- [ ] Replacement of a region in a ROP, before scrubber.
Create an instance but don’t start it. For pantry, never start it.
Figure out the ROP region you want to replace, get the replace command ready.
Start the instance, then wait 10 or so seconds, then use omdb to submit the replace.
- [ ] Replacement of a region in a ROP during scrubber (attached only).
- [ ] Replacement of a region in a ROP after scrubber (attached only).
- [ ] Replacement of a region in a ROP that has been replaced once already.
- [ ] Replace multiple regions on multiple disks attached to a single instance (only attached)

leftwo:
Testing has sadly confirmed that Omicron does not bump generation for SV when the ROP changes:
```
18:20:44.476Z INFO propolis-server (vm_state_driver): request to replace Crucible VCR
    backend_id = 76487e36-b8fa-4cfa-b69b-b5fb9c584356
18:20:44.476Z INFO propolis-server (vm_state_driver): compare result is Volume { sub_compares: [Region { delta: Same }], read_only_parent_compare: Volume { sub_compares: [Region { delta: Target { old: [fd00:1122:3344:101::16]:19008, new: [fd00:1122:3344:101::12]:19003 } }], read_only_parent_compare: NewMissing } }
18:20:44.476Z INFO propolis-server (vm_state_driver): state driver handled event
    outcome = Continue
18:20:44.476Z INFO propolis-server: request completed
    error_message_external = Invalid downstairs replace SubVolumes don't have generation bump
    error_message_internal = Invalid downstairs replace SubVolumes don't have generation bump
    latency_us = 188
    local_addr = [fd00:1122:3344:101::1:9]:12400
    method = PUT
    remote_addr = [fd00:1122:3344:101::b]:49419
    req_id = 4456a878-e235-4608-ade6-b9153d08389d
    response_code = 400
    uri = /instance/disk/76487e36-b8fa-4cfa-b69b-b5fb9c584356/vcr
 ```
 
 And, I think that is the right call, so I'm going to change this PR (and crucible) to adopt the same behavior.

PR #1552, opened 2024-11-05
Remove `ackable_work`; ack immediately instead

mkeeter:
A long time ago, one task accumulated ackable jobs into `Downstairs::ackable_work` and a different task periodically read out that set and sent acks back to the guest.

The tasks have long since been consolidated, but the `ackable_work` set remained.

This PR removes the `ackable_work` set, acking jobs immediately when they are ready.  It's a significant LOC simplification, because it's removing intermediate state (which previously had to be checked in unit tests).

There's one subtlety in handling live-repair: previously, the live-repair check would look in `ackable_work` to see if the next job was ackable, which indicates that live-repair can continue.  Now, we update the `LiveRepairState` directly in `Downstairs::ack`, storing a result for the pending live-repair job.

### DIFF
--- a/upstairs/src/downstairs.rs
+++ b/upstairs/src/downstairs.rs
@@ -1014,8 +1014,8 @@ impl Downstairs {
 
     /// Pushes live-repair forward, if possible
     ///
-    /// It's possible that handling the current live-repair job will make
-    /// subsequent live-repair jobs ackable immediately
+    /// It's possible that handling the current live-repair job will cause
+    /// subsequent live-repair jobs to be completed immediately
     /// (`test_repair_extent_fail_noop_out_of_order` exercises this case).  As
     /// such, this function will continue running until the next live-repair job
     /// is not ready.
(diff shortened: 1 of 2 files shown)


leftwo (approved):
We had a separate task (way way back) in part to make progress if queues were full and we needed to make forward progress.  I don't believe that situation still exists.

upstairs/src/downstairs.rs
```diff
@@ -2140,7 +2178,7 @@ impl Downstairs {
     ///
     /// - enqueue the job in each of [Self::clients] (clients may skip the job)
     /// - add the job to [Self::ds_active]
-    /// - Mark the job as ackable if it was skipped by all downstairs
+    /// - Ackable the job if it was skipped by all downstairs
```
jmpesp, line 2181: typo here?

jmpesp (approved):

upstairs/src/downstairs.rs
mkeeter: Yup, fixed

PR #1554, opened 2024-11-08
Skip backpressure guard when job is skipped; fix log

mkeeter:
We normally use `set_job_state` when changing job state, which handles dropping backpressure guards when a job is skipped.  However, this wasn't being handled correctly if a job was skipped when initially created.


mkeeter:
Note that this is also fixed with #1515, which may be a better way to move forward

leftwo (approved):

mkeeter:
Closed in favor of #1515

Issue #1390, opened 2024-07-22
dsc failed to honor a client stop request


leftwo:
In a CI run, the test-up-encrypted.sh hung wile doing the replace-reconcile test
(original CI link, may be broken: https://github.com/oxidecomputer/crucible/pull/1387/checks?check_run_id=27758312216
)

Copy of logs from the buildomat instance are here: /staff/tmp/buildomat-crucible-20240722/
Looking at the dsc log:
```
/var/tmp/bins/dsc start --region-count 4 --output-dir /tmp/test_up/dsc --ds-bin /var/tmp/bins/crucible-downstairs --region-dir /var/tmp/test_up
Using existing output directory "/tmp/test_up/dsc"
start ds: 8810
start access at:127.0.0.1:9998
start ds: 8820
start ds: 8830
start ds: 8840
Starting downstairs at port 8820
Starting downstairs at port 8810
Make output file at "/tmp/test_up/dsc/downstairs-8810.txt"
Make output file at "/tmp/test_up/dsc/downstairs-8820.txt"
Starting downstairs at port 8830
Make output file at "/tmp/test_up/dsc/downstairs-8830.txt"
Starting downstairs at port 8840
Make output file at "/tmp/test_up/dsc/downstairs-8840.txt"
[8820][1] reports Starting
[8810][0] reports Starting
[8830][2] reports Starting
[8840][3] reports Starting
```

Normal so far, we are starting up 4 downstairs at the given locations,

```
Downstairs /var/tmp/test_up/8830 port 8830 PID:Some(1233)
[8830][2] reports Running
Downstairs /var/tmp/test_up/8840 port 8840 PID:Some(1235)
[8840][3] reports Running
Downstairs /var/tmp/test_up/8820 port 8820 PID:Some(1236)
[8820][1] reports Running
Jul 22 16:05:29.547 INFO listening, local_addr: 127.0.0.1:9998
Control access at:127.0.0.1:9998
```
However, only three downstairs make it to running, 8810 (client 0) we don't get a message for.

A client requests a stop of 0:
```
Jul 22 16:05:34.651 INFO accepted connection, remote_addr: 127.0.0.1:36794, local_addr: 127.0.0.1:9998
Main task has work to do, go find it
got dsc Stop(0)
stop 0
[...]

PR #1561, opened 2024-11-12
Consolidate ack checking

mkeeter:
(staged on top of #1552)

There are a bunch of places where we check to see if jobs should be acked and/or retired:

- `Downstairs::enqueue` (if skipped on all three Downstairs)
- `Downstairs::skip_all_jobs` (for skipped jobs only)
- `DownstairsClient::process_io_completion` (lots of conditionals to decide if we're ready to ack; returns a flag to `Downstairs::process_io_completion_inner`)

This PR consolidates ack and retire checks into a single place: `Downstairs::ack_check`.  This new function is responsible for deciding when to ack and retire jobs; it's called in all of the places where you'd expect.

`DownstairsClient::process_io_completion` is reorganized, with one minor logical change: non-fatal read errors now increment the `downstairs_errors` counter.  It's no longer responsible for deciding when to ack a job; that's now in `Downstairs::ack_check`.

The `up-to-ds-*-done` DTrace probes are removed, because (after #1552) they fire immediately before the equivalent `gw-*-done` probes; there's no delay waiting for ackable jobs to be acked.

### DIFF
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -1337,7 +1337,13 @@ impl DownstairsClient {
                     // snapshot exists already but we should not increment
                     // `downstairs_errors` and transition that Downstairs to
                     // Failed - that downstairs is still able to serve IO.
-                    (_, CrucibleError::SnapshotExistsAlready(..)) => (),
+                    (
+                        IOop::Flush {
+                            snapshot_details: Some(..),
+                            ..
+                        },
+                        CrucibleError::SnapshotExistsAlready(..),
+                    ) => (),
 
                     // If a read job fails, we sometimes need to panic
                     (IOop::Read { .. }, CrucibleError::HashMismatch) => {
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -1232,34 +1232,97 @@ impl DownstairsClient {
 
     /// Handles a single IO operation
     ///
-    /// Returns `true` if the job is now ackable, `false` otherwise
-    ///
     /// If this is a read response, then the values in `responses` must
-    /// _already_ be decrypted (with corresponding validation results stored in
-    /// `read_validations`).
+    /// _already_ be decrypted, with validated contexts stored in
+    /// `responses.blocks`.
     pub(crate) fn process_io_completion(
         &mut self,
         ds_id: JobId,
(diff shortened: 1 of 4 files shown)


upstairs/src/client.rs
```diff
@@ -1232,341 +1232,149 @@ impl DownstairsClient {
...
+                    self.log,
+                    "DS Reports error {e:?} on job {}, {:?} EC", ds_id, job,
+                );
+                match (&job.work, &e) {
+                    // Some errors can be returned without considering the
+                    // Downstairs bad. For example, it's still an error if a
+                    // snapshot exists already but we should not increment
+                    // `downstairs_errors` and transition that Downstairs to
+                    // Failed - that downstairs is still able to serve IO.
+                    (_, CrucibleError::SnapshotExistsAlready(..)) => (),
```
jmpesp, line 1340: Not matching on Flush here?

upstairs/src/downstairs.rs
```diff
@@ -474,12 +474,77 @@ impl Downstairs {
         }
     }
 
+    /// Checks whether an ack and/or a retire check is necessary for the job
```
jmpesp, line 477: This is exactly the follow up to 1552 I was going to chat to you about, but it was next in the queue :) :+1:

jmpesp (approved):

upstairs/src/client.rs
mkeeter: This matches the previous behavior ([L1296](https://github.com/oxidecomputer/crucible/pull/1561/files#diff-df991e179692a549eace0bb472a6b3c07483ab18527ee9b30f433142a95fd89fL1296)), but I agree that we could be more specific; fixed!

(Note that this only handles error counters, the actual [error checking in the Downstairs](https://github.com/oxidecomputer/crucible/blob/f3d07aa414081b77e13a2f5f0da7bfbaa1749372/upstairs/src/downstairs.rs#L3361-L3363) remains agnostic to `IOop` type)

faithanalog (approved):

faithanalog:
it seems CI may not agree with us on these changes

mkeeter:
CI weirdness appears to be https://github.com/oxidecomputer/crucible/issues/1390 (or similar)

PR #1565, opened 2024-11-19
Add early rejection of IOs if too many Downstairs are inactive

mkeeter:
This aborts the IO before passing it to the Downstairs, so it's not assigned a `JobId` or put into the `ActiveJobs` map.  The most noticeable change is that writes are now fast-err'd instead of fast-acked if > 1 Downstairs is inactive.

### DIFF
--- a/upstairs/src/downstairs.rs
+++ b/upstairs/src/downstairs.rs
@@ -3330,12 +3330,7 @@ impl Downstairs {
     ///
     /// A client can accept IO if it is in the `Active` or `LiveRepair` state.
     pub fn active_client_count(&self) -> usize {
-        self.clients
-            .iter()
-            .filter(|c| {
-                matches!(c.state(), DsState::Active | DsState::LiveRepair)
-            })
-            .count()
+        self.clients.iter().filter(|c| c.is_accepting_io()).count()
     }
 
     /// Wrapper for marking a single job as done from the given client
--- a/upstairs/src/downstairs.rs
+++ b/upstairs/src/downstairs.rs
@@ -3326,6 +3326,18 @@ impl Downstairs {
         }
     }
 
+    /// Returns the number of clients that can accept IO
+    ///
+    /// A client can accept IO if it is in the `Active` or `LiveRepair` state.
+    pub fn active_client_count(&self) -> usize {
+        self.clients
+            .iter()
+            .filter(|c| {
+                matches!(c.state(), DsState::Active | DsState::LiveRepair)
+            })
+            .count()
+    }
+
     /// Wrapper for marking a single job as done from the given client
     ///
     /// This can be used to test handling of job acks, etc
--- a/upstairs/src/dummy_downstairs_tests.rs
+++ b/upstairs/src/dummy_downstairs_tests.rs
@@ -3191,7 +3191,7 @@ async fn read_with_one_fault() {
     h.await.unwrap(); // we have > 1x reply, so the read will return
     harness.ds3.ack_read().await;
 
-    // Take out DS1 next
+    // Take out DS2 next
     harness
         .guest
         .write(BlockIndex(0), write_buf.clone())
--- a/upstairs/src/dummy_downstairs_tests.rs
+++ b/upstairs/src/dummy_downstairs_tests.rs
@@ -12,6 +12,8 @@ use crate::guest::Guest;
 use crate::up_main;
 use crate::BlockIO;
 use crate::Buffer;
+use crate::ClientFaultReason;
+use crate::ClientStopReason;
 use crate::ConnectionMode;
 use crate::CrucibleError;
 use crate::DsState;
@@ -3103,6 +3105,19 @@ async fn test_bytes_based_barrier() {
     harness.ds3.ack_flush().await;
 }
 
+fn assert_faulted(s: &DsState) {
+    match s {
+        DsState::Stopping(ClientStopReason::Fault(
+            ClientFaultReason::RequestedFault,
+        ))
+        | DsState::Connecting {
+            mode: ConnectionMode::Faulted,
+            ..
+        } => (),
+        _ => panic!("invalid state: expected faulted, got {s:?}"),
+    }
+}
+
 /// Test for early rejection of writes if > 1 Downstairs is unavailable
 #[tokio::test]
 async fn fast_write_rejection() {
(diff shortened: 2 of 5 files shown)


leftwo:
Some random musings around this.

If we send the write to one downstairs, it could land on disk, but the IO will sit waiting for another downstairs to write it before we could ACK back to the guest.

Any flush is going to stop all future IOs, as they will all build up behind it.  This is actually a reason to fast fail writes/flushes as I believe that would enable reads to still work, right?

What should we do with in-flight IOs?  Do we now want to discard them as well?  If we do fail outstanding IOs, outstanding writes we would have already fast-acked, but flushes will now go to failed.

By doing this, are we cutting off a chance for a replay to catch things up?

mkeeter:
> If we send the write to one downstairs, it could land on disk, but the IO will sit waiting for another downstairs to write it before we could ACK back to the guest.

No, writes are either fast-acked to the guest or (after changes made in this PR) fast-errored.  In the latter case, they aren't sent to any Downstairs.

> Any flush is going to stop all future IOs, as they will all build up behind it. This is actually a reason to fast fail writes/flushes as I believe that would enable reads to still work, right?

Yes, with caveats – if the Guest is waiting for a flush, then sending a read, it would previously stall if 2x Downstairs are offline.  Now, the flush will return an error immediately, and the read can proceed (or not, depending on how the Guest handles error codes).

> What should we do with in-flight IOs? Do we now want to discard them as well? If we do fail outstanding IOs, outstanding writes we would have already fast-acked, but flushes will now go to failed.

This PR hasn't changed any behavior here.  If a Downstairs goes to the Faulted state, all of its IOs are skipped, which may cause them to be acked to the Guest.

> By doing this, are we cutting off a chance for a replay to catch things up?

No, any IO rejected here is never added to the active jobs list, so it's never sent to _any_ downstairs.  It's rejected before being assigned a `JobId`.

leftwo:
> > If we send the write to one downstairs, it could land on disk, but the IO will sit waiting for another downstairs to write it before we could ACK back to the guest.
> 
> No, writes are either fast-acked to the guest or (after changes made in this PR) fast-errored. In the latter case, they aren't sent to any Downstairs.

Ah yes, fast ack will ack while it thinks things are good.  If one downstairs is online, it could get the data and the other two would not, but we have already acked it back to the guest.

 
> > Any flush is going to stop all future IOs, as they will all build up behind it. This is actually a reason to fast fail writes/flushes as I believe that would enable reads to still work, right?
> 
> Yes, with caveats – if the Guest is waiting for a flush, then sending a read, it would previously stall if 2x Downstairs are offline. Now, the flush will return an error immediately, and the read can proceed (or not, depending on how the Guest handles error codes).
> 
> > What should we do with in-flight IOs? Do we now want to discard them as well? If we do fail outstanding IOs, outstanding writes we would have already fast-acked, but flushes will now go to failed.
> 
> This PR hasn't changed any behavior here. If a Downstairs goes to the Faulted state, all of its IOs are skipped, which may cause them to be acked to the Guest.

The transition to faulted here, that's what I think I needed to see, so that will skip IO to any downstairs in `Faulted` state.

> > By doing this, are we cutting off a chance for a replay to catch things up?
> 
> No, any IO rejected here is never added to the active jobs list, so it's never sent to _any_ downstairs. It's rejected before being assigned a `JobId`.

So, your right that a replay does not happen here. I'm thinking about `Offline` state, vs. `Faulted`.  Once we have Faulted, then it's only LiveRepair to bring a downstairs back.

upstairs/src/downstairs.rs
```diff
@@ -3261,6 +3261,18 @@ impl Downstairs {
...
     }
 
+    /// Returns the number of clients that can accept IO
+    ///
+    /// A client can accept IO if it is in the `Active` or `LiveRepair` state.
+    pub fn active_client_count(&self) -> usize {
+        self.clients
+            .iter()
+            .filter(|c| {
+                matches!(c.state(), DsState::Active | DsState::LiveRepair)
```
leftwo, line 3271: I think we need to allow additional states here.  `Offliine` for instance is one where we have not yet decided what the situation is, and it could just be a downstairs rebooting and soon enough (fingers crossed) it will come back.

There may be other transitory states to consider as well.

mkeeter: Good point, I'll revisit this.  In parallel, I've been working on simplifying the state machine; see https://github.com/oxidecomputer/crucible/pull/1568 and https://github.com/oxidecomputer/crucible/pull/1570

faithanalog (commented):
I believe this change does what it intends to do, and I am pretty sure it's the right thing to do broadly, but I'm deferring to alan's comments on the details

upstairs/src/downstairs.rs
leftwo: If this goes in before 1568/1570, then we want the updated list of `DsState`'s

mkeeter: At long last, #1577 is merged and this is rebased.  I've moved this `match` into a standalone function (`DownstairsClient::is_accepting_io`) and added `DsState::Connecting { mode: ConnectionMode::Offline, .. }` as a valid case.

leftwo (commented):
Will this block internal flushes for read only upstairs?

mkeeter:
> Will this block internal flushes for read only upstairs?

@leftwo I don't think so – internal flushes call `Upstairs::submit_flush` directly, which doesn't check for active clients (flushes initiated by the guest come in through `BlockOp::Flush`, which _does_ check).

leftwo (approved):

upstairs/src/dummy_downstairs_tests.rs
```diff
@@ -2950,3 +2999,204 @@ async fn test_bytes_based_barrier() {
...
+    // Check that reads still work
+    let h = harness.spawn(|guest| async move {
+        let mut buffer = Buffer::new(1, 512);
+        guest.read(BlockIndex(0), &mut buffer).await.unwrap();
+    });
+    harness.ds2.ack_read().await;
+    h.await.unwrap(); // we have > 1x reply, so the read will return
+    harness.ds3.ack_read().await;
+
+    // Take out DS1 next
```
jmpesp, line 3089: s/ds1/ds2/

jmpesp (approved):

Issue #1498, opened 2024-10-07
dsc failed to start a downstairs


leftwo:
In CI (link may be bad) job: https://buildomat.eng.oxide.computer/wg/0/artefact/01J9KZDBT7Q76BMAZ3NFT2EM6G/JHuMk44VV03fuovquSYDTdL9uCEruARhbYRD7WhRUTif7Lam/01J9KZDV6QXGVBQG0ZHA6JNXTH/01J9M3TRMJHBD7KTGG9QWWK022/dsc-out.txt

We had the test timeout.
Looking at test log output, we can see a new check we added is waiting for all the downstairs to respond:
```

120 | 2024-10-07T17:27:21.088Z | + ptime -m bash /input/build/work/scripts/test_up.sh -r 2 encrypted
-- | -- | --
121 | 2024-10-07T17:27:21.099Z | /input/build/work
122 | 2024-10-07T17:27:21.111Z | Using 2 region sets
123 | 2024-10-07T17:27:21.126Z | Upstairs using key: 1gql5EL5hhzi5hY5rXSvSfN31jxffOX/9IhgN0kkwMI=
124 | 2024-10-07T17:27:21.129Z | dsc output goes to /tmp/test_up-build/dsc-out.txt
125 | 2024-10-07T17:27:21.129Z | Creating 6 downstairs regions
126 | 2024-10-07T17:27:21.586Z | Starting 6 downstairs
127 | 2024-10-07T17:27:21.589Z | dsc started at PID: 1191
128 | 2024-10-07T17:27:26.657Z | Downstairs client 0 not running, waiting for it
129 | 2024-10-07T17:27:31.718Z | Downstairs client 0 not running, waiting for it
130 | 2024-10-07T17:27:36.778Z | Downstairs client 0 not running, waiting for it
```

The waiting for it waits till the test hits a timeout and we abort it.
In the logs for `dsc` we can see it got to `Starting` on client 0, but never to `Running`:

```
/var/tmp/bins/dsc create --encrypted --cleanup --region-count 6 --extent-size 10 --extent-count 5 --output-dir /tmp/test_up-build/dsc --ds-bin /var/tmp/bins/crucible-downstairs --region-dir /var/tmp/test_up-build
Removing existing dsc directory "/tmp/test_up-build/dsc"
Creating region directory at: /var/tmp/test_up-build
Creating dsc directory at: /tmp/test_up-build/dsc
[...]

PR #1566, opened 2024-11-20
Remove unprinted header item from sled_upstairs_info

leftwo:
The BAKPR is not printed, so let's take it out the header.


leftwo:
Filed https://github.com/oxidecomputer/crucible/issues/1569 for the test flake, but I now suspect it's another case of: https://github.com/oxidecomputer/crucible/issues/1498

PR #1568, opened 2024-11-20
Simplify fault and restart code (1/3)

mkeeter:
The conditions under which we restart a client IO task are somewhat baroque:

- It may halt spontaneously (e.g. upon a socket error)
- We may **fault** it explicitly (e.g. upon IO error).  In this case, we want to skip all work associated with this client.
- We may restart the IO task if initial negotiation fails
- ...and when replacing a Downstairs
- ...and when deactivated

This PR adds more specific types: `enum ClientNegotiationFailed` and `enum ClientFaultReason`.  Mid-negotiation restarts must provide a `ClientNegotiationFailed`; faulting the IO task must provide a `ClientFaultReason`.  Both of these types are then converted into a `ClientStopReason` for logging.

In addition, faulting a client is now done through `Downstairs::fault_client` instead of calling both `Downstairs::skip_all_jobs` and `DownstairsClient::fault`.  Having a single function makes this harder to mess up!

### DIFF
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -318,6 +318,7 @@ impl DownstairsClient {
     }
 
     fn halt_io_task(&mut self, r: ClientStopReason) {
+        info!(self.log, "halting IO task due to {r:?}");
         if let Some(t) = self.client_task.client_stop_tx.take() {
             if let Err(_e) = t.send(r) {
                 warn!(self.log, "failed to send stop request")
@@ -847,22 +848,12 @@ impl DownstairsClient {
         self.state
     }
 
-    pub(crate) fn restart_connection(
+    pub(crate) fn abort_negotiation(
         &mut self,
         up_state: &UpstairsState,
-        reason: ClientStopReason,
+        reason: ClientNegotiationFailed,
     ) {
         let new_state = match self.state {
-            DsState::Active | DsState::Offline
-                if matches!(
-                    reason,
-                    ClientStopReason::Fault(
-                        ClientFaultReason::IneligibleForReplay
-                    )
-                ) =>
-            {
-                DsState::Faulted
-            }
             DsState::Active | DsState::Offline => DsState::Offline,
             DsState::Faulted => DsState::Faulted,
             DsState::Deactivated => DsState::New,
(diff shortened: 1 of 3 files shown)


upstairs/src/downstairs.rs
```diff
@@ -2705,30 +2720,23 @@ impl Downstairs {
...
-                }
-                DsState::Faulted => {
-                    // Jobs were already skipped when we hit the IO error that
-                    // marked us as faulted
-                    self.clients[i].abort_repair(up_state, false);
+                    self.fault_client(
+                        i,
+                        up_state,
+                        ClientFaultReason::FailedLiveRepair,
+                    );
```
leftwo, line 2727: If we no longer need the `Faulted` check here, it means when this DS went to Faulted, it must have triggered the `fault_client()` path.  That would triggered the `skip_all_jobs()` call.

I don't believe that was always true, but it (as far as I can tell) is true now.

leftwo (approved):
I think this is good, one nit and one question/confirmation.

upstairs/src/client.rs
```diff
@@ -2202,25 +2180,53 @@ pub(crate) struct DownstairsStats {
 #[derive(Debug)]
 pub(crate) enum ClientStopReason {
     /// We are about to replace the client task
-    Replacing,
+    Replaced,
```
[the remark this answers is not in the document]
mkeeter: I agree that `ClientStopReason::Replacing` makes more sense here.

> If we request to replace a downstairs during negotiation, that's ClientNegotiationFailed(Replacing)
If we request to replace a downstairs after activation, does that become ClientStopReason(Replaced)?

Close, the two options are `ClientStopReason::NegotiationFailed(ClientNegotiationFailed::Replaced))` and `ClientStopReason::Replaced`.

upstairs/src/downstairs.rs
mkeeter: I think I agree with you – the `DsState::Faulted` check is now in the catchall `_ => { .. }` branch below.  You don't want anything to change here, right?

jmpesp (approved):

PR #1570, opened 2024-11-21
Add explicit Stopping state (2/3)

mkeeter:
Staged on top of #1568

There's an unfortunate ambiguity in certain states (e.g. `DsState::Faulted`), which represent two different things:

- We've stopped the IO task due to a fault, and are waiting for it to restart
- The IO task has restarted, and we're doing negotiation from a faulted state (i.e. will do live-repair)

This PR adds a new `DsState::Stopping(ClientStopReason)` state, which represents the former.  The previous states (`DsState::Faulted`) now _only_ mean that we're doing negotiation.

The new state subsumes `DsState::Deactivated`, `DsState::Replacing`, `DsState::Disabled`, which were specialized states that waited for the IO task to exit.  Each of those states is now `DsState::Stopping(..)` with an appropriate `ClientStopReason`.

The vast majority of this PR is automatic OpenAPI changes.  I don't think anyone is relying on the specific shape of `DsState` (which is only used in `UpstairsInfo` / the `info` endpoint), but please let me know if I'm wrong!

### DIFF
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -319,14 +319,13 @@ impl DownstairsClient {
         });
     }
 
-    fn halt_io_task(&mut self, r: ClientStopReason) {
+    fn halt_io_task(&mut self, up_state: &UpstairsState, r: ClientStopReason) {
         info!(self.log, "halting IO task due to {r:?}");
         if let Some(t) = self.client_task.client_stop_tx.take() {
             if let Err(_e) = t.send(r) {
                 warn!(self.log, "failed to send stop request")
             }
-            // XXX use checked_state_transition?
-            self.state = DsState::Stopping(r);
+            self.checked_state_transition(up_state, DsState::Stopping(r));
         } else {
             warn!(self.log, "client task is already stopping")
         }
@@ -550,7 +549,7 @@ impl DownstairsClient {
 
     /// Switches the client state to Deactivated and stops the IO task
     pub(crate) fn deactivate(&mut self, up_state: &UpstairsState) {
-        self.halt_io_task(ClientStopReason::Deactivated)
+        self.halt_io_task(up_state, ClientStopReason::Deactivated)
     }
 
     /// Resets this Downstairs and start a fresh connection
@@ -864,7 +863,7 @@ impl DownstairsClient {
         up_state: &UpstairsState,
         reason: ClientNegotiationFailed,
     ) {
-        self.halt_io_task(reason.into());
+        self.halt_io_task(up_state, reason.into());
     }
 
     /// Sets the current state to `DsState::Active`
@@ -974,7 +973,7 @@ impl DownstairsClient {
         self.region_metadata = None;
         self.stats.replaced += 1;
 
-        self.halt_io_task(ClientStopReason::Replacing);
+        self.halt_io_task(up_state, ClientStopReason::Replacing);
     }
 
     /// Sets `self.state` to `new_state`, with logging and validity checking
@@ -1201,8 +1200,7 @@ impl DownstairsClient {
         up_state: &UpstairsState,
         reason: ClientFaultReason,
     ) {
-        self.checked_state_transition(up_state, DsState::Faulted);
-        self.halt_io_task(reason.into());
+        self.halt_io_task(up_state, reason.into());
     }
 
     /// Finishes an in-progress live repair, setting our state to `Active`
@@ -1373,7 +1371,7 @@ impl DownstairsClient {
     ///
     /// The IO task will automatically restart in the main event handler
     pub(crate) fn disable(&mut self, up_state: &UpstairsState) {
-        self.halt_io_task(ClientStopReason::Disabled);
+        self.halt_io_task(up_state, ClientStopReason::Disabled);
     }
 
     /// Skips from `LiveRepairReady` to `Active`; a no-op otherwise
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -21,6 +21,8 @@ use std::{
 };
 
 use futures::StreamExt;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
 use slog::{debug, error, info, o, warn, Logger};
 use tokio::{
     net::{TcpSocket, TcpStream},
@@ -323,6 +325,8 @@ impl DownstairsClient {
             if let Err(_e) = t.send(r) {
                 warn!(self.log, "failed to send stop request")
             }
+            // XXX use checked_state_transition?
+            self.state = DsState::Stopping(r);
         } else {
             warn!(self.log, "client task is already stopping")
         }
@@ -546,7 +550,6 @@ impl DownstairsClient {
 
     /// Switches the client state to Deactivated and stops the IO task
     pub(crate) fn deactivate(&mut self, up_state: &UpstairsState) {
-        self.checked_state_transition(up_state, DsState::Deactivated);
         self.halt_io_task(ClientStopReason::Deactivated)
     }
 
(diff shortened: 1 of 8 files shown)


leftwo:
> 
> The vast majority of this PR is automatic OpenAPI changes. I don't think anyone is relying on the specific shape of `DsState` (which is only used in `UpstairsInfo` / the `info` endpoint), but please let me know if I'm wrong!

The `tools/test_fail_live_repair.sh` test uses /info (poorly)

Here is a diff to fix it.  I can also just push this to your branch directly (I thinik..)

```
diff --git a/tools/test_fail_live_repair.sh b/tools/test_fail_live_repair.sh
index 333f95a..0a71a87 100755
--- a/tools/test_fail_live_repair.sh
+++ b/tools/test_fail_live_repair.sh
@@ -53,6 +53,12 @@ for bin in $cds $crucible_test $dsc; do
     fi
 done
 
+# The jq program is required for processing the /info endpoint
+if ! jq --version > /dev/null; then
+    echo "Can't find jq program, required for this test"
+    exit 1
+fi
+
 # Verify there is not a downstairs already running.
 if pgrep -fl -U "$(id -u)" "$cds"; then
     echo "Downstairs already running" >&2
@@ -160,13 +166,7 @@ while [[ $count -le $loops ]]; do
     choice_state="undefined"
     while [[ "$choice_state" != "faulted" ]]; do
         sleep 3
-        if [[ $choice -eq 0 ]]; then
-            choice_state=$(curl -s http://127.0.0.1:7890/info | awk -F\" '{print $8}')
-        elif [[ $choice -eq 1 ]]; then
-            choice_state=$(curl -s http://127.0.0.1:7890/info | awk -F\" '{print $10}')
-        else
-            choice_state=$(curl -s http://127.0.0.1:7890/info | awk -F\" '{print $12}')
-        fi
+        choice_state=$(curl -s http://127.0.0.1:7890/info | jq -r .ds_state[$choice].type)
     done
 
     if [[ $pstop -eq 0 ]]; then
@@ -180,13 +180,7 @@ while [[ $count -le $loops ]]; do
     choice_state="undefined"
     while [[ "$choice_state" != "live_repair" ]]; do
         sleep 2
-        if [[ $choice -eq 0 ]]; then
-            choice_state=$(curl -s http://127.0.0.1:7890/info | awk -F\" '{print $8}')
-        elif [[ $choice -eq 1 ]]; then
-            choice_state=$(curl -s http://127.0.0.1:7890/info | awk -F\" '{print $10}')
-        else
-            choice_state=$(curl -s http://127.0.0.1:7890/info | awk -F\" '{print $12}')
-        fi
+        choice_state=$(curl -s http://127.0.0.1:7890/info | jq -r .ds_state[$choice].type)
     done
 
     # Give the live repair between 5 and 10 seconds to start repairing.
@@ -204,13 +198,7 @@ while [[ $count -le $loops ]]; do
     choice_state="undefined"
     while [[ "$choice_state" != "faulted" ]]; do
         sleep 3
-        if [[ $choice -eq 0 ]]; then
-            choice_state=$(curl -s http://127.0.0.1:7890/info | awk -F\" '{print $8}')
-        elif [[ $choice -eq 1 ]]; then
-            choice_state=$(curl -s http://127.0.0.1:7890/info | awk -F\" '{print $10}')
-        else
-            choice_state=$(curl -s http://127.0.0.1:7890/info | awk -F\" '{print $12}')
-        fi
+        choice_state=$(curl -s http://127.0.0.1:7890/info | jq -r .ds_state[$choice].type)
     done
 
     sleep 2
@@ -223,10 +211,11 @@ while [[ $count -le $loops ]]; do
 
     # Now wait for all downstairs to be active
     echo Now wait for all downstairs to be active | tee -a "$test_log"
-    all_state=$(curl -s http://127.0.0.1:7890/info | awk -F\" '{print $8","$10","$12}')
-    while [[ "${all_state}" != "active,active,active" ]]; do
+    all_state=$(curl -s http://127.0.0.1:7890/info | jq -r .ds_state[].type | tr '\n' ',')
+    # The trailing comma here is required
+    while [[ "${all_state}" != "active,active,active," ]]; do
         sleep 5
-        all_state=$(curl -s http://127.0.0.1:7890/info | awk -F\" '{print $8","$10","$12}')
+        all_state=$(curl -s http://127.0.0.1:7890/info | jq -r .ds_state[].type | tr '\n' ',')
     done
 
     echo All downstairs active, now stop IO test and wait for it to finish | tee -a "$test_log"
```

upstairs/src/client.rs
```diff
@@ -938,8 +922,11 @@ impl DownstairsClient {
             // We never send jobs if we're in certain inactive states
             DsState::Faulted
             | DsState::Replaced
-            | DsState::Replacing
-            | DsState::LiveRepairReady => EnqueueResult::Skip,
+            | DsState::LiveRepairReady
+            | DsState::Stopping(..) => EnqueueResult::Skip,
+            // XXX there are some `Stopping` cases which should never happen,
+            // should we panic on them (like we do for negotiation states
+            // below)?
```
jmpesp, line 929: I'd say yes but I think that should be addressed in the next PR in the series (?)

upstairs/src/downstairs.rs
```diff
@@ -2706,17 +2708,21 @@ impl Downstairs {
...
+                DsState::LiveRepairReady |
                 // If just a single IO reported failure, we will fault this
                 // downstairs and it won't yet have had a chance to move back
                 // around to LiveRepairReady yet.
-                c.state() == DsState::Faulted
-        }));
+                DsState::Faulted |
+                // It's also possible for a Downstairs to be in the process of
+                // stopping, due a fault or disconnection
+                DsState::Stopping(..) // XXX should we be more specific here?
```
jmpesp, line 2724: maybe?

also, slight nit: I think we should do away with this assert, move the comments to the match on client state below, and panic on the catch-all case. We're missing both Faulted and Stopping(..) enum variants in the below match, and I'd rather be explicit there that we don't have to do anything in those cases.

jmpesp (approved):

mkeeter:
@leftwo I applied your patch, thanks!

upstairs/src/downstairs.rs
leftwo, line 2724: Would `ClientStopReason::NegotiationFailed` be valid here?  That would be the only one I might not expect to see.

leftwo (commented):
Some simple questions for you

leftwo:
Looking at 3/3 in this series, some of the questions/comments here might be obsolete with what happens in 3/3.
If so, just note that and ignore the question/comment.

upstairs/src/client.rs
```diff
@@ -968,9 +955,7 @@ impl DownstairsClient {
             DsState::New
             | DsState::WaitActive
             | DsState::WaitQuorum
-            | DsState::Reconcile
-            | DsState::Deactivated
```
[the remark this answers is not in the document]
mkeeter: Agreed that this behavior changed slightly; I think we can ignore it here, because #1577 adds explicit logic to [bail out early](https://github.com/oxidecomputer/crucible/blob/43a1e3098befd1944f90426663dbe06bdf1e2a1c/upstairs/src/client.rs#L1273-L1283) if we receive a negotiation message while not in the `Connecting` state.

upstairs/src/client.rs
```diff
@@ -1834,16 +1839,6 @@ impl DownstairsClient {
                             NegotiationState::GetExtentVersions;
                         self.send(Message::ExtentVersionsPlease);
                     }
-                    DsState::Replacing => {
```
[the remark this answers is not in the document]
mkeeter: Also fixed in #1577 (see above)

upstairs/src/client.rs
```diff
@@ -938,8 +922,11 @@ impl DownstairsClient {
             // We never send jobs if we're in certain inactive states
             DsState::Faulted
             | DsState::Replaced
-            | DsState::Replacing
-            | DsState::LiveRepairReady => EnqueueResult::Skip,
+            | DsState::LiveRepairReady
+            | DsState::Stopping(..) => EnqueueResult::Skip,
+            // XXX there are some `Stopping` cases which should never happen,
```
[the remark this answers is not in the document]
mkeeter: The panics here are last-chance invariant maintenance.  I've split up `DsState::Stopping` into multiple different substates in 3eb259f1 ; opinions are welcome as to which should panic versus skipping the IOs (the only obvious one is `Deactivated`, which should not send further IO).

upstairs/src/client.rs
mkeeter: I tried both ways, but because it's just affecting `DsState::Stopping(..)`, it seems reasonable to do here.

leftwo (approved):
I think I just have one question left.

upstairs/src/downstairs.rs
mkeeter: Updated in c5f27dd89d9cd5d5b20a58b2990a1859d008e3f5, taking your suggestion of removing the assert in favor of building it into the `match` below.  We can't panic in the catch-all case, because we're checking the state for all 3x Downstairs (some of which may not be in live-repair), but I'm setting a flag and panicking if _none_ of the Downstairs are in the right state.

mkeeter: Also fixed in c5f27dd89d9cd5d5b20a58b2990a1859d008e3f5

leftwo:
> I think I just have one question left.

I now have zero questions left.

Issue #1571, opened 2024-11-26
Propolis panic potentially related to some crucible operations


askfongjojo:
I have a group of VMs provisioned with identical config using Terraform. Some VM stay running while the others all went into `failed` state after running for a while and then hitting propolis panic. Here is what I saw in one of the core files:
```
BRM42220014 # mdb /pool/ext/8be8c577-23ac-452e-a205-6d9c95088f61/crypt/debug/core.oxz_propolis-server_828f7793-0916-43db-924a-da37dc1be196.propolis-server.28428.1732575554
Loading modules: [ libumem.so.1 libnvpair.so.1 libc.so.1 ld.so.1 ]
> ::status
debugging core file of propolis-server (64-bit) from oxz_propolis-server_828f7793-0916-43db-924a-da37dc1be196
initial argv: /opt/oxide/propolis-server/bin/propolis-server run /var/svc/manifest/site/propo
threading model: native threads
status: process terminated by SIGABRT (Abort), pid=28428 uid=0 code=-1
> $C ! demangle
fffff5ffeb5fe770 libc.so.1`_lwp_kill+0xa()
fffff5ffeb5fe7a0 libc.so.1`raise+0x22(6)
fffff5ffeb5fe7f0 libc.so.1`abort+0x58()
fffff5ffeb5fe800 0x28b5f19()
fffff5ffeb5fe810 0x28b5f09()
fffff5ffeb5fe870 rust_panic+0xd()
fffff5ffeb5fe950 std::panicking::rust_panic_with_hook::hfbc2313a4516d465+0x8bf()
fffff5ffeb5fe990 std::panicking::begin_panic_handler::{{closure}}::h8b784919ae9f75f5+0x95()
fffff5ffeb5fe9a0 0x289c0c9()
fffff5ffeb5fe9f0 0x289ea54()
fffff5ffeb5fea40 0x28e3452()
fffff5ffeb5feaa0 0x18587ae()
fffff5ffeb5feba0 crucible::downstairs::Downstairs::enqueue::_$u7b$$u7b$closure$u7d$$u7d$::hf6ede5fa3a4ef3cb (.llvm.13013778811916355444)+0x55()
fffff5ffeb5ff010 crucible::downstairs::Downstairs::enqueue::hfc0405a64f20e3d4 (.llvm.12065988445570034316)+0x22d()
fffff5ffeb5ff190 crucible::downstairs::Downstairs::submit_flush::h46a2044513dd81a7+0x38f()
[...]

PR #1573, opened 2024-11-26
Send deactivate flush if `need_flush` is set

mkeeter:
Otherwise, deactivate can fail in a case where there are no active jobs but `need_flush` is set: it will deactivate right away (because there are no active jobs), then the automatic flush will be sent to a Downstairs in an invalid state (`Deactivated` or `New`).

This PR will hopefully fix https://github.com/oxidecomputer/crucible/issues/1571

### DIFF
--- a/integration_tests/src/lib.rs
+++ b/integration_tests/src/lib.rs
@@ -5832,7 +5832,7 @@ mod test {
 
         // Send exactly IO_CACHED_MAX_JOBS to force a Barrier to be sent.  The
         // barrier empties out the active list, so deactivation may proceed.
-        for _ in 0..10000 {
+        for _ in 0..crucible::testing::IO_CACHED_MAX_JOBS {
             let mut buf = Buffer::new(2, 512);
             volume.read(BlockIndex(0), &mut buf).await.unwrap();
         }
--- a/integration_tests/src/lib.rs
+++ b/integration_tests/src/lib.rs
@@ -5809,4 +5809,41 @@ mod test {
             }
         ));
     }
+
+    #[tokio::test]
+    async fn test_auto_flush_deactivate() {
+        let log = csl();
+        let child = TestDownstairsSet::small(false).await.unwrap();
+        let vcr = VolumeConstructionRequest::Volume {
+            id: Uuid::new_v4(),
+            block_size: 512,
+            sub_volumes: vec![VolumeConstructionRequest::Region {
+                block_size: 512,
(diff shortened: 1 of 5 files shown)


leftwo (approved):
Assuming this is what james has been testing in CA, yeah.

integration_tests/src/lib.rs
```diff
@@ -5809,4 +5809,41 @@ mod test {
...
+            }],
+            read_only_parent: None,
+        };
+
+        let volume = Volume::construct(vcr, None, log.clone()).await.unwrap();
+        volume.activate().await.unwrap();
+
+        // Send exactly IO_CACHED_MAX_JOBS to force a Barrier to be sent.  The
+        // barrier empties out the active list, so deactivation may proceed.
+        for _ in 0..10000 {
```
jmpesp, line 5835: nit: should this instead be `IO_CACHED_MAX_JOBS + N`

mkeeter: `IO_CACHED_MAX_JOBS` isn't public in the `crucible` crate, because it's (mostly) an implementation detail.  I'm ambivalent about making it public versus hard-coding it here, let me know if you have strong opinions.

mkeeter: ...and it must be an exact multiple of `IO_CACHED_MAX_JOBS`, because otherwise there would be leftover jobs in `can_deactivate_immediately` and deactivation wouldn't skip the final flush.

jmpesp: if it must be exactly that value, I'd say make it public or have a way to return it (less excited about that option)

jmpesp: alternatively, mimic what omicron does and use

```
#[cfg(any(test, feature = "testing"))]
```

to guard if it's pub or not?

mkeeter: Okay, done in 7dc0be16221d

There's not a particularly clean way to enable / disable the `pub` visibility qualifier, so I added a dedicated module that re-exports the constant.

jmpesp (approved):
I'm running this right now, but I think the test is definitive. :ship:

leftwo:
I've also been running this on the dublin racklett setup.  I (accidentally) verified I could reproduce the problem, which is good.  I'm now at 17 instances created and no issues.

leftwo:
Overnight testing of 40 more instance starts, still good.

PR #1575, opened 2024-11-26
Send all flushes through the deferred jobs queue (if present)

mkeeter:
Unclear if this actually matters, but I'm worried that flushes and barriers can skip the deferred job queue.

### DIFF
--- a/upstairs/src/upstairs.rs
+++ b/upstairs/src/upstairs.rs
@@ -211,7 +211,7 @@ pub(crate) struct Upstairs {
     /// issue a flush of its own to be sure that data is pushed to disk.  Note
     /// that this is not an indication of an ACK'd flush, just that the last IO
     /// command we put on the work queue was not a flush.
-    need_flush: bool,
+    need_flush: Option<AutoFlush>,
 
     /// Statistics for this upstairs
     ///
@@ -255,6 +255,11 @@ pub(crate) struct Upstairs {
 
     /// Thread pool for doing heavy CPU work outside the Tokio runtime
     pool: rayon::ThreadPool,
+}
+
+enum AutoFlush {
+    Needed,
+    Submitted,
 }
 
 /// Action to be taken which modifies the [`Upstairs`] state
@@ -411,7 +416,7 @@ impl Upstairs {
             guest,
             guest_dropped: false,
             ddef: rd_status,
-            need_flush: false,
+            need_flush: None,
             stats,
             counters,
             log,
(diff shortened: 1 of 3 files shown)


upstairs/src/upstairs.rs
```diff
@@ -615,36 +623,40 @@ impl Upstairs {
...
-                    self.downstairs.clients[i].state(),
-                    DsState::Deactivated | DsState::New
-                ) {
-                    debug!(self.log, "already deactivated {i}");
-                } else if self.downstairs.try_deactivate(i, &self.state) {
-                    info!(self.log, "deactivated client {i}");
-                } else {
-                    info!(self.log, "not ready to deactivate client {i}");
+            if !self.deferred_ops.is_empty() {
+                info!(self.log, "waiting for deferred ops...");
```
jmpesp, line 627: this is a behaviour change, no?

I think we should allow the guest to deactivate at any time. I imagine in a scenario where the Upstairs / Propolis is stuck, we would still want to shut down or migrate a instance, and if Propolis calls deactivate that could block those.

upstairs/src/upstairs.rs
```diff
@@ -1278,10 +1294,12 @@ impl Upstairs {
             }
             UpstairsState::Active => (),
         }
-        if !self.downstairs.can_deactivate_immediately() {
+        if self.need_flush.is_some()
+            || !self.deferred_ops.is_empty()
```
jmpesp, line 1298: same question here

jmpesp (commented):
> Unclear if this actually matters, but I'm worried that flushes and barriers can skip the deferred job queue.

I don't think I understand the concern - my mental model of the deferred job queue is that it offloads the encryption and decryption from the main loop?

mkeeter:
My concern is the following sequence of events:

- The guest submits a read, which is sent to the Downstairs.
- The guest submits a large write.  It is moved to the deferred queue for processing
- The guest submit deactivate.  In the code on `main`:
  - This checks `can_deactivate_immediately`, which returns `false` (because the read is pending)
  - Then, it calls `submit_flush`, which submits a flush _immediately_
- The read returns from each Downstairs
- That flush returns from eachDownstairs
  - The Upstairs calls `try_deactivate` on each Downstairs, which returns true
  - The Upstairs deactivates
- The deferred write encryption finishes
- The Upstairs tries to send the deferred write to the Downstairs, and panics because we're in an invalid state

However, after trying to build a unit test for this, I couldn't trigger any issues:

- The _deactivate_ message from the Guest is deferred if submitted after the write, so that implicitly puts an order on the final flush
- If instead we submit the deactivation immediately _before_ the write, then the Upstairs is marked as `Deactivating`, and so the write returns with an error immediately

In other words, this is safe because the deactivation request goes through the same queued as deferred writes, even though the final flush does **not** go through that queue.

Automatically generated flushes and barriers are still unqueued, so I need to think more about whether this could cause problems, but my immediate concerns are lessened.

jmpesp:
> Then, it calls submit_flush, which submits a flush immediately

As I understand it this flush would be submitted to each downstairs immediately, but wouldn't complete due to the job dependency on the write?

mkeeter:
> As I understand it this flush would be submitted to each downstairs immediately, but wouldn't complete due to the job dependency on the write?

I don't think so, because the write hasn't actually been enqueued (so the flush won't see it as a dependency); the write is only enqueued after the off-thread encryption finishes.  In other words, the write will end up depending on the flush, not the other way around.

PR #1584, opened 2024-12-05
Move Nexus notification to standalone task

mkeeter:
Previously, we would spawn a new `tokio::task` in the `Downstairs` to do an async notification to Nexus, iff the `notify-nexus` feature was enabled.  This is awkward for a few reasons:

- Spawning tasks in an otherwise-synchronous function can be surprising (since it implicitly requires a tokio runtime)
- It requires importing a bunch of Nexus types into the `mod downstairs`
- This code was only compiled if `notify-nexus` is enabled, which means that it's easy to have something that builds locally but fails in CI
- There's a bunch of duplicate logging, because each function notifies Nexus separately

This PR moves Nexus notifications to a separate `notify` module, which runs a single task to do these notifications in one place.  It communicates to the rest of Crucible through a queue; if the queue fills up, we log and discard the notification request (since this is best-effort).

All of this code is _always_ compiled; we check the feature (using `cfg!(feature = "notify-nexus")` to determine whether it's actually enabled.  This module is responsible for translating into Nexus types, so they don't leak into the rest of the codebase.


upstairs/src/downstairs.rs
```diff
@@ -3722,562 +3672,196 @@ impl Downstairs {
...
-                }
-            }
-        });
+        if let Some(notify) = &self.notify {
+            notify.send(NotifyRequest::ReconcileProgress {
+                upstairs_id: self.cfg.upstairs_id,
+                repair_id: reconcile_id,
+                // surely we won't have usize::MAX extents
+                current_item: current_task as i64,
+                // i am serious, and don't call me shirley
```
jmpesp, line 3842: :airplane:

upstairs/src/downstairs.rs
```diff
@@ -3722,562 +3672,196 @@ impl Downstairs {
...
-                Err(e) => {
-                    error!(
-                        log,
-                        "failed to notify Nexus of client stopped: {e}"
-                    );
-                }
-            }
-        });
+            // Spawn a task so we don't block the main loop talking to
+            // Nexus.
```
jmpesp, line 3858: rm -rf

upstairs/src/downstairs.rs
```diff
@@ -3722,562 +3672,196 @@ impl Downstairs {
...
-                continue;
-            };
+            for cid in &repair.repair_downstairs {
+                let Some(region_uuid) = self.clients[*cid].id() else {
+                    // A downstairs doesn't have an id but is being repaired...?
+                    warn!(log, "downstairs {cid} has a None id?");
+                    continue;
+                };
 
-            repairs.push(DownstairsUnderRepair {
```
jmpesp, line 3753: It's a bit of a shame to change this struct into a tuple, but I wouldn't block the PR for it

upstairs/src/notify.rs
```diff
@@ -0,0 +1,354 @@
...
+
+            Err(e) => {
+                error!(log, "failed to notify Nexus of {s}: {e}");
+            }
+        }
+    }
+    info!(log, "notify_task exiting");
+}
+
+/// Gets a Nexus client based on any IPv6 address
```
jmpesp, line 321: this comment should also specify that this is a rack-internal IPv6 address

jmpesp (approved):

upstairs/src/downstairs.rs
mkeeter: Agreed, but it _is_ intentional; I wanted to remove Nexus-flavored types from every module except `notify.rs`.  We could add a new intermediate-type, but probably not worth it...

leftwo (approved):

PR #1595, opened 2025-01-09
Update rust toolchain to 1.84

leftwo:
A bunch of just clippy and warning changes that came with the 1.84 version.

### DIFF
--- a/measure_iops/src/main.rs
+++ b/measure_iops/src/main.rs
@@ -64,7 +64,7 @@ pub fn opts() -> Result<Opt> {
 
 macro_rules! ceiling_div {
     ($a: expr, $b: expr) => {
-        ($a + ($b - 1)) / $b
+        $a.div_ceil($b)
     };
 }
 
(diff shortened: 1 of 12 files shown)


leftwo:
Fixes https://github.com/oxidecomputer/crucible/issues/1498
and https://github.com/oxidecomputer/crucible/issues/1390 (both the same root cause)

leftwo:
This should also unlock turning back on tests for VCRs with multiple sub-volumes.

measure_iops/src/main.rs
```diff
@@ -64,7 +64,7 @@ pub fn opts() -> Result<Opt> {
 
 macro_rules! ceiling_div {
```
mkeeter, line 65: ...can we just delete this macro?

mkeeter (approved):

measure_iops/src/main.rs
leftwo: Yes, done in 56068760

PR #1601, opened 2025-01-13
Reset completed work Downstairs on a `Barrier` operation

mkeeter:
A `Barrier` operation should also clear the Downstair's list of complete jobs, because it's guaranteed to be the sole dependency for all future jobs.  The current implementation is correct but unnecessarily conservative, and will accumulate arbitrary amounts of complete jobs if IO is constantly streaming without a flush.

I believe this was an oversight in #1494 – we handled it correctly in `test::complete`, but not in the actual non-test code!

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -1597,13 +1597,16 @@ impl ActiveConnection {
             // The job completed successfully, so update our stats
             dss.on_complete(&m);
 
-            // Notify the upstairs before completing work, which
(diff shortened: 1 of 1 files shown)


leftwo (approved):
Seems a simple enough change.

Won't a flush come along eventually, even if the guest does not send one?

mkeeter:
In most cases, yes – but it requires that the guest stops doing IO for 500 ms.

PR #1602, opened 2025-01-13
DTrace cleanup.

leftwo:
Update all the places where we print downstairs status able to handle the new format.  
Standardize the three letter name printed.

Use 10 for column width for next job ID
Use 6 for column width for job ID delta

For places where we print a header, print that header right away.

Updated `cmon` to use the same three letter strings as all the DTrace scripts.


mkeeter (approved):
Thanks for cleaning up my mess!

PR #1604, opened 2025-01-15
Always remove skipped jobs from dependencies

leftwo:
Here is a possible solution on handling skipped jobs for returning RO downstairs.

If there are skipped jobs for a downstairs client, we remove them. No questions asked.

### DIFF
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -373,17 +373,26 @@ impl DownstairsClient {
         mut job: IOop,
         repair_min_id: Option<JobId>,
     ) -> IOop {
-        if self.dependencies_need_cleanup() {
-            match &mut job {
-                IOop::Write { dependencies, .. }
-                | IOop::WriteUnwritten { dependencies, .. }
-                | IOop::Flush { dependencies, .. }
-                | IOop::Barrier { dependencies, .. }
-                | IOop::Read { dependencies, .. }
-                | IOop::ExtentFlushClose { dependencies, .. }
-                | IOop::ExtentLiveRepair { dependencies, .. }
-                | IOop::ExtentLiveReopen { dependencies, .. }
-                | IOop::ExtentLiveNoOp { dependencies } => {
+        match &mut job {
+            IOop::Write { dependencies, .. }
+            | IOop::WriteUnwritten { dependencies, .. }
+            | IOop::Flush { dependencies, .. }
+            | IOop::Barrier { dependencies, .. }
+            | IOop::Read { dependencies, .. }
+            | IOop::ExtentFlushClose { dependencies, .. }
+            | IOop::ExtentLiveRepair { dependencies, .. }
+            | IOop::ExtentLiveReopen { dependencies, .. }
+            | IOop::ExtentLiveNoOp { dependencies } => {
+                debug!(
+                    self.log,
+                    "{} Check/remove skipped:{:?} from deps:{:?}",
+                    ds_id,
+                    self.skipped_jobs,
+                    deps
+                );
+                deps.retain(|x| !self.skipped_jobs.contains(x));
+
+                if self.dependencies_need_cleanup() {
                     self.remove_dep_if_live_repair(
                         dependencies,
                         ds_id,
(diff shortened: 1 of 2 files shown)


mkeeter:
I'd vote to remove `dependencies_need_cleanup` and `remove_dep_if_live_repair`, which are both only ever called from this one place (and have lost most of their original meaning).  How about something like this?

```rust
    pub(crate) fn prune_deps(
        &self,
        ds_id: JobId,
        mut job: IOop,
        repair_min_id: Option<JobId>,
    ) -> IOop {
        if matches!(self.state, DsState::LiveRepair) {
            assert!(repair_min_id.is_some());
            // If our downstairs is under repair, then include any extent limit
            // sent in the IOop; otherwise, clear it out
            if let IOop::Flush { extent_limit, .. } = &mut job {
                *extent_limit = None;
            }
        } else {
            assert!(repair_min_id.is_none());
        }
        if !self.skipped_jobs.is_empty() || repair_min_id.is_some() {
            match &mut job {
                IOop::Write { dependencies, .. }
                | IOop::WriteUnwritten { dependencies, .. }
                | IOop::Flush { dependencies, .. }
                | IOop::Barrier { dependencies, .. }
                | IOop::Read { dependencies, .. }
                | IOop::ExtentFlushClose { dependencies, .. }
                | IOop::ExtentLiveRepair { dependencies, .. }
                | IOop::ExtentLiveReopen { dependencies, .. }
                | IOop::ExtentLiveNoOp { dependencies } => {
                    debug!(
                        self.log,
                        "{} Check/remove skipped:{:?}, repair min id {:?} \
                         from deps:{:?}",
                        ds_id,
                        self.skipped_jobs,
                        repair_min_id,
                        dependencies,
                    );
                    dependencies.retain(|x| {
                        !self.skipped_jobs.contains(x)
                            && repair_min_id.map(|r| *x >= r).unwrap_or(true)
                    });
                    info!(
                        self.log,
                        " {} final dependency list {:?}", ds_id, dependencies
                    );
                }
            }
        }
        job
    }
```

leftwo:
> I'd vote to remove `dependencies_need_cleanup` and `remove_dep_if_live_repair`, which are both only ever called from this one place (and have lost most of their original meaning). How about something like this?
> 
> ```rust
>     pub(crate) fn prune_deps(
>         &self,
>         ds_id: JobId,
>         mut job: IOop,
>         repair_min_id: Option<JobId>,
>     ) -> IOop {
>         if matches!(self.state, DsState::LiveRepair) {
>             assert!(repair_min_id.is_some());
>             // If our downstairs is under repair, then include any extent limit
>             // sent in the IOop; otherwise, clear it out
>             if let IOop::Flush { extent_limit, .. } = &mut job {
>                 *extent_limit = None;
>             }
>         } else {
>             assert!(repair_min_id.is_none());
>         }
> ```

I'm trying this (and the other cleanup with it) but I think we can't assert:
```
assert!(repair_min_id.is_none());
```

If the specific client is not in `LiveRepair` but the overall system is.  We set/send `repair_min_id` at the downstairs level, so all clients arrive here with it set.

mkeeter:
Ah okay; I think it's fine to only have the assertion in the positive case then.

leftwo:
Ah, another twist that I missed that we had a test for that caught.

```
                     dependencies.retain(|x| {
                        !self.skipped_jobs.contains(x)
                          && repair_min_id.map(|r| *x >= r).unwrap_or(true)
```

Because `repair_min_id` is set for all downstairs, we can't use it here to exclude dependencies because it will pull them out of the `Active` downstairs, where we actually want the IO to go out.

mkeeter:
How about clearing it if we're not in live-repair, e.g.
```rust
    pub(crate) fn prune_deps(
        &self,
        ds_id: JobId,
        mut job: IOop,
        mut repair_min_id: Option<JobId>,
    ) -> IOop {
        if matches!(self.state, DsState::LiveRepair) {
            assert!(repair_min_id.is_some());
        } else {
            repair_min_id = None;
            // If our downstairs is under repair, then include any extent limit
            // sent in the IOop; otherwise, clear it out
            if let IOop::Flush { extent_limit, .. } = &mut job {
                *extent_limit = None;
            }
        }
```
(also I think clearing `extent_limit` was in the wrong branch?)

leftwo:
> (also I think clearing extent_limit was in the wrong branch of the conditional, oops)

Yes, and I'm surprised we don't have a test for this.  I'll see about adding one.

mkeeter (approved):

PR #1606, opened 2025-01-16
Simplify mend region selection

mkeeter:
I spent a while staring at `find_source`, convincing myself that it was just picking the highest `(gen, flush, dirty)` tuple.  We can express it more compactly by leaning into `#[derive(Ord)]`.

Similarly, `find_dest` is picking clients which are not the source client, picking all such clients if the source is dirty, and clients which differ from the source otherwise.


mkeeter:
I found some more cleanup:

- Picking which extents need reconciliation was... complicated.  It's now simpler (027cdc5a43903df08e787aeda236ea866aa5bfe6).
- `RegionMetadata` previously maintained three separate `Vec<..>`, and we had to check that they were the same length whenever we used them.  Now, it has a single `Vec<ExtentMetadata>`, which is correct by construction (9ddc9a53a25af8d8e865aee166c9f4978d563e43)

upstairs/src/mend.rs
```diff
@@ -327,49 +181,20 @@ fn find_dest(
...
-     */
-    let rec = ClientData([c0, c1, c2]);
+    let out = ClientId::iter()
+        .filter(|&i| i != source && (s.dirty || c[i] != s))
+        .collect::<Vec<_>>();
 
     info!(
         log,
-        "find dest for source {} for extent at index {}", source, i
+        "find dest for source {source} for extent at index {i} => {out:?}"
```
leftwo, line 195: I think we can update this message to indicate we have found our result.  This was printing before we did the search, but now this log comes after we have found our answer.

leftwo (approved):
Changing this makes me a little nervous as it has been working, but the tests we have should have found any issues if their were some.

upstairs/src/mend.rs
```diff
@@ -177,148 +197,38 @@ fn make_repair_list(
...
- */
+/// Find the client ID which should be the source for reconciliation
+///
+/// This function chooses the source with the
+///
+/// - Highest generation number
+/// - Highest flush number (if generation numbers are equal)
+/// - Dirty bit set (if generation and flush numbers are equal)
+///
+/// If there is still a tie at the end, then the earliest source is chosen
```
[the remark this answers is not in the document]
mkeeter: Numerically lowest, I updated the comment

upstairs/src/mend.rs
mkeeter: Sure, changed to `"found dest ..."`

jmpesp (approved):

Issue #1599, opened 2025-01-10
Support activating a read only upstairs with less than 3 downstairs


leftwo:
Right now, crucible requires all three downstairs to be present before it will activate.
When we have a read only upstairs, we could activate the upstairs with less than three downstairs present, as there is no reconciliation that can happen (everyone is read only) and there is no actual work we could do, though it does provide a sanity check.

When we do a downstairs replacement on a read only upstairs, we don't do any live repair, and we don't even verify the flush/gen numbers are the same on the new replacement downstairs, so this behavior already exists (adding without checking).

We could also add a "scanner" or something similar that would (on a read only upstairs) record the flush/gen for all extents for the first downstairs that we find.  When a 2nd/3rd/replacement downstairs shows up, we could compare it with the recorded versions and panic if things don't match.

leftwo:
With 1608, this is fixed.

Issue #1593, opened 2025-01-07
Pantry region replacement can only handle a single failed region


leftwo:
When we have a disk that contains a downstairs that needs repair, and that disk is not attached to a running propolis, we will spin up an upstairs off the pantry and send that upstairs the VCR with the "fix" in it.  We then rely on the upstairs to repair the new downstairs during the initial reconciliation process that happens on activation at startup.

This process is fine when there is just one bad downstairs in a VCR, and once the pantry upstairs has completed activation, we can tell nexus that all is good and the disk is repaired.

However, we don't have a simple disk in this case.  We have a VCR with a RW sub-volume, and a tree of read only parents.
 
Consider this disk:
```
root@oxz_switch1:~# omdb db disks info bef4889d-e590-458e-81ec-b997e18f3760
HOST_SERIAL DISK_NAME INSTANCE_NAME PROPOLIS_ZONE VOLUME_ID                            DISK_STATE
-           disk-20   -             -             c9fd9286-4528-4212-be60-070178b3e093 detached
HOST_SERIAL REGION                               ZONE                                              PHYSICAL_DISK
BRM42220062 f153ee6c-910d-4628-8350-ba9022bb5158 oxz_crucible_31942354-90cc-4f03-86c8-31b5c8ab33a7 591b6235-01c7-41bc-ba37-03d3c1b1e9dc
BRM42220030 452b7095-312e-4d1a-9285-4ef344d09df4 oxz_crucible_3b82acd4-61e0-421b-921d-44b042254c69 544ed702-38fe-4314-b7c5-8de94cc19304
BRM42220030 ccc72284-7f4e-4ae6-acd3-3edbef635fff oxz_crucible_ddb565c5-7cdc-406c-9abc-b396bb5effd3 e2f72669-bb24-4f3d-bc1a-cebc7f9d2aad

VCR from volume ID c9fd9286-4528-4212-be60-070178b3e093
ID                                   BS  SUB_VOLUMES READ_ONLY_PARENT
bef4889d-e590-458e-81ec-b997e18f3760 512 1           true

SUB VOLUME 0
    ID                                   BS  BPE    EC   GEN READ_ONLY
    bef4889d-e590-458e-81ec-b997e18f3760 512 131072 3200 5   false
    [fd00:1122:3344:102::e]:19007
    [fd00:1122:3344:102::8]:19006
    [fd00:1122:3344:103::7]:19003

READ ONLY PARENT:
[...]

PR #1608, opened 2025-01-16
Allow read only activation with less than three downstairs

leftwo:
Allow a read only upstairs to activate with one a single downstairs present.                                                                          

In upstairs/src/upstairs.rs, I've added a check when a downstairs transitions to `WaitQuorum`.
If we are read-only, then we can skip reconciliation and activate the upstairs.
If we are already active (and read only), then a new downstairs can go to active.

Added some tests and a bit of additional test framework to verify                 
an upstairs can activate with only a single downstairs ready.

There are a few places where I have XXX question comments that I'm not sure about the answers.

This "fixes" the feature request in https://github.com/oxidecomputer/crucible/issues/1599
and may help with https://github.com/oxidecomputer/crucible/issues/1593

### DIFF
--- a/tools/test_read_only.sh
+++ b/tools/test_read_only.sh
@@ -37,7 +37,7 @@ if [[ -d "$testdir" ]]; then
     rm -rf "$testdir"
 fi
 
-# Store log files we want to keep in /tmp/test_read_only/.txt as this is what
+# Store log files we want to keep in /tmp/test_read_only-$user/.txt as this is what
 # buildomat will look for and archive
 test_output_dir="/tmp/test_read_only-$user"
 rm -rf "$test_output_dir" 2> /dev/null
@@ -121,7 +121,7 @@ while [[ "$loop" -lt 10 ]]; do
     "$dsc" cmd start-all | tee -a "$test_log"
 
     # In this First loop, we just turn off one downstairs and then verify our
-    # volume # with two downstairs running.
+    # volume with two downstairs running.
     for cid in {0..2}; do
 
         while "$dsc" cmd all-running | grep false > /dev/null ; do
--- a/tools/test_read_only.sh
+++ b/tools/test_read_only.sh
@@ -1,0 +1,202 @@
+#!/bin/bash
+#
+# Tests of allowing a read only Upstairs with < 3 running downstairs
+# We do an initial RW fill, and record what data we expect
+# Then, stop and restart the downstairs as read only.
+# Loop over 1 missing downstairs, activate and verify our volume.
+# Loop over 2 missing downstairs, activate and verify our volume.
+set -o pipefail
+SECONDS=0
+
+ROOT=$(cd "$(dirname "$0")/.." && pwd)
+BINDIR=${BINDIR:-$ROOT/target/debug}
+
+echo "$ROOT"
+cd "$ROOT" || (echo failed to cd "$ROOT"; exit 1)
+
+if pgrep -fl crucible-downstairs; then
+    echo 'Downstairs already running?' >&2
+    exit 1
+fi
+
+cds="$BINDIR/crucible-downstairs"
+crutest="$BINDIR/crutest"
+dsc="$BINDIR/dsc"
+for bin in $cds $crutest $dsc; do
+    if [[ ! -f "$bin" ]]; then
+        echo "Can't find crucible binary at $bin" >&2
+        exit 1
+    fi
+done
+
+# Use the id of the current user to make a unique path
+user=$(id -u -n)
+# Downstairs regions go in this directory
+testdir="/var/tmp/test_read_only-$user"
+if [[ -d "$testdir" ]]; then
+    rm -rf "$testdir"
+fi
+
+# Store log files we want to keep in /tmp/test_read_only/.txt as this is what
+# buildomat will look for and archive
+test_output_dir="/tmp/test_read_only-$user"
+rm -rf "$test_output_dir" 2> /dev/null
+mkdir -p "$test_output_dir"
+log_prefix="${test_output_dir}/test_read_only"
+fail_log="${log_prefix}_fail.txt"
+rm -f "$fail_log"
+verify_file="${log_prefix}_verify"
+rm -f "$verify_file"
+test_log="${log_prefix}_log"
+rm -f "$test_log"
+dsc_output_dir="${test_output_dir}/dsc"
+mkdir -p "$dsc_output_dir"
+dsc_output="${test_output_dir}/dsc-out.txt"
+echo "dsc output goes to $dsc_output"
+
+region_count=3
+args=()
+dsc_args=()
+dsc_create_args=()
+upstairs_key=$(openssl rand -base64 32)
+echo "Upstairs using key: $upstairs_key" | tee -a "$test_log"
+
+args+=( --key "$upstairs_key" )
+args+=( --dsc "127.0.0.1:9998" )
+dsc_create_args+=( --encrypted )
+dsc_create_args+=( --cleanup )
+dsc_args+=( --output-dir "$dsc_output_dir" )
+dsc_args+=( --ds-bin "$cds" )
+dsc_args+=( --region-dir "$testdir" )
+
+# Control-C to cleanup.
+trap ctrl_c INT
+function ctrl_c() {
+	echo "Stopping at your request" | tee -a "$test_log"
+    ${dsc} cmd shutdown
+    exit 1
+}
+
+echo "Creating $region_count downstairs regions"
+echo "${dsc}" create "${dsc_create_args[@]}" --region-count "$region_count" --extent-size 10 --extent-count 5 "${dsc_args[@]}" > "$dsc_output"
+"${dsc}" create "${dsc_create_args[@]}" --region-count "$region_count" --extent-size 10 --extent-count 5 "${dsc_args[@]}" >> "$dsc_output" 2>&1
+
+echo "Starting $region_count downstairs"
+echo "${dsc}" start "${dsc_args[@]}" --region-count "$region_count" >> "$dsc_output"
+"${dsc}" start "${dsc_args[@]}" --region-count "$region_count" >> "$dsc_output" 2>&1 &
+dsc_pid=$!
+echo "dsc started at PID: $dsc_pid"
+
+while "$dsc" cmd all-running | grep false > /dev/null ; do
+    echo "Wait for all clients to be running."
+    sleep 3
+done
+
+echo ""
+echo "Begin tests, output goes to $test_log"
+gen=1
+
+# Initial fill and creation of verify file for future use.
+echo "$crutest" generic -g $gen "${args[@]}" --stable --verify-out "$verify_file" -c 250 | tee -a "$test_log"
+if ! "$crutest" generic -g $gen "${args[@]}" --stable --verify-out "$verify_file" -c 250 >> "$test_log"; then
+    echo "Failed initial fill"
+    exit 1
+fi
+(( gen += 1 ))
+
+echo "Shutdown dsc" | tee -a "$test_log"
+"$dsc" cmd shutdown
+wait $dsc_pid
+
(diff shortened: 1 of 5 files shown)


tools/test_read_only.sh
```diff
@@ -0,0 +1,202 @@
...
+
+# Use the id of the current user to make a unique path
+user=$(id -u -n)
+# Downstairs regions go in this directory
+testdir="/var/tmp/test_read_only-$user"
+if [[ -d "$testdir" ]]; then
+    rm -rf "$testdir"
+fi
+
+# Store log files we want to keep in /tmp/test_read_only/.txt as this is what
```
mkeeter, line 40: This comment says `/tmp/test_read_only/`, but the actual directory is `/tmp/test_read_only-$user`.  I appreciate not having to fight over folder permissions on Atrium; will this be an issue on Buildomat?

tools/test_read_only.sh
```diff
@@ -0,0 +1,202 @@
...
+
+# Control-C to cleanup.
+trap ctrl_c INT
+function ctrl_c() {
+	echo "Stopping at your request" | tee -a "$test_log"
+    ${dsc} cmd shutdown
+    exit 1
+}
+
+echo "Creating $region_count downstairs regions"
```
mkeeter, line 80: Nit: it would be shorter to create + start the regions in a single `dsc` call, using `dsc start --create`

tools/test_read_only.sh
```diff
@@ -0,0 +1,202 @@
...
+    echo "Wait for all clients to be running."
+    sleep 3
+done
+
+echo ""
+echo "Begin tests, output goes to $test_log"
+gen=1
+
+# Initial fill and creation of verify file for future use.
+echo "$crutest" generic -g $gen "${args[@]}" --stable --verify-out "$verify_file" -c 250 | tee -a "$test_log"
```
mkeeter, line 100: Why does this line specifically use `tee -a`, versus `>>` for other "log the command before running it" lines?

upstairs/src/dummy_downstairs_tests.rs
```diff
@@ -2885,6 +2885,97 @@ async fn test_no_send_offline() {
...
+        handles.push(tokio::spawn(async move {
+            guest.activate().await.unwrap();
+        }));
+    }
+
+    // Connect two Downstairs
+    //
+    // We're using a VecDeque instead of FuturesOrdered so that if a future
+    // panics, we get a meaningful backtrace with a correct line number.
+    let mut fut = VecDeque::new();
```
mkeeter, line 2932: Do we get anything from running these futures in a spawned task, versus doing it in order?
```
for mut ds in [ds1, ds2] {
    ds.negotiation_start().await;
    ds.negotiation_step_extent_versions_please().await;
}
```

upstairs/src/upstairs.rs
```diff
@@ -1756,11 +1759,38 @@ impl Upstairs {
...
+                        // if we are RW
+                        if self.cfg.read_only {
+                            if self.connect_ro_region_set() {
+                                // We are the first one to arrive, so we
+                                // can't be needing any live repair yet.
+                                self.repair_check_deadline = None;
+
+                                // XXX If we are trying to replace a region that
+                                // had never joined, do we need to do anything
+                                // different here with the repair_check_deadline?
```
mkeeter, line 1772: I have a Modest Proposal:

https://github.com/oxidecomputer/crucible/pull/1610

tools/test_read_only.sh
leftwo: I'll fix the comment.

This test does not (yet) run in buildomat.  When it does, if it's like the others, it will run in its own instance so there is not any other test (or user) that it has to fight with.   So, no, it should not be an issue in buildomat.

tools/test_read_only.sh
leftwo: It is shorter, yes.  And, that's how we did it originally in these tests.  But we ran into some issues where the creation would fail or take long enough (when creating large regions) that the test would fail.  If we create first, and wait for dsc to exit we know that the creation was successful.  I just took that create first then start model and applied it everywhere we have dsc.

However, this was all before dsc got smarter and can now tell us when everything is ready, so perhaps it's time to think about a one step DSC.

It's also probably time to make some "common" area that all these scripts can use, there is a bunch of copied code that is basically the same in every test.

tools/test_read_only.sh
leftwo: It was my personal preference based on yesterdays mood on what I wanted to look at while watching the test run. :)

I, somewhat arbitrarily decided I only wanted to see the summary of what the test is doing and not the full crutest output, but I do want to log everything so there is a full record if there is a problem.

upstairs/src/dummy_downstairs_tests.rs
leftwo: I don't actually know Mr. Chesterton.  I just copied the fence from another place where I found it doing a similar thing.

upstairs/src/upstairs.rs
```diff
@@ -1962,6 +2019,60 @@ impl Upstairs {
...
+                // If we are not active yet (this is the first downstairs) then
+                // go ahead and set ourselves active.
+                info!(self.log, "Set Downstairs and Upstairs active");
+                self.downstairs.on_reconciliation_skipped(true);
+
+                // Swap out the state for UpstairsState::Active
+                let UpstairsState::GoActive(res) =
+                    std::mem::replace(&mut self.state, UpstairsState::Active)
+                else {
+                    unreachable!(); // We just matcheed!
```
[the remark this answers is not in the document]
leftwo: matcheeed!  Done

upstairs/src/upstairs.rs
leftwo: Modest proposal approved.  I'll take this comment out and that will possibly refactor things a little if I no longer need to know first vs. second/third joining (remove the `-> bool` result).

upstairs/src/dummy_downstairs_tests.rs
leftwo: Copied from: 
```
8da7be776 (Matt Keeter       2024-04-08 12:34:36 -0400  598)         // We're using a VecDeque instead of FuturesOrdered so that if a future 
```

I've also tested with just this:
```
    for ds in [&mut ds1, &mut ds2] {
        ds.negotiate_start().await;
        ds.negotiate_step_extent_versions_please().await;
    }
```
And, it works.  I did not see what happens on a panic though.

tools/test_read_only.sh
leftwo: Updated in 84ede33e

tools/test_read_only.sh
```diff
@@ -0,0 +1,202 @@
...
+echo "dsc started at PID: $dsc_pid" | tee -a "$test_log"
+
+loop=0
+while [[ "$loop" -lt 10 ]]; do
+    echo "$(date) Begin loop $loop" | tee -a "$test_log"
+    # Begin with all downstairs running
+    "$dsc" cmd start-all | tee -a "$test_log"
+
+    # In this First loop, we just turn off one downstairs and then verify our
+    # volume # with two downstairs running.
```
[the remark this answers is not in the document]
leftwo: Updated in 84ede33e

upstairs/src/dummy_downstairs_tests.rs
leftwo: Updated in 84ede33e

upstairs/src/upstairs.rs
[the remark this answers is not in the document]
leftwo: Updated in 84ede33e

upstairs/src/client.rs
```diff
@@ -854,8 +854,13 @@ impl DownstairsClient {
     ) -> EnqueueResult {
         match self.state {
             // We never send jobs if we're in certain inactive states
+            // XXX ConnectioMode::New only valid for read only upstairs, can
+            // we check this here?
```
jmpesp, line 858: Maybe? Something like

```rust
            DsState::Connecting {
                mode: ConnectionMode::Faulted | ConnectionMode::Replaced,
                ..
            }
            | DsState::Stopping(
                ClientStopReason::Fault(..)
                | ClientStopReason::Disabled
                | ClientStopReason::Replacing
                | ClientStopReason::NegotiationFailed(..),
            ) if !self.cfg.read_only => EnqueueResult::Skip,
            
            DsState::Connecting {
                mode:
                    ConnectionMode::Faulted
                    | ConnectionMode::Replaced
                    | ConnectionMode::New,
                ..
            }
            | DsState::Stopping(
                ClientStopReason::Fault(..)
                | ClientStopReason::Disabled
                | ClientStopReason::Replacing
                | ClientStopReason::NegotiationFailed(..),
            ) if self.cfg.read_only => EnqueueResult::Skip,
```

jmpesp (approved):

upstairs/src/upstairs.rs
leftwo: 1610 has landed

upstairs/src/dummy_downstairs_tests.rs
```diff
@@ -2885,6 +2885,85 @@ async fn test_no_send_offline() {
     }
 }
 
+#[tokio::test]
+async fn test_ro_activate_with_two() {
```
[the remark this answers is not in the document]
leftwo: In 792c252 I added a `with_one` and updated `with_two` so we check all possible combinations.

upstairs/src/client.rs
leftwo: I tried something, see what you think

upstairs/src/client.rs
```diff
@@ -896,7 +895,7 @@ impl DownstairsClient {
             // client back into compliance by replaying jobs).
             DsState::Active => EnqueueResult::Send,
             DsState::Connecting {
-                mode: ConnectionMode::Offline,
+                mode: ConnectionMode::Offline | ConnectionMode::New, // RO client checked above
```
jmpesp, line 898: I think this comment is saying that if we're in this match, it's a read-write config right? Because we'd have executed the first match if it was read-only?

leftwo: Yes exactly.
I wanted to help if someone noticed that `ConnectionMode::New` appeared in two places.

upstairs/src/client.rs
```diff
@@ -857,6 +857,10 @@ impl DownstairsClient {
     ) -> EnqueueResult {
         match self.state {
             // We never send jobs if we're in certain inactive states
+            DsState::Connecting {
+                mode: ConnectionMode::New,
+                ..
+            } if self.cfg.read_only => EnqueueResult::Skip,
```
mkeeter, line 863: Is this necessary?  I feel like it should bail out earlier, but am not 100% confident

mkeeter: Okay, I've convinced myself that this makes sense.  You should put a comment saying that we can be connected in read-only mode with only a single Downstairs, so the other Downstairs should skip jobs in that case.

It seems like we _could_ also pass in `UpstairsState` to distinguish here?

upstairs/src/client.rs
mkeeter: There's a behavior change here for the read-write case.

Previously, it would panic if you tried to send a message to a downstairs in `Connecting { mode: New, .. }`.  Now, it will either return `Skip` (if read-only) or `Hold` (if read-write).  I can believe that the read-only behavior change is intended; is the read-write behavior change also intentional?

leftwo: > is the read-write behavior change also intentional?

No!

I put this back in the wrong place, let me fix.

mkeeter (approved):

upstairs/src/client.rs
leftwo: Comment added in f212546c

upstairs/src/client.rs
leftwo: Fixed in f212546c

PR #1609, opened 2025-01-17
Generic read test for crutest

leftwo:
Made a generic read test for crutest, just do reads (and an occasional flush).  
It makes use of the generic test that is already present.

This is paving the way for a read only stress test with downstairs coming and going.

Added the ability for the crutest cli to call this new read only test.

### DIFF
--- a/crutest/src/main.rs
+++ b/crutest/src/main.rs
@@ -84,7 +84,7 @@ enum Workload {
         skip_verify: bool,
     },
     Generic,
-    /// Do random read IOs
+    /// Do random read and flush IOs
     GenericRead,
     Nothing,
     One,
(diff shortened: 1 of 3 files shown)


crutest/src/main.rs
```diff
@@ -84,6 +84,8 @@ enum Workload {
         skip_verify: bool,
     },
     Generic,
+    /// Do random read IOs
```
mkeeter, line 87:
```suggestion
    /// Do random read and flush IOs
```
(to distinguish between `RandRead` below)

mkeeter (approved):

crutest/src/main.rs
leftwo: Done

PR #1610, opened 2025-01-17
Make live-repair part of invariants checks

mkeeter:
Right now, the Crucible upstairs has a timer which periodically checks whether live-repair should start.  Maintaining this timer is tricky: we set it to `None` if live-repair isn't possible, or `Some(t)` if live-repair _is_ possible.  We have to manually keep the timer in sync with the rest of the system's state.

In https://github.com/oxidecomputer/crucible/pull/1608, @leftwo expresses concern about whether he's maintaining the timer correctly; it's certainly not obvious.

This PR removes the timer, moving checking for live-repair into the "invariant maintenance" category of work.  In other words, every time `Upstairs::apply` is called, it will also check whether live-repair needs to start.  This is a cheap check, and doing it every time means we don't have to think about keeping the timer correctly enabled / disabled.

There's one edge case of behavior change: we will no longer perform live-repair on two Downstairs simultaneously.  One of them will necessarily enter `LiveRepairReady` first, since events are handled in a single task; it will then immediately enter `LiveRepair` in the same call to `Upstairs::apply`.

### DIFF
--- a/upstairs/src/upstairs.rs
+++ b/upstairs/src/upstairs.rs
@@ -599,7 +599,7 @@ impl Upstairs {
         }
 
         // Check whether we need to start live-repair
-        self.on_repair_check();
+        self.check_live_repair_start();
 
         // Check whether we need to mark an offline Downstairs as faulted
         // because too many jobs have piled up.
@@ -886,7 +886,7 @@ impl Upstairs {
     /// any Downstairs from
     /// `DsState::Connecting { state:  NegotiationState::LiveRepairReady, .. }`
     /// back to [DsState::Active] without actually performing any repair.
-    pub(crate) fn on_repair_check(&mut self) {
+    pub(crate) fn check_live_repair_start(&mut self) {
         if !matches!(self.state, UpstairsState::Active) {
             return;
         }
@@ -2128,7 +2128,7 @@ pub(crate) mod test {
         to_live_repair_ready(&mut up, or_ds);
 
         // Assert that the repair started
-        up.on_repair_check();
+        up.check_live_repair_start();
         assert!(up.downstairs.live_repair_in_progress());
 
         // The first thing that should happen after we start repair_extent
@@ -3413,13 +3413,13 @@ pub(crate) mod test {
 
         // Before we are active, we have no need to repair or check for future
         // repairs.
-        up.on_repair_check();
+        up.check_live_repair_start();
         assert!(!up.downstairs.live_repair_in_progress());
 
         up.force_active().unwrap();
 
         // No need to repair or check for future repairs here either
-        up.on_repair_check();
+        up.check_live_repair_start();
         assert!(!up.downstairs.live_repair_in_progress());
 
         // No downstairs should change state.
@@ -3442,7 +3442,7 @@ pub(crate) mod test {
 
         // Force client 1 into LiveRepairReady
         to_live_repair_ready(&mut up, ClientId::new(1));
-        up.on_repair_check();
+        up.check_live_repair_start();
         assert!(up.downstairs.live_repair_in_progress());
         assert_eq!(up.ds_state(ClientId::new(1)), DsState::LiveRepair);
         assert!(up.downstairs.repair().is_some());
@@ -3462,7 +3462,7 @@ pub(crate) mod test {
         up.ds_transition(ClientId::new(1), DsState::LiveRepair);
 
         // Start the live-repair
-        up.on_repair_check();
+        up.check_live_repair_start();
         assert!(up.downstairs.live_repair_in_progress());
 
         // Pretend that DS 0 faulted then came back through to LiveRepairReady;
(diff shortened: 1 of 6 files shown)


upstairs/src/upstairs.rs
```diff
@@ -901,71 +880,53 @@ impl Upstairs {
...
     pub(crate) fn on_repair_check(&mut self) {
-        info!(self.log, "Checking if live repair is needed");
         if !matches!(self.state, UpstairsState::Active) {
-            info!(self.log, "inactive, no live repair needed");
-            self.repair_check_deadline = None;
             return;
         }
 
         if self.cfg.read_only {
-            info!(self.log, "read-only, no live repair needed");
```
leftwo, line 920: At first I was, wait, no I want my messages.
But, then I remembered this now fires every loop, so, yeah, no need to spam the logs..

upstairs/src/upstairs.rs
```diff
@@ -901,71 +880,53 @@ impl Upstairs {
...
-            info!(self.log, "No Live Repair required at this time");
-        } else if !self.downstairs.start_live_repair(
+        // Try to start live-repair, logging if it fails
+        if !self.downstairs.start_live_repair(
             &self.state,
             self.ddef.get_def().unwrap().extent_count(),
         ) {
             // It's hard to hit this condition; we need a Downstairs to be in
             // LiveRepairReady, but for no other downstairs to be in Active.
             warn!(self.log, "Could not start live repair, trying again later");
```
leftwo, line 929: How fast will this check loop?
Could this warning (which I do like to see) end up spamming the log file?

upstairs/src/upstairs.rs
```diff
@@ -1757,18 +1718,14 @@ impl Upstairs {
...
-                        }
+                        self.connect_region_set();
                     }
                     Ok(NegotiationResult::Replay) => {
                         self.downstairs.replay_jobs(client_id);
                     }
                     Ok(NegotiationResult::LiveRepair) => {
-                        // Immediately check for live-repair
-                        self.repair_check_deadline = Some(Instant::now());
+                        // We will immediately check for live-repair as part of
```
leftwo, line 1727: Would this place be good to log a message that someone wants live repair?
Meaning, we could get one message and one message only?  Verify range to target?

leftwo (approved):
Just some questions about logs

upstairs/src/upstairs.rs
mkeeter: Good idea, done in de089ab41583

upstairs/src/upstairs.rs
mkeeter: Yes, this will spam the logs if it fails.  Frequency depends on how often events are hitting `Upstairs::apply`; minimum of 1/sec for stat updates, but it could be more if the Upstairs is spamming IO.  I don't have strong opinions on the best way to handle this, any ideas?

My weak inclination is to remove all logging from the failure path; we'll see enough other information in the logs to triangulate what's going wrong in the rare case where we want to live-repair but there are no Active downstairs.

upstairs/src/upstairs.rs
mkeeter: Yup, exactly.

jmpesp:
> There's one edge case of behavior change: we will no longer perform live-repair on two Downstairs simultaneously. One of them will necessarily enter LiveRepairReady first, since events are handled in a single task; it will then immediately enter LiveRepair in the same call to Upstairs::apply.

1. Does this mean that the states would remain Active + LiveRepair + LiveRepairReady for the duration of the repair?
2. This doubles the time it takes to repair the two because they'd be done back to back. This is a bit unfortunate

mkeeter:
> 1. Does this mean that the states would remain Active + LiveRepair + LiveRepairReady for the duration of the repair?

Yes.

> 2. This doubles the time it takes to repair the two because they'd be done back to back. This is a bit unfortunate

Good catch, but after staring at the code, I don't think it's actually any different from our current situation!

Right now, [when we reach `LiveRepairReady` during negotiation](https://github.com/oxidecomputer/crucible/blob/c0b8bc44ba114fa6ef08d34b2fcd1028df73fa35/upstairs/src/upstairs.rs#L1769-L1772), we set `repair_check_deadline` to `Instant::now()`.  This means that the next call to `Upstairs::select` will almost certainly choose `UpstairsAction::RepairCheck` and begin live-repair of just that Downstairs.

The only way we'd repair two Downstairs simultaneously is if we get very lucky and they (1) complete negotiation at exactly the same time and (2) the second negotiation message wins the race with `tokio::time::sleep_until(Instant::now())`.

upstairs/src/upstairs.rs
leftwo: I agree with your weak inclination here, at least for this message.  Any message spamming the log is almost always worse, and in this case there should be enough other information (seeing DS transitions) that we can determine what happened.

mkeeter: Okay, I've removed the logging - this is now just normal invariant maintenance, so it doesn't need to be treated specially!

PR #1613, opened 2025-01-23
Tweaks to automatic flush

mkeeter:
There was previously an edge case where we'd submit automatic flushes 500ms after the last Upstairs event, _even if_ we had IO pending.  We normally don't see this in practice, because events are constantly happening (and therefore resetting the flush timer).

However, @jmpesp saw it while testing in the Canada region; we're not yet sure why!

This PR makes two changes: (1) do not send flushes if jobs are pending, even if the timeout has expired, and (2) log a specific warning message about this unusual situation.

### DIFF
--- a/upstairs/src/upstairs.rs
+++ b/upstairs/src/upstairs.rs
@@ -567,13 +567,26 @@ impl Upstairs {
                 self.on_client_message(m);
             }
             UpstairsAction::FlushCheck => {
-                self.counters.action_flush_check += 1;
-                cdt::up__action_flush_check!(|| (self
-                    .counters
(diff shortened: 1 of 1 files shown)


leftwo (approved):
I think the warning is fine for us looking for stuff.

I also wonder, if we have conditions like this that we don't expect, should we have a stat counter that we bump and remains around for us to look for later in a running system?
I could imagine this warning getting lost in the logs  on a system that has been up
for a few months.  But, if we had a counter we could quickly scan all propolis running instances and see if it had changed.

Not something for this PR, but something I'm thinking about.

Issue #1615, opened 2025-01-27
Upstairs write stats are no longer sent to oximeter.


leftwo:
I noticed on dogfood that read/flush metrics are showing updates, but write metrics are empty.

<img width="1264" alt="Image" src="https://github.com/user-attachments/assets/1b7dd3e7-da81-4be5-b1db-e2ae73aa16f7" />

Over in `ack_job()` we have this:

```
            IOop::Write { .. } => {
                cdt::gw__write__done!(|| (ds_id.0));
                self.stats.add_write(io_size as i64);
            }
```

Checking on a running propolis instance that is doing writes, I see the probe "gw__write__done" is no longer firing:

```
BRM44220011 # dtrace -n 'crucible_upstairs*:::gw-write-done { trace("fds"); }'
dtrace: description 'crucible_upstairs*:::gw-write-done ' matched 8 probes
^C
```

The read/flush probes do fire.

leftwo:
The ack for a write now happens in `compute_deferred_write()`, so it never travels through the `ack_job()` in `upstairs/src/downstairs.rs`.  So, the accounting never happens.

Also, it seems like the `gw__*__start` dtrace probes fire after a bit of work has happened already in the upstairs, so those probes may have lost some of their accuracy.  In the case of a write, we have already acked that back to the guest before we even get to the place where we set `gw__write__start`

PR #1617, opened 2025-01-28
Fix missing write stats in Oximeter.

leftwo:
Write stats were not making it to Oximeter any longer.  Sometime in the recent changes we stopped calling the method that updated both DTrace probes for write completion and calling the `add_write` that would eventually make it to Oximeter.

To do this, I made a common "stat update" function and added a call to the stat update function when a write "would have" been acked had we not fast-acked it already.

While I was here, I also updated some paths for dtrace scripts and updated column width for upstairs_count.d script.

Fixes https://github.com/oxidecomputer/crucible/issues/1615

After the changes on my bench gimlet you can see writes now:
WritesNow

### DIFF
--- a/upstairs/src/downstairs.rs
+++ b/upstairs/src/downstairs.rs
@@ -537,6 +537,14 @@ impl Downstairs {
                 _ => (),
             }
             self.ack_job(ds_id);
+        } else if ack_ready && job.work.is_write() {
+            // We already acked this job, but, we should update dtrace probes
+            Self::update_io_done_stats(
+                &self.stats,
+                job.work.clone(),
+                ds_id,
+                job.io_size(),
+            );
         }
 
         if complete {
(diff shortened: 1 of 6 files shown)


upstairs/src/downstairs.rs
```diff
@@ -558,14 +566,52 @@ impl Downstairs {
 
         // Fire DTrace probes and update stats
         let io_size = done.io_size();
-        match &done.work {
+        let work = done.work.clone();
+        Self::update_io_done_stats(&self.stats, work, ds_id, io_size);
```
leftwo, line 570: Ahh. github diff made a mess of this.

I pulled the stat update part of `ack_job()` out to a stand alone function, but
the diff thinks I changed more of `ack_job()` than I really did.  There should be
almost nothing else different

mkeeter (approved):

PR #1622, opened 2025-02-11
Select a random Nexus from DNS

jmpesp:
When a sled drops off the network, there's a delay before an expunged Nexus is no longer returned in the list of records (due to a blueprint update). Crucible will try a few times to send high priority messages but will eventually give up.

If the expunged Nexus is the first in the list of DNS records returned, this unfortunately won't change: the DNS server does not randomize the returned records. Crucible's notification will be continuously sent to the same expunged Nexus and won't arrive.

If the "started" message for a repair or reconciliation is not received by a Nexus, it will drop progress and finish messages with the same upstairs ID. This can cause stuck repairs if Nexus can't poll for a repair's progress for some reason.

### DIFF
--- a/upstairs/src/notify.rs
+++ b/upstairs/src/notify.rs
@@ -5,6 +5,7 @@
 //! Nexus-flavored types internally.
 
 use chrono::{DateTime, Utc};
+use rand::prelude::SliceRandom;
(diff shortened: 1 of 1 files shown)


leftwo (approved):
Probably in the limit we should be using quorb for these kinds of things

PR #1626, opened 2025-02-19
Fix a test flake

### LINKED ISSUE #1623: test flake in dummy_downstairs_tests::test_ro_activate_with_two
@mkeeter reported the following test flake:

cargo test --release -pcrucible --lib -- test_ro_activate_with_two

```
test dummy_downstairs_tests::test_ro_activate_with_two ... FAILED

failures:

---- dummy_downstairs_tests::test_ro_activate_with_two stdout ----
thread 'dummy_downstairs_tests::test_ro_activate_with_two' panicked at upstairs/src/dummy_downstairs_tests.rs:415:18:
saw non read Flush { upstairs_id: 3ca17ccf-3d16-46ba-b0c2-be01f2296960, session_id: d101246d-352c-434d-87cb-13ab38e199b2, job_id: JobId(1001), dependencies: [], flush_number: 0, gen_number: 1, snapshot_details: None, extent_limit: None }
```

I can also get this failure on some systems.  It appears to be a timing issue somewhere in
the test, as the panic is receiving a flush when expecting a read.

leftwo:
Avoid a race in a test of read only downstairs that can connect before or after IO is sent.  
We can do this by only expecting IOs to land on downstairs that are present before we send the IO. 

Fixes #1623

### DIFF
--- a/upstairs/src/dummy_downstairs_tests.rs
+++ b/upstairs/src/dummy_downstairs_tests.rs
@@ -2988,14 +2988,24 @@ async fn test_ro_activate_from_list(activate: [bool; 3]) {
         guest,
     };
 
-    // Get downstairs state before we send IO.
-    // If a downstairs is not present before we send IO, then we won't bother
-    // calling ack_read() for that downstairs.  It's possible the downstairs
-    // joins just after  we check, and that is okay for this test.  The race
-    // here is when a downstairs should join, and eventually will but has not
-    // joined quite yet and the IO for that downstairs is skipped, then we go
-    // on an expect it to have arrived and fail the ack_read() below.
-    let ds = harness.guest.downstairs_state().await.unwrap();
+    // Wait for all downstairs we expect to be active to actually be active.
+    // If we don't do this, it's possible for a downstairs to join after
+    // we send the IO, which will cause the read to be skipped, and we
+    // will end up acking the following flush.
+    loop {
+        tokio::time::sleep(Duration::from_secs(1)).await;
+        let ds = harness.guest.downstairs_state().await.unwrap();
+        if activate[0] && ds[ClientId::new(0)] != DsState::Active {
+            continue;
+        }
+        if activate[1] && ds[ClientId::new(1)] != DsState::Active {
+            continue;
+        }
+        if activate[2] && ds[ClientId::new(2)] != DsState::Active {
+            continue;
+        }
+        break;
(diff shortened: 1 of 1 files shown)


upstairs/src/dummy_downstairs_tests.rs
```diff
@@ -2988,21 +2988,31 @@ async fn test_ro_activate_from_list(activate: [bool; 3]) {
         guest,
     };
 
+    // Get downstairs state before we send IO.
+    // If a downstairs is not present before we send IO, then we won't bother
+    // calling ack_read() for that downstairs.  It's possible the downstairs
+    // joins just after  we check, and that is okay for this test.  The race
```
mkeeter, line 2994:
```suggestion
    // joins just after we check, and that is okay for this test.  The race
```

upstairs/src/dummy_downstairs_tests.rs
```diff
@@ -2988,21 +2988,31 @@ async fn test_ro_activate_from_list(activate: [bool; 3]) {
...
+    // here is when a downstairs should join, and eventually will but has not
+    // joined quite yet and the IO for that downstairs is skipped, then we go
+    // on an expect it to have arrived and fail the ack_read() below.
+    let ds = harness.guest.downstairs_state().await.unwrap();
+
```
mkeeter, line 2999: Is there not still a race condition here?  What if the Downstairs is not active when we call `downstairs_state`, but becomes active before spawning the read?

leftwo: Its is still a race, but if we lose this path (it becomes active afterwards) it just means we won't check for and respond to the read on that downstairs.

We must have at least one downstairs that has gone active because we `guest.query_is_active()`, so that one will be responsible for sending the ack.

Hmm. 
We could update the test to wait for as many downstairs as are "true" before proceeding with the read, that would avoid hitting the situation where we activate before the 2nd downstairs has joined.

mkeeter: >  We could update the test to wait for as many downstairs as are "true" before proceeding with the read, that would avoid hitting the situation where we activate before the 2nd downstairs has joined.

Yeah, I think it makes sense to wait until all of the expected Downstairs are active before sending the read – that way, everything is in a known state.

mkeeter (approved):

PR #1663, opened 2025-03-05
Use `RangeSet` instead of tracking every complete job

mkeeter:
(staged on top of #1662)

Right now, we store each completed `JobId` and search linearly through the list when checking for dependencies.  This means that a flush, which is dependent on all previous jobs, takes `O(N^2)` time to check its dependencies.

This PR switches to a [`RangeSet`](https://docs.rs/rangemap/latest/rangemap/set/struct.RangeSet.html), which provides `O(log N)` lookup.  In practice, we expect it to be `O(1)`: jobs are completed in order, so the `RangeSet` will only contain a single contiguous range.

I did a quick audit of the `rangemap` crate, and it seems reasonable:

- No dependencies
- `proptest` based testing, Github CI
- Consistent docstrings

There are two small behavior changes:

- `completed()` now allocates, expanding the `RangeSet` into a `Vec<JobId>` (instead of a `&[JobId]`).  It's only used in unit tests and our debug work-printing function, so I think that's fine.
- The output of `completed()` is now sorted.  This required changing one test assertion.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1098,6 +1098,7 @@ dependencies = [
  "oximeter-producer",
  "rand 0.8.5",
  "rand_chacha 0.3.1",
+ "rangemap",
  "rayon",
  "repair-client",
  "reqwest",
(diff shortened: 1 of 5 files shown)


faithanalog (approved):
if you like this library, we could probably replace ImpactedBlocks with it too.

mkeeter:
> if you like this library, we could probably replace ImpactedBlocks with it too.

Not quite – `ImpactedBlocks` stores a single block range, not a set of (possibly-discontinuous) ranges.

We _almost_ could replace `BlockMap`, except that our `ImpactedAddr` is kinda awkward (it's a tuple of `(ExtentId, BlockId)`)

leftwo (approved):

PR #1665, opened 2025-03-06
nightly test polish

leftwo:
Updates to the nightly tests and the tests it calls.

All the nightly tests have their own named directory for test logs
```
WORK_ROOT=${WORK_ROOT:-/tmp}
TEST_ROOT="$WORK_ROOT/<test_name>"
```

This makes cleanup at the end easier.
In all tests, dsc now logs to this or a specified directory.

All the nightly tests now have a `REGION_ROOT`, and this is where dsc or the crucible downstairs will create region directories.  This makes it easier for the caller to direct the tests to use a specific location instead of everyone piling in to `/var/tmp`

If a test finishes without error, it will clean up logs and regions that were created as part of the test.  If there is an error, then we leave things behind.  We clean up the contents of `REGION_ROOT` but only what each test creates itself.

Updated `tools/test_nightly.sh` to print the name of the test it is running, and just cleaned up the output a bit.


tools/hammer_loop.sh
```diff
@@ -29,6 +29,24 @@ if pgrep -fl -U "$(id -u)" "$cds"; then
...
+REGION_ROOT=${REGION_ROOT:-/var/tmp/hammer_loop}
+if [[ ! -d "$TEST_ROOT" ]]; then
+    mkdir -p "$TEST_ROOT"
+    if [[ $? -ne 0 ]]; then
+        echo "Failed to make test root $TEST_ROOT"
+        exit 1
+    fi
+else
+    # Delete previous test data
+    rm -r "$TEST_ROOT"
```
mkeeter, line 43: Should this be `rm -r "$TEST_ROOT/*"`, to avoid deleting the directory itself?

Alternatively, we could do this first if `$TEST_ROOT` exists, then **unconditionally** create the directory and check the `mkdir` return code.

tools/hammer_loop.sh
```diff
@@ -138,12 +157,21 @@ printf "[%03d] %d:%02d  ave:%d:%02d  total:%d:%02d errors:%d last_run_seconds:%d
...
     pkill -f -U "$(id -u)" "$cds"
 fi
 
+if [[ $err -eq 0 ]]; then
+    # No errors, then cleanup all our logs and the region directories.
+    rm -r "$TEST_ROOT"
+    rm -r "$REGION_ROOT"/8810
+    rm -r "$REGION_ROOT"/8820
+    rm -r "$REGION_ROOT"/8830
+    # If empty, remove the region directory
```
mkeeter, line 174: Are there cases where this wouldn't be empty?  If not, should we do `rm -r "$REGION_ROOT"` versus deleting region dirs individually?

tools/test_live_repair.sh
```diff
@@ -26,11 +26,22 @@ mkdir -p "$REGION_ROOT"
...
+TEST_ROOT="$WORK_ROOT/test_live_repair"
+if [[ ! -d "$TEST_ROOT" ]]; then
+    mkdir -p "$TEST_ROOT"
+    if [[ $? -ne 0 ]]; then
+        echo "Failed to make test root $TEST_ROOT"
+        exit 1
+    fi  
+else
+    # Delete previous test data
+    rm -r "$TEST_ROOT"
```
mkeeter, line 38: Same concern as above (presumably this will also repeat below, so I'm not going to comment on subsequent files)

tools/hammer_loop.sh
leftwo: So, what I want here was something I don't think this is going to give me, and you are 
right to call out what this half measure was doing.

What I want is for the caller to be able to supply a directory and have this (and any
other tests) use that directory as a place where the regions are created.
The problem arises if that directory is not empty.  I don't want this test to blindly
destroy whatever exists.   My half measure was to only remove directories that I
knew this tests created.

However, I now think a better solution is to, like we do with `TEST_ROOT`, is to create
a specific subdirectory inside `REGION_ROOT`, and put all region directories inside that.

This way I can always remove `$REGION_ROOT/<my_test_unique_subdir>` and not have
to care what is in that directory, or even know what exact directories exist inside it, which
will be a problem when I make these tests have multiple region sets.

I'll refactor this and all the places where we use `REGION_ROOT`

tools/hammer_loop.sh
```diff
@@ -42,7 +42,7 @@ else
     rm -r "$TEST_ROOT"
 fi
 REGION_ROOT=${REGION_ROOT:-/var/tmp}
-MY_REGION_ROOT=${REGION_ROOT/hammer_loop}
```
leftwo: This is wrong, and should be `MY_REGION_ROOT=${REGION_ROOT}/hammer_loop`
Should be fixed in newer commits.

tools/hammer_loop.sh
leftwo: I've refactored all the test logs and region directories to always be in a sub-directory, so all this should be consistent now.

tools/hammer_loop.sh
```diff
@@ -29,6 +29,34 @@ if pgrep -fl -U "$(id -u)" "$cds"; then
...
+fi
+mkdir -p "$TEST_ROOT"
+if [[ $? -ne 0 ]]; then
+    echo "Failed to make test root $TEST_ROOT"
+    exit 1
+fi
+
+REGION_ROOT=${REGION_ROOT:-/var/tmp}
+MY_REGION_ROOT=${REGION_ROOT}/hammer_loop
+if [[ ! -d "$MY_REGION_ROOT" ]]; then
```
mkeeter, line 46: Should this set of conditionals be organized in the same way as `TEST_ROOT` (delete old directory, then make new directory and check success)?  Right now, if `MY_REGION_ROOT` already exists, then we delete it and _don't_ recreate the directory.

(also applicable in many other scripts)

tools/test_live_repair.sh
```diff
@@ -21,16 +21,35 @@ function ctrl_c() {
...
+mkdir -p "$TEST_ROOT"
+if [[ $? -ne 0 ]]; then
+    echo "Failed to make test root $TEST_ROOT"
+    exit 1
+fi
 
-loop_log="$WORK_ROOT"/test_live_repair_summary.log
-test_log="$WORK_ROOT"/test_live_repair.log
-verify_log="$WORK_ROOT/test_live_repair_verify.log"
+loop_log="$TEST_ROOT"/test_live_repair_summary.log
```
mkeeter, line 49: Nit: style is inconsistent across these lines (quotes are sometimes around the entire path, sometimes around just `$TEST_ROOT`)

mkeeter (approved):

tools/hammer_loop.sh
leftwo: Yeah, that's what I had locally, just failed to push it up.

tools/test_live_repair.sh
leftwo: I went and made everything (I could find)   `new_variable="${old_variable}/new_path" ` style.

Issue #1667, opened 2025-03-12
pantry should log better what it has activated


leftwo:
When digging through pantry logs, it's difficult to know what exactly it has activated.
We could do a better job of producing logs that tie together which upstairs is logging, and when an activation happens, we should display both a session ID and the UUID of the upstairs.

PR #1670, opened 2025-03-14
Log when pantry/volume layers activate things

leftwo:
While debugging other problems, we found it difficult to determine when a pantry had activated
which part of a multi level VCR.

This adds some log messages showing when each piece of the VCR is activated.


leftwo:
Fix for https://github.com/oxidecomputer/crucible/issues/1667

jmpesp (approved):

PR #1674, opened 2025-03-17
Log loops during crutest replay

leftwo:


mkeeter (approved):
controversial, but I'll allow it

PR #1685, opened 2025-03-18
Make `ImpactedBlocks` an absolute range

mkeeter:
The current implementation of `ImpactedBlocks` stores an inclusive range of `ImpactedAddr`, which are themselves `(ExtentId, BlockOffset)` tuples.  This makes life awkward, because we don't know how to step through a range of `ImpactedBlocks` without knowing the blocks-per-extent count: is `(ExtentId(0), BlockOffset(7))` adjacent to `(ExtentId(1), BlockOffset(0))`?  Without additional information, there's no way of knowing!

This ambiguity adds a bunch of boilerplate, and specifically complicates the implementation of `ActiveJobs`.

The only time when we actually _need_ the extent ID is when doing live-repair activities, which forgo the block abstraction and care about individual extent files.  It makes much more sense to have `ImpactedBlocks` be an absolute block range (using `BlockIndex` instead of `BlockOffset`), then tweak the small number of sites which care about extents.  These sites are:

- `Downstairs::reserve_repair_ids_for_extent`
- `Downstairs::get_repair_ids`
- `Downstairs::check_repair_ids_for_range`
- `ActiveJobs::deps_for_repair`
- `ImpactedBlocks::extents`

### DIFF
--- a/upstairs/src/active_jobs.rs
+++ b/upstairs/src/active_jobs.rs
@@ -3,7 +3,10 @@
 use crucible_common::{BlockIndex, ExtentId, RegionDefinition};
 use crucible_protocol::JobId;
 
-use crate::{DownstairsIO, ExtentRepairIDs, IOop, ImpactedBlocks};
+use crate::{
+    extent_to_impacted_blocks, DownstairsIO, ExtentRepairIDs, IOop,
+    ImpactedBlocks,
+};
 use std::collections::{BTreeMap, BTreeSet};
 
 /// `ActiveJobs` tracks active jobs (and associated metadata) by job ID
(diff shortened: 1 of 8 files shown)


faithanalog (approved):

upstairs/src/active_jobs.rs
```diff
@@ -173,16 +163,14 @@ impl ActiveJobs {
         &mut self,
         repair_ids: ExtentRepairIDs,
         extent: ExtentId,
+        ddef: &RegionDefinition,
     ) -> Vec<JobId> {
+        let blocks_per_extent = ddef.extent_size().value;
         let blocks = ImpactedBlocks::InclusiveRange(
```
jmpesp, line 169: `extent_to_impacted_blocks` here?

jmpesp (approved):

upstairs/src/active_jobs.rs
mkeeter: Good catch, fixed

PR #1687, opened 2025-03-19
update omicron

iliana:


crutest/src/stats.rs
```diff
@@ -34,7 +34,7 @@ pub fn client_oximeter(
     let config = Config {
         server_info,
         registration_address: Some(registration_address),
-        request_body_max_bytes: 2048,
+        default_request_body_max_bytes: 2048,
```
leftwo, line 37: I updated dropshot to 16 and made this change in a bunch of other places.
 I wonder how these still compiled.  strange.

leftwo (approved):

crutest/src/stats.rs
iliana: I think this is from a re-exported type in Oximeter, which was on the Omicron clone still pulling in Dropshot 0.12.

PR #1693, opened 2025-04-02
Make `IOop` use absolute block indices

mkeeter:
Following up from #1685, the upstairs only cares about the block → extent mapping when checking live-repair stuff.  However, all of our `IOop` variants store `(ExtentId, BlockOffset)` tuples.

This PR makes the `IOop` more compact by using an absolute `BlockIndex` instead, then unpacks it into an `(ExtentId, BlockOffset)` only when needed for live-repair checks.

### DIFF
--- a/upstairs/src/test.rs
+++ b/upstairs/src/test.rs
@@ -473,8 +473,6 @@ pub(crate) mod up_test {
     }
 
     fn test_ddef(blocks_per_extent: u64) -> RegionDefinition {
-        // Check the send_io_live_repair for a read below extent limit,
-        // at extent limit, and above extent limit.
         let mut region_options = RegionOptions::default();
         region_options.set_block_size(512);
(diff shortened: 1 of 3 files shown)


upstairs/src/test.rs
```diff
@@ -472,69 +472,85 @@ pub(crate) mod up_test {
         Ok(())
     }
 
-    #[test]
-    fn send_io_live_repair_read() {
+    fn test_ddef(blocks_per_extent: u64) -> RegionDefinition {
         // Check the send_io_live_repair for a read below extent limit,
```
leftwo, line 476: I think this comments belongs with `send_io_live_repair_read()` test below.

leftwo (approved):

upstairs/src/test.rs
mkeeter: Good catch, fixed!

jmpesp (approved):

PR #1712, opened 2025-05-02
bump Progenitor to 0.10.0

iximeow:
my true goal is to bump Progenitor in Nexus to 0.10.0, but because Nexus uses `crucible-agent-client` from this repo in the `nexus` crate while _also_ having its own reference to `progenitor` where the types are used interchangeably, the Progenitor versions in Nexus and the rev of Crucible that Nexus depends on must be the same.

so, bump it here, then bump Crucible there. question for reviewers here: do i want to also bump Crucible in Propolis as well? meaning i'd merge this, then repeat this against Propolis, then update both in Omicron?

Progenitor 0.10.0 is interesting because it adds a new `api-version` header to requests with whatever version is in the OpenAPI spec. this is suitable for, among other things, use with Dropshot API versioning!

i did check that Dropshot acts reasonably when an unexpected `api-version` is present, so this doesn't necessitate changes to HTTP servers in Crucible.

i'd like to untangle Nexus from Crucible's Progenitor version. i *think* we can use omicron-common here, export a `ProgenitorOperationRetry`, and use the `crucible_agent_client::ProgenitorOperationRetry`. this would be marginally better than the status quo of expecting Nexus and Crucible to agree on `progenitor` version so they don't conflict on the `progenitor_client::Error` type.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -179,7 +179,7 @@ dependencies = [
  "omicron-workspace-hack",
  "proc-macro2",
  "quote",
- "syn 2.0.99",
+ "syn 2.0.101",
 ]
 
 [[package]]
@@ -214,7 +214,7 @@ checksum = "3b43422f69d8ff38f95f1b2bb76517c91589a924d1559a0e935d7c8ce0274c11"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.99",
+ "syn 2.0.101",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


leftwo (approved):

leftwo:
We usually bump crucible, then propolis, then Omicron so they all match, but they don't always need to depending on
what the actual change it.

Does Omicron complain if you just try to bump crucible and not propolis?

We (greg) also just updated propolis and omicron to latest crucible, so I don't think there are other differences besides this one.

iximeow:
Omicron does not complain with a bumped Crucible but same Propolis, and yeah if it was recently bumped that would explain why (i was a _bit_ surprised too)

Issue #1716, opened 2025-05-05
region replacement can trigger a panic in crucible


leftwo:
While running `tools/test_replace_special.sh` with two region sets (6 downstairs) I can get this
panic in crucible to hit:

```
thread 'tokio-runtime-worker' panicked at upstairs/src/client.rs:786:18:
invalid state for set_active_request: Stopping(Replacing)
```

leftwo:
Here is the play by play

We are running crutest `replace_while_reconcile` test on a volume that has two sub-volumes.
This test is trying to have a replacement request land in the middle of a reconciliation.

I'm only showing the output from crutest, the guest layer, and the specific session id for the
sub-volume that has the problem.

We start in crutest by sending the volume a request to activate, we spawn a task here because this
command will not return until the activation has completed:
```
let gc = volume.clone();
        let handle =
            tokio::spawn(async move { gc.activate_with_gen(gen).await });
```

We log a message and pause for `6` seconds to give the upstairs time to get going, here are
those logs:
```
20:59:11.246Z INFO crucible: [6] Request the upstairs activate
    task = crutest
20:59:11.246Z INFO crucible: [6] wait 6 for reconcile to start
    task = crutest
```

We see the guest layer receive the message
```
20:59:11.246Z INFO crucible: The guest has requested activation with gen:8
```

We see some messages from the other sub-volume which starts up, but before
it can come all the way online the crutest timeout hits and the test resumes

We now send call the volume `replace_downstairs()`
```
20:59:17.247Z INFO crucible: [6] activate should now be waiting false
    task = crutest
20:59:17.247Z INFO crucible: [6] Replacing DS 5:127.0.0.1:8860 with 4:127.0.0.1:8850
    task = crutest
```

We see a message that indicates our first sub-volume has activated, and that the
request to activate the second sub-volume is being sent:
```
20:59:17.247Z INFO crucible: The guest has finished waiting for activation with:8
20:59:17.247Z INFO crucible: Activated sub_volume 263b1ab7-ac98-4403-9007-99ce5522fd61
20:59:17.247Z INFO crucible: The guest has requested activation with gen:8
```

Also near the same time, we have the replacement request landing at that second sub-volume:
```
20:59:17.247Z WARN crucible: eeae0b6a-bfbb-48d8-aa1e-fa8fc36435a4 request to replace downstairs 127.0.0.1:8860 with 127.0.0.1:8850
     = downstairs
[...]

PR #1717, opened 2025-05-09
Fix panic in `set_active_request` when client is in Stopping(Replacing) state

nrempel:
Closes https://github.com/oxidecomputer/crucible/issues/1716

Don't have a ton of context on how this service is actually used in a larger context but whipped up a simple fix based on the above issue.

Instead of panicking, we ignore activation requests in the `Stopping` state and log a message.

### Changes
- Add explicit handling for the `DsState::Stopping(ClientStopReason::Replacing)` state in the `set_active_request` method
- Rather than panicking, log a message and gracefully ignore the activation request
- Add tests that verify this race condition is handled correctly

### DIFF
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -783,6 +783,37 @@ impl DownstairsClient {
                     mode: ConnectionMode::New,
                 };
             }
+            // This client is currently being stopped to be replaced. If an activation
+            // request (from `UpstairsState::GoActive`) arrives now, we cannot
+            // process it immediately and must avoid panicking.
+            //
+            // This is safe due to the following recovery path:
+            // 1. The client will complete its stopping sequence.
+            // 2. `DownstairsClient::reinitialize` will be called.
+            // 3. `reinitialize` inspects the current `UpstairsState`:
+            //    - If `UpstairsState` is `GoActive` or `Active` at the time of
+            //      `reinitialize`, then `auto_promote = true` is set for the
+            //      new connection. This ensures the client attempts to activate
+            //      upon reconnecting.
+            //    - If `UpstairsState` is not `GoActive` at that specific moment
+            //      (e.g., if this client is lagging or the `GoActive` call hasn't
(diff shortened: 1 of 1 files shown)


upstairs/src/client.rs
```diff
@@ -783,6 +783,14 @@ impl DownstairsClient {
                     mode: ConnectionMode::New,
                 };
             }
+            DsState::Stopping(reason) => {
```
leftwo, line 786: I think we want to limit this case to just this one reason:
```
DsState::Stopping(ClientStopReason::Replacing)
```
I've not fully determined that any other stopping state is okay to move forward.

leftwo: Oh, and thanks for submitting the PR!

nrempel: Good point. I made this change and also pulled in @mkeeter's suggested integration test in [this comment](https://github.com/oxidecomputer/crucible/issues/1716#issuecomment-2873255432).

I leveraged the `integration-test` feature to add the 100ms delay.

Thanks for the review!

nrempel: Oh and I added some comments as suggested. I did a bit of digging to try and understand what's going on there. Do those comments make sense?

upstairs/src/client.rs
```diff
@@ -2478,6 +2478,11 @@ impl ClientIoTask {
                     self.log,
                     "client IO task stopped before connecting: {s:?}"
                 );
+
+                // Used in integration_test_volume_inactive_replace_race
+                #[cfg(feature = "integration-tests")]
+                tokio::time::sleep(Duration::from_millis(100)).await;
```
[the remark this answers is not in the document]
nrempel: makes sense to me!

nrempel:
> Just one last request :)

done!

leftwo (approved):
This is good to go now (at least as far as what you need to do @nrempel ).
I'll figure out how to get the buildomat tests to run (our internal only CI) and
assuming that all passes I'll get it merged.

Thanks for the work and addressing our PR comments.

PR #1718, opened 2025-05-14
Add AllStopped command and endpoint for downstairs status

### LINKED ISSUE #1660: dsc cmd should have a "are all downstairs stopped" check
Just like there is a `cmd` for verifying all downstairs are started, we should
have one for "all downstairs have stopped".  This would avoid hacks where we
sleep and hope things have stopped in time.  The sleep method works most of
the time, but can make a test flake hard to catch.

nrempel:
Closes #1660

This PR introduces the `dsc cmd all-stopped` command to reliably check if all "downstairs" services are stopped. This replaces previous `sleep`-based checks in tests, reducing potential flakiness.

### Changes

*   Added `AllStopped` command to `dsc/src/client.rs` and corresponding handling.
*   Implemented the `/allstopped` GET endpoint in `dsc/src/control.rs`.
*   Added the `all_stopped` logic to `DscInfo` in `dsc/src/main.rs`, considering `Running`, `Starting`, or `Stopping` states as not stopped.
*   Updated `openapi/dsc-control.json` with the new endpoint definition.
*   Modified `tools/test_dsc.sh` to utilize the new `all-stopped` command.

All tests pass with these changes.

### DIFF
--- a/dsc/src/client.rs
+++ b/dsc/src/client.rs
@@ -10,6 +10,8 @@ use anyhow::Result;
 pub enum ClientCommand {
     /// Returns true if all downstairs are running
     AllRunning,
+    /// Returns true if all downstairs are stopped
+    AllStopped,
     /// Disable random stopping of downstairs
     DisableRandomStop,
     /// Disable auto restart on the given downstairs client ID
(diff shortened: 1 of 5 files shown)


leftwo (approved):
Sorry for the delay on this one, we got busy and I did not get time to look at it.

I was worried that the old test_dsc.sh had a retry loop that we no longer have (to check for stopped) so I ran this code in a test loop for the past few weeks to see if it ever failed to stop, and it never did.    So, I'm no longer worried about that.

PR #1720, opened 2025-05-22
Update tests to honor REGION_SETS env if provided.

leftwo:
Some tests can make use of multiple region sets.  
For those tests that do, use the env REGION_SETS if it is set, otherwise use a default. 
Updated the nightly test to print env variables on start as well.

Run a few more loops of the test_replace_special.

### DIFF
--- a/tools/test_replace_special.sh
+++ b/tools/test_replace_special.sh
@@ -64,7 +64,7 @@ region_sets=${REGION_SETS:-1}
 usage () {
     echo "Usage: $0 [-l #] [-r #]" >&2
     echo " -l loops       Number of test loops to perform (default 5)" >&2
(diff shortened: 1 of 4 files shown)


tools/test_replace_special.sh
```diff
@@ -149,6 +149,8 @@ ${dsc} cmd shutdown
 wait "$dsc_pid"
 
 echo "$(date) Test ends with $result" | tee -a "$test_log"
+cp /tmp/test_replace_special/test_replace_special.log ~
```
mkeeter, line 152: Did you mean to leave this here?

leftwo: Nope, remove in 163b9bd, thanks

mkeeter (approved):

PR #1722, opened 2025-05-22
Remove `WaitActive` state during negotiation

mkeeter:
Staged on top of #1721

The `WaitActive` state has outlived its usefulness.

Right now, there are two wait points in the Upstair's client task:

- It waits for a oneshot (`client_connect_tx/rx`) to be fired before connecting to the Downstairs
- After receiving `YesItsMe` from the downstairs, it waits for the upstairs to be activated before sending `PromoteToActive`.  This is implemented through a runtime check when it reaches the `WaitForActive` state

It turns out that these two things always happen together; there's no case where we want to connect to a Downstairs, get to `YesItsMe`, then _not_ promote it to active.

This PR removes the `WaitActive` state, moving straight from `Start` to `WaitForPromote` once the connection oneshot fires.  I believe this fixes the (hypothetical) panic proposed in #1721, although I haven't written out a unit test for it.

Most of the LOC changes are removing the increasingly-outdated block comment about negotiation; it's all nicely represented the `NegotiationState` docstring, which has been kept up to date.

### DIFF
--- a/upstairs/src/client.rs
+++ b/upstairs/src/client.rs
@@ -467,7 +467,7 @@ impl DownstairsClient {
         match &self.state {
             DsState::Connecting {
                 mode: ConnectionMode::New,
-                state: NegotiationState::Start | NegotiationState::WaitActive,
+                state: NegotiationState::Start,
             } => {
                 info!(
                     self.log,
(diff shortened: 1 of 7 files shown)


upstairs/src/client.rs
```diff
@@ -1429,10 +1267,13 @@ impl DownstairsClient {
...
-                // to promote if it's time.
-                *state = NegotiationState::WaitActive;
-                Ok(NegotiationResult::WaitActive)
+                *state = NegotiationState::WaitForPromote;
+                self.send(Message::PromoteToActive {
+                    upstairs_id: self.cfg.upstairs_id,
+                    session_id: self.cfg.session_id,
+                    gen: self.cfg.generation(),
+                });
+                Ok(NegotiationResult::NotDone)
```
jmpesp, line 1276: non blocking: 

I was about to leave a comment in the previous PR about how this return value changed from `NotDone` to `WaitActive`, but see that it's reverted here :)

Although I'm now wondering if `NotDone` can be removed, it doesn't seem like a useful variant. Though that'd likely require more work than I think.

upstairs/src/client.rs
```diff
@@ -1308,97 +1237,6 @@ impl DownstairsClient {
         up_state: &UpstairsState,
         ddef: &mut RegionDefinitionStatus,
     ) -> Result<NegotiationResult, NegotiationError> {
-        /*
```
jmpesp, line 1311: It's a little surprising to see this entire block comment get deleted, at least it could be broken up and the correct bits moved somewhere appropriate?

mkeeter: Yeah, it's tangential to the main body of the PR, but the comment was significantly out of date / incorrect.  We've basically moved all of the relevant info into the [`NegotiationState` docstring](https://github.com/oxidecomputer/crucible/blob/34a5d8722fdaaef364aed07b58f0c9dc712fe06f/upstairs/src/client.rs#L1901-L1942) instead.

leftwo (approved):
I think the new states for a downstairs have broken some of the DTrace scripts that show downstairs states.  It's probably not worth fixing them till we get through this set of PRs.

leftwo:
Here is an example, the `connecting` state has sub-states, but not the active state:

From `pfexec dtrace -s tools/dtrace/upstairs_raw.d | jq -r`

```
  "ds_state": [
    {
      "type": "connecting",
      "value": {
        "state": {
          "type": "start"
        },
        "mode": "faulted" 
      }
    },
    {
      "type": "active"
    },                                   
    { 
      "type": "active"
    }
  ]
```

jmpesp (approved):

PR #1730, opened 2025-06-23
Make Downstairs stoppable

mkeeter:
Right now, `start_downstairs` returns a `RunningDownstairs` handle that includes a `JoinHandle` for the network listener task.  It also spawns several other tasks:

- The task running `Downstairs::run`
- The repair server
- An optional Oximeter server
- Per-connection tasks (spawned on demand)

The lifetime of these tasks is unclear, and it's hard to stop them all.  Aborting the network listener task doesn't actually stop any of the other tasks!

When writing unit tests (not currently pushed), I wanted to be able to stop and restart an entire Downstairs task group.  This would make it easier to write integration tests in Rust, instead of using separate processes orchestrated by Bash.

This PR renames `RunningDownstairs` to `DownstairsClient`, and gives it the power to halt the full set of Downstairs tasks.  In addition, `start_downstairs` is moved to `DownstairsClient::spawn`, and takes a new `DownstairsClientSettings` object (instead of 7 individual arguments).

Finally, certificate info is moved into a `Option<DownstairsClientCerts>` object, providing all-or-none behavior; previously, we could panic if only some certificate were provided.  This is an internal change; `CrucibleOpts` preserves the old behavior, because it has been serialized to disk.

### DIFF
--- a/downstairs/src/admin.rs
+++ b/downstairs/src/admin.rs
@@ -73,15 +73,17 @@ pub async fn run_downstairs_for_region(
         .map_err(|e| HttpError::for_internal_error(e.to_string()))?;
 
     let handle = d.handle();
-    let _join_handle = start_downstairs(
+    let _join_handle = DownstairsClient::spawn(
         d,
-        run_params.address,
-        run_params.oximeter,
-        run_params.port,
-        run_params.rport,
-        run_params.cert_pem,
-        run_params.key_pem,
-        run_params.root_cert_pem,
+        DownstairsClientSettings {
+            address: run_params.address,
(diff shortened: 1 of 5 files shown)


leftwo (approved):
Yeah, I've wanted to do this as well, so looking forward to being able to do so from inside tests.  

This is test only though, right?  I don't see that we would ever call this from anywhere else other than tests.

mkeeter:
> This is test only though, right? I don't see that we would ever call this from anywhere else other than tests.

I pushed another commit to make `stop` only available in `test` or `feature = "integration-tests"` (since `crucible-integration-tests` is a separate crate)

PR #1732, opened 2025-06-25
Split "check reconciliation state" from "start reconciliation"

mkeeter:
(staged on top of #1731)

This is a small cleanup: previously, `collate` called `collate_inner`, then asserted that nothing changed if `collate_inner` returned an error.

However, it turns out that `collate_inner` can be split into two independent pieces:

- A fallible function taking `&self` which checks whether the state is valid and can fail (`check_region_metadata`)
- An infallible function which begins reconciliation, given state from the checks (`start_reconciliation`)

This division makes "assert nothing changed" unnecessary: the checking function can't modify `&self`, and starting reconciliation can't fail.

### DIFF
--- a/upstairs/src/downstairs.rs
+++ b/upstairs/src/downstairs.rs
@@ -273,6 +273,12 @@ impl LiveRepairData {
 }
 
 #[derive(Debug)]
+pub(crate) struct CollateData {
+    max_flush: u64,
+    max_gen: u64,
+}
+
+#[derive(Debug)]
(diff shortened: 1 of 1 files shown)


leftwo (approved):
I was on the fence about the name of `start_reconciliation()` as we are just checking to see if we need to do reconciliation.  But, really, that is part of the reconciliation process, to first determine if additional work is what we need to do, so I think I'm fine with the name.

PR #1733, opened 2025-06-25
Min-quorum reconciliation

### LINKED ISSUE #1690: Support activation with 2/3 read/write downstairs present.
In the event that an instance is started and only 2 of the 3 downstairs are available, we should support activation with a way to allow the third downstairs to join at a later date.

See also discussion around solutions in https://github.com/oxidecomputer/omicron/issues/7826
As well as [RFD 564](https://rfd.shared.oxide.computer/rfd/0542)
PR for that [here](https://github.com/oxidecomputer/rfd/pull/875/files#top)

mkeeter:
(staged on #1732)

This implements [RFD 542](https://rfd.shared.oxide.computer/rfd/542) and closes #1690.

Here's the quick version:

- Once we have two downstairs in `WaitQuorum`, we schedule an event to fire after `NEGOTIATION_DELAY` (currently 500 ms)
- If the third Downstairs arrives before this event fires, then we do full-quorum reconciliation (our usual path)
- Otherwise, then we enter min-quorum reconciliation, marking the third Downstairs as faulted (so it must rejoin through live-repair)

I'm opening this as a draft because I want to see how the CI tests go.  The PR includes integration tests for common and uncommon orderings of events, but it's hard to hit every possible path due to specific timing requirements; I'm very open to suggestions for other tests.

Before merging, we also need to figure out https://github.com/oxidecomputer/crucible/pull/1661#issuecomment-2836793959

### DIFF
--- a/crutest/src/main.rs
+++ b/crutest/src/main.rs
@@ -2516,9 +2516,17 @@ async fn replace_before_active(
     // that the initial downstairs are all synced up on the same flush and
     // generation numbers.
     fill_workload(volume, di, true).await?;
+
+    // Track which SocketAddr corresponds to which region.  This shifts over
+    // time as the test runs, so we have to track it correctly disable 2
+    // downstairs for a given region.
+    let mut regions = vec![];
+    for i in 0..targets.len() - 1 {
+        regions.push(Some(i as u32 / 3));
+    }
+    regions.push(None);
     let ds_total = targets.len() - 1;
-    let mut old_ds_a = 0;
-    let mut old_ds_b = 1;
+    let mut old_ds = 0;
     let mut new_ds = targets.len() - 1;
     for c in 1.. {
         info!(log, "[{c}] Touch every extent");
@@ -2534,11 +2542,24 @@ async fn replace_before_active(
             tokio::time::sleep(tokio::time::Duration::from_secs(4)).await;
         }
 
-        // Stop two downstairs, wait for dsc to confirm they are stopped.
-        for old_ds in [old_ds_a, old_ds_b] {
-            dsc_client.dsc_stop(old_ds).await.unwrap();
+        // Pick a second downstairs that's in the same region, so we can stop
+        // two downstairs and prevent activation.  This is linear-time with the
+        // number of targets, but that's fine (so is writing to every block).
+        assert!(regions[new_ds].is_none());
+        let region = regions[old_ds].unwrap();
+        let (other_ds, _) = regions
+            .iter()
+            .enumerate()
+            .find(|(i, d)| *i != old_ds && **d == Some(region))
+            .unwrap();
+
+        // Stop two downstairs in the same region, then wait for dsc to confirm
+        // they are stopped.  Having two downstairs stopped blocks activation.
+        for old_ds in [old_ds, other_ds] {
+            dsc_client.dsc_stop(old_ds as u32).await.unwrap();
             loop {
-                let res = dsc_client.dsc_get_ds_state(old_ds).await.unwrap();
+                let res =
+                    dsc_client.dsc_get_ds_state(old_ds as u32).await.unwrap();
                 let state = res.into_inner();
                 if state == DownstairsState::Exit {
                     break;
@@ -2564,14 +2585,14 @@ async fn replace_before_active(
 
         info!(
             log,
-            "[{c}] Replacing DS {old_ds_a}:{} with {new_ds}:{}",
-            targets[old_ds_a as usize],
+            "[{c}] Replacing DS {old_ds}:{} with {new_ds}:{}",
+            targets[old_ds],
             targets[new_ds],
         );
         match volume
             .replace_downstairs(
                 Uuid::new_v4(),
-                targets[old_ds_a as usize],
+                targets[old_ds],
                 targets[new_ds],
             )
             .await
@@ -2610,8 +2631,8 @@ async fn replace_before_active(
 
         // Start up all the stopped downstairs so they are ready for the next
         // loop.
-        for old_ds in [old_ds_a, old_ds_b] {
-            let res = dsc_client.dsc_start(old_ds).await;
+        for old_ds in [old_ds, other_ds] {
+            let res = dsc_client.dsc_start(old_ds as u32).await;
             info!(log, "[{c}] Replay: started {old_ds}, returned:{:?}", res);
         }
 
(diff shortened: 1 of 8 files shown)


mkeeter:
Quick notes on failing CI tests:

## `test-repair`
```
EXT  BLOCKS GEN0 GEN1 GEN2  FL0 FL1 FL2  D0 D1 D2 DIFF
  0 000-019    5    5    2   10  10   5   F  F  T <---
  1 020-039    5    5    1   10  10   1   F  F  T <---
  2 040-059    5    5    1   10  10   1   F  F  T <---
  3 060-079    1    1    1    1   1   1   F  F  F
  4 080-099    4    4    4    9   9   9   F  F  F
  5 100-119    5    5    1   10  10   1   F  F  T <---
  6 120-139    5    5    1   10  10   1   F  F  T <---
  7 140-159    5    5    2   10  10   6   F  F  T <---
  8 160-179    2    2    2    3   3   3   F  F  F
  9 180-199    5    5    1   10  10   1   F  F  T <---
 10 200-219    5    5    1   10  10   1   F  F  T <---
 11 220-239    5    5    4   10  10   9   F  F  T <---
 12 240-259    5    5    1   10  10   1   F  F  T <---
 13 260-279    5    5    1   10  10   1   F  F  T <---
 14 280-299    5    5    1   10  10   1   F  F  T <---
 15 300-319    5    5    1   10  10   1   F  F  T <---
 16 320-339    5    5    2   10  10   6   F  F  T <---
 17 340-359    1    1    1    1   1   1   F  F  F
 18 360-379    5    5    1   10  10   1   F  F  T <---
 19 380-399    5    5    2   10  10   6   F  F  T <---
 20 400-419    2    2    2    4   4   4   F  F  F
 21 420-439    1    1    1    1   1   1   F  F  F
 22 440-459    1    1    1    1   1   1   F  F  F
 23 460-479    1    1    1    1   1   1   F  F  F
 24 480-499    5    5    2   10  10   6   F  F  T <---
 25 500-519    5    5    2   10  10   6   F  F  T <---
 26 520-539    5    5    1   10  10   1   F  F  T <---
 27 540-559    3    3    3    7   7   7   F  F  F
 28 560-579    1    1    1    1   1   1   F  F  F
 29 580-599    2    2    2    3   3   3   F  F  F
Max gen: 5,  Max flush: 10
Error: Difference in extent metadata found!
```

This is probably because it's easier to leave the Downstairs in mismatched states, e.g. activation no longer guarantees that they're all identical.

## `test-up-*`
```
thread 'main' panicked at crutest/src/main.rs:2559:9:
assertion failed: !is_active
{"msg":"[1] activate should now be waiting true","v":0,"name":"crucible","level":30,"time":"2025-06-25T19:31:21.840999142Z","hostname":"w-01JYM8FQQ54H9R8BDCCJ9EAVVD","pid":1304,"task":"crutest"}
Failed crutest replace-before-active
```

`crutest` assumes that activation will wait until all 3x Downstairs are available, which is no longer true

leftwo (commented):
A few comments to start, I'm still looking though.

crutest/src/main.rs
```diff
@@ -2622,12 +2631,13 @@ async fn replace_before_active(
             tokio::time::sleep(tokio::time::Duration::from_secs(4)).await;
         }
 
-        old_ds = (old_ds + 1) % (ds_total as u32 + 1);
+        old_ds_a = (old_ds_a + 1) % (ds_total as u32 + 1);
```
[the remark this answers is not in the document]
mkeeter: Good question!  I notice a CI failure in [`test-up-2region-encrypted`](https://github.com/oxidecomputer/crucible/pull/1733/checks?check_run_id=45076504047), so I suspect that you're correct 😄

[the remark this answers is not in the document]
mkeeter: I think this is fixed in d2d168e9436f.

It's somewhat subtle: the unused downstairs is always right before the downstairs we're about to replace, but the _second_ downstairs in that set could either be before or after (and shifts over time).  That commit tracks which downstairs is in which region set, so that we can always disable 2 in the same region set (and prevent activation, which is the whole point).

upstairs/src/upstairs.rs
```diff
@@ -1858,7 +1885,7 @@ impl Upstairs {
         self.on_reconciliation_skipped()
     }
 
-    /// Checks whether we can connect all three regions
+    /// Response to a client entering `NegotiationState::WaitQuorum`
     ///
     /// Returns `false` if we aren't ready, or if things failed.  If there's a
     /// failure, then we also update the client state.
```
jmpesp, line 1891: I'd vote for returning `Result<bool, some error>` instead - having `false` cover both cases seems wrong

upstairs/src/upstairs.rs
```diff
@@ -1988,6 +2032,76 @@ impl Upstairs {
...
+        else {
+            warn!(
+                self.log,
+                "min-quorum negotiation found upstairs state {:?}; cancelling",
+                self.state
+            );
+            return;
+        };
+        assert!(min_quorum_deadline.is_some());
+        *min_quorum_deadline = None;
```
jmpesp, line 2051: I think clearing `min_quorum_deadline` here means that if `ready_count` != 2 (checked below), control will leave this function and not attempt min quorum activation again?

upstairs/src/downstairs.rs
```diff
@@ -1943,13 +1961,37 @@ impl Downstairs {
...
-        for c in self.clients.iter_mut() {
-            c.set_active();
+        for (i, c) in self.clients.iter_mut().enumerate() {
+            match c.state() {
+                DsState::Connecting {
+                    state: NegotiationState::WaitQuorum,
+                    ..
+                } => {
+                    assert!(!did_work);
+                    c.set_active();
```
jmpesp, line 1980: not faulted?

jmpesp (commented):
Some comments, and also: there should be some additional downstairs tests for the non-"all clients are participating" cases.

upstairs/src/upstairs.rs
mkeeter: Counterproposal: let's remove the return value entirely!  It's only ever used in one unit test, so I've updated that test to look at states instead.

﻿fb2b63285050

upstairs/src/upstairs.rs
mkeeter: I agree, but I think this is fine: next time we end up with 2 ready Downstairs, then we'll reschedule `min_quorum_deadline`.

mkeeter:
closing in favor of #1908

PR #1761, opened 2025-08-07
Add back job delays in the downstairs with the --lossy flag

leftwo:
Somewhere in the past we lost the ability for the `--lossy` flag in the downstairs to make some jobs take longer.  The test test_repair.sh was counting on this behavior.  Things still generally behaved the same, but the volume of repair that the test would do was greatly reduced.  This change adds back the delay and will result in whatever downstairs has the lossy flag falling much further behind the other downstairs, which is what we want for that test.

### DIFF
--- a/downstairs/src/lib.rs
+++ b/downstairs/src/lib.rs
@@ -1617,6 +1617,10 @@ impl ActiveConnection {
     ) -> Message {
         let upstairs_connection = self.upstairs_connection;
 
+        if flags.lossy && random() && random() {
+            info!(self.log, "lossy pause {:?}", job_id);
+            tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+        }
         match &work {
             IOop::Read {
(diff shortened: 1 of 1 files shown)


jmpesp (approved):

downstairs/src/lib.rs
```diff
@@ -1617,6 +1617,10 @@ impl ActiveConnection {
     ) -> Message {
         let upstairs_connection = self.upstairs_connection;
 
+        if flags.lossy && random() && random() {
```
mkeeter, line 1620: Nit: the `random() && random()` is non-obvious, maybe something like
```suggestion
        if flags.lossy && random_range(0..4) == 0 { // 25% failure rate
```

mkeeter (approved):
We are stretching the meaning of the word "lossy", but I guess it's baked into the serialized-to-disk VCRs at this point...

PR #1768, opened 2025-08-28
[crucible-downstairs] migrate to API traits

sunshowers:
Migrate the downstairs admin and repair APIs to being API traits.

Part of https://github.com/oxidecomputer/omicron/issues/8922.


downstairs-types/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+use std::net::{IpAddr, SocketAddr};
+use std::path::PathBuf;
+
+use schemars::JsonSchema;
+use serde::Deserialize;
+
+// Repair API types
+#[derive(Deserialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct Eid {
```
jmpesp, line 12: Is this needed any more? It looks like it was replaced with `ExtentPath` in `extent_repair_ready` and `get_files_for_extent`?

jmpesp (changes requested):
Some questions:

downstairs-types/src/lib.rs
sunshowers: Ah yeah -- would you prefer I rename this back to `Eid` or keep it as `ExtentPath`?

jmpesp: Either works, whatever matches our convention - `ExtentPath` has less of a naming collision than `Eid` haha.

Cargo.toml
```diff
@@ -136,6 +139,9 @@ crucible-control-client = { path = "./control-client" }
 # importantly, don't use features = ["zfs_snapshot"] here, this will cause
 # cleanup issues in the integration tests!
 crucible-downstairs = { path = "./downstairs" }
+crucible-downstairs-admin-api = { path = "./downstairs-admin-api" }
+crucible-downstairs-repair-api = { path = "./downstairs-repair-api" }
```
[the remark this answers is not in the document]
sunshowers: > It might be an oversight to not have checked in the admin document, but: if both traits are combined that'll mean that the combined document should cover both right?

So yes, but at the moment it isn't possible to combine multiple traits into one document -- this is work that we deferred, there isn't anything fundamental about traits that makes them impossible to combine.

jmpesp (approved):

PR #1774, opened 2025-09-27
migrate Crucible APIs to Dropshot API manager

sunshowers:
Migrate the three Crucible APIs that are part of the Omicron service graph to
the Dropshot API manager:

* crucible-agent
* crucible-pantry
* downstairs-repair

Also add an xtask framework so you can run `cargo xtask openapi`.


jmpesp:
does the Dropshot API manager also perform the openapi linting checks?

sunshowers:
Oh I missed that, whoops! The API manager doesn't do it by default but can be configured to do it. Will put up a PR shortly.

sunshowers:
Readded validation -- should work now.

jmpesp (approved):

Issue #1776, opened 2025-09-29
IO can get stuck if a downstairs goes offline during LiveRepair


leftwo:
We no longer have a time based dismissal of a downstairs.  We now will only dismiss a downstairs if enough IO has accrued that we have decided we don't want to hold any more in memory, and so replay is no longer an option.

This process runs into trouble if we are in the middle of a LiveRepair.  Live Repair IOs require all three downstairs to go through the IO at the same time, if a single downstairs is present when the LR starts, but goes missing, we will wait forever for it to return so we can complete the repair IO.

This is tangentially related somewhat to https://github.com/oxidecomputer/crucible/issues/1194, where we should not even try to replay IOs to downstairs if a LiveRepair has happened.

It's easy enough to reproduce this hang.

Start IO to all downstairs.
Take downstairs 0 offline, wait for enough IO that it goes to faulted.
Bring downstairs 0 back, let LR start.
Once LR is started, stop downstairs 1 or 2.

IOs will now stop, as we are now going to wait for an IO from the offline downstairs.

Issue #1556, opened 2024-11-08
If three downstairs have all faulted, the upstairs can't self recover


leftwo:
After the upstairs is active, if all three downstairs go to faulted, we have no path back to a working upstairs.
The current path back to working requires having a good downstairs to `LiveRepair` from.

In this situation, we need to revert back to reconciliation and use that to return ourselves to service.

If you have an instance in this state, stopping it, then starting it will cause it to create a new propolis instance and that will go through reconciliation.

Issue #1194, opened 2024-03-07
Don't try to replay when in LiveRepair


leftwo:
If we are in LiveRepair, we need to handle `Replay` differently.
I.e. I think we can't replay, as that requires all the magic for lock step stuff, and all the deps are going to be.. weird.

Here is a panic seen while in LR, then pstop'ing a downstairs not in LR, then letting it come back after it has moved
to `Offline`

```
Mar 07 19:55:12.790 INFO Returning UUID:12345678-0000-0000-0000-000000018820 matches, client: 1, : downstairs, session_id: ed3a306d-263f-4433-b910-7014a50e9f72
Mar 07 19:55:12.791 INFO send last flush ID to this DS: 3139936, client: 1, : downstairs, session_id: ed3a306d-263f-4433-b910-7014a50e9f72
Mar 07 19:55:12.791 INFO Replied this last flush ID: 3139936, client: 1, : downstairs, session_id: ed3a306d-263f-4433-b910-7014a50e9f72
Mar 07 19:55:12.791 INFO ds_transition from Offline to Replay, client: 1, : downstairs, thread 'session_idtokio-runtime-worker: ed3a306d-263f-4433-b910-7014a50e9f72
Mar 07 19:55:12.791' panicked at  upstairs/src/client.rs:INFO1461: [1] Transition from Offline to Replay21:
, [1] Bad job received in process_ds_completion: DownstairsIO { ds_id: JobId(3139937), guest_id: GuestWorkId(3138938), work: ExtentFlushClose { dependencies: [JobId(3139936)], extent: 5, flush_number: 1630, gen_number: 2, repair_downstairs: [ClientId(2)] }, state: ClientData([Done, Done, Done]), acked: true, replay: true, data: None, read_response_hashes: [] }client:
 note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
1, : downstairs, session_id: ed3a306d-263f-4433-b910-7014a50e9f72
Mar 07 19:55:12.791 Abort
```

PR #1777, opened 2025-10-01
Offline downstairs go to faulted when LR starts

leftwo:
If a downstairs goes offline before, or during a Live Repair, we must not allow any replay of IO to that downstairs.  The Live Repair process requires all downstairs to go through specific steps at the same time to be sure all three will agree, if there are repair IOs outstanding, the will be blocked while the missing downstairs is away, and all IOs will come to a halt backing up behind them.

To handle this, we move any downstairs that is Offline to faulted and have it come back through LiveRepair once it returns.

Added a new tests to crutest that will allow IO to keep flowing even if there is an error.  It also does a better job of tracking what we expect to have written to each block.

Added two tests to check that offline downstairs are handled correctly during Live Repair situations.

### DIFF
--- a/crutest/src/main.rs
+++ b/crutest/src/main.rs
@@ -139,6 +139,8 @@ enum Workload {
     /// Select a random offset/length, then Write/Flush/Read that
     /// offset/length.
     WFR,
+    /// Do reads and writes, keep going if there are errors.
+    Yolo,
 }
 
 #[derive(Debug, Parser)]
@@ -204,7 +206,7 @@ pub struct Opt {
     #[clap(short, global = true, long, action, conflicts_with = "stable")]
     quit: bool,
 
-    /// For the verify test, if this option is included we will allow
+    /// For the verify and yolo tests, if this option is included we will allow
     /// the write log range of data to pass the verify_volume check.
     #[clap(long, global = true, action)]
     range: bool,
@@ -514,7 +516,12 @@ impl WriteLog {
     pub fn update_wc(&mut self, index: usize) {
         assert!(self.count_cur[index] >= self.count_min[index]);
         // TODO: handle more than u32 max writes to the same location.
-        self.count_cur[index] += 1;
+        self.count_cur[index] = self.count_cur[index].wrapping_add(1);
+    }
+    pub fn undo_update_wc(&mut self, index: usize) {
+        assert!(self.count_cur[index] >= self.count_min[index]);
+        // TODO: handle more than u32 max writes to the same location.
+        self.count_cur[index] = self.count_cur[index].wrapping_sub(1);
     }
 
     // This returns the value we should expect to find at the given index,
@@ -604,6 +611,7 @@ impl WriteLog {
             let shift = self.count_min[index] / 256;
 
             let s_value = value as u32 + (256 * shift);
+            /*
             println!(
                 "Shift {}, v:{} sv:{} min:{} cur:{}",
                 shift,
@@ -612,7 +620,7 @@ impl WriteLog {
                 self.count_min[index],
                 self.count_cur[index],
             );
-
+            */
             res = s_value >= self.count_min[index]
                 && s_value <= self.count_cur[index];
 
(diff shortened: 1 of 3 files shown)


leftwo:
Fix for: https://github.com/oxidecomputer/crucible/issues/1776
Fix for: https://github.com/oxidecomputer/crucible/issues/1194

This work will enable a fix for: https://github.com/oxidecomputer/crucible/issues/1556

leftwo:
We hit this during upgrade testing on dublin.

crutest/src/main.rs
```diff
@@ -2223,6 +2245,163 @@ async fn generic_workload(
...
+
+        if op == 0 {
+            match volume.flush(None).await {
+                Ok(_) => {
+                    flushes += 1;
+                    di.write_log.commit();
+                }
+                Err(e) => {
+                    flush_errors += 1;
+                    println!("Flush error: {:?}", e);
```
mkeeter, line 2296: Elsewhere in this function, we use `eprintln!` for errors

crutest/src/main.rs
```diff
@@ -2223,6 +2245,163 @@ async fn generic_workload(
...
+                di.volume_info.block_size,
+            );
+            assert_eq!(data[1], di.write_log.get_seed(block_index));
+            match volume.write(offset, data).await {
+                Ok(_) => {
+                    writes += 1;
+                }
+                Err(e) => {
+                    write_errors += 1;
+                    println!("Write error: {:?}", e);
```
mkeeter, line 2318: (same here, should this be `eprintln!`?)

mkeeter:
## Architectural vibes

Right now, we want to avoid a situation where we're performing live-repair and there's a downstairs that's `Offline`.  The PR implements this in two ways:

- When we begin live-repair, we move any other `Offline` downstairs to `Faulted`
- Whenever the main event loop processes an event, if live-repair is happening, we move `Offline` downstairs to `Faulted` (this is added to `check_gone_too_long`)

The first one seems reasonable, but I'm wondering if we could avoid the second chunk of logic by **never** moving a Downstairs to `Offline` if live-repair is in progress.  I think this could be done with a one-line change:
```diff
--- a/upstairs/src/downstairs.rs
+++ b/upstairs/src/downstairs.rs
@@ -680,7 +680,8 @@

         // Restart the IO task for that specific client, transitioning to a new
         // state.
-        self.clients[client_id].reinitialize(up_state, self.can_replay);
+        let can_replay = self.can_replay && !self.live_repair_in_progress();
+        self.clients[client_id].reinitialize(up_state, can_replay);

         // If the IO task stops on its own, then under certain circumstances,
         // we want to skip all of its jobs.  (If we requested that the IO task
```

(then the changes in `check_gone_too_long` are not necessary)

This requires changing `test_downstairs_goes_offline_during_live_repair` to deliver a disconnection event, instead of poking the internal state of the system directly:

```rust
        // Now simulate client 2 spontaneously going offline during live repair
        // This is what would happen when a connection drops unexpectedly
        up.apply_client_action(
            ClientId::new(2),
            ClientAction::TaskStopped(ClientRunResult::Timeout),
        );

        // After check_gone_too_long, client 2 should be faulted (not offline)
        assert_eq!(
            up.ds_state(ClientId::new(2)),
            DsState::Connecting {
                state: NegotiationState::Start,
                mode: ConnectionMode::Faulted
            }
        );
```

upstairs/src/downstairs.rs
```diff
@@ -71,7 +71,9 @@ pub(crate) struct Downstairs {
     /// Indicates whether we are eligible for replay
     ///
     /// We are only eligible for replay if all jobs since the last flush are
-    /// buffered (i.e. none have been retired by a `Barrier` operation).
+    /// buffered (i.e. none have been retired by a `Barrier` operation), or we
```
mkeeter, line 74: I'd suggest leaving this comment as-is; we don't actually change the value of `self.can_replay` when live-repair starts, so the new comment is misleading.  You could move the new part of the comment below (before calling `reinitialize`), e.g. something like
```rust
        // Restart the IO task for that specific client, transitioning to a new
        // state.  We can replay as long as all jobs since the last flush are
        // buffered (indicated by `self.can_replay`), and there's no live-repair
        // in progress.
        let can_replay = self.can_replay && !self.live_repair_in_progress();
        self.clients[client_id].reinitialize(up_state, can_replay);
```

crutest/src/main.rs
```diff
@@ -2223,6 +2245,163 @@ async fn generic_workload(
...
+    let mut reads = 0;
+    let mut writes = 0;
+    let mut flushes = 0;
+    let mut read_errors = 0;
+    let mut write_errors = 0;
+    let mut verify_errors = 0;
+    let mut flush_errors = 0;
+    let mut last_print = Instant::now();
+    for c in 1.. {
+        let op = rng.random_range(0..100);
```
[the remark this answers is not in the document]
leftwo: done

Issue #1788, opened 2025-10-11
Upstairs panics with decryption failure: DecryptionError resulting from a bad extent


leftwo:
On an upgrade of dogfood, we found a pile of propolis cores left behind as sleds rebooted.
All of the cores had this same stack backtrace:

```
fffff6ffef08ef62 raise (6) + 22
 fffff6ffef068628 abort () + 58
 0000000002bf8179 ???????? ()
 0000000002bf8169 ???????? ()
 0000000002be29dd __rustc::rust_panic () + d
 0000000002be2882 std::panicking::rust_panic_with_hook::hb8223ccd7c528a7f () + 22f
 0000000002be24a8 std::panicking::begin_panic_handler::{{closure}}::hcec802fa388f0165 () + 98
 0000000002bde1f9 ???????? ()
 0000000002be213c ???????? ()
 0000000002c24bdf ???????? ()
 000000000187e55d crucible::client::DownstairsClient::process_io_completion::h774e9f695bd028d9 () + 96d
 00000000017919fc crucible::downstairs::Downstairs::process_io_completion::hfd4beb5a9673e22c () + f1c
 0000000001838dcf crucible::upstairs::Upstairs::on_client_message::h0d05f243f21facc8 () + bf
 0000000001833afd crucible::upstairs::Upstairs::apply::h57a5c23e3793cdae () + 29d
 00000000018cfea2 crucible::up_main::_$u7b$$u7b$closure$u7d$$u7d$::h150f4fa6ce5c8862 (.llvm.8401839033468049112) () + 5a2
 00000000018c19d5 tokio::runtime::task::core::Core<T,S>::poll::h8198b5a1284cb828 () + 35
 00000000018a77eb tokio::runtime::task::harness::Harness<T,S>::poll::h21f4d96260aa377a () + 5b
```

leftwo:
After the 2nd upgrade was done, we stopped all VMs and ran verify on all the downstairs.  Here are the results:

No problems found on cubby 17 (the first sled we upgrade)

**Cubby 14 BRM42220081**
Zone oxz_crucible_6685867d-de03-45fc-84e0-907fe19feab4
Region 27332c8d-cfd9-4799-a61c-5b4145267adb at Thu Oct 16 19:37:29 UTC 2025 with 1600 extents
validation failed for extent 397: GenericError("block 0 has an active slot with mismatched hash")
validation failed for extent 590: GenericError("block 0 has an active slot with mismatched hash")

[...]

PR #1791, opened 2025-10-14
Check hashes during repair and upon startup

mkeeter:
⚠️  Do not actually merge this, it's for debugging #1788

This calls `Extent::validate` after repair and when opening an region, panicking on failures.

See https://github.com/oxidecomputer/crucible/pull/1792 for details about extent validation.

### DIFF
--- a/downstairs/src/extent_inner_raw.rs
+++ b/downstairs/src/extent_inner_raw.rs
@@ -92,11 +92,14 @@ pub struct RawInner {
     /// This allows us to only write the flag when the value changes
     dirty: bool,
 
-    /// Marks whether the given context slot is dirty
+    /// Marks whether the given block is dirty
     ///
-    /// A dirty context slot has not yet been saved to disk, and must be
-    /// synched before being overwritten.
-    context_slot_dirty: ContextSlotsDirty,
+    /// A dirty block has not yet been saved to disk, and must be synched before
+    /// being overwritten.  In other words, if we write to the same block twice
+    /// (without a flush in between), we must add a sync.
+    ///
+    /// See `crucible#1788` for details on why this is necessary
+    block_dirty: BlockBitArray,
 
     /// Total number of extra syscalls due to context slot fragmentation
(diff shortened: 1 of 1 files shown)


leftwo:
I pushed an update, I hope that was okay.
I wanted to make a proper command out of this, so you can just call verify and pass a directory and not have to have dump do anything else.

downstairs/src/region.rs
```diff
@@ -400,6 +400,30 @@ impl Region {
...
+            })
+            .collect();
+        if !errors.is_empty() {
+            for (number, err) in &errors {
+                warn!(
+                    self.log,
+                    "validation falied for extent {number}: {err:?}"
+                );
+            }
+            panic!("validation failed");
```
jmpesp: I worry about this panic because we can't recover from it like we can the "check during repair" panic. Eventually the downstairs service will be in maintenance and the Upstairs can't do anything to fix this.

downstairs/src/extent.rs
```diff
@@ -251,7 +251,22 @@ pub fn extent_dir<P: AsRef<Path>>(dir: P, number: ExtentId) -> PathBuf {
  * anchored under "dir".
  */
 pub fn extent_path<P: AsRef<Path>>(dir: P, number: ExtentId) -> PathBuf {
-    extent_dir(dir, number).join(extent_file_name(number, ExtentType::Data))
+    let e = extent_file_name(number, ExtentType::Data);
+
+    // XXX terrible hack: if someone has already provided a full directory tree
```
leftwo: We need this so we can verify the incoming copy of an extent?

leftwo (approved):

mkeeter:
The CI failure is because checking downstairs on startup is slower, so there's a race between `dsc` and the downstairs actually starting up.  In this case, `dsc` tried to talk to the downstairs at 20:02:30, but it didn't start serving until 20:02:42.  We can fix this, but it's not a red flag about the rest of the checking.

PR #1793, opened 2025-10-15
Add two repair tests to agent-antagonist

jmpesp:
Discussing possible causes for the zeroes appearing at the beginning of an extent file (#1788), one theory that came up was that the Extent was being repaired, and those zeroes came from reading the Extent from the repair API, not from the disk itself: only _one_ Region's Extents were bad, not all of them.

In order to test this theory, add two new tests, both of which will contact an Agent to get a Region's details, then use the repair API to read Extent data files and either look for a block of zeroes at the beginning or write that to a temporary file and use the new `Extent::validate` routine.

### DIFF
--- a/agent-antagonist/src/main.rs
+++ b/agent-antagonist/src/main.rs
@@ -2,12 +2,15 @@ use anyhow::bail;
 use anyhow::Result;
 use clap::Parser;
 use crucible_common::build_logger;
+use crucible_common::mkdir_for_file;
 use futures::StreamExt;
 use rand::random;
 use signal_hook::consts::signal::*;
 use signal_hook_tokio::Signals;
 use slog::Logger;
 use slog::{info, warn};
+use std::fs::File;
+use std::io::Write;
 use std::net::SocketAddr;
 use std::process::Command;
 use std::sync::{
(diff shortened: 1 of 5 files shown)


agent-antagonist/src/main.rs
```diff
@@ -60,6 +72,56 @@ enum Args {
...
+    RepairTest {
+        /// Address of the crucible agent
+        #[clap(short, long)]
+        agent: SocketAddr,
+
+        /// Region id
+        #[clap(short, long)]
+        region_id: Uuid,
+
+        /// Allow reading extents from a read-write downstairs that may not be
```
leftwo: This seems fraught with peril.  Can we expect this data to be good?

leftwo (commented):
These changes are really just around looking for and checking around the #1788 all zeros situation, right?   We might want to document somewhere how to do a setup where someone could run this, so that information is not lost to time.

agent-antagonist/src/main.rs
jmpesp: Right, we can't expect it to be good no, if it's accepting IO - but in cases where it's RW downstairs that's not currently doing anything, we should expect the data to be good.

agent-antagonist/src/main.rs
```diff
@@ -838,6 +993,98 @@ async fn main() -> Result<()> {
...
+            println!("{} candidate regions", regions.len());
+
+            let dur = std::time::Duration::from_secs(60);
+
+            let client = reqwest::ClientBuilder::new()
+                .connect_timeout(dur)
+                .build()
+                .unwrap();
+
+            loop {
```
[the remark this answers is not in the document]
jmpesp: yep!

PR #1800, opened 2025-10-23
Add `crucible-verify-raw` and `crucible-raw-extent packages`

mkeeter:
This PR extracts the `crucible-verify-raw` tool from #1791.  The on-disk data structures for raw extents are moved to a new `crucible-raw-extent` package, which is used by both `crucible-downstairs` and `crucible-verify-raw`.


leftwo:
I was thinking about the "one binary to rule them all" vs. having separate tools.

Having one binary means it's already available on the sled, which I liked, and if we did not make it part of crucible-downstairs, we would have to add framework to build and install the new binary into the image so we would have it installed without having to copy it in from outside the rack.  

However, it's actually not that hard to solve this another way.  We already have the crucible-dtrace package that is part of every global zone, so we can stuff these tool binaries in this same package without too much additional work.

PR #1846, opened 2025-11-18
Remove cargo nextest from github workflow, out of space

leftwo:
The github provided ubuntu image/host/container has a limited amount of space, and we have apparently hit that limit.  We could figure it out and update it, but this test is already a duplicate of what we do in buildomat, so I just pulled out the test part, so now it just does cargo fmt/build/check.


leftwo:
This should unblock all the dependabot PRs that are failing

PR #1851, opened 2025-12-23
update dropshot to 0.16.7, dropshot-api-manager to 0.5.2

sunshowers:
Contains some improvements required for RFD 619.

### DIFF
--- a/dropshot-apis/src/main.rs
+++ b/dropshot-apis/src/main.rs
@@ -81,6 +81,7 @@ pub fn all_apis() -> anyhow::Result<ManagedApis> {
 
     let apis = ManagedApis::new(apis)
         .context("error creating ManagedApis")?
+        .with_git_stub_storage()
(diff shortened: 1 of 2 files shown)


dropshot-apis/src/main.rs
```diff
@@ -78,12 +76,12 @@ pub fn all_apis() -> anyhow::Result<ManagedApis> {
             },
             api_description:
                 crucible_downstairs_repair_api_mod::stub_api_description,
-            extra_validation: None,
         },
     ];
 
     let apis = ManagedApis::new(apis)
         .context("error creating ManagedApis")?
+        .with_git_stub_storage()
```
sunshowers: I also turned on Git stub storage per RFD 634. This doesn't have any effect in Crucible today since all APIs have just a single version, but it will make Git treat future versions as diffs.

PR #1899, opened 2026-03-04
[crucible-agent-types] migrate to RFD 619 pattern

sunshowers:


agent/src/main.rs
```diff
@@ -1,10 +1,8 @@
-// Copyright 2026 Oxide Computer Company
+// Copyright 2021 Oxide Computer Company
```
jmpesp: `s/2021/2026/` I think, though I'm not exactly sure when these copyright notices should be updated, duplicated for additional years, etc?

jmpesp (approved):

agent/src/main.rs
sunshowers: Yeah -- I think we should just use the latest year.

PR #1907, opened 2026-03-16
Fix warnings

mkeeter:
- `ringbuffer::RingBuffer::push` is deprecated
- Fix warning about returned values being less public than functions, by making the functions less public


leftwo (approved):
Yeah.

See also: https://github.com/oxidecomputer/crucible/pull/1904

PR #1908, opened 2026-03-17
Min-quorum negotiation (again)

### LINKED ISSUE #1690: Support activation with 2/3 read/write downstairs present.

mkeeter:
(This is a revival of #1733, rebased onto `main`)

This implements [RFD 542](https://rfd.shared.oxide.computer/rfd/542) and closes #1690.

Here's the quick version:

- Once we have two downstairs in `WaitQuorum`, we schedule an event to fire after `NEGOTIATION_DELAY` (currently 500 ms)
- If the third Downstairs arrives before this event fires, then we do full-quorum reconciliation (our usual path)
- Otherwise, then we enter min-quorum reconciliation, marking the third Downstairs as faulted (so it must rejoin through live-repair)

I'm opening this as a draft because I want to see how the CI tests go.  The PR includes integration tests for common and uncommon orderings of events, but it's hard to hit every possible path due to specific timing requirements; I'm very open to suggestions for other tests.

Before merging, we also need to figure out https://github.com/oxidecomputer/omicron/issues/8528


crutest/src/main.rs
```diff
@@ -2764,6 +2789,9 @@ async fn replace_before_active(
             }
         }
 
+        // At this point, we've got two Downstairs (one of which was provided
```
leftwo: I think this test is invalid.
Or, at least we should not support replacement of a downstairs unless two other downstairs
are present (or the upstairs is still active).  If we do, we can lose new data.

Let me see if I can explain why.
Let's say all three downstairs are active, and we write A to block one.
```
DS0 | DS1 | DS2
 A  |  A  | A
```
Then, say DS2 goes offline (live update maybe).
Then, we write B to block 1.
Now we have:
```
DS0 | DS1 | DS2
 B  |  B  | A(OFF)
```

Next let's say the VM goes away, so no upstairs is connected.
Then, let's say DS2 comes back.
Then the host with DS0 goes offline (live update again maybe)
For the final piece, say disk DS1 fails, so we want to replace it.
We start up the VM (or the pantry).
DS0 is still offline, DS1 has failed, DS2 is present.
So this is our current state:
```
  DS0  | DS1 | DS2
 B(OFF)| XXX | A
```

If we allow a new disk to replace DS1, it will have no data, and we will repair
that downstairs with the contents of DS2.
If DS0 comes back now, it will be "behind" the other downstairs and because
we now have a newer generation number, it will appear like it has the old
data and will be overwritten.

So, I believe this means the upstairs has to inforce replacement of a downstairs
only when two other downstairs are present and possibly after activation (if we
support 2/3 activation that is).  I'm going to ponder this point a little more though.

leftwo (changes requested):

crutest/src/main.rs
leftwo: In addition, this test from crutest:
```
    /// Test replacement of a downstairs while doing the initial reconciliation.
      ReplaceReconcile {
          /// The address:port of a running downstairs for replacement
          #[clap(long, action)]
          replacement: SocketAddr,
      },  
```
It might be that we don't allow replacement during reconciliation, or don't if all three downstairs are not present?  This gets tricky as once we replace a downstairs, we have to wait for that downstairs to be all synced back up with the other two before we can replace on that region set again.

leftwo:
I've started on removing the "replace before reconcile" ability for this PR.
More changes coming soon

crutest/src/main.rs
leftwo: Okay, I pulled all this out and made a new 2/3 test.

leftwo:
I've removed the tests and ability to replace a downstairs before activation, which I believe is what we want now that we have 2/3 activation.

I also added a specific crutest test for 2/3 activation.  I'll most likely build on that test and roll something into the nightly test set to do a 2/3 activation test.

crutest/src/main.rs
```diff
@@ -3014,6 +2586,237 @@ async fn replace_workload(
...
+
+        for t in targets {
+            if t.state != DownstairsInfoStatus::Active {
+                return false;
+            }
+        }
+    }
+
+    true
+}
```
leftwo, line 2612: The new VolumeInfo struct and this function we can use in a bunch of other places and make better tests out of it.  This can replace some bad bash scripts that do the same thing, but poorly.

leftwo:
Before this goes in (or before propolis/omicron point to it) we must have https://github.com/oxidecomputer/omicron/issues/8528 sorted.

PR #1909, opened 2026-03-17
Bump dropshot to 0.17.0

david-crespo:
Bumping all the things in Omicron that depend on Dropshot.

### DIFF
--- a/agent/src/server.rs
+++ b/agent/src/server.rs
@@ -282,6 +282,7 @@ pub async fn run_server(
         default_request_body_max_bytes: 1024 * 10,
         default_handler_task_mode: HandlerTaskMode::Detached,
         log_headers: vec![],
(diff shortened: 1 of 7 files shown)


agent/src/server.rs
```diff
@@ -281,7 +281,7 @@ pub async fn run_server(
         bind_address,
         default_request_body_max_bytes: 1024 * 10,
         default_handler_task_mode: HandlerTaskMode::Detached,
-        log_headers: vec![],
+        ..Default::default()
```
david-crespo: Happy to change these to

```rs
log_headers: vec![],
compression: CompressionConfig::None,
```

if explicit is preferred.

david-crespo: Eh, I will change it. I prefer explicit myself.

leftwo (approved):

PR #1912, opened 2026-03-18
verify extent after LR

leftwo:
This is not going back to main

Just wanted some extra eyes on it to verify (haha) that it's doing what I want it to do, which is re-compute the hash for all blocks in a newly opened extent and confirm that we have a match in the slot that is active.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -192,6 +192,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "96d30a06541fbafbc7f82ed10c06164cfbd2c401138f6addd8404629c4b16711"
 
 [[package]]
+name = "async-compression"
+version = "0.4.33"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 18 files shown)


mkeeter:
I agree that this does what you want it to do, although you may want to log the full result instead of just `result.is_ok()`.

I'm not sure what the upstairs will do if it gets an `ExtentLiveAckId` with an error code, but it will hopefully be loud.

leftwo:
I want the downstairs to panic and dump core, and I'm looking for cores as a sign of a problem.

mkeeter:
Okay, in that case I _don't_ think it will do what you want 😄 

As written, it's going to return the error to the upstairs.  You want something like `extent.validate().unwrap()` instead.

leftwo:
> For ExtentLiveAckId

I believe the upstairs will abort the repair and start over, but really that's not quite loud enough for me.
I'll also go double check that the dtrace probe for "aborted repair" is still working.  I don't see it on any of my instances, but I should verify it still works.

PR #1913, opened 2026-03-23
Use separate in-memory types

jmpesp:
Serializing and deserializing raw JSON to the Crucible Agent's data file creates the potential for a versioning problem that would be hidden from the normal validation performed to make sure that our API surface doesn't change.

Consider the following:

- a normal update occurs that changes the Crucible Agent's version
  - that newer version's serialized data file is not compatible with the older version

- some Bad Situation occurs and an operator decides to MUPdate back to the previous version

In this scenario the older Crucible Agent will not be able to deserialize what the newer Agent serialized, adding to the existing Bad Situation. In a future when Nexus pays attention to Crucible health and there are Agents that cannot start, this would result in potentially stuck Upstairs and that may cascade to interrupt other operations while Nexus waits for repairs or reconciliations.

This commit separates the in-memory types used in the DataFile struct from the types committed to disk (and adds a comment that warns against changing said type), and further distinguishes between Region states and RunningSnapshot states.

This work is a prerequisite to getting the Agent to work on multiple read-only region clones concurrently as that work adds a new State, and internally discussing the Bad Situation scenario referenced earlier in this message lead to the inspiration to use separate in-memory types.

Additionally, fix a bug where the `key_pem` field stored on disk would be output in the error message that would result from requesting a region that matches an existing one except for that field. Luckily no production system uses Crucible's support for X509 yet and this endpoint is not exposed to users.

### DIFF
--- a/agent-types/src/region.rs
+++ b/agent-types/src/region.rs
@@ -1,13 +1,9 @@
 // Copyright 2025 Oxide Computer Company
 
 use std::net::SocketAddr;
-use std::path::Path;
 
-use crucible_smf::scf_type_t::*;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
-
-use crate::smf::SmfProperty;
 
 #[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(Serialize, Deserialize, JsonSchema, Debug, PartialEq, Clone)]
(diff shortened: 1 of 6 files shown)


leftwo (approved):
While predicting the future is hard, I don't think it's wrong to at least try.

PR #1914, opened 2026-03-23
Reject snapshot requests when read-only

### LINKED ISSUE #1856: Return an error from the Upstairs if trying to snapshot a read-only region set
The root of the problem behind https://github.com/oxidecomputer/omicron/issues/9855 was that the downstairs was retrying an job that would never succeed, and notifying the upstairs every time it failed in a hot loop.

One thought to fix this in the downstairs is that certain errors should be considered transient, and others are permanent. `ModifyingReadOnlyRegion` is almost certainly a condition that will never change and should not be retried. The downstairs will have to notify the upstairs that this condition is true, and the upstairs will have to consider this job failed right away.

The larger problem here is that changing the downstairs to not retry all jobs until completion is a modification of the original design of the upstairs and downstairs, and would entail changing how the upstairs works in a big way. We rely on this retry behaviour for correctness.

Instead: change the upstairs to not even submit these jobs in the first place, and return an error upstack.

jmpesp:
The root of the problem behind oxidecomputer/omicron#9855 was that the downstairs was retrying an job that would never succeed, and notifying the upstairs every time it failed in a hot loop. This consumed memory until it was entirely exhausted.

Fix this by rejecting snapshot requests when read-only in the Upstairs. Note we don't need to check this in Volume because each Upstairs in a Volume should be read-only if any of them are.

Fixes #1856

### DIFF
--- a/upstairs/src/upstairs.rs
+++ b/upstairs/src/upstairs.rs
@@ -1336,15 +1336,23 @@ impl Upstairs {
 
         self.need_flush = false;
 
-        /*
-         * Get the next ID for our new guest work job. Note that the flush
-         * ID and the next_id are connected here, in that all future writes
-         * should be flushed at the next flush ID.
-         */
-
         if snapshot_details.is_some() {
+            if self.cfg.read_only {
+                info!(
+                    self.log,
+                    "rejecting flush with snap request for read-only upstairs",
+                );
+
+                if let Some(res) = res {
+                    res.send_err(CrucibleError::ModifyingReadOnlyRegion);
+                }
+
+                return;
+            }
+
             info!(self.log, "flush with snap requested");
         }
+
         let ds_id =
             self.downstairs
                 .submit_flush(snapshot_details, res, io_guard);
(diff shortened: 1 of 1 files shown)


upstairs/src/upstairs.rs
```diff
@@ -1336,15 +1336,23 @@ impl Upstairs {
...
         self.need_flush = false;
 
-        /*
-         * Get the next ID for our new guest work job. Note that the flush
-         * ID and the next_id are connected here, in that all future writes
-         * should be flushed at the next flush ID.
-         */
-
         if snapshot_details.is_some() {
+            if self.cfg.read_only {
```
leftwo, line 1340: Do we expect to be calling this if things are behaving as expected?
If yes, then maybe we don't want to log it.

If it's not expected, then it's fine to leave the log.

leftwo (approved):

upstairs/src/upstairs.rs
jmpesp: It's definitely not expected - if anything it'd be great to package this up as a fault soemhow (aka https://github.com/oxidecomputer/omicron/issues/10118)

leftwo: Yeah, we should start making a list of these places in crucible where we want to tell someone about it, but don't know how to.

PR #1915, opened 2026-03-31
Retry data writes that return an IO error

jmpesp:
The downstairs should retry all the extent file modifications that are part of a higher level Crucible write until they succeed. Fix a bug where this does not occur for data writes, and add a test to validate the fix.

### DIFF
--- a/downstairs/src/extent_inner_raw.rs
+++ b/downstairs/src/extent_inner_raw.rs
@@ -1167,7 +1167,6 @@ impl RawInner {
     ) -> Result<Vec<T>, CrucibleError>
     where
         F: Fn(Option<OnDiskDownstairsBlockContext>, u64) -> T,
-        T: std::fmt::Debug,
     {
         let mut out = Vec::with_capacity(count as usize);
         let mut reads = 0u64;
--- a/downstairs/src/extent_inner_raw.rs
+++ b/downstairs/src/extent_inner_raw.rs
@@ -19,6 +19,8 @@ use itertools::Itertools;
 use slog::{Logger, error};
 
 use std::collections::HashSet;
+#[cfg(test)]
+use std::collections::VecDeque;
 use std::fs::{File, OpenOptions};
 use std::io::{BufReader, Read};
 use std::os::fd::{AsFd, AsRawFd};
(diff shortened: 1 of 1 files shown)


downstairs/src/extent_inner_raw.rs
```diff
@@ -1498,7 +1555,12 @@ impl RawLayout {
     ) -> Result<(), CrucibleError>
     where
         F: Fn(Option<OnDiskDownstairsBlockContext>, u64) -> T,
+        T: std::fmt::Debug,
     {
+        eprintln!(
```
mkeeter, line 1560: Leftover debugging code?  We shouldn't print this on every call.

downstairs/src/extent_inner_raw.rs
```diff
@@ -1516,8 +1578,10 @@ impl RawLayout {
                     CrucibleError::BadContextSlot(e.to_string())
                 })?;
             let v = f(ctx, block_start + i as u64);
+            eprintln!("v {v:?}");
```
mkeeter, line 1581: likewise

jmpesp: yep, see 1400606

downstairs/src/extent_inner_raw.rs
jmpesp: yep, see 1400606

downstairs/src/extent_inner_raw.rs
```diff
@@ -108,6 +110,23 @@ pub struct RawInner {
 
     /// Most recent flush number, written to context slots for debug purposes
     last_flush: u64,
+
+    /// For testing purposes, change this field to simulate the pwrite
+    /// responsible for writing out the block data succeeding or failing based
+    /// on some configuration.
+    #[cfg(test)]
```
mkeeter, line 117: I don't love putting (even test-gated) debug shenanigans into this struct, but until we live in a world with mockable IO abstractions, I guess it's _fine_...

downstairs/src/extent_inner_raw.rs
```diff
@@ -108,6 +110,23 @@ pub struct RawInner {
...
+    simulate_data_write_error:
```
leftwo, line 118: I'm not in love with the test code being stuffed in here like this, but I can't think of a better way to get the coverage we want from this.

Are we sure that when we compile, we won't get this struct in the release image?

leftwo (approved):

PR #1919, opened 2026-04-09
Background read-only region creation

jmpesp:
When the Crucible Agent is requested to create a read-only region from a remote Downstairs source, this currently blocks the worker thread as region creation is performed in the worker loop, and it cannot respond to other state changes.

This commit spawns region creation threads that the main worker thread can send requests to, and sends all read-only region creation requests there.

This builds on the previous work to separate the serialized on-disk types from the in-memory types: a `Creating` state is added to the in-memory type and used while this background creation is occurring.

### DIFF
--- a/agent/src/datafile.rs
+++ b/agent/src/datafile.rs
@@ -980,7 +980,7 @@ impl DataFile {
                             // requested and may exist. According to zfs list,
                             // it does not yet. How was a snapshot taken?
                             bail!(
-                                "region {} (maybe?) does not exist yet! {e}",
+                                "region {} not found in zfs list! {e}",
                                 request.id.0
                             );
                         }
--- a/agent/src/datafile.rs
+++ b/agent/src/datafile.rs
@@ -1,6 +1,6 @@
 // Copyright 2021 Oxide Computer Company
 
-use anyhow::{Result, anyhow, bail};
+use anyhow::{Result, bail};
 use crucible_agent_types::region::CreateRegion;
 use crucible_agent_types::region::RegionId;
 use crucible_agent_types::snapshot::CreateRunningSnapshotRequest;
@@ -9,8 +9,9 @@ use crucible_agent_types::snapshot::DeleteSnapshotRequest;
 use crucible_agent_types::snapshot::Snapshot;
 use crucible_common::write_json;
 use crucible_smf::scf_type_t::*;
+use dropshot::HttpError;
 use serde::{Deserialize, Serialize};
-use slog::{Logger, crit, error, info};
+use slog::{Logger, crit, error, info, warn};
 use std::collections::BTreeMap;
 use std::net::SocketAddr;
 use std::path::Path;
@@ -38,6 +39,8 @@ pub struct DataFile {
 #[derive(Debug, PartialEq, Clone)]
 pub enum RegionState {
     Requested,
+    /// Region creation is occuring in the background
+    Creating,
     Created,
     Tombstoned,
     Destroyed,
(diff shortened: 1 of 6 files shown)


agent/src/datafile.rs
```diff
@@ -945,7 +975,17 @@ impl DataFile {
                             return Ok(());
                         }
 
-                        RegionState::Requested | RegionState::Created => {
+                        RegionState::Requested | RegionState::Creating => {
+                            // According to the agent's datafile, the region was
+                            // requested and may exist. According to zfs list,
+                            // it does not yet. How was a snapshot taken?
+                            bail!(
+                                "region {} (maybe?) does not exist yet! {e}",
```
leftwo: I think I might change this error message to say "region {} not found in zfs list".  The "maybe" and "yet" imply things that I don't know if they are always true.

agent/src/datafile.rs
```diff
@@ -50,6 +53,33 @@ impl From<RegionState> for crucible_agent_types::region::State {
             RegionState::Requested => {
                 crucible_agent_types::region::State::Requested
             }
+            RegionState::Creating => {
+                // The Crucible agent will only read the data file on startup,
+                // either after being updated or after crashing. There's a few
+                // considerations here that all conclude that Creating should be
+                // converted here to Requested:
+                //
+                // If this is a newer Crucible Agent that supports background
```
leftwo: I think the older/newer comment here is going to confuse the future readers of this section.
It will always be a newer crucible agent from the point of view of the person reading the code.
The comment should be written from the point of view that we are a "newer Crucible Agent" and
describe what happens.

If we do have to roll back, then I feel like it's up to us to dig into the differences between what the current and rolled back software has.  So, while I like the information in this comment, I think we should put it in the PR or somewhere else.  At minimum, we could say that the previous version of Crucible Agent will behave differently, but I think once we get past one upgrade with this as the default, that comment will no longer be needed.

agent/src/main.rs
```diff
@@ -37,6 +37,11 @@ const SERVICE: &str = "oxide/crucible/downstairs";
 const RESERVATION_FACTOR: f64 = 1.25;
 const QUOTA_FACTOR: u64 = 3;
 
+// The number of tasks spawned to handle region clone operations.
+//
+// TODO does this need to be variable?
+const WORKER_COUNT: usize = 64;
```
leftwo, line 43: I wonder if we need to tune this at all?  How many parallel clones could an agent actually perform?  Is requesting 64 from any single agent something we believe could happen?

agent/src/main.rs
```diff
@@ -315,7 +320,7 @@ async fn main() -> Result<()> {
              */
             let log0 = log.new(o!("component" => "worker"));
             let df0 = Arc::clone(&df);
-            std::thread::spawn(|| {
+            tokio::spawn(async {
```
leftwo, line 323: If we are going from a real thread to a tokio task, could a long running region create trip us up here?  The old way was with a thread which seemed like it could go off and do whatever for an hour and the rest of the agent could continue working.  Do we run any risk of that here?

agent/src/main.rs
```diff
@@ -858,102 +897,50 @@ fn worker(
...
-                            Err(e) => {
-                                error!(
-                                    log,
-                                    "Failed to find path for dataset {}: {}",
-                                    &r.id.0,
-                                    e,
-                                );
-                                df.fail(&r.id);
-                                break 'requested;
+                                std::process::exit(1);
```
leftwo, line 929: what happens to the agent if we fail like this?  Is it going to crash and restart?

leftwo (commented):
Thanks for the work here, I have some questions for you.

agent/src/main.rs
jmpesp: Yep!

agent/src/main.rs
jmpesp: I'm not sure about all the differences between threads and tasks, but I don't think there's a risk. With `worker` running in a thread or with a task, the read/write region creation occurs separately from the dropshot server and datafile manipulation logic.

agent/src/main.rs
jmpesp: Pathologically, depending on the disk size, we could have hundreds or even thousands of small regions managed by a single agent, and there's nothing I'm aware of currently limiting the number of read-only region creations that can occur at the same time. I also wonder about tuning this and don't have a good answer.

agent/src/datafile.rs
jmpesp: agreed, done in d482c4a8e46b9a0a9c300f347e25e6ed886d4000.

agent/src/datafile.rs
jmpesp: done in bacd7020, I agree that is confusing

agent/src/main.rs
leftwo: Yeah, there is already a TODO here, so I guess I don't know what more I wanted :)

agent/src/main.rs
leftwo: I'm a little concerned that, if we fail here we restart the whole agent.  If it's a persistant failure, then we get ourselves into a crash loop?

Looking at the error though, it's a failure to send the request over the channel to one of our worker threads, and that should be a difficult situation to reach correct?   And, if we do see it, a restart of the process is likely to behave differently?  I just want to avoid setting ourselves up for crash loop.

jmpesp: If we're crashing here then all the worker tasks have crashed or otherwise closed their receivers, closing the channel, leading to a `SendError`. In this case we should crash the agent.

leftwo (approved):

PR #1927, opened 2026-04-21
Remove bonus Volume layer

jmpesp:
`Volume::construct` was unintentionally adding another layer of Volume when constructing from a request. Remove this, ensuring that what is constructed matches the request.

### DIFF
--- a/upstairs/src/guest.rs
+++ b/upstairs/src/guest.rs
@@ -171,7 +171,7 @@ impl Guest {
         (guest, io)
     }
 
-    pub async fn create_and_up_main(
+    pub fn create_and_up_main(
         log: Logger,
         opts: CrucibleOpts,
         extent_info: RegionExtentInfo,
--- a/upstairs/src/guest.rs
+++ b/upstairs/src/guest.rs
@@ -1,4 +1,5 @@
-// Copyright 2024 Oxide Computer Company
+// Copyright 2026 Oxide Computer Company
+
 use std::{
     net::SocketAddr,
     sync::atomic::{AtomicU64, Ordering},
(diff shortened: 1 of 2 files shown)


leftwo (approved):
I assume (as tests pass) that the volume replacement stuff can still handle and we did not accidentally require multiple volume layers to make things work?

Also, this should not impact what Omicron has in the json blob for a VCR, right?

jmpesp:
> I assume (as tests pass) that the volume replacement stuff can still handle and we did not accidentally require multiple volume layers to make things work?

If anything it's a testament to our code quality that the replacement code worked before and after this change :) I have the same fear here too, but will also trust the tests here.

> Also, this should not impact what Omicron has in the json blob for a VCR, right?

I also had this same fear, but no: the request that's stored in the database doesn't change or need to change here. It's `Volume::construct`'s interpretation of the request that is changing.

upstairs/src/volume.rs
```diff
@@ -261,6 +265,101 @@ impl From<VolumeBuilder> for Volume {
     }
 }
 
+fn to_arc_block_io<T>(b: T) -> Arc<dyn BlockIO + Send + Sync>
```
mkeeter, line 265: I'm mad that this is necessary (and that we can't use `Arc::new` or `.from()` directly), but haven't found a way to avoid it.

upstairs/src/guest.rs
```diff
@@ -166,6 +171,22 @@ impl Guest {
         (guest, io)
     }
 
+    pub async fn create_and_up_main(
```
mkeeter: This function doesn't need `async` (clippy complains in my editor, dunno why it passes in CI)

mkeeter (approved):

upstairs/src/volume.rs
jmpesp: :cry:

upstairs/src/guest.rs
jmpesp: thanks, did that in ae5f3ee0

jmpesp:
I spun up the Canada region and performed a bunch of replacements, so we're good here. Going to merge!

PR #1929, opened 2026-04-23
Document nightly test assumptions

leftwo:
Added some documentation around what the nightly class tests expect for incoming environmental variables and how the tests should use them.  These are things we already did, just writing them down somewhere.

Other work coming later in this area.

Added a bit more info about the system we are running on for the nightly test.

### DIFF
--- a/tools/test_nightly.sh
+++ b/tools/test_nightly.sh
@@ -17,6 +17,11 @@ export BINDIR=${BINDIR:-$ROOT/target/release}
 echo "Nightly starts at $(date)" | tee "$output_file"
 echo "Running on $(git log -1 --no-color | head -20)" | tee -a "$output_file"
 echo "" >> "$output_file"
+echo "System info:" | tee -a "$output_file"
+psrinfo -v | grep "operates at" | head -1 | tee -a "$output_file"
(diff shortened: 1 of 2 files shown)


tools/test_nightly.sh
```diff
@@ -17,6 +17,11 @@ export BINDIR=${BINDIR:-$ROOT/target/release}
 echo "Nightly starts at $(date)" | tee "$output_file"
 echo "Running on $(git log -1 --no-color | head -20)" | tee -a "$output_file"
 echo "" >> "$output_file"
+echo "System info:" | tee -a "$output_file"
+psrinfo -v | grep "operates at" | head -1 | tee -a "$output_file"
+cpu_count=$(psrinfo  | wc -l | tr -d ' ')
+echo "CPU count: $cpu_count" | tee -a "$output_file"
```
jmpesp, line 23: what about `nproc`?

jmpesp (approved):

tools/test_nightly.sh
leftwo:
```
BRM42220009 # nproc
-bash: nproc: command not found
```

PR #1933, opened 2026-05-08
Update omicron/oximeter

ahrens:
Now that Omicron uses reqwest 0.13, we can update omicron/oximeter, and remove the explicit dependency on reqwest 0.12.

`cargo update oximeter oximeter-producer progenitor@0.13.0 crossterm@0.28`


sunshowers (approved):
Thanks for fixing this up!

leftwo (approved):
Just noticed you had tagged me on this Friday.
Not that you need it, as you have an approval already, but you have mine now too :)

PR #1936, opened 2026-05-14
Use an explicit rev for oxidecomputer git deps

jmpesp:
Chasing down a cargo build failure in propolis caused by multiple versions of the same package lead to the discovery that rev of omicron that is in propolis' lock file is old, about 700 commits behind main. Once propolis was updated to the same rev of omicron that crucible uses, cargo no longer hit that same build error.

Use explicit git revs in Cargo.toml: this doesn't change the version in the lock file but does make it explicit what is being used. Propolis will be updated shortly to pull this commit in (a no-op, as the lockfile version doesn't change), and will be updated to use explicit revs in the same way.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -159,7 +159,7 @@ checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 [[package]]
 name = "api_identity"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/omicron?branch=main#fa779b29af7a8656657ccca6ec531446188d916c"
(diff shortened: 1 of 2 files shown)


leftwo (approved):
Does this just make it obvious that we are not updating these dependencies?
because of the lock file, it means that we get "main" from whatever time the lockfile has and it won't be updated unless someone does something to update it?
Also, renovate, should it be tracking this?

jmpesp:
> Does this just make it obvious that we are not updating these dependencies?
> because of the lock file, it means that we get "main" from whatever time the lockfile has and it won't be updated unless someone does something to update it?

yeah, it's basically moving the information into the much more readable Cargo.toml.

> Also, renovate, should it be tracking this?

Maybe! Not sure :)

PR #1939, opened 2026-05-20
[meta] update dropshot-api-manager to 0.7.2

sunshowers:
Has a neat new UI for reporting incompatibilities (see https://github.com/oxidecomputer/dropshot-api-manager/pull/105).


leftwo (approved):
I spent the last two months digging deep deep into this just to be sure it's good.

PR #1942, opened 2026-06-01
dsc: add --oximeter flag to publish downstairs metrics

gburd:
## Summary

`dsc` spawns `crucible-downstairs` instances but never passes `--oximeter`, so a `dsc`-managed fleet emits no Oximeter metrics — there's no way to wire the downstairs producer endpoints up to a collector.

This adds an optional `--oximeter <addr>` flag to `dsc start`. When set, the address is threaded through to every downstairs `dsc` spawns (`crucible-downstairs run ... --oximeter <addr>`), mirroring how the existing `read_only` option is propagated through `DownstairsInfo`/`DscInfo`. When unset, behavior is unchanged.

This makes it possible to point a `dsc` test fleet at an Oximeter collector (or a Nexus-compatible registration endpoint) for metrics collection.

## Implementation

- `DownstairsInfo` and `DscInfo` gain an `oximeter: Option<SocketAddr>` field, propagated from the `Start` subcommand.
- `DownstairsInfo::start()` appends `--oximeter <addr>` to the spawned command's args only when the option is set.
- Non-`Start` paths (`Create`, `RegionPerf`) pass `None`.

## Test plan

- [x] `cargo fmt -p dsc --check` — clean
- [x] `cargo clippy -p dsc --all-targets -- -D warnings` — clean
- [x] `cargo test -p dsc` — all 16 tests pass
- [x] Manually verified end-to-end: `dsc start --create --oximeter <addr>` causes each spawned downstairs to register with and publish metrics to the collector at `<addr>`.

### DIFF
--- a/dsc/src/main.rs
+++ b/dsc/src/main.rs
@@ -188,6 +188,12 @@ enum Action {
         #[clap(long, default_value = "15", action)]
         extent_count: u32,
 
+        /// Registration address of an Oximeter collector. When set, each
+        /// downstairs is started with `--oximeter <addr>` and publishes
+        /// metrics to that collector.
+        #[clap(long, action)]
+        oximeter: Option<SocketAddr>,
+
         /// Downstairs will all be started read only (default: false)
         #[clap(long, action, default_value = "false")]
         read_only: bool,
@@ -234,6 +240,9 @@ struct DownstairsInfo {
     output_file: PathBuf,
     client_id: usize,
     read_only: bool,
+    /// Registration address of an Oximeter collector. When set, the spawned
+    /// downstairs is told to register and publish metrics (`--oximeter`).
+    oximeter: Option<SocketAddr>,
 }
 
 impl DownstairsInfo {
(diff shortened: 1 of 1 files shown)


leftwo:
This is cool, thanks for doing this.

How did you test the oximeter side?  Like, did you run a server and point the downstairs at it?  Were you able to read back the metrics that were sent to it?  I'm curious to hear more about the steps here.  Thanks.

gburd:
Hello, thanks.  Crucible is good work so I thought I'd add a block device driver to OSv for it.
https://codeberg.org/gregburd/osv/commit/5294ad9083e2316304f99534ae5112ee6d923c9d

So I can use ZFS to knit together a few volumes into one.  I have some code that slurps up the metrics and re-publishes them via Grafana and OpenTelemetry.  It's not nearly as cool as Oximeter, but it's all I've got right now.
https://codeberg.org/gregburd/osv/commit/5294ad9083e2316304f99534ae5112ee6d923c9d

Issue #1947, opened 2026-06-10
A read only upstairs can get a downstairs stuck in a state the upstairs can't handle.


leftwo:
It's possible for the upstairs to put a read only downstairs in a state where the upstairs can't make forward progress with it.

We saw this in https://github.com/oxidecomputer/crucible/issues/1945, where a quick activation/deactivation resulted in the downstairs in this state and then the pantry could make no further progress with it.

The invalid state looks like:

```
00:01:20.120Z WARN crucible-pantry (datafile): Invalid upstairs state Deactivating(BlockRes(Some(Sender { inner: Some(Inner { state: State { is_complete: false, is_closed: false, is_rx_task_set: true, is_tx_task_set: false } }) }))) for connecting when read only                                                  
    session_id = 467934bc-b4f1-475c-a72a-7836cd8cbdd4                     
```

leftwo:
Fixed with https://github.com/oxidecomputer/crucible/pull/1948

PR #1948, opened 2026-06-11
Handle read-only downstairs reaching WaitQuorum during deactivation

leftwo:
When a read-only upstairs is deactivating and a downstairs reconnects and negotiates to WaitQuorum,
on_reconciliation_skipped() previously hit a catch-all arm that logged a warning but left the client stuck in WaitQuorum. Since ready_to_deactivate() only returns true for WaitConnect, deactivation could never complete.

Deactivate any client in WaitQuorum when the upstairs is in the Deactivating state. This moves them through
Stopping(Deactivated) and back to WaitConnect, allowing deactivation to finish.

Fix for https://github.com/oxidecomputer/crucible/issues/1947

### DIFF
--- a/upstairs/src/upstairs.rs
+++ b/upstairs/src/upstairs.rs
@@ -2082,7 +2082,7 @@ impl Upstairs {
                             self.log,
                             "deactivating client {i} in \
                              WaitQuorum during read-only \
-                             deactivation"
+                             reconciliation skip"
                         );
                         self.downstairs.clients[i].deactivate(&self.state);
                     }
@@ -4984,7 +4984,8 @@ pub(crate) mod test {
     #[tokio::test]
     async fn deactivate_ro_with_late_third_downstairs() {
         // Activate with only two downstairs
-        let mut up = Upstairs::test_default(None, true);
+        let ddef = RegionDefinition::default();
+        let mut up = Upstairs::test_default(Some(ddef), true);
         for cid in [ClientId::new(0), ClientId::new(1)] {
             for state in [
                 NegotiationStateData::Start,
@@ -5042,7 +5043,6 @@ pub(crate) mod test {
             NegotiationStateData::WaitForPromote,
             NegotiationStateData::WaitForRegionInfo,
             NegotiationStateData::GetExtentVersions,
-            NegotiationStateData::WaitQuorum(Default::default()),
         ] {
             up.ds_transition(
                 cid2,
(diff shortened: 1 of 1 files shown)


upstairs/src/upstairs.rs
```diff
@@ -2065,6 +2066,28 @@ impl Upstairs {
...
+                        DsState::Connecting {
+                            state: NegotiationState::WaitQuorum,
+                            ..
+                        }
+                    ) {
+                        info!(
+                            self.log,
+                            "deactivating client {i} in \
+                             WaitQuorum during read-only \
+                             deactivation"
```
mkeeter:
```suggestion
                             reconciliation skip"
```

upstairs/src/upstairs.rs
```diff
@@ -4955,4 +4978,108 @@ pub(crate) mod test {
...
+                    mode: ConnectionMode::New,
+                    state,
+                },
+            );
+        }
+
+        // Client 2 is now in WaitQuorum during deactivation. Call
+        // connect_ro_region_set which is the path triggered when a
+        // read-only downstairs reaches WaitQuorum.
+        up.connect_ro_region_set();
```
mkeeter: I'd suggest writing the test using `up.apply(..)`, so that we don't have to manually call internal functions (which is fragile if the implementation changes).  Here's a patch:

```diff
diff --git a/upstairs/src/upstairs.rs b/upstairs/src/upstairs.rs
index aa6eec8018..e022eec6b1 100644
--- a/upstairs/src/upstairs.rs
+++ b/upstairs/src/upstairs.rs
@@ -4984,7 +4984,8 @@
     #[tokio::test]
     async fn deactivate_ro_with_late_third_downstairs() {
         // Activate with only two downstairs
-        let mut up = Upstairs::test_default(None, true);
+        let ddef = RegionDefinition::default();
+        let mut up = Upstairs::test_default(Some(ddef), true);
         for cid in [ClientId::new(0), ClientId::new(1)] {
             for state in [
                 NegotiationStateData::Start,
@@ -5042,7 +5044,6 @@
             NegotiationStateData::WaitForPromote,
             NegotiationStateData::WaitForRegionInfo,
             NegotiationStateData::GetExtentVersions,
-            NegotiationStateData::WaitQuorum(Default::default()),
         ] {
             up.ds_transition(
                 cid2,
@@ -5053,10 +5054,18 @@
             );
         }
 
-        // Client 2 is now in WaitQuorum during deactivation. Call
-        // connect_ro_region_set which is the path triggered when a
-        // read-only downstairs reaches WaitQuorum.
-        up.connect_ro_region_set();
+        // Simulate an ExtentVersions message, which should put Client 2 into
+        // WaitQuorum.  It should then immediately shift to deactivating.
+        warn!(up.log, "about to send ExtentVersions message");
+        up.apply(UpstairsAction::Downstairs(DownstairsAction::Client {
+            client_id: ClientId::new(2),
+            action: ClientAction::Response(Message::ExtentVersions {
+                gen_numbers: vec![],
+                flush_numbers: vec![],
+                dirty_bits: vec![],
+            }),
+        }));
+        warn!(up.log, "done sending ExtentVersions message");
 
         // Client 2 should now be deactivating, not stuck.
         assert_eq!(
```

(feel free to remove the `warn!` messages, although I found them helpful when scrolling through the logs)

upstairs/src/upstairs.rs
leftwo: done

upstairs/src/upstairs.rs
leftwo: Also done.

mkeeter (approved):

Issue #1945, opened 2026-06-08
The pantry holds locks while waiting for activation and/or deactivation


askfongjojo:
During rack2 online update, a certain snapshot request couldn't run to completion even though the snapshot was successfully created. The disk used to create the snapshot was found to be in `faulted` state.

Here is the saga that fails to be quiesced. Notice that the saga has been pending for 57+ minutes.
```
root@oxz_switch0:~# omdb nexus quiesce show
note: Nexus URL not specified.  Will pick one from DNS.
note: using DNS from system config (typically /etc/resolv.conf)
note: (if this is not right, use --dns-server to specify an alternate DNS server)
note: using Nexus URL http://[fd00:1122:3344:10b::79]:12232
quiescing since 2026-06-08T22:11:50.766Z (57m 12s 792ms ago)
details: waiting for running sagas to finish
saga quiesce:
    new sagas: DisallowedQuiesce
    drained as of blueprint: none
    blueprint for last completed recovery pass: f0bc3c73-356f-4db3-9fa8-0ae3467abd78
    blueprint for last reassignment pass: f0bc3c73-356f-4db3-9fa8-0ae3467abd78
    reassignment generation: 1 (pass running: no)
    recovered generation: 1
    recovered at least once successfully: yes
    recovery pending: no
    sagas running: 1
        saga a2071c6e-5422-4173-bd8e-dccdbdb4382c pending since 2026-06-08T21:25:19.519Z (snapshot-create)
database connections held: 1
    claim 633281 held since 2026-06-08 23:08:52.664049482 UTC (10s 922ms ago)
```
Here are the saga steps:
```
root@oxz_switch1:~# omdb db saga show a2071c6e-5422-4173-bd8e-dccdbdb4382c
note: database URL not specified.  Will search DNS.
note: (override with --db-url or OMDB_DB_URL)
note: using DNS from system config (typically /etc/resolv.conf)
note: (if this is not right, use --dns-server to specify an alternate DNS server)
note: using database URL postgresql://root@[fd00:1122:3344:109::3]:32221,[fd00:1122:3344:105::3]:32221,[fd00:1122:3344:10b::3]:32221,[fd00:1122:3344:107::3]:32221,[fd00:1122:3344:108::3]:32221/omicron?sslmode=disable
WARN: found schema version 261.0.0, expected 263.0.0
It's possible the database is running a version that's different from what this
tool understands.  This may result in errors or incorrect output.
[...]

PR #1949, opened 2026-06-12
Handle locks better in the pantry.

leftwo:
Improve the lock handling around pantry actions that require outside activity to make progress.  We were holding a lock and then activating or deactivating a volume, and this can take unknown amount of time.

Now we take a lock and leave a state to indicate someone is working on the activation but give back the lock so other actions can continue. A new PantryEntryState enum tracks whether a volume is Attaching (construct/activate in progress) or Attached (ready for use). This prevents duplicate concurrent attach attempts while allowing other volumes or other actions to proceed independently.

entry_get() now returns 503 for Attaching volumes so callers retry. 
This might require some updates on the Nexus/Omicron side.

detach() rejects Attaching volumes until attach completes, but this was the same behavior as before, but just because we held the lock for the entire attach.

More unit tests added, as well as a specific pantry test, but I'm not sure if this tools/test_pantry_hang.sh is worth keeping or not.  I think growing the pantest into a more functional/stress test of the pantry might be a better way to go.  I also might consider putting something into the CI tests for the pantry, maybe this crude script is good enough?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1747,8 +1747,10 @@ dependencies = [
  "serde_json",
  "sha2",
  "slog",
+ "slog-term",
  "static_assertions",
  "subprocess",
+ "tempfile",
  "tokio",
  "uuid",
 ]
--- a/pantry/Cargo.toml
+++ b/pantry/Cargo.toml
@@ -35,4 +35,6 @@ crucible-workspace-hack.workspace = true
 
 [dev-dependencies]
 expectorate.workspace = true
+slog-term.workspace = true
 subprocess.workspace = true
+tempfile.workspace = true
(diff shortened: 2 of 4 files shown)


leftwo:
Fix for https://github.com/oxidecomputer/crucible/issues/1945

jmpesp:
I disagree with having GET for entry return a 503 during activation, it seems wrong to me. Previously the endpoint was returning a 200 and the caller would look at the return struct to see what the activation status is - a 503 could mean anything.

I agree with the shape of the changes here, in that we hold the entries lock for as short a time as we can, and would like to chat next week: I think we can get to the same spot without the new enum.

leftwo:
This PR obsoleted by https://github.com/oxidecomputer/crucible/pull/1952/

PR #1962, opened 2026-07-23
Fast ack a flush to a RO volume.

leftwo:
Added a test to verify that a read only volume that can connect to just one downstairs will be able to activate and read from that volume.

Shade was being thrown on Crucible, and this test was written in part to confirm that a single downstairs is all you need to talk to a RO volume.

quotes from the chat:

> but we get vcpus going at 19:20:57.697 and then at 19:21:07.691 propolis logs about timeouts two downstairses
> so that seems funky

And
> "msg":"client task is sending Done(ConnectionFailed(Os { code: 146, kind: ConnectionRefused, message: \"Connection refused\" }))"
> after some seconds when the upstairs tried again, something was quite sad

And
> it does seem odd to have failed to connect to a downstairs though
> this doesn't explain why we'd fail to boot the RO volume for sure

And
> because if that means that downstairs 0 and 1 were active, but 2 wasn't, and then we lost 0 and 1 .. that sounds like it would be a problem
> but i don't know if that's possible

These slights were handled, and Crucible's good standing was returned.

### DIFF
--- a/integration_tests/src/lib.rs
+++ b/integration_tests/src/lib.rs
@@ -1211,6 +1211,18 @@ mod integration_tests {
 
         assert_eq!(vec![0x00; BLOCK_SIZE], &buffer[..]);
 
+        // Manually send a flush.  With only one downstairs running, the flush
+        // still completes: the two stopped downstairs have their jobs moved to
+        // Skipped, so the flush is complete on all clients and acks back to us.
+        // This should not hang.
+        volume.flush(None).await?;
+
+        // A second read after the flush should also complete successfully.
+        let mut buffer = Buffer::new(1, BLOCK_SIZE);
+        volume.read(BlockIndex(0), &mut buffer).await?;
+
+        assert_eq!(vec![0x00; BLOCK_SIZE], &buffer[..]);
+
         Ok(())
     }
 
--- a/integration_tests/src/lib.rs
+++ b/integration_tests/src/lib.rs
@@ -487,6 +487,17 @@ mod integration_tests {
             downstairs.clone_region(source).await
         }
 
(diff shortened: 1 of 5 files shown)


integration_tests/src/lib.rs
```diff
@@ -1160,6 +1171,61 @@ mod integration_tests {
...
+        let opts = tds.opts();
+
+        tds.downstairs2.stop().await?;
+        tds.downstairs3.stop().await?;
+
+        let vcr = VolumeConstructionRequest::Volume {
+            id: Uuid::new_v4(),
+            block_size: BLOCK_SIZE as u64,
+            sub_volumes: vec![],
+            read_only_parent: Some(Box::new(
```
jmpesp: Flushes won't be sent to read-only parents, you'll have to change the structure of the VCR to put this under sub volumes instead

upstairs/src/upstairs.rs
```diff
@@ -1172,7 +1172,10 @@ impl Upstairs {
                     done.send_err(CrucibleError::UpstairsInactive);
                     return;
                 }
-
+                if self.cfg.read_only {
```
leftwo, line 1175: This is the real change, the rest is to support the test changes.

jmpesp, line 1175: Maybe put a comment in for future us about how this doesn't affect the auto flush that the upstairs sends as part of the flush check?

jmpesp (approved):

upstairs/src/upstairs.rs
leftwo: Done

PR #1970, opened 2026-09-09
Add pid to the raw dtrace output, update cmon

leftwo:
upstairs_raw.d matches every upstairs on the system, so it will print records interleaved, but nothing in a record said which upstairs it came from.  To fix this we now wrap each line with the pid:

  {"pid":<pid>,"status":{ ...DtraceInfo... }}

Fixe a panic in cmon where the job delta was tracked in a single u64, which meant "the last job ID from whichever session printed last", so with two upstairs running the subtraction underflowed and cmon aborted with 'attempt to subtract with overflow'.  Track the last job ID per session instead, and report --- for a session's first record, which has nothing to compare against.

Changing the format the script emits means an older cmon reading the new output fails with 'missing field pid'.

Make dtrace-decode print names that -o accepts.  Display was seventeen hand written strings with nothing tying them to the values clap derives from the variant names.
Delegate Display to clap's PossibleValue instead, so the label is by construction a string you can hand back to -o, and test the round trip.


cmon-common/src/lib.rs
```diff
@@ -9,12 +9,29 @@
...
+pub struct DtraceWrapper {
+    pub pid: u32,
+    pub status: DtraceInfo,
+}
+
 /// The possible fields we will display when receiving DTrace output.
 #[derive(Debug, Copy, Clone, PartialEq, Eq, ValueEnum, EnumIter)]
 pub enum DtraceDisplay {
+    Pid,
+    Session,
```
jmpesp: this should probably be SessionId?

jmpesp (approved):

cmon-common/src/lib.rs
leftwo: It did get changed in a later PR to SessionId, yeah, but no reason not to make it that now.

PR #1971, opened 2026-09-09
cmon: Add the remaining dtrace display fields and a default set.

leftwo:
Added eight more fields from DtraceInfo that had no way to be displayed:
ds_reconciled and ds_reconcile_needed on their own rather than only as part of the reconcile trio.
write_bytes_out, ds_ro_lr_skipped, and the four ds_io_count categories split out so a row can show just the ones you care about.

Give -o a default set worth looking at, and share it as default_display_fields() so anything else displaying this data starts
from the same place.

Three columns were wrong.  ds_delay printed three per-client values under a single DELAY heading, so everything to its right was shifted. Delay now gets DLY0 DLY1 DLY2 like the other per-client fields.  
live_repair labelled its third column LRC0 instead of LRC2.  
next_job_id was seven wide, which a long lived upstairs overflows, so it goes to ten.

With ds_delay fixed, every field's header and row are the same width, so the known-broken exemption in that test goes away.


cmon-common/src/lib.rs
```diff
@@ -319,20 +391,63 @@ pub fn format_row(
...
+                result.push_str(&format!(" {:>10}", d_out.next_job_id));
             }
             DtraceDisplay::JobDelta => match delta {
                 Some(delta) => result.push_str(&format!(" {delta:>5}")),
                 None => result.push_str(&format!(" {:>5}", "---")),
             },
             DtraceDisplay::DsDelay => {
                 result.push_str(&format!(
-                    " {:5} {:5} {:5}",
+                    " {:>5} {:>5} {:>5}",
```
jmpesp, line 402: a lot of these other ones don't have the right justification either

cmon-common/src/lib.rs
```diff
@@ -271,6 +337,12 @@ pub fn format_row(
...
             }
+            DtraceDisplay::DsReconciled => {
+                result.push_str(&format!(" {:>4}", d_out.ds_reconciled));
+            }
+            DtraceDisplay::DsReconcileNeeded => {
+                result.push_str(&format!(" {:>4}", d_out.ds_reconcile_needed));
+            }
             DtraceDisplay::LiveRepair => {
                 result.push_str(&format!(
                     " {:4} {:4} {:4}",
```
jmpesp: eg ^

jmpesp (approved):

cmon-common/src/lib.rs
leftwo: done

cmon-common/src/lib.rs
leftwo: done

PR #1975, opened 2026-09-14
Alan/ctop select

leftwo:
Arrow keys move a cursor to allow selection of a specific session.
The row selected adds a '>' and the footer updates as [n/total].
    
The cursor records a session rather than a row number.  Sessions arrive
and expire underneath it, and if we use a row number it would move
around if sessions come and go.
    
When the selected session is itself the one that goes away, the cursor
goes back to the top.
    
The table is rendered with its state so it scrolls itself once there
are more sessions than rows.


ctop/src/main.rs
```diff
@@ -264,6 +268,76 @@ fn is_quit(key_event: KeyEvent) -> bool {
...
+    let mut sessions: Vec<&SessionData> = state.sessions.values().collect();
+    sessions.sort_by_key(|s| (s.pid, &s.dtrace_info.session_id));
+    sessions
+        .into_iter()
+        .map(|s| s.dtrace_info.session_id.clone())
+        .collect()
+}
+
+/// Move the cursor one row, stopping at either end.
+fn move_selection(state: &mut CtopState, down: bool) {
```
jmpesp, line 302: using bool here feels weird, what about a `Navigation` enum, plus we may want to add things in the future, like left, right, top, bottom, etc

ctop/src/main.rs
```diff
@@ -264,6 +268,76 @@ fn is_quit(key_event: KeyEvent) -> bool {
...
+    let ids = sorted_session_ids(state);
+
+    let current = state
+        .selected_session
+        .as_ref()
+        .and_then(|id| ids.iter().position(|s| s == id))
+        .unwrap_or(0);
+
+    let next = if down {
+        current + 1
```
jmpesp, line 312: optional: we could wrap around back to the top here instead of having the cursor stay put?

ctop/src/main.rs
```diff
@@ -301,10 +376,20 @@ fn render_table_view(
...
 
+        let selected = table_state.selected();
+
         let rows: Vec<Row> = sessions
             .iter()
-            .map(|s| {
-                let indicator = if is_stale(s, now) { '*' } else { ' ' };
+            .enumerate()
+            .map(|(idx, s)| {
+                // Cursor location will override the stale flag.
```
jmpesp: it may actually be worth having both here, I can see being surprised if the selected session suddenly goes away

jmpesp (approved):

ctop/src/main.rs
leftwo: ha!  It actually was a delta originally.
The only place that used it was a bunch of code that would allow the cursor to stay on the same session
when other sessions come and go, and I tore it all out as it seemed like a bunch of complexity for features
that we were not using yet, and wanted to ease the reviewer burden.

ctop/src/main.rs
leftwo: Interesting.

Though, if you have more sessions than will fit inside the window, it will scroll things, if you got to the end and went back to the top, it might not be obvious you had scrolled too far.  I think I would prefer to keep the selection from looping.

ctop/src/main.rs
leftwo: What if it was a third indicator?  if I do both then I have to do more column stuff,  If I made it '#' would that work?

jmpesp: How much more column stuff would be required? `>` and `*` and `#` is not as immediately intuitive as ` >` and `*>`

leftwo: Well, you called my bluff, it was not that much code to show both.

ctop/src/main.rs
leftwo: I'll add a Navigation in a later PR if we want more movement.

jmpesp (approved):

